Issue #12, opened 2018-06-26
Routing app refinements


lovelydinosaur:
* Include `HEAD` if `GET` is in `methods`.
* Set `Allow` header on responses.
* Respond to OPTIONS with 200 responses.

jordaneremieff:
Would it make sense to extend the routing app with a simple convenience decorator? I replaced the routing in Afiqah using the Starlette routing components - could we have something like [this](https://github.com/erm/afiqah/blob/master/afiqah/routing.py) and [this](https://github.com/erm/afiqah/blob/master/afiqah/app.py#L33) here?

Example usage:

```python
@app.route("/")
def home(scope):
    return HTMLResponse("<h1>Hello</h1><p>World</p>")
```

lovelydinosaur:
@erm I reckon so, yup.

Issue #3, opened 2018-06-25
Starlette's scope


jordaneremieff:
I am very interested in contributing to some common middleware for ASGI, and this seems like it is the perfect place to do that, however I have a few questions about how much Starlette is aiming to achieve:

How much functionality will this project implement?

How much functionality does it intend to leave up to particular frameworks?

Will it potentially become a more traditional framework itself, or is the intention to remain a small library?

lovelydinosaur:
The set of stuff that it makes sense to provide for here is ~anything that is composable.

Would be in scope:

* A routing middleware, where each target is itself an ASGI app.

Would not be in scope:

* An application instance that configures templates, parsers, renders etc...

Think along similar lines to werkzeug for scope.

lovelydinosaur:
Okay, so there's a bit of an case-in-point in https://github.com/encode/starlette/pull/5

It adds a router, but doesn't include anything like Django's "reverse()".

Instead, I think it should stick with simple, composible units, that make it easy to glue things together, and that also serve as a reference point for how to build more fully featured equivalents.

jordaneremieff:
Ah yes, that clears things up nicely. Thanks.

PR #17, opened 2018-07-11
Routing decorator

jordaneremieff:
Refs https://github.com/encode/starlette/issues/12

Implements a method on the `Router` class to allow a simple means of routing ASGI applications to `Path` objects via a decorator.

Example:

```python
from starlette import Router, Response


app = Router()


@app.route('/', methods=['GET'])
def homepage(scope):
    return Response('Hello, world', media_type='text/plain')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,20 @@ app = Router([
 The router will respond with "404 Not found" or "406 Method not allowed"
 responses for requests which do not match.
 
+Additionally, the router offers a convenience decorator method for defining ASGI application routes.
+
+```python
+from starlette import Router, Response
+
+
+app = Router()
+
+
+@app.route('/', methods=['GET'])
+def homepage(scope):
+    return Response('Hello, world', media_type='text/plain')
+```
+
 ---
 
 ## Test Client
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ app = Router([
 The router will respond with "404 Not found" or "406 Method not allowed"
 responses for requests which do not match.
 
-Additionally, the router offers a convenience decorator method for defining ASGI application routes.
+Additionally, the router offers a convenience method for defining ASGI application routes.
 
 ```python
 from starlette import Router, Response
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -69,7 +69,7 @@ class PathPrefix(Route):
 
 
 class Router:
-    def __init__(self, routes: typing.List[Route], default: ASGIApp = None) -> None:
+    def __init__(self, routes: typing.List[Route] = [], default: ASGIApp = None) -> None:
         self.routes = routes
         self.default = self.not_found if default is None else default
 
@@ -82,3 +82,10 @@ class Router:
 
     def not_found(self, scope: Scope) -> ASGIInstance:
         return Response("Not found", 404, media_type="text/plain")
+
+    def route(self, path: str, methods: typing.Sequence[str] = ()) -> None:
+        def _route(app: ASGIApp) -> ASGIApp:
+            route = Path(path=path, app=app, methods=methods)
+            self.routes.append(route)
+            return app
+        return _route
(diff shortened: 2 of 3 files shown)


jordaneremieff:
I understand what you mean, I think fleshing out a more traditional view pattern in a pure-ASGI context would make things quite a bit more clear.

Two question I have at this point:

1) Should we also make a more explicit distinction between application/protocol routing vs. url/path routing? 

2) Would it make sense to introduce a view pattern into Starlette, or would that be out of scope? 

Because how I am understanding the current routing components _seems_ consistent with how the decorator is being used here, and I would think any confusion using the decorator would be experienced when using the current router functionality also.

I could continue implementing this outside of Starlette and work the routing components into a more appropriate view pattern (eventually), however, I think the convenience of this routing style could be desirable to enough people to include here somehow.

jordaneremieff:
Thinking about this a bit more, perhaps the `asgi_application` decorator could be the answer to my second question (re: introducing a view pattern)? I wasn't clear on its purpose initially and ignored it when I was working on this - maybe refactoring that method with routing in mind is the answer here, not sure.

jordaneremieff:
Somewhat related to this: 

I started on a new project using Starlette, and I'm currently trying to figure out a good way of handling the routing/templating/views. Below is my first attempt to come up with a CBV pattern using the functionality `asgi_application` provides. 


```python
import asyncio
from starlette import Request, Response
from starlette.routing import Router, Path
from starlette.types import Receive, Send, Scope


class View:
    def __init__(self, scope: Scope):
        self.scope = scope

    async def __call__(self, receive: Receive, send: Send) -> None:
        request = Request(self.scope, receive)
        request_method = request.method if request.method != "HEAD" else "GET"
        func = getattr(self, request_method.lower(), None)
        if func is None:
            raise Exception(
                f"Method {request_method} is not implemented for this view handler."
            )
        if asyncio.iscoroutinefunction(func):
        else:
            response = func(request)
        await response(receive, send)


class HomepageView(View):
    # async def get(self, request):
    #     response = Response("Hello async world!", media_type="text/plain")
    #     return response

    def get(self, request):
        response = Response("Hello sync world!", media_type="text/plain")
        return response

app = Router([Path("/", app=HomepageView, methods=["GET"])])
```

It would need to be finessed a bit and the `asgi_application` logic could be pulled out of the view class itself and implemented using a `make_asgi` method in the way you described above - think something along the lines of Django's `as_view()`.

Would something like this be in the scope of Starlette? If it is too much to include here, I could see about doing a PR for just the `make_asgi` bit and implement the view pattern in Afiqah.

lovelydinosaur:
Exactly what I was thinking, yes. Perhaps @asgi_application could be renamed to ‘@view’ too.

lovelydinosaur:
One other thing here: we *could* run sync views inside a threadpool, always.

For static views that just return some templates data or a fixed response, that’s just unnecessary overhead, but for most teal view cases it’d be what you want.

That’d allow you to use either standard thread-concurrency code at the view level, or asyncio-concurrency code instead if needed.

jordaneremieff:
What are your thoughts on creating a distinction between application and view routing? Also about including class-based views? The answers in https://github.com/encode/starlette/issues/3 made things generally clear, but there are still a few grey areas in regards to the scope of Starlette that are relevant here.

lovelydinosaur:
> What are your thoughts on creating a distinction between application and view routing?

I think we should only provide routing to ASGI applications. For class based views that works seemlessly, since View *is* an ASGI app.

For function based views, we’ve got a request/response interface tho. Either you can wrap that function up in a decorator to make it an ASGI app (That’s what we currently have documented, tho I don’t like that it changes the signature, makes the interface less directly testable) or instead wrap the adapter around the function within the routing, something like ‘Path(‘/blah’, app=asgi_application(my_view))’

lovelydinosaur:
In other words: I think it’d make sense to start by adding CBVs, with our existing routing. I like decorator routing, but still not sure how to best apply it.

jordaneremieff:
Thanks for clarifying. I'll close this PR now and see about introducing CBVs first with the existing routing then approach the decorator routing later.

PR #24, opened 2018-07-13
Proxy

lovelydinosaur:


akx:
For people possibly finding this for inspiration for a Starlette-based proxy, please note that it's missing a fair few important things such as

* query-string passing
* `allow_redirects=False` in the proxy request call (so cookies set in redirect responses get passed through)
* ... and possibly other things.

**EDIT:** https://github.com/valohai/asgiproxy is an ASGI toolkit that can be used in conjunction with Starlette.

PR #28, opened 2018-07-17
Fixed encoding issues on windows

gbozee:
Attempting to install the package on windows throws 
```UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 1549: character maps to <undefined>```

This is an attempted fix


lovelydinosaur:
Thanks!
Released as 0.1.13.
Want to give it a try?

Issue #27, opened 2018-07-16
Class Based Views


jordaneremieff:
This relates to discussions in https://github.com/encode/starlette/pull/17 - figured an issue better than that PR for working this out.

lovelydinosaur:
Closing this, as we've got 'em now. Tho Class Based Websocket endpoints is still desirable.

PR #35, opened 2018-07-30
Implement CBV pattern

jordaneremieff:
Refs #27 

This is an attempt to define a CBV pattern in Starlette.

Example usage:

```python
from starlette import PlainTextResponse
from starlette.app import App
from starlette.views import View


app = App()


class HomepageView(View):
    async def get(self, request, **kwargs):
        response = PlainTextResponse(f"Hello, world!")
        return response


class UserView(View):
    async def get(self, request, **kwargs):
        username = kwargs.get("username")
        response = PlainTextResponse(f"Hello, {username}")
        return response


app.add_route("/", HomepageView())
app.add_route("/user/{username}", UserView())
```


starlette/decorators.py
```diff
@@ -1,10 +1,9 @@
 import asyncio
 from starlette.request import Request
-from starlette.response import Response
 from starlette.types import ASGIInstance, Receive, Send, Scope
 
 
-def asgi_application(func):
+def make_asgi(func):
```
lovelydinosaur: I'm kinda thinking that we'll drop the `asgi_application`/`make_asgi` decorator altogether - I think it just muddies things up.

Let's leave this alone for now, but perhaps don't use it in the PR, and include any logic directly in the class.

starlette/views.py
```diff
@@ -0,0 +1,25 @@
...
+            raise Exception(
+                f"Method {request_method} is not implemented for this view."
+            )
+        return make_asgi(func)(scope)
+
+    @classmethod
+    def as_view(cls) -> ASGIApp:
+        def view(scope: Scope):
+            self = cls()
+            return self.dispatch(scope)
```
lovelydinosaur: Do we need to do the same closure `.as_view()` dance that Django does?

Rather than using `.as_view()` I figure that we just implement the endpoint/view/whatever class as an ASGI application. We're not using multi-threading so I don't think we need the same "put stuff in a closure for safety" dance, but maybe I'm not thinking it through enough.

jordaneremieff:
I made some changes based on your comments - removed the `as_view()` and `make_asgi` methods from the view class. Still a bit more to do here when I have time, but those changes were simple enough to push for now.

lovelydinosaur (commented):
I’d also suggest we drop support for the handler methods being non-awaitable. Let’s just cover async stuff in Starlette.

jordaneremieff:
I've made the following updates to this PR:

- View class now only supports async methods
- Returns '404 Not found' if an allowed method is unimplemented on the view class rather than raising exception
- Intended to be used with App() class `add_route` method, previous`add_route` method on Router removed
- Documentation with brief description and example code added
- Tests added

lovelydinosaur:
Coolio!

One observation: We’ll want to make sure that each request gets a new instance, rather than routing to an instance of the class.

Also, use 406 for the method-does-not-exist-on-this-class case.

Good stuff!

lovelydinosaur:
Added some further commits on top of this, in #52. Thanks so much for this!

delijati:
hi @tomchristie can this be also added to the docs ... i found only this here

ghost:
@delijati that was the case before the doc changed  https://github.com/encode/starlette/compare/0.12.0...master

lovelydinosaur:
@foxmask @delijati https://www.starlette.io/endpoints/ (Or are we talking about something else?)

ghost:
@tomchristie if that were endpoints here 
https://github.com/encode/starlette/compare/0.12.0...master#diff-1a523bd9fa0dbf998008b37579210e12 
then yes ;)

**edit** the pasted URL does not work fine so ... I wanted to show you : 

https://github.com/encode/starlette/compare/0.12.0...master 

then > "file changed" then we can see the docs/index.md

lovelydinosaur:
We're not documenting a pure ASGI class, nope. We don't really need to in ASGI 3, since it's easier to write them as plain functions.

What we *are* documenting is the class-based `HTTPEndpoint` and `WebSocketEndpoint` implementations, which is *actually* what you want if you're looking for an CBV implementation.

PR #40, opened 2018-08-27
Add a TOC to the repo README

jeffbuttars:


lovelydinosaur:
Thanks!

lovelydinosaur:
I think we're probably at the point that we should move away from a single README and start using a proper doc pages.

I'd like us to use the same setup as uvicorn's mkdocs.

Happy to either take PRs for this, or else get onto it myself shortly.

jeffbuttars:
Yeah, it's at the point, which is great!

PR #43, opened 2018-08-28
Add http and websocket status codes with descriptive variables

jeffbuttars:
* updated websockets and response tests with status module usage


lovelydinosaur:
Yup, I like the style of this.

My thoughts here are:

1. Should we try to make sure that http status codes here match what's supported in `http.HTTPStatus`?
2. Do we need to be concerned that we're replicating `http.HTTPStatus`?

I guess my initial answers would be:

1. Yeah but let's start by accepting this as-is.
2. No, the usage of the `http.HTTPStatus` enum in source code doesn't include both the status code and reason phrase being visible, which they are here, so let's prefer this.

Last thing is that there's a conflict to resolve.

jeffbuttars:
I agree with both of your thoughts and conclusions. I would in principal prefer to use the `http` module, but I've found using the `enum`s for status codes to be, cumbersome I'd say. In fact, the first time I coded up websocket status codes I used `enum`  like in `http`, but disliked using it. I remembered liking how it was done in DRF, so I've gone with that :)

tests/test_websockets.py
```diff
@@ -148,39 +149,39 @@ def app(scope):
         return asgi
 
     client = TestClient(app)
-    with client.websocket_connect("/") as session:
-        session.close(code=1001)
-    assert close_code == 1001
+    with client.wsconnect("/") as session:
```
lovelydinosaur: This is out of date with master.

lovelydinosaur (commented):
One minor thing that's out of date.

lovelydinosaur:
Seems like we probably ought to go with this. Any last comments anyone?

marcosschroh:
Awesome!! I like the same styling as `DRF`!

auvipy (commented):
Looks good to get in.

lovelydinosaur:
Thanks @jeffbuttars!

PR #44, opened 2018-08-28
Initial conversion of README.md to mkdocs

jeffbuttars:
* This is a straight conversion of the README.md to mkdocs
* This PR does _not_ alter README.md

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -1,0 +1,43 @@
+Starlette also includes an `App` class that nicely ties together all of
+its other functionality.
+
+```python
+from starlette.app import App
+from starlette.response import PlainTextResponse
+from starlette.staticfiles import StaticFiles
+
+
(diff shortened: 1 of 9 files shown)


lovelydinosaur:
@jeffbuttars Brilliant! Going to take a look at this in a bit.
Note that the websockets section is very slightly out of date in this PR, given some pushes I've made recently today. Also the `Applications` section needs adding. (I'm very happy to take on either/both of those two if that helps.)

jeffbuttars:
@tomchristie I saw applications go in, I just added it!

lovelydinosaur:
Ace, pulled these in now.

There's some issue with the DNS conflicting with the org github page, so I'll need to sort out a domain name before this'll be up and running. We'll also want to strip down the README once that's all done.

Issue #51, opened 2018-08-29
Version 1.0


lovelydinosaur:
## Introduce `App` first in README / Docs intro.

* Promote the `app = App()` style, but then go on to demonstrate plain ASGI style example, and discuss design philosophy of "everything is just ASGI" , interoperable, eg. use `TestClient` with Channels.

## Naming

* `WebSocketSession` as just `WebSocket`? Rename `session` variable as `websocket` in docs?
* Drop `StaticFile`, and just leave `StaticFiles` and `FileResponse`.
* Module naming - should it be `starlette.requests`/`starlette.responses` instead of `starlette.request`/`starlette.response`? Would fit `websockets` and align better with documentation titles.
* `App` should perhaps be `Starlette` instead. Less likely to be confused with plain ASGI `App` class examples, and mirrors eg. `flask.Flask` and `sanic.Sanic`.
* Class based views instead as `HTTPEndpoint` and `WebSocketEndpoint`?

- Zero-dependencies

Ensure that `requests` is only required if `TestClient` is used. Ensure that `aiofiles` is only required if `FileResponse`/`StaticFiles` is used. We *could* choose to push `FileResponse` into the `starlette.staticfiles` module, and only ever import `aiofiles` from there.

## Features

See https://github.com/encode/starlette/milestone/1

* URL Reversing
* Cookies
* Form support
* Class based views
* Exception handling
* Middleware
* StartUp / Shutdown events
* Background tasks

## Later

* Server Sent Events
* HTTP/2 (and server push)

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/51">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/51/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/51/pledge.svg">
</picture>
</a>

lovelydinosaur:
We have everything that I'd originally wanted for 1.0 support.

The *only* bit of API that's giving me any pause on a 1.0 jump is the async bits of requests/websockets. eg. request parsing. Ideally we'd like to eventually be able to use
fully sync views too, and still have all functionality available to us. Haven't yet figured
[...]

PR #52, opened 2018-08-30
Class based views

lovelydinosaur:
Taken from #35, with a little bit of extra work on top. Huge thanks to @erm for breaking the back of this one!

Views are now instantiated per-request. Use them either as regular ASGI apps, or route to them with `@app.route(...)`.

We'll also want to add an equivalent for websockets that looks like...

```python
class App(...):
    async def on_connect(self, websocket, **kwargs):
        ...
    async def on_receive(self, websocket, data):
        ...
    async def on_disconnect(self, websocket):
        ...
```

I'm not totally sure about the naming. We could go for `HTTPEndpoint`/`WebSocketEndpoint` instead of "views".


jeffbuttars:
I like the idea of using '*Endpoint' over 'views'. That's how I talk about API implementations in IRL.

lovelydinosaur:
@jeffbuttars Indeed. I've included it in the list of potential renames in https://github.com/encode/starlette/issues/51

PR #54, opened 2018-08-31
Exception handling

lovelydinosaur:
Adds an ExceptionHandler middleware, that allows you to delegate classes of exceptions to given handlers.

- [x] Add into `App`, with functions for adding new handlers.
- [x] Ordering of handlers so that subclasses are ordered by specificity 
- [x] Use `HTTPException` in `View` and `Router`.
- [ ] Smarter behavior for exceptions on websockets. (Perhaps `request` should be `connection` instead?)
- [x] Consider interaction with `DebugMiddleware`.
- [x] Consider interaction when response has already started.
- [x] Document `TestClient(app, raise_exceptions=False)` (and possible also add a 500 fallback when no response is sent).
- [x] Support both async and regular functions.

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -29,6 +29,10 @@ async def websocket_endpoint(session):
     await session.close()
 ```
 
+### Instantiating the application
+
+* `App(debug=False)` - Create a new Starlette application.
+
 ### Adding routes to the application
 
 You can use any of the following to add handled routes to the application:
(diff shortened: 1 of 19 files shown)


lovelydinosaur:
* DebugMiddleware needs to run *inside* in order to function. (It needs to catch the error and respond *before* the exception middleware can do so.)
* View and Router will now raise `HTTPException` if run within an `App`, or return a response otherwise. Allows for customized exc handling with `App`, but still allows them to be used as stand-alone ASGI apps otherwise.

PR #55, opened 2018-08-31
Name changes

jordaneremieff:
Refs https://github.com/encode/starlette/issues/51

Renames `View` class -> `HTTPEndpoint` and `App` class -> `Starlette`. 

There were some other name change suggestions on the ticket as well, but these two seemed like the most obvious ones to start with. I can add the other name changes to this PR as well, but figured it would be better to open it up to discussion first.

> Module naming - should it be starlette.requests/starlette.responses instead of starlette.request/starlette.response? Would fit websockets and align better with documentation titles.

I think this makes sense.

> WebSocketSession as just WebSocket? Rename session variable as websocket in docs?
> Drop StaticFile, and just leave StaticFiles and FileResponse.

These I'm not as sure about - haven't looked at them enough to have an opinion either way at this point.


docs/views.md
```diff
@@ -1,44 +1,44 @@
 
-Starlette includes a `View` class that provides a class-based view pattern which
```
lovelydinosaur, line 2: Great. Best rename the docs file too.

lovelydinosaur (commented):
Great stuff.

Let's rename the `views.md` file, and also bring in the other changes discussed at the same time. `WebSocketSession` -> `WebSocket`, `session` -> `websocket` when it is used as a variable. `requests.py`, `responses.py`

We'll want to make all these in a single version bump.

lovelydinosaur:
Something else I'd also like to see us do as part of this, would be to *always* use full module imports in the source and tests, and drop imports from `__init__`.

lovelydinosaur:
Okay, let's go for it.

PR #56, opened 2018-08-31
WebSocket : normalize errors when sending on a closed websocket and receive_json with bytes or text

jeffbuttars:
* Catch exception and re-raise as WebSocketDisconnect when trying to
  send on a WebSocket that has been closed, typically by the client.
* receive_json can parse bytes or text received since a client can send
  JSON as either.

In a pub/sub scenario where a client consumes a stream of data from a websocket until the client disconnects the websocket publisher will(should) receive an exception from the underlying websocket  implementation. We need to expose the error/disconnect in a uniform way regardless of the underlying
ASGI server.


lovelydinosaur:
There's two different parts, so let's discuss those separately.

* receive_json can parse bytes or text received since a client can send
JSON as either.

Okay, we *could* do that. I think it's more correct to send/receive as binary, since JSON is a binary encoding that's defined to always be utf-8 encoded. I like the existing symmetry of receive/send both being expected to be over binary, but it's possible that we could just switch one, and make the receive case more lenient.

* Catch exception and re-raise as WebSocketDisconnect when trying to
send on a WebSocket that has been closed, typically by the client.

It's not obvious that we should treat that as an exception case. In the HTTP case, an ASGI server that's had the client close early will just allow the server to send the remaining part of the message without raising any error (it's valid for the client to close early, and it's not an error as far as the server should be concerned). I *think* that we probably want the same for websockets (don't error if the client has closed but just allow the server to send any remaining messages into the void, and then complete whatever else normally eg. receiving and handling the disconnect message.)

(*) Tho server implementations will typically want to deal with enforced timeout behavior once the client has disconnected, and kill the task forcibly for apps that haven't responded to a disconnect within a reasonable time-frame.

jeffbuttars:
On the websocket scenario I have a particular case in mind that I run into which is an infinite, as infinite as we get, amount of data coming from a websocket service that does not wait to receive. A one way data stream. The problem being we don't want it to send infinitely into a void because it doesn't ever receive after it's started it's send stream so it won't know it's disconnected unless it's interrupted. In this case, with uvicorn and likely others, an Exception will be thrown. The "infinite" stream server will need to know that and stop sending.

Here's a real scenario I use websockets for:

-----------------                                                                   ------------------                                                 ------------
| Sensor(s) | == endless one way data stream => |API service| == websocket stream ==> |client|
-----------------                                                                   ------------------                                                 ------------
The stream data is time sensitive and needs to be as close to real time as possible

The server side stays open and sends data for as long as the client is connected.

I hope that helps explain the issue I was going after with the exception on send stuff.

* On the JSON, looking back I'll side with you on that and revert it. I think I over thought that and jumped the gun there. If that's ever an issue it will show itself.

jeffbuttars:
I intend to close this soon, after I test out a better solution from @tomchristie and post it back here for proper closure.

lovelydinosaur:
Sure thing! There is some stuff that we *could* do here, eg. raise errors against incorrect state transitions (eg. sending a message without having accepted, or sending any kind of message after sending a close) but I guess treat that separately.

I'd expect that uvicorn and perhaps also hypercorn and daphne might need a bit of refinement around "don't raise exceptions on sending to a closed connection" since I think we've only really *just* thrashed out that it's probably how the ASGI spec ought to behave there.

jeffbuttars:
Using the pattern from @tomchristie I augmented an existing 'producer' test I have to this:

NOTE: this is with python3.6
```python
class Producer(WebSocketEndpoint):
    async def on_connect(self, websocket):
        await websocket.accept()
        self.ticker_task = loop.create_task(self.produce(websocket))

    async def on_disconnect(self, websocket, close_code):
        self.ticker_task.cancel()

    async def produce(self, websocket):
        while True:
            await websocket.send_text('%s' % random.randint(0, 1000000))
            await asyncio.sleep(0.001)
```

This closes gracefully. Much nicer! And can be easily abstracted into Producer centric class, a great start.

jeffbuttars:
@tomchristie I'd rather use a separate PR for the other error condition handling changes and get a cleaner PR for those.
I'll open issue for it: #66 

> "I'd expect that uvicorn and perhaps also hypercorn and daphne might need a bit of 
> refinement around "don't raise exceptions on sending to a closed connection" since
> I think we've only really just thrashed out that it's probably how the ASGI spec 
> ought to behave there."

I couldn't agree more. Having a consistent, ASGI way, of handling that condition is needed.

I'm really happy to have good solution to this for now. Thank you @tomchristie for your help on this.

jeffbuttars:
Closing, with one last note.

In response to progress here and more directly to the issue https://github.com/django/asgiref/issues/66 :
_my response to https://github.com/django/asgiref/issues/66 is very Starlette specific, so I'm putting it here_

After having used two methods to deal with the "websocket sends on a closed connection" scenario:
1. Wait for 'an exception' from the underlying ASGI server
2. Wait for a disconnect message on the receive from server

I very much prefer the second scenario. This allows for much cleaner and explicit code. You don't have to catch `Exception` because you don't know what may be raised. And your not using an error mechanism to drive valid behavior, as @tomchristie has noted.  

I'm very much in line with the last two points on server behavior. The second point, how to handle a closed websocket, is something I've dealt with and thought about a lot lately. I like the proposed behavior. Especially "ensure that disconnect is sent on the "receive" channel, and **should forcibly kill the task if it has not completed after a reasonable timeout**". I was concerned how to handle a misbehaving (ignoring disconnect) websocket endpoint and I think this does so in a good way.

Following recent discussions on this and development of `WebSocketEndpoint`, I've been testing this:

_using python 3.6_
```python
class WebSocketPublishEndpoint(WebSocketEndpoint):
    async def on_connect(self, websocket, *args, **kwargs):
        await super().on_connect(websocket, *args, **kwargs)
        self.ticker_task = loop.create_task(self.produce(websocket, *args, **kwargs))

    async def produce(self, websocket):
        raise NotImplementedError()

    async def on_disconnect(self, websocket, close_code):
        self.ticker_task.cancel()
```

I then implement a subclass and override `produce` and it's nice:

```python
class Timer(WebSocketPublishEndpoint):
    async def produce(self, websocket):
        while True:
            await websocket.send_text(datetime.datetime.now().isoformat())
            await asyncio.sleep(1.0)
```


I guess this is just a wordy way of saying "I agree with your conclusion on https://github.com/django/asgiref/issues/66 @tomchristie ".
This is a corner of ASGI I use everyday and I think you have a good direction here.

lovelydinosaur:
Great, thanks for the real-usage feedback!

PR #65, opened 2018-09-22
Add support for background tasks

### LINKED ISSUE #59: Background tasks
* Add a `background=...` keyword argument to all responses. It should be either `None` or a coroutine with no arguments.
* The response `__call__` should run the background coroutine if it exists (after sending the response).
* Provide a `BackgroundTask()` class that wraps a callable and some arguments. This can then be used to pass a function *or* coroutine with args/kwargs to a response.

```python
class BackgroundTask:
    def __init__(self, func, *args, **kwargs):
        self.func = func
        self.args = args
        self.kwargs = kwargs

    async def __call__(self):
        if asyncio.iscoroutine(self.func):
            await self.func(*self.args, **self.kwargs)
        else:
            self.func(*self.args, **self.kwargs)
```

Then we can do...

```
return Response(..., background=BackgroundTask(my_func, a=123)
```

Naming might need some thinking about - not sure that conflicting with asyncio's `Task` is ideal.

This'd be a good feature to get in, since it's a low amount of work, but provides functionality that WSGI is unable to do.

alexbotello:
Closes #59


starlette/responses.py
```diff
@@ -96,6 +99,9 @@ def headers(self):
         )
         await send({"type": "http.response.body", "body": self.body})
 
+        if isinstance(self.background, BackgroundTask):
```
lovelydinosaur: I think just `if self.background is not None`

starlette/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @property
+    def id(self):
+        if not hasattr(self, "_id"):
+            self._id = str(uuid.uuid4())
+        return self._id
+
+    @property
+    def result(self):
+        return getattr(self, "_result", None)
```
lovelydinosaur: I'm not convinced we should have an interface for `result` or `id`. Those seem superfluous.

starlette/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+
+    async def __call__(self):
+        if asyncio.iscoroutinefunction(self.func):
+            self._result = await asyncio.ensure_future(
+                self.func(*self.args, **self.kwargs)
+            )
+        else:
+            fn = functools.partial(self.func, *self.args, **self.kwargs)
+            loop = asyncio.get_event_loop()
+            self._result = await loop.run_in_executor(None, fn)
```
lovelydinosaur: Bit of a question here - we may, or may not want to do this.

For synchronous routes we don't currently run them in an executor (we just assume that you're only going to do some non-blocking stuff if you're not using `async`.) running in an executor here isn't consistent with *that* set of behavior, but it's possible that we want to switch things around anyways.

One option for now would be to *force* background tasks to only ever use `async` functions. If users want to run something in a thread pool then they can, explicitly, but we keep consistency better consistency with how we handle routes. (Not 100% either way on this tho)

lovelydinosaur (changes requested):

starlette/tasks.py
alexbotello: My thinking here was to support if a client needed access to the status or result of a task directly. Once a task is initiated the task.id could be appended to `content` and sent along with the response. 
So then a client could access like so..

`@app.route('/status/{task_id}')`

Might be too niche though

lovelydinosaur: That's a reasonable pattern, yup, I just don't think it's absolute enough that we should include it ourselves. Leave that part up to developers (or higher-level frameworks).

starlette/tasks.py
alexbotello: I think the option should exist either way. The docs could warn users that any sync tasks would be running in a thread pool. This allows more flexibility, especially if you need to use any third party code (most often synchronous) inside your background task.

I'd be interested to hear other opinions though. For now, would you just like me to force the `async` option only?

lovelydinosaur: I'm not sure, either we:

* Leave it as you have it now.
* Force it to `async` only, and at some point add threadpools for *both* background tasks, *and* for routes/websockets, making sure to support them everywhere, rather than in just one particular place.

starlette/tasks.py
```diff
@@ -0,0 +1,30 @@
+import uuid
```
lovelydinosaur: We *might?* want to call it `starlette/background.py` to conflict a little less with asyncio tasks?

starlette/tasks.py
alexbotello: I got you. I guess I'm just unsure how a future developer would access the return value of the task if we aren't setting it to a property on the class.

lovelydinosaur: I suppose generate the uuid and pass it to the background task as an argument.

lovelydinosaur: That we you can include it in the response content, and the background task has it available for updating any state it needs to.

lovelydinosaur (approved):

alexbotello:
@tomchristie Is this a better documentation example? Thoughts? Changes?

lovelydinosaur:
Nice one! 👍

PR #67, opened 2018-09-26
Missing annotations added on responses module.

marcosschroh:
command to run: `mypy starlette/responses.py --disallow-untyped-defs --ignore-missing-imports --follow-imports=silent`


starlette/responses.py
```diff
@@ -19,9 +19,9 @@
 try:
     import ujson as json
 
-    JSON_DUMPS_OPTIONS = {"ensure_ascii": False}
+    JSON_DUMPS_OPTIONS: typing.Dict[str, typing.Any] = {"ensure_ascii": False}
 except ImportError:  # pragma: nocover
-    import json
+    import json  # type: ignore
```
lovelydinosaur: If mypy is forcing us to add this kind of noise then I'm :-/

starlette/responses.py
```diff
@@ -19,9 +19,9 @@
 try:
     import ujson as json
 
-    JSON_DUMPS_OPTIONS = {"ensure_ascii": False}
+    JSON_DUMPS_OPTIONS: typing.Dict[str, typing.Any] = {"ensure_ascii": False}
```
lovelydinosaur: Can we avoid this?

starlette/responses.py
marcosschroh: Yes, me too. Seems that is the only solution. You can take a look [here](https://github.com/python/mypy/issues/1153)

starlette/responses.py
marcosschroh: We need to define the type. I can do this if you want:

```python
JSON_DUMPS_OPTIONS = {"ensure_ascii": False}  # type: typing.Dict[str, typing.Any]
```

lovelydinosaur:
Good stuff. This'd need updating to the latest master particularly since #99 made some changes here.

marcosschroh:
Hi,

Now is updated with `master`!!

PR #69, opened 2018-09-26
Missing annotations on websockets module.

marcosschroh:
Command to run: `mypy starlette/websockets.py --disallow-untyped-defs --ignore-missing-imports --follow-imports=silent`

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -1,9 +1,12 @@
-from collections.abc import Mapping
-from enum import Enum
-from starlette.datastructures import URL, Headers, QueryParams
-from urllib.parse import unquote
 import enum
 import json
+import typing
+
+from collections.abc import Mapping
+from urllib.parse import unquote
+
+from starlette.datastructures import URL, Headers, QueryParams
+from starlette.types import Scope, Receive, Send, Message
 
 
 class WebSocketState(enum.Enum):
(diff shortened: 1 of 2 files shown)


starlette/websockets.py
```diff
@@ -110,52 +113,54 @@ def query_params(self) -> QueryParams:
         else:
             raise RuntimeError('Cannot call "send" once a close message has been sent.')
 
-    async def accept(self, subprotocol=None):
+    async def accept(
+        self, subprotocol: typing.Optional[typing.List[str]] = None
```
lovelydinosaur: Let's use implied optionals `typing.List[str]= None`

lovelydinosaur:
Cool - there's only one very minor comment on this. If we can use implied optionals in that case, then I think this is good to go.

starlette/websockets.py
```diff
@@ -13,26 +16,26 @@ class WebSocketState(enum.Enum):
...
 
 class WebSocketDisconnect(Exception):
-    def __init__(self, code=1000):
+    def __init__(self, code: int = 1000) -> None:
         self.code = code
 
 
 class WebSocket(Mapping):
-    def __init__(self, scope, receive=None, send=None):
+    def __init__(self, scope: Scope, receive: Receive, send: Send) -> None:
```
lovelydinosaur, line 24: Is it possible to leave `send` and `receive` as optionals for now, or does that introduce errors?

marcosschroh: If we allow optional like before we have ` error: "None" not callable`. Anyway, does it make sense that `receive` and `send` can be optional? If for some reason any of them is `None` the server will crash.

PR #70, opened 2018-09-26
Missing annotations added on staticfiles module

marcosschroh:
Command: `mypy starlette/staticfiles.py --disallow-untyped-defs --ignore-missing-imports --follow-imports=silent`

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -86,9 +86,9 @@ class _StaticFilesResponder:
             response = PlainTextResponse("Not Found", status_code=404)  # type: Response
         else:
             mode = stat_result.st_mode
-
-            response = FileResponse(self.path, stat_result=stat_result)
             if not stat.S_ISREG(mode):
                 response = PlainTextResponse("Not Found", status_code=404)
+            else:
+                response = FileResponse(self.path, stat_result=stat_result)
 
         await response(receive, send)
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -1,20 +1,17 @@
 import os
 import stat
-import typing
 
 from aiofiles.os import stat as aio_stat
 
-from starlette.responses import PlainTextResponse, FileResponse
-from starlette.types import Send, Receive, Scope
+from starlette.responses import PlainTextResponse, FileResponse, Response
+from starlette.types import Send, Receive, Scope, ASGIInstance
 
 
 class StaticFile:
     def __init__(self, *, path: str) -> None:
         self.path = path
 
-    def __call__(
-        self, scope: Scope
-    ) -> typing.Union[PlainTextResponse, "_StaticFileResponder"]:
+    def __call__(self, scope: Scope) -> ASGIInstance:
         if scope["method"] not in ("GET", "HEAD"):
             return PlainTextResponse("Method Not Allowed", status_code=405)
         return _StaticFileResponder(scope, path=self.path)
@@ -25,9 +22,7 @@ class StaticFiles:
         self.directory = directory
         self.config_checked = False
 
-    def __call__(
-        self, scope: Scope
-    ) -> typing.Union[PlainTextResponse, "_StaticFilesResponder"]:
+    def __call__(self, scope: Scope) -> ASGIInstance:
         if scope["method"] not in ("GET", "HEAD"):
             return PlainTextResponse("Method Not Allowed", status_code=405)
         path = os.path.normpath(os.path.join(*scope["path"].split("/")))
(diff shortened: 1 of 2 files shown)


starlette/staticfiles.py
```diff
@@ -1,25 +1,33 @@
...
 
 class StaticFile:
-    def __init__(self, *, path):
+    def __init__(self, *, path: str) -> None:
         self.path = path
 
-    def __call__(self, scope):
+    def __call__(
+        self, scope: Scope
+    ) -> typing.Union[PlainTextResponse, "_StaticFileResponder"]:
```
lovelydinosaur: We ought to be able to use `-> ASGIInstance` here

starlette/staticfiles.py
```diff
@@ -1,25 +1,33 @@
...
 class StaticFiles:
-    def __init__(self, *, directory):
+    def __init__(self, *, directory: str) -> None:
         self.directory = directory
         self.config_checked = False
 
-    def __call__(self, scope):
+    def __call__(
+        self, scope: Scope
+    ) -> typing.Union[PlainTextResponse, "_StaticFilesResponder"]:
```
lovelydinosaur: Again, using `-> ASGIInstance` here would be appropriate.

starlette/staticfiles.py
```diff
@@ -54,12 +62,14 @@ def __init__(self, scope, path):
 
 
 class _StaticFilesResponder:
-    def __init__(self, scope, path, check_directory=None):
+    def __init__(
+        self, scope: Scope, path: str, check_directory: typing.Optional[str] = None
```
lovelydinosaur: Let's just use implied optionals, which I belive mypy supports. `check_directory: str = None`

starlette/staticfiles.py
```diff
@@ -73,19 +83,21 @@ def __init__(self, scope, path, check_directory=None):
...
         if self.check_directory is not None:
             await self.check_directory_configured_correctly()
 
         try:
             stat_result = await aio_stat(self.path)
         except FileNotFoundError:
-            response = PlainTextResponse("Not Found", status_code=404)
+            response = PlainTextResponse(
+                "Not Found", status_code=404
+            )  # type: typing.Union[PlainTextResponse, FileResponse]
```
lovelydinosaur: Can we avoid having these noisy comments? Is there an option so that mypy will allow us to not error on missing variable types?

lovelydinosaur: I get why it's there, but I don't think they add much personally. If we *do* need it then `type: Response` would be preferable.

marcosschroh: Yes, totally right. Should be `type: Response`. It does not add anything, but if we don't add it `Incompatible types in assignment` error is raised. I can not find any option to allow it, and I think make sense because we would be breaking the `Liskov principle`, I guess.

starlette/staticfiles.py
```diff
@@ -73,19 +76,19 @@ def __init__(self, scope, path, check_directory=None):
...
-    async def __call__(self, receive, send):
+    async def __call__(self, receive: Receive, send: Send) -> None:
         if self.check_directory is not None:
             await self.check_directory_configured_correctly()
 
         try:
             stat_result = await aio_stat(self.path)
         except FileNotFoundError:
-            response = PlainTextResponse("Not Found", status_code=404)
+            response = PlainTextResponse("Not Found", status_code=404)  # type: Response
```
lovelydinosaur, line 86: I think this is the only thing I’d like to change. Ideally I think we should ensure the our function definitions are entirely typed, but I’m not convinced I like the variable typing, as it’s quite noisy.

Either we could just remove the variable typing, or else we could add a `response = None  # Type: Response` to the top of the method in cases like this?

lovelydinosaur (commented):
This is looking great! Just one more query on this. Happy to take you feedback on which way we should go there, but I don’t especially like the asymmetry and noise that the variable annotation currently adds in.

starlette/staticfiles.py
marcosschroh: I agree. I am not a big fan of it as well. We can not do `response = None # type: Response` because we will have an error, but we can do `response = None # type:  typing.Any`

Another solution could be rethinking a little bit this block of code and declare `response = PlainTextResponse("Not Found", status_code=404)  # type: Response` since this is should be de default value.

starlette/staticfiles.py
```diff
@@ -73,19 +76,19 @@ def __init__(self, scope, path, check_directory=None):
...
         else:
             mode = stat_result.st_mode
+
+            response = FileResponse(self.path, stat_result=stat_result)
```
lovelydinosaur: I'd prefer that we don't move this logic around. It's better to keep this in the `else` block and only instantiate one or the other case.

marcosschroh: Perfect!

PR #72, opened 2018-09-26
Missing annotations added on endpoints module.

marcosschroh:
Command: `mypy starlette/endpoints.py --disallow-untyped-defs --ignore-missing-imports --follow-imports=silent`


lovelydinosaur:
Great, thanks!

Issue #11, opened 2018-06-26
Cookies


lovelydinosaur:
Add `request.cookies` and `response.cookies`.

Not much work to do, since https://docs.python.org/3/library/http.cookies.html supports everything we'd need.

PR #75, opened 2018-09-29
Add cookies support.

songww:
#11

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -1,6 +1,7 @@
 from starlette.datastructures import URL, Headers, QueryParams
 from collections.abc import Mapping
 from urllib.parse import unquote
+from http.cookies import SimpleCookie, CookieError
 import json
 import typing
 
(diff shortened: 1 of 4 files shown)


lovelydinosaur:
Good work both!

Things to watch out for:

* `cookies` header not in request.
* Multiple cookies headers.
* `request.cookies` should prob be a key: value dict, as @erm’s implementation.

starlette/requests.py
```diff
@@ -25,6 +26,22 @@ def __iter__(self):
     def __len__(self):
         return len(self._scope)
 
+    @property
+    def cookies(self) -> SimpleCookie:
```
lovelydinosaur, line 30: I'd like us to return `typing.Dict[str, str]` instead of `SimpleCookie`

lovelydinosaur:
@erm Very welcome to submit a PR from your branch too - not sure which one of these two gets us over the line first, but happy to have 'em both and then figure which to work from. (Or eg. pull in the case from one, and implementation from another)

lovelydinosaur:
Good work on this! #86 cleans up a few extra bits, so have merged that one in instead. Thanks so much!

Issue #34, opened 2018-07-29
CORS support


jordaneremieff:
Related to https://github.com/encode/starlette/issues/1

I'm having trouble figuring out an appropriate pattern for including CORS support. My use-case is that I am connecting a React app to an ASGI backend that uses Starlette's routing and response classes, but enabling CORS support would require altering the response class headers which cannot be accomplished simply by wrapping the ASGI app at this point.

What should CORS support look like in Starlette?

An example of my use-case can be found [here](https://github.com/erm/apistar-react/blob/master/app.py), it uses APIStar and a CORS extension to achieve what I want to do with Starlette as well. Below is my work-in-progress attempt at an ASGI CORS middleware that I would ultimately like to work into Starlette somehow.

```python3
from starlette import PlainTextResponse
from starlette.datastructures import Headers

ACCESS_CONTROL_ALLOW_ORIGIN = b"Access-Control-Allow-Origin"
ACCESS_CONTROL_EXPOSE_HEADERS = b"Access-Control-Expose-Headers"
ACCESS_CONTROL_ALLOW_CREDENTIALS = b"Access-Control-Allow-Credentials"
ACCESS_CONTROL_ALLOW_HEADERS = b"Access-Control-Allow-Headers"
ACCESS_CONTROL_ALLOW_METHODS = b"Access-Control-Allow-Methods"
ACCESS_CONTROL_MAX_AGE = b"Access-Control-Max-Age"
ACCESS_CONTROL_REQUEST_METHOD = b"Access-Control-Request-Method"
ACCESS_CONTROL_REQUEST_HEADERS = b"Access-Control-Request-Headers"


DEFAULT_OPTIONS = {
    "allow_origin": [b"*"],
    "allow_credentials": False,
    "allow_headers": [b"*"],
    "allow_methods": [
        b"GET",
        b"HEAD",
        b"POST",
        b"OPTIONS",
        b"PUT",
        b"PATCH",
        b"DELETE",
    ],
    "expose_headers": [b""],
    "max_age": b"86400",
}


class CORSMiddleware:
    def __init__(self, app, options=None):
        self.app = app
        self.options = options or DEFAULT_OPTIONS

    def __call__(self, scope):
        headers = Headers(scope.get("headers", []))
        origin = headers.get("origin", None)
        if origin is None:
            return self.app(scope)

        allow_origin = self.options["allow_origin"]
        if origin not in allow_origin and b"*" not in allow_origin:
[...]

PR #76, opened 2018-09-29
Add support for CORS

alexbotello:
This implements the ideas tossed around in #34.

Wasn't sure if we should have a `middlewares.py` to place all future middleware in one place. But for the time being, I've placed this in `cors.py`

Basic usage:
```python
class App:
    def __init__(self, scope):
        self.scope = CORSMiddleware(scope)

    async def __call__(self, receive, send):
        response = Response("Hello, world!", media_type='text/plain')
        await response(receive, send)
```


lovelydinosaur:
Cool, thanks for this! Used it to help the work on #83, which is now released... https://www.starlette.io/middleware/

Note that theres a few behavioral differences in the implementation (eg. cases where we need to mirror back headers that the request included, plus I've made the CORS preflight responses include a bit of informational detail.)

PR #79, opened 2018-10-01
Add pipfile

tombasche:
`pip install pipenv`
Run `pipenv install`


lovelydinosaur:
Thanks but going to pass on this. Not really getting us anything beyond the existing requirements.txt.

PR #80, opened 2018-10-03
Missing annotations testclient

marcosschroh:
Command: `mypy starlette/testclient.py --disallow-untyped-defs --ignore-missing-imports --follow-imports=silent`


starlette/testclient.py
```diff
@@ -197,49 +204,49 @@ def _run(self):
...
-    def send(self, message):
+    def send(self, message: Message) -> None:
         self._receive_queue.put(message)
 
-    def send_text(self, data):
+    def send_text(self, data: typing.Any) -> None:
         self.send({"type": "websocket.receive", "text": data})
 
-    def send_bytes(self, data):
+    def send_bytes(self, data: typing.Any) -> None:
```
lovelydinosaur: Let's use `data: bytes` here

starlette/testclient.py
```diff
@@ -197,49 +204,49 @@ def _run(self):
...
+    def _raise_on_close(self, message: Message) -> None:
         if message["type"] == "websocket.close":
             raise WebSocketDisconnect(message.get("code", 1000))
 
-    def send(self, message):
+    def send(self, message: Message) -> None:
         self._receive_queue.put(message)
 
-    def send_text(self, data):
+    def send_text(self, data: typing.Any) -> None:
```
lovelydinosaur: Let's use `data: str` here.

starlette/testclient.py
```diff
@@ -162,31 +169,31 @@ def send(self, request, *args, **kwargs):
...
 class WebSocketTestSession:
-    def __init__(self, app, scope):
+    def __init__(self, app: ASGIApp, scope: Scope) -> None:
         self.accepted_subprotocol = None
         self._loop = asyncio.new_event_loop()
         self._instance = app(scope)
-        self._receive_queue = queue.Queue()
-        self._send_queue = queue.Queue()
+        self._receive_queue = queue.Queue()  # type: queue.Queue
+        self._send_queue = queue.Queue()  # type: queue.Queue
```
lovelydinosaur, line 206: Are these two needed? They're surprising.

starlette/testclient.py
```diff
@@ -58,11 +65,11 @@ def send(self, request, *args, **kwargs):
 
         # Include the 'host' header.
         if "host" in request.headers:
-            headers = []
+            headers = []  # type: list
         elif port == 80:
             headers = [[b"host", host.encode()]]
         else:
-            headers = [[b"host", ("%s:%d" % (host, port)).encode()]]
+            headers = [[b"host", ("%s:%d" % (host, port)).encode()]]  # type: ignore
```
lovelydinosaur: How come we need `type: ignore` here? Also should it be a list of two-tuples instead?

starlette/types.py
```diff
@@ -9,3 +9,9 @@
 
 ASGIInstance = typing.Callable[[Receive, Send], typing.Awaitable[None]]
 ASGIApp = typing.Callable[[Scope], ASGIInstance]
+
+StrDict = typing.Mapping[str, str]
+BytesPairs = typing.List[typing.Tuple[bytes, bytes]]
+
+Methods = typing.Sequence[str]
+Subprotocols = typing.Sequence[str]
```
lovelydinosaur: Let's avoid these extra indirections, and just use the full type annotations directly in these 4 cases.

starlette/testclient.py
```diff
@@ -259,13 +266,15 @@ def __init__(
...
+        self, method: str, url: str, **kwargs: typing.Any
+    ) -> requests.Response:
         url = urljoin(self.base_url, url)
         return super().request(method, url, **kwargs)
 
     def websocket_connect(
-        self, url: str, subprotocols=None, **kwargs
-    ) -> WebSocketTestSession:
+        self, url: str, subprotocols: Subprotocols = None, **kwargs: typing.Any
+    ) -> typing.Any:
```
lovelydinosaur, line 339: I'll have a think about how we can change this around so that we've got the correct `WebSocketTestSession` annotation. Leaving this comment as a reminder to myself.

starlette/testclient.py
```diff
@@ -58,11 +65,11 @@ def send(self, request, *args, **kwargs):
 
         # Include the 'host' header.
         if "host" in request.headers:
-            headers = []
+            headers = []  # type: list
```
lovelydinosaur: Can we use something like `typing.Sequence[typing.Tuple[bytes, bytes]]` here?

starlette/testclient.py
```diff
@@ -22,34 +23,40 @@ class _MockOriginalResponse(object):
...
+    def send(  # type: ignore
+        self, request: requests.PreparedRequest, *args: typing.Any, **kwargs: typing.Any
+    ) -> requests.Response:
+        scheme, netloc, path, params, query, fragement = urlparse(  # type: ignore
+            request.url
+        )
+
         if ":" in netloc:
+            port: typing.Any
             host, port = netloc.split(":", 1)
```
lovelydinosaur: Let's use `host, port_string` here, then `port = int(port_string)` in order to avoid the `Any` annotation.

lovelydinosaur (commented):
Good stuff, thanks! A few comments to look into here.

starlette/testclient.py
songww: Is the `return` was missing, at line 289?

lovelydinosaur: No, it's just that the codepath *always* follows the except case.

starlette/testclient.py
```diff
@@ -197,49 +206,49 @@ def _run(self):
...
 
-    def receive(self):
+    def receive(self) -> Message:
         message = self._send_queue.get()
         if isinstance(message, BaseException):
             raise message
         return message
 
-    def receive_text(self):
+    def receive_text(self) -> typing.Any:
```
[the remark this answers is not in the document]
lovelydinosaur: You may need to use `return typing.cast(message["text"], str)` here.

starlette/testclient.py
```diff
@@ -197,49 +206,49 @@ def _run(self):
...
         message = self.receive()
         self._raise_on_close(message)
         return message["text"]
 
-    def receive_bytes(self):
+    def receive_bytes(self) -> typing.Any:
```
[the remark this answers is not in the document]
lovelydinosaur: You may need to use `return typing.cast(message["text"], bytes)` here.

starlette/testclient.py
```diff
@@ -22,58 +26,63 @@ class _MockOriginalResponse(object):
...
 class _ASGIAdapter(requests.adapters.HTTPAdapter):
-    def __init__(self, app: typing.Callable, raise_server_exceptions=True) -> None:
+    def __init__(self, app: ASGIApp, raise_server_exceptions: bool = True) -> None:
         self.app = app
         self.raise_server_exceptions = raise_server_exceptions
 
-    def send(self, request, *args, **kwargs):
-        scheme, netloc, path, params, query, fragement = urlparse(request.url)
+    def send(  # type: ignore
+        self, request: requests.PreparedRequest, *args: typing.Any, **kwargs: typing.Any
```
lovelydinosaur, line 68: Interesting that this is marked `type: ignore` - are we not adhering to the same interface as `requests.adapters.HTTPAdapter` here?

starlette/testclient.py
```diff
@@ -22,58 +26,63 @@ class _MockOriginalResponse(object):
...
+    ) -> requests.Response:
+        scheme, netloc, path, params, query, fragement = urlparse(  # type: ignore
+            request.url
```
lovelydinosaur, line 71: Why are we having to ignore this?

starlette/testclient.py
```diff
@@ -259,13 +268,15 @@ def __init__(
         self.headers.update({"user-agent": "testclient"})
         self.base_url = base_url
 
-    def request(self, method: str, url: str, **kwargs) -> requests.Response:
+    def request(  # type: ignore
+        self, method: str, url: str, **kwargs: typing.Any
```
lovelydinosaur: I guess we *could* drop `**kwargs` and instead follow the complete `requests` signature here, as per https://github.com/requests/requests/blob/master/requests/sessions.py#L457

Though it might end up being a bit of work to figure all the correct annotations to use? Any thoughts?

lovelydinosaur (changes requested):
This one is getting really close. 👍

lovelydinosaur:
Looks like it’s just `send(...)` that needs resolving on this one now.

marcosschroh:
I am having some troubles with `send(...)`. I know that the signature should be something like:

```python
from urllib3.util import Timeout as TimeoutSauce

def send(self, request: requests.PreparedRequest, stream: bool = False,
                timeout: typing.Union[typing.Tuple, TimeoutSauce, float] = None,
                verify: typing.Union[bool, str] = True, cert: typing.Any = None,
                proxies: typing.Dict[str, str] = None) -> requests.Response: ...
```

But the thing is we are only using the `request` argument. Also, if I complete the signature is valid for `BaseAdapter ` but not for `HTTPAdapter` :-(

lovelydinosaur:
Finalizing this in #115, and turning on `--disallow-untyped-defs` 🎉

lovelydinosaur:
Thanks for all your work on getting us to this point! 😄

marcosschroh:
You are welcome! I continue on #115 if you want

marcosschroh:
You are welcome!!

PR #86, opened 2018-10-05
Cookies

jordaneremieff:
Refs https://github.com/encode/starlette/issues/11 & https://github.com/encode/starlette/pull/75


starlette/requests.py
```diff
@@ -50,6 +50,21 @@ def query_params(self) -> QueryParams:
...
+            cookie_header = self.headers.get("cookie")
+            if cookie_header:
+                cookie = http.cookies.SimpleCookie()
+                cookie.load(cookie_header)
+                for key, morsel in cookie.items():
+                    self._cookies[key] = morsel.value
+        return self._cookies
+
+    def get_cookie(self, key: str) -> typing.Union[str, None]:
+        return self.cookies.get(key, None)
```
lovelydinosaur: Not sure we really need this method, since we have `request.cookies.get()` anyways.

lovelydinosaur (changes requested):
Looks spot-on. Just one method that I reckon we can do without.

lovelydinosaur:
At some point in the future we might also want to look into making it an immutable dict interface, but let’s treat that separately if we do.

starlette/requests.py
lovelydinosaur: Ok, I’d suggest drop it *for now* since the request/response cookies interfaces aren’t at all symmetrical in any case.

At a later point we should then add a smart `response.cookies` dict-like interface, and start to bring the two in-line, except for immutable interfaces on the request.

lovelydinosaur:
Great, let’s get this in! Welcome to issue a docs PR, or else I’ll get into it next week!

PR #88, opened 2018-10-06
Fix runtime error within CORSMiddleware

### LINKED ISSUE #87: CORSMiddleware is sending an extra 'http.response.body'
It seems that even with all tests passing and cors being successfully applied, CORSMiddleware still raises a runtime error.

Code being tested:
```python
app = Starlette()

app.add_middleware(CORSMiddleware, allow_origins=["*"])

@app.route("/")
async def homepage(request):
    return PlainTextResponse('Hello', status_code=200)

if __name__ == "__main__":
```

And the error being produced:
```
ERROR: Exception in ASGI application
  File "/home/alexbotello/.local/share/virtualenvs/starlette-dshJy1CJ/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 384, in run_asgi
    result = await asgi(self.receive, self.send)
  File "/home/alexbotello/Code/starlette/starlette/exceptions.py", line 60, in app
    raise exc from None
  File "/home/alexbotello/Code/starlette/starlette/exceptions.py", line 52, in app
    await instance(receive, sender)
  File "/home/alexbotello/Code/starlette/starlette/middleware/cors.py", line 116, in simple_response
    await inner(receive, send)
  File "/home/alexbotello/Code/starlette/starlette/applications.py", line 26, in awaitable
    await response(receive, send)
  File "/home/alexbotello/Code/starlette/starlette/responses.py", line 100, in __call__
    await send({"type": "http.response.body", "body": self.body})
  File "/home/alexbotello/Code/starlette/starlette/middleware/cors.py", line 130, in send
    await send(message)
  File "/home/alexbotello/Code/starlette/starlette/exceptions.py", line 47, in sender
    await send(message)
  File "/home/alexbotello/.local/share/virtualenvs/starlette-dshJy1CJ/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 518, in send
    raise RuntimeError(msg % message_type)
RuntimeError: Unexpected ASGI message 'http.response.body' sent, after response already completed.
```
It seems the issue is originating from `send`. Specifically:
```python
if message["type"] != "http.response.start":
    await send(message)
```
Removing this fixes the issue and does not break any tests.

alexbotello:
Fixes #87

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -116,9 +116,6 @@ class CORSMiddleware:
         await inner(receive, send)
 
     async def send(self, message, send=None, origin=None):
-        if message["type"] != "http.response.start":
-            await send(message)
-
         message.setdefault("headers", [])
         headers = MutableHeaders(message["headers"])
 
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -117,7 +117,8 @@ class CORSMiddleware:
 
     async def send(self, message, send=None, origin=None):
         if message["type"] != "http.response.start":
-            return await send(message)
+            await send(message)
+            return
 
         message.setdefault("headers", [])
         headers = MutableHeaders(message["headers"])
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -116,6 +116,9 @@ class CORSMiddleware:
         await inner(receive, send)
 
     async def send(self, message, send=None, origin=None):
+        if message["type"] != "http.response.start":
+            return await send(message)
+
         message.setdefault("headers", [])
         headers = MutableHeaders(message["headers"])
 


lovelydinosaur (changes requested):
Correct resolution to this will be to make sure we `return` after the `await send(message)`.

Checking the response body in all our CORS test cases would pick this error up.

The TestClient could also be catching this for us.

lovelydinosaur (changes requested):
That’ll work but it’s misleading as send always just returns `None`

Would be better as...

```
await send(message)
return
```

lovelydinosaur:
> Could you clarify how you would check in the tests for this? I was unsuccessful in intentionally breaking the tests.

Sure. TestClient isn’t yet checking that valid message types are sent in each state. Let’s treat that as a separate issue tho.

lovelydinosaur:
Ace, thanks. Will roll a release as soon as poss.

lovelydinosaur:
Released as 0.3.5

PR #92, opened 2018-10-08
Support `allow_origin_regex` in CORSMiddleware

### LINKED ISSUE #84: Add `allow_origin_regex` to CORSMiddleware.
It'd be helpful if `CORSMiddleware` supported an `allow_origin_regex`, so that users could do...

```python
# Enforce a subdomain CORS policy
app.add_middleware(CORSMiddleware, allow_origin_regex="(http|https)://*.example.com")
```

Or...

```python
# Enforce an HTTPS-only CORS policy.
app.add_middleware(CORSMiddleware, allow_origin_regex="https://*")
```

The string should be compiled to a regex by the middleware and matches should be anchored to the start/end of the origin string.

alexbotello:
Closes #84
- Adds support for optional keyword argument `allow_origin_regex`

- Adds `is_allowed_origin` method to either verify regex match or check in `self.allow_origins`

- Adds test `test_cors_allow_origin_regex`

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -3,6 +3,7 @@ from starlette.responses import PlainTextResponse
 from starlette.types import ASGIApp, ASGIInstance, Scope
 import functools
 import typing
+import re
 
 
 ALL_METHODS = ("DELETE", "GET", "OPTIONS", "PATCH", "POST", "PUT")
@@ -16,12 +17,17 @@ class CORSMiddleware:
         allow_methods: typing.Sequence[str] = ("GET",),
         allow_headers: typing.Sequence[str] = (),
         allow_credentials: bool = False,
+        allow_origin_regex: str = None,
         expose_headers: typing.Sequence[str] = (),
         max_age: int = 600,
     ):
 
         if "*" in allow_methods:
             allow_methods = ALL_METHODS
+
+        if allow_origin_regex is not None:
+            regex = re.compile(allow_origin_regex)
+            allow_origin_regex = regex
 
         simple_headers = {}
         if "*" in allow_origins:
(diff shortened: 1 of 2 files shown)


starlette/middleware/cors.py
```diff
@@ -16,13 +17,18 @@ def __init__(
...
         expose_headers: typing.Sequence[str] = (),
         max_age: int = 600,
     ):
 
         if "*" in allow_methods:
             allow_methods = ALL_METHODS
 
+        if allow_origin_regex is not None:
+            regex = re.compile(allow_origin_regex)
+            allow_origin_regex = regex
```
lovelydinosaur, line 30: We're changing the type of this variable at this point, so I'd expect `mypy` will complain about this. I'll have a think about just tweaking how we set that.

starlette/middleware/cors.py
```diff
@@ -66,12 +73,22 @@ def __call__(self, scope: Scope):
...
                     return self.preflight_response(request_headers=headers)
                 else:
-                    return functools.partial(
-                        self.simple_response, scope=scope, origin=origin
-                    )
+                    if self.is_allowed_origin(origin=origin):
+                        return functools.partial(
+                            self.simple_response, scope=scope, origin=origin
+                        )
+                    return PlainTextResponse("Disallowed CORS origin", status_code=400)
```
lovelydinosaur, line 80: An interesting point to make here: We *don't* currently disallow the requests on the server side, as CORS is enforced by the browser rather than by the server.

We *could* choose to implement server-side enforcement *too*, but it'd involve a little more work than *just* checking the origin, since we'd also want to eg. check if allowed headers have been sent, check if allow_credentials is being respected etc.etc.

I'm not against us doing that, or doing that as an optional point of configuration, but let's treat it as a separate issue, since there's more to consider than just origin checks.

lovelydinosaur (approved):
Great stuff! I'm going to pull this in as it is, and then make two very minor tweaks, as per the comments.

PR #99, opened 2018-10-09
Add explicit UJSONResponse

lovelydinosaur:
Needs docs updating.
Would need README.md and index.md notes on micro-performance nonsense.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -5,6 +5,7 @@ from starlette.datastructures import MutableHeaders, URL
 from starlette.types import Receive, Send
 from urllib.parse import quote_plus
 import hashlib
+import json
 import os
 import typing
 import http.cookies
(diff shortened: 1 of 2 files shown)


Midnighter:
I was wondering about this one, what about a general response class that you can instantiate with just the `dumps` equivalent of your package of choice (or a partial with the preferred arguments set). Or set `dumps` as a class variable.

lovelydinosaur:
@Midnighter I'd kinda considered that yup, tho it gets a bit more icky in that we'd quite like different arguments in the UJSON case. Happy to consider alternatives, but it's not obvious to me that there's better answers than simply subclassing JSONResponse with whatever `render` method you want.

Midnighter:
That's why a partial might be a good solution.

lovelydinosaur:
Let me put it this way around instead: I’m not convinced that `return JSONResponse(..., dumps=ujson.dumps)` is better than `return UJSONResponse(...)`

(Or equivalently for alternative subclasses)

Midnighter:
Fair enough :smiley: what about:

```python
from starlette import JSONResponse

JSONResponse.dumps = partial(...)
```
Although, as you say the classes are very thin anyway and it's actually very explicit then what you return. My only suggestion is then to move `UJSONResponse` (and each sibling like it) into its own module that is not automatically imported. A developer would then get it `from starlette.ujson_response import UJSONResponse`, for example, and if `ujson` is not available the resulting `ImportError` is for the developer to handle.

alexbotello:
I don’t think it’s use case warrants it’s own module. In fact, I don’t really think it’s neccessary to have more than two subclasses.

lovelydinosaur:
Published as 0.4.0 (Version bump because JSONResponse is stricter than UJSONResponse for some edge-cases)

PR #101, opened 2018-10-10
Added init file so autodetect will pick up middleware

bendog:
Middleware was not being included in the build for PyPI
adding the __init__.py file fixed this.


lovelydinosaur:
Thanks!

PR #105, opened 2018-10-11
Respond to credentialed requests with specific origin

### LINKED ISSUE #95: Credentialed CORS standard requests should not respond with wildcard origins
See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Credentialed_requests_and_wildcards 

If a standard request is made, that includes any cookie headers, then CORSMiddleware *ought* to strictly respond with the requested origin, rather than a wildcard.

This is actually potentially a bit fiddly since we maybe also need to make sure to *set or add* Vary: Origin in those cases, in order to ensure correct cacheability.

alexbotello:
Closes #95


lovelydinosaur:
Nice one!

PR #116, opened 2018-10-17
Add documentation for cookies

### LINKED ISSUE #89: Document cookies support
The request and response docs should document cookie usage. 🍪

alexbotello:
Closes #89 :cookie: !

### DIFF
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -55,6 +55,16 @@ Headers are exposed as an immutable multi-dict.
 
 For example: `request.query_params['abc']`
 
+#### Cookies
+
+Cookies are exposed as an immutable dict.
+
+**Note**: Exception `KeyError` will be raised if nonexistant key is  accessed as `request.cookies['key']`.
+Instead cookies should be accessed with the `.get` method.
+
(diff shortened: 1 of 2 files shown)


docs/requests.md
```diff
@@ -55,6 +55,16 @@ Headers are exposed as an immutable multi-dict.
 
 For example: `request.query_params['abc']`
 
+#### Cookies
+
+Cookies are exposed as an immutable dict.
```
lovelydinosaur, line 60: It's not actually immutable at the moment, tho it'd be nice if it was.
I'll merge this as-is and then update this.

docs/requests.md
```diff
@@ -55,6 +55,16 @@ Headers are exposed as an immutable multi-dict.
...
+
+**Note**: Exception `KeyError` will be raised if nonexistant key is  accessed as `request.cookies['key']`.
+Instead cookies should be accessed with the `.get` method.
```
lovelydinosaur, line 63: I don't think we need to say this - it's just a regular dict interface, so that's expected behavior.

lovelydinosaur:
Nice one!

PR #120, opened 2018-10-17
Add WSGI middleware

lovelydinosaur:
Taken from Uvicorn's existing implementation, with a little light refinement.

We could potentially use this to all `app.mount("/example", some_wsgi_app)`, although I haven't implemented that in this current pass.


blaggacao:
Yeah! `... groovy jingle ...`

PR #121, opened 2018-10-17
Create MANIFEST.in

synapticarbors:
The BSD license requires that a copy be included in the source distribution. This file should ensure that:

https://docs.python.org/3.7/distutils/sourcedist.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE.md


lovelydinosaur:
Any thoughts on how we could do this without introducing another file into the repo?

Can we include it in the `setup.py` somehow instead?

carlodri:
you can use `data_files` like [this](https://stackoverflow.com/a/9977956/4511441)

tiangolo:
If changing or adding setup / dist files and configurations, I would suggest switching from `setup.py` and all the sibling files to the new (PEP official) `pyproject.toml`, and a tool like [Flit](https://flit.readthedocs.io/en/latest/index.html) or [Poetry](https://poetry.eustace.io/).

Given that Starlette doesn't have any C extensions and is only for Python 3.6+, it would make sense and simplify things. And all these configurations would live in the same file.

BTW, Flit includes stuff from git automatically, so, the license is included by default.

And the publishing process is simplified to:

* Bump the version in `starlette/__init__.py`
* `flit publish`

---

I have done some packages based on `setuptools` before. I felt the process was a bit confusing and cumbersome.

Like:

* Having to install and use twine for publishing
* Removing `__pycache__` dirs by hand
* Reading the `long_description` from the `README.md` in code
* Parsing the version inside of the `setup.py`
* etc.

Flit has worked surprisingly well up to now (I'm using it in [FastAPI](https://github.com/tiangolo/fastapi)).

---

More info about the `pyproject.toml` vs `setup.py`: https://snarky.ca/clarifying-pep-518/

lovelydinosaur:
Okay, I'd accept a pull request which added `data_files` to the existing `setup.py`.

Don't particularly want to change to pyproject.toml, since setup.py is the lowest common demoninator, but it's not impossible I'd consider it at a later date. (But in any case let's treat that as a seperate issue.)

PR #124, opened 2018-10-17
Support a function-based interface for adding events on LifespanHandler

alexbotello:
For example:
```python
app = Starlette()

async def open_database_connections():
    ...
async def close_database_connections():
    ...

# Use as a function interface
app.add_event_handler("startup", open_database_connections)
app.add_event_handler("cleanup", close_database_connections)
```


lovelydinosaur:
Perfect! A docs update when you get a chance would be great, too!
The [events docs](https://www.starlette.io/events/) ought to mention it after the decorator syntax example.
We're also missing a brief "Adding event handlers" section in the [application docs](https://www.starlette.io/applications/).

PR #135, opened 2018-10-20
Add support for empty `request.json()`.

olivertso:
The app throws a `json.decoder.JSONDecodeError` if I try to access `request.json()` when the request has an empty body.

Hope this is helpful :)

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -103,6 +103,8 @@ class Request(Mapping):
     async def json(self) -> typing.Any:
         if not hasattr(self, "_json"):
             body = await self.body()
+            if not body:
+                body = b"{}"
             self._json = json.loads(body)
         return self._json
 
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -160,6 +160,10 @@ def test_request_json():
         return asgi
 
     client = TestClient(app)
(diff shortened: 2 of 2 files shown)


lovelydinosaur:
Thanks! It's not obvious that this is the behavior that we want.
I'm inclined to think that we should error with 400 if we expected JSON content, but there is an empty body.

starlette/requests.py
```diff
@@ -103,6 +103,8 @@ def cookies(self) -> typing.Dict[str, str]:
     async def json(self) -> typing.Any:
         if not hasattr(self, "_json"):
             body = await self.body()
+            if not body:
+                body = b"{}"
```
alexbotello, line 107: You could probably do something like:
```suggestion
                if "app" in self._scope:
                    raise HTTPException(status_code=400)
                return PlainTextResponse("Bad Request", status_code=400)
```

alexbotello:
@olivertso you still up for making this change? I’d actually really like to get this behavior in. If not, i’ll go ahead and submit a PR

olivertso:
@alexbotello Sorry for the delay. I'm still up but I'm really busy lately. Please submit your PR. Thanks.

Issue #139, opened 2018-10-22
MultiDict __iter__ returns tuples instead of just keys


ryankask:
The `QueryParams` and `Headers` multidict classes' `__iter__` method returns key/value pairs instead of just keys which is what I would expect from a Python mapping.

I haven't seen an implementation like this and checked a few from Django, Werkzeug, urllib3, and aio-lib's multidict.

Was this intentional or should it be changed?

Thanks!

lovelydinosaur:
Should probably be changed if it doesn’t match standard dict behaviour.
It’d be interesting to know what eg. werkzueg does here, since keys() and values() aren’t super meaningful in a multidict context. set(d.keys()) is pretty much the only context I can see it as being meaningful.

ryankask:
I've posted a few examples below. The behaviour of these libraries differs but in both cases, `list(instance) == list(instance.keys())`.

[Werkzeug's `MultiDict`](http://werkzeug.pocoo.org/docs/0.14/datastructures/#werkzeug.datastructures.MultiDict) returns the first value seen:

```python
In [68]: d = MultiDict([('a', 1), ('b', 2), ('a', 3)])

In [69]: list(d.keys())
Out[69]: ['a', 'b']

In [70]: list(d.values())
Out[70]: [1, 2]

In [71]: list(d)
Out[71]: ['a', 'b']
```

[aio-lib's `MultiDict`](https://multidict.readthedocs.io/en/stable/multidict.html#multidict.MultiDict.keys) returns all possible values:

```python
In [2]: d = MultiDict([('a', 1), ('b', 2), ('a', 3)])

In [3]: list(d.keys())
Out[3]: ['a', 'b', 'a']

In [4]: list(d.values())
Out[4]: [1, 2, 3]

In [5]: list(d)
Out[5]: ['a', 'b', 'a']
```

Starlette:

```python
In [10]: params = QueryParams(items=[('a', 1), ('b', 2), ('a', 3)])

In [11]: params
Out[11]: QueryParams(query_string='a=1&b=2&a=3')

In [12]: list(params)
Out[12]: [('a', 1), ('b', 2), ('a', 3)]

In [13]: list(params.keys())
Out[13]: ['a', 'b', 'a']

In [14]: list(params) == list(params.keys())
Out[14]: False
```

I'm happy to help if this is something you'd like to change.

ryankask:
[...]

PR #142, opened 2018-10-23
Return an iterator over keys for Headers and QueryParams

ryankask:
The mapping data structures (`QueryParams` and `Headers`) now return an iterable over keys instead of items.

This addresses the first part of the solution mentioned in #139.


lovelydinosaur:
Ace, thanks!

PR #143, opened 2018-10-24
Support a routing interface for `Router`

### LINKED ISSUE #125: Routing interfaces for `Router` and `Starlette` should match.
We should probably hide away `Path` and `Include` as implementation details for now, and instead focus on making the app and router interfaces match up.

Eg:

```python
app = Router()

@app.route('/')
async def homepage(...):
    ...
```

or...

```python
app = Router()

app.mount('/admin', AdminApp)
app.mount('/articles', BlogArticles)
```

This'll mean that a `Router` instance strictly gives you a subset of the functionality that a `Starlette` instance offers.

Refs #122

alexbotello:
Closes #125 

`Router` class now has functionality for:
* `add_route`
* `add_graphql_route`
* `add_websocket_route`
* `mount`
* `route` decorator
* `websocket_route` decorator

Essentially, all function logic now resides inside `Router` and `Starlette`'s interface simply wraps what the `Router` provides.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,53 +1,9 @@
-import asyncio
-import inspect
 import typing
-from concurrent.futures import ThreadPoolExecutor
 
 from starlette.exceptions import ExceptionMiddleware
 from starlette.lifespan import LifespanHandler
-from starlette.graphql import GraphQLApp
-from starlette.requests import Request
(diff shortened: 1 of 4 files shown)


lovelydinosaur:
Amazing. Yes please!

PR #145, opened 2018-10-26
Drop support for `StaticFile`

### LINKED ISSUE #50: Drop `StaticFile` app.
We have `FileResponse` and `StaticFiles`.

I think that including the `StaticFile` ASGI app complicates things unnecessarily, and that we should probably remove it.

* Drop `StaticFile` app.
* Put runtime checks that file exists, and file is a regular file in `FileResponse`.

alexbotello:
Closes #50 
* `FileResponse` now contains `StaticFile`'s runtime checks
* Remove `StaticFile` and `_StaticFileResponsder`

### DIFF
--- a/docs/staticfiles.md
+++ b/docs/staticfiles.md
@@ -1,20 +1,17 @@
 
-As well as the `FileResponse` class, Starlette also includes ASGI applications
-for serving a specific file or directory:
+Starlette also includes an `StaticFiles` class for serving a specific directory:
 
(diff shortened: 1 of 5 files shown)


lovelydinosaur (changes requested):
Nice one! Just a docs update too, then this is good to go!

lovelydinosaur:
Perfect. I’ll release a 0.6 tomorrow.

PR #147, opened 2018-10-27
Only check the host in TrustedHostMiddleware

taoufik07:
Now we ignore the port and only check against the host


taoufik07:
Part of PR #148 now

PR #148, opened 2018-10-27
Wildcard domains and TrustedhostMiddleware

taoufik07:
Support for wildcard domains


taoufik07:
@tomchristie Tests passed but the check fails

Issue #138, opened 2018-10-21
Consolidating HTTP status codes


berislavlopac:
I have noticed that there is `starlette.status`, which groups all status codes in one place. However, there are several places in the code where it's using a number directly instead of the `starlette.status`, e.g. when [returning errors](https://github.com/encode/starlette/blob/49f76ab5e9ef0ce5d966485553ad6019a9d37da5/starlette/endpoints.py#L40).

That said, I'm a big fan of the standard library's nice little `http.HTTPStatus` enum, which neatly groups all the status codes and their standard messages and descriptions, while at the same time keeps the integer functionality (truth be told, HTTP status codes are not _really_ integers; we just treat them as such as it's easy and clean).

So I'd like to propose doing a few minor refactorings in the Starlette code; I can easily make a PR myself, but I'd like to hear any comments on the idea first, in particular if there is any concrete reason for not using `http.HTTPStatus`. My ideas are:

1. Replace all integer values in `starlette.status` with the equivalent `HTTPStatus` value; or even, possibly, completely replace them in the manner that `starlette.status.HTTP is http.HTTPStatus` (this needs applying the change everywhere, but it shouldn't be a problem if its just internal).
2. Replace all the integers used for status code with the corresponding `startlette.status` values.
3. Create a `HTTPStatus`-like enum for all the codes it does not cover -- this especially goes for the WebSocket Close Codes -- and apply them in the same manner as above.

Optionally, I would like to suggest a new `PlainTextResponse` subclass specifically for errors, so in the case [linked above](https://github.com/encode/starlette/blob/49f76ab5e9ef0ce5d966485553ad6019a9d37da5/starlette/endpoints.py#L40) it would be something like:

    return HTTPErrorResponse(status.HTTP_405_METHOD_NOT_ALLOWED)

or, if the second approach in no. 1 above is used:

    return HTTPErrorResponse(status.HTTP.METHOD_NOT_ALLOWED)

My main problem here is that it can be used for any type of response and not just errors; the solutions I can think of are:

1. Trust the user and let them use it as they see fit (I would prefer this).
2. Not trust the user and explicitly check if the status is >= 400; I'm not sure what is the general policy for Starlette in such cases.

OK, that's it from me for now, looking forward to any thoughts and comments.

alexbotello:
As far as I am aware the decision not to use `http.HTTPStatus` was a conscious one. See #43 for the discussion on that.

[...]

PR #152, opened 2018-10-29
Plaintext response with status

berislavlopac:
As discussed in #138 I have replaced explicit integer status codes with the `status` consonants, and added the option for `PlainTextResponse` to automatically add content if just the status code is passed (specifically, if the content passed is an integer that exists in `http.HTTPStatus`.

### DIFF
--- a/starlette/debug.py
+++ b/starlette/debug.py
@@ -2,6 +2,7 @@ import html
 import traceback
 import typing
 
+from starlette import status
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, PlainTextResponse, Response
 from starlette.types import Scope, Receive, Send, Message, ASGIApp, ASGIInstance
@@ -15,9 +16,9 @@ def get_debug_response(request: Request, exc: Exception) -> Response:
         content = (
             f"<html><body><h1>500 Server Error</h1><pre>{exc_html}</pre></body></html>"
         )
-        return HTMLResponse(content, status_code=500)
+        return HTMLResponse(content, status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
     content = "".join(traceback.format_tb(exc.__traceback__))
-    return PlainTextResponse(content, status_code=500)
+    return PlainTextResponse(content, status_code=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
 class DebugMiddleware:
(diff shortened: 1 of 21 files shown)


lovelydinosaur:
> and added the option for PlainTextResponse to automatically add content if just the status code is passed

I think we should possibly treat these two separately.
Definitely want `PlainTextResponse(status_code=404)` to set the content automatically.
Not sure if we want to replace integer status codes everywhere in the codebase.

lovelydinosaur:
*Maybe* we'll take 'em both in one, just not sure yet.

lovelydinosaur:
@berislavlopac We *could* subclass `__init__` (eg. see RedirectResponse) but 🤷‍♂️

berislavlopac:
Of course; I was concerned about the `PlainTextResponse()` use case. But since `200` is the default for `status_code` anyway, I suppose it would just return `OK`. Good point.

alexbotello:
What would you think about overriding the `render` method of `PlainTextResponse` instead of the `__init__`? This way class signatures can remain consistent as well as being cleaner I feel.

Maybe something like:
```python
    if isintance(content, http.HTTPStatus):
        self.status_code = content
        content = http.HTTPStatus(content).phrase
    elif isinstance(content, bytes):
        return content
    return content.encode(self.charset)
```
So that `PlainTextResponse(404) -> "Not Found"` 

Thoughts?

berislavlopac:
@alexbotello Possibly, but I might have approached this the wrong way in the first place; please check my [latest comment](https://github.com/encode/starlette/issues/138#issuecomment-433870739) on the original issue.

alexbotello:
@berislavlopac I think the approach is fine. `HTTPException` is for use with "app" in scope and with the ExceptionMiddlware. But Starlette is also modular so it makes sense that we should have this manual functionality as well in case when a user wants to use `PlainTextResponse` independently.  If a user is performing validation  or etc. and what’s to manually send back a status message, this is a good use case.

@tomchristie feel free to correct me if I am off base

lovelydinosaur:
Starting to use `isort` has added a lot of simple conflicts here.
I think best to treat this as 2 PRs in any case - not fully decided on each part yet.

berislavlopac:
@tomchristie No problem, I can separate them. What is your opinion on `__init__` vs `render`?

lovelydinosaur:
@berislavlopac Not sure.

lovelydinosaur:
Let's treat these two issues separately. Fairly sure we want the PlainTextResponse behavior. Less obvious if we want to use verbose status codes throughout.

PR #161, opened 2018-10-30
Session property asserts that SessionMiddleware is installed

alexbotello:
This will prevent a server crash in the event that `session` is accessed without `SessionMiddlware` being initialized.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -79,7 +79,7 @@ class Request(Mapping):
 
     @property
     def session(self) -> dict:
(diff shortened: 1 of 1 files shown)


lovelydinosaur (commented):
I think we should probably raise an error instead. “SessionMiddleware must be installed to access request.session”

alexbotello:
I went ahead and used an `assert` statement to handle the error

lovelydinosaur:
Thanks again!

PR #162, opened 2018-10-30
Fixes to static files documentation.

johanzietsman-em:
I fixed a grammar error. The example code did not work, since there are no Path and PathPrefix classes in starlette.routing. I replaced PathPrefix with Mount to get it to work.


lovelydinosaur:
Thank you!

PR #168, opened 2018-11-01
Fix the return value of `url_path_for` 

alexbotello:
Current `Starlette.url_path_for` returns the `URL` instance instead of `path` string. This example:
```python
@app.route("/")
async def homepage(request):
    return JSONResponse({"path": app.url_path_for("func", param="model")})
```
returns `TypeError: Object of type URL is not JSON serializable`.

`url_path_for` will now just return the `path` string

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -73,7 +73,7 @@ class Starlette:
 
         return decorator
 
-    def url_path_for(self, name: str, **path_params: str) -> URL:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Actually not as simple as this if we want consistent `url_path_for` signatures at the router and app levels. We were returning `URL` instances because we need to also attach protocol information (eg. this was a websocket match) I've changed it around in #169 so that `url_path_for` now returns a str subclass `URLPath`.

PR #173, opened 2018-11-01
Finesse mount routing

### LINKED ISSUE #160: [question] how to mount router into Starlette app? `app.mount('/', router)` does't work
Example:

```python
from starlette.routing import Router, Route
import uvicorn

from views import FooView


app = Starlette(debug=True)
# need middleware, so I use Starlette app
app.add_middleware(
    CORSMiddleware, allow_origins=["*"], allow_headers=["*"], allow_methods=["*"]
)
router = Router([
    Route('/foo', endpoint=FooView),
])
app.mount('/', router)  # << doesn't work for some reason

if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=4000)
```

lovelydinosaur:
Closes #160.

* Mounted routes will now only match if the submount is properly directory separated.
* Adds append slash redirect behaviors when a directory URL would have matched.
* Support `url_path_for('mounted', path=...)`, eg. for use with static files.
* Support `url_path_for('mount_name:sub_name', **params)`

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -8,7 +8,7 @@ from enum import Enum
 from starlette.datastructures import URL, URLPath
 from starlette.exceptions import HTTPException
 from starlette.requests import Request
-from starlette.responses import PlainTextResponse
+from starlette.responses import PlainTextResponse, RedirectResponse
 from starlette.types import ASGIApp, ASGIInstance, Receive, Scope, Send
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
It'd be nice to also ensure that `url_path_for('mounted', path=...)` works as desired.

PR #179, opened 2018-11-03
Improve output of DebugMiddleware

marcosschroh:
First attempt to improve `DebugMiddleware` output.

I guess we should create a `debug` folder and put all the `css` and `js`.

screen shot 2018-11-03 at 15 17 40

The idea is to show `locals` and add some basic `js` to collapse some blocks.

### DIFF
--- a/starlette/debug.py
+++ b/starlette/debug.py
@@ -1,22 +1,76 @@
-import html
 import traceback
-import typing
 
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, PlainTextResponse, Response
 from starlette.types import ASGIApp, ASGIInstance, Message, Receive, Scope, Send
 
+STYLES = """\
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Loving it!

lovelydinosaur:
Awesome 🎉

lovelydinosaur:
Really happy with how that's looking now, nice one @marcosschroh!

Sentry gives us a good guideline for how we could improve things even further...

screen shot 2018-11-07 at 10 06 41

* Reverse the order of the traceback by default.
* Collapsible traceback items, with the top one open by default.
* 5 lines of code context on either side.

We'll also want to include frame locals, but I'd suggest we treat that as a later iteration.

lovelydinosaur:
Related: In debug mode I think we should also handle 404s as a special case, and return a similar page that shows what routes you have configured.

PR #183, opened 2018-11-06
Add `@app.middleware(...)` decorator

lovelydinosaur:
Adds a decorator syntax, which uses `BaseHTTPMiddleware` under the hood:

```python
@app.middleware("http")
async def add_custom_header(request, call_next):
    response.headers['Custom'] = 'Example'
    return response
```

The response instance here will always be a StreamingResponse (We've captured it from the ASGI send channel, rather than getting the raw instance returned by the view. This means we can use any other standard ASGI middleware between this and the view, and everything will continue to work just fine.)

At a later date I think we might want to also support:

```python
@app.middleware("request")
async def add_custom_header(request, call_next) -> None:
    ...  # Response is not captured, call_next just runs the rest of the stack and returns None.
```

```python
@app.middleware("response")
async def add_custom_header(request, response) -> Response:
    ...  # Next in stack has already been called, we're passed `request`/`response`
```

We could also consider "connection" (websocket + http) and/or "websocket" middlewares.
Any new cases there would probably also involve an equivalent base class


jhallard:
@tomchristie was support ever added for "connection" or "websocket"?

PR #186, opened 2018-11-07
Implement session expiry with configurable maximum age

### LINKED ISSUE #178: Session middleware is highly insecure
The session middleware does not have any timestamp component to it, which means that if a session happens to be somehow leaked, it can be reused any time in the future.

Can we have a timestamp component added to the session, so that the session middleware can reject such sessions after a reasonable period of time (e.g. 24 hours)?

jessebraham:
Resolves #178 

This involved minimal changes; it should all be straight forward. Essentially I just swapped out the `Signer` instance for one of type `TimestampSigner` and added an optional `max_age` argument to the `SessionMiddleware` constructor. Then when we attempt to unsign the session, a `SignatureExpired` exception is raised if the age of the session exceeds `max_age`, and the scope's `session` attribute is cleared. I'm not thrilled about how I specified the default `max_age` value, but also can't really think of any reasonable alternatives.

I also had to refactor the tests a bit. To avoid interfering the the existing session-related tests, I created a simple app factory function to allow us to configure the middleware separately.

Let me know if you'd like any changes made :)

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -3,6 +3,7 @@ import json
 from base64 import b64decode, b64encode
 
 import itsdangerous
+from itsdangerous.exc import SignatureExpired
 
 from starlette.datastructures import MutableHeaders
 from starlette.requests import Request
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Looks great! I'll make sure we're ready to bump to 0.8 before pulling this in.

It might be worth us double checking what the behavior of this will be if there's a pre-existing old style session. (It'd be nice if it just treated them as blank in that case.)

PR #193, opened 2018-11-07
Rejig ExceptionMiddleware and ServerErrorMiddleware

### LINKED ISSUE #190: Exceptions and middleware
Currently our middleware stack looks like this:

* Server
* ExceptionMiddleware # Generates 500 pages, and deals with handled exceptions.
* User Middleware
* Router
* Endpoint

Having implemented https://github.com/encode/sentry-asgi I've realised that this isn't quite right. We don't want handled exceptions to be captured by sentry or other exception logging middleware. Instead what we *ought* to have is this...

* Server
* ServerErrorMiddleware  # Generates 500 pages.
* User Middleware
* ExceptionMiddleware  # Deals with handled exceptions.
* Router
* Endpoint

Handled exceptions may be raised inside endpoints (or the router), but should not be available to regular middleware, which should just return responses directly. We want this behavior anyway because middleware classes that raise exceptions to be handled will not be portable across to other ASGI frameworks (without also including an exception handling middleware.)

lovelydinosaur:
Closes #190

* Rejigs exception handling, so managed exceptions do not get seen as errors by the middleware stack.
* Adds support for status code handlers

```python
@app.exception_handler(404)
async def not_found(request, exc):
    ...
```

TODO:

- [x] Update docs.
- [x] DebugMiddleware -> ServerErrorMiddleware(app, debug: bool)


vishes-shell:
You left [commented out](https://github.com/encode/starlette/blob/master/tests/test_exceptions.py#L46-L62) test cases for exceptions. That code could probably start to smell in future...

lovelydinosaur:
@vishes-shell Thanks! Now resolved.

PR #194, opened 2018-11-07
fixed issue 188,trailing spaces not allowed in header

Witekkq:
http.cookies output is created in following way:
```
def output(self, attrs=None, header="Set-Cookie:"):
	return "%s %s" % (header, self.OutputString(attrs)
```

our goal is to get rid standard Set-Cookie header, and create a tuple instead of string, but cookie_val is created with leading space if header is empty ("") by the http lib.

Based on https://tools.ietf.org/html/rfc2616 leading space is not allowed (HTTP/1.1)


lovelydinosaur:
Looks good to me. Thanks!

PR #195, opened 2018-11-08
StaticFiles will check if directory exists upon instantiation

### LINKED ISSUE #185: Check directory exists when instantiating `StaticFiles`
The `StaticFiles` application should ensure that the directory exists at the point it is instantiated.

(With an optional switch to turn this behavior off)

alexbotello:
Closes #185 

StaticFiles will now preform a runtime check once instantiated and `check_dir` has been added as a switch.


lovelydinosaur:
Perfect!

Issue #189, opened 2018-11-07
Support an `APISpec` SchemaGenerator


lovelydinosaur:
So @kennethreitz's Responder uses APISpec for generating the API schemas.
That's actually really nice, as it includes various plugins including marshmallow support etc.

We should have a [SchemaGenerator](https://www.starlette.io/schemas/) implementation that uses APISpec. I'm guess ideally it should be a third party implementation, rather than included in the main package.

This would make a great little project for a contributor to jump onto.

See also #172

lovelydinosaur:
Note that we'll kept the schema generation strictly separate from any API docs generation.

woile:
Some questions

1) What would be the idea for now? Do you have any rough example using APISpec?

 Would something like this be okay?

```python
from starlette.applications import Starlette
from apispec import APISpec

app = Starlette()
app.schema_generator = APISpec(
    title='Gisty',
    version='1.0.0',
    openapi_version='3.0.0',
    info=dict(
        description='A minimal gist API'
    )
)

@app.route("/users", methods=["POST"])
def create_user(request):
    """
    responses:
      200:
        description: A user.
        examples:
          {"username": "tom"}
    """
    raise NotImplementedError()

@app.route("/schema", methods=["GET"], include_in_schema=False)
def schema(request):
    return OpenAPIResponse(app.schema)
```

Of course this would mean that you could use marshmallow plugin for example

And make starlette understand when an `APISpec` instance is received?

And regarding this:

> Note that we'll kept the schema generation strictly separate from any API docs generation.

2) It means apistar could be used to generate the docs based on starlette's schema, right?

Cheers!

lovelydinosaur:
@Woile That's about right yes.

You wouldn't be able to *quite* do this: `app.schema_generator = APISpec` since the schema generator needs to fulfil the `BaseSchemaGenerator` interface.

https://github.com/encode/starlette/blob/master/starlette/schemas.py#L24

Ie. implement `get_schema` so that it passes each route to `spec.add_path`

woile:
Yeah, true. So we should have a wrapper over APISpec, maybe called `APISpecSchemaGenerator`, right?

Should this schema generator live inside Starlette or should it be a third party package?

lovelydinosaur:
> Should this schema generator live inside Starlette or should it be a third party package?

Probably a separate package, but could go either way. I guess start out by demonstrating it in code, then we can figure out where it lives afterwards.

woile:
[...]

PR #197, opened 2018-11-08
refactor: schemas now have better support for third party packages

woile:
Hey @tomchristie this is my initial proposal for apispec support. I want to know what do you think. If you give me the green these are the next steps:

* tests
* docs

Also:

renamed: SchemaGenerator -> DictSchemaGenerator

relates: #189

Current schema generator would be used the same, but the name would change, and for APISpec here's an example:

```python
import pprint
from starlette.schemas import (
    OpenAPIResponse,
    APISpecSchemaGenerator,
)
from marshmallow import Schema, fields
from apispec.ext.marshmallow import MarshmallowPlugin
from apispec import APISpec

app = Starlette()
app.schema_generator = APISpecSchemaGenerator(APISpec(
    title="Example API",
    version="1.0",
    openapi_version="3.0.0",
    info={"description": "best api every"},
    plugins=[MarshmallowPlugin()],
))


class UserSchema(Schema):
    id = fields.Int(dump_only=True)
    name = fields.Str(description="The user's name")


@app.route("/users", methods=["GET"])
def list_users(request):
    """
    responses:
      200:
        description: A list of users.
        schema:
            schema: User
        examples:
          [{"username": "tom"}, {"username": "lucy"}]
    """
    raise NotImplementedError()


@app.route("/users", methods=["POST"])
def create_user(request):
    """
    responses:
      200:
        description: A user.
        examples:
          {"username": "tom"}
    """
    raise NotImplementedError()


@app.route("/schema", methods=["GET"], include_in_schema=False)
def schema(request):
    return OpenAPIResponse(app.schema)


app.schema_generator.spec.definition("User", schema=UserSchema)
pp = pprint.PrettyPrinter(indent=2)
pp.pprint(app.schema)
```

### DIFF
--- a/starlette/schemas.py
+++ b/starlette/schemas.py
@@ -21,42 +21,73 @@ class OpenAPIResponse(Response):
         return yaml.dump(content, default_flow_style=False).encode("utf-8")
 
 
+class EndpointInfo(typing.NamedTuple):
+    path: str
+    http_method: str
+    func: typing.Callable
+
+
 class BaseSchemaGenerator:
     def get_schema(self, routes: typing.List[BaseRoute]) -> dict:
         raise NotImplementedError()  # pragma: no cover
 
+    def get_endpoints(
+        self, routes: typing.List[BaseRoute]
+    ) -> typing.List[EndpointInfo]:
+        """
+        Given the routes, yields the following information:
 
-class SchemaGenerator(BaseSchemaGenerator):
-    def __init__(self, base_schema: dict) -> None:
-        assert yaml is not None, "`pyyaml` must be installed to use SchemaGenerator."
-        self.base_schema = base_schema
-
-    def get_schema(self, routes: typing.List[BaseRoute]) -> dict:
-        paths = {}  # type: dict
+        - path
+            eg: /users/
+        - http_method
+            one of 'get', 'post', 'put', 'patch', 'delete', 'options'
+        - func
+            method ready to extract the docstring
+        """
+        endpoints_info: list = []
 
         for route in routes:
             if not isinstance(route, Route) or not route.include_in_schema:
                 continue
 
             if inspect.isfunction(route.endpoint) or inspect.ismethod(route.endpoint):
-                docstring = route.endpoint.__doc__
                 for method in route.methods or ["GET"]:
                     if method == "HEAD":
                         continue
-                    if route.path not in paths:
-                        paths[route.path] = {}
-                    data = yaml.safe_load(docstring) if docstring else {}
-                    paths[route.path][method.lower()] = data
+                    endpoints_info.append(
+                        EndpointInfo(route.path, method.lower(), route.endpoint)
+                    )
             else:
                 for method in ["get", "post", "put", "patch", "delete", "options"]:
                     if not hasattr(route.endpoint, method):
                         continue
-                    docstring = getattr(route.endpoint, method).__doc__
-                    if route.path not in paths:
-                        paths[route.path] = {}
-                    data = yaml.safe_load(docstring) if docstring else {}
-                    paths[route.path][method] = data
+                    func = getattr(route.endpoint, method)
+                    endpoints_info.append(
+                        EndpointInfo(route.path, method.lower(), func)
+                    )
+        return endpoints_info
 
+    def parse_docstring(self, func_or_method: typing.Callable) -> dict:
+        """
+        Given a function, parse the docstring as YAML and return a dictionary of info.
+        """
+        docstring = func_or_method.__doc__
+        return yaml.safe_load(docstring) if docstring else {}
+
+
+class SchemaGenerator(BaseSchemaGenerator):
+    def __init__(self, base_schema: dict) -> None:
+        self.base_schema = base_schema
+
+    def get_schema(self, routes: typing.List[BaseRoute]) -> dict:
         schema = dict(self.base_schema)
-        schema["paths"] = paths
+        schema.setdefault("paths", {})
+        endpoints_info = self.get_endpoints(routes)
+
+        for endpoint in endpoints_info:
+            if endpoint.path not in schema["paths"]:
+                schema["paths"][endpoint.path] = {}
+            schema["paths"][endpoint.path][endpoint.http_method] = self.parse_docstring(
+                endpoint.func
+            )
         return schema
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@
 
 import os
 import re
-import sys
 
 from setuptools import setup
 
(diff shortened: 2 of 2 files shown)


lovelydinosaur:
Oh that's exciting, good stuff!

So, let's not rename SchemaGenerator, or change it's interface.

We might want to pull some of the logic in `get_schema` down into a couple of different methods on the base class, so that it is more easily reusable.

```python
class BaseSchemaGenerator:
    def get_route_map(self, routes: typing.List[BaseRoute]) -> typing.Dict[str, typing.Dict[str, typing.Callable]]:
        """
        Given a list of routes, returns a structure like:

        {
            "/": {
                 "get": <func or method>,
                 "put": <func or method>
            }
        }
        """
        ...

    def parse_docstring(self, func_or_method: typing.Callable) -> dict:
        """
        Given a function, parse the docstring as YAML and return a dictionary of info.
        """
        ...

    def get_schema(self, routes):
        raise NotImplemented()

class SchemaGenerator(BaseSchemaGenerator):
    def __init__(self, base_schema):
        self.base_schema = base_schema

    def get_schema(self, routes):
        schema = dict(self.base_schema)
        route_map = self.get_route_map(routes)
        for path, operations in route_map.items():
            schema['paths'][path] = {}
            for operation, func in operations.items():
                schema['paths'][path][operation] = self.parse_docstring(func)
        return schema

class APISpecSchemaGenerator(BaseSchemaGenerator):
    def __init__(self, sepc):
        self.spec = spec

    def get_schema(self, routes):
        route_map = self.get_route_map(routes)
        for path, operations in route_map.items():
            operations_dict = {
                operation: self.parse_docstring(func) for operation, func in operations.items()
            }
            self.spec.add_path(path=path, operations=operations_dict)
        return spec.to_dict()
```

woile:
ohhh nice nice!! I like it!! I'll push an update tomorrow with tests + docs

requirements.txt
```diff
@@ -14,6 +14,8 @@ isort
 mypy
 pytest
 pytest-cov
+marshmallow
```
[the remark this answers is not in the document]
woile: I will skip it for now. The yaml generated for the schemas does not keep the order, making the tests fail randomly.

woile:
Hey @tomchristie here an update with the tests. 

I decided to go for a generator yielding the endpoints found, instead of collecting all of the operations, because there was no need to collect them all first. Both schemas generator can handle the operations properly.

I will add the docs next

starlette/schemas.py
```diff
@@ -18,45 +24,90 @@ def render(self, content: typing.Any) -> bytes:
         assert isinstance(
             content, dict
         ), "The schema passed to OpenAPIResponse should be a dictionary."
+
+        def represent_dict(dumper, instance):
+            return dumper.represent_mapping("tag:yaml.org,2002:map", instance.items())
+
+        yaml.add_representer(collections.OrderedDict, represent_dict)
+        if LazyDict is not None:
```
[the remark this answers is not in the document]
lovelydinosaur: I'd suggest:

```python
def make_plain_datastructure(data):
    if isinstance(data, list):
        return [make_plain_datastructure(item) for item in data]
    elif isinstance(data, dict):
        return {key: make_plain_datastructure(value) for key, value in data.items()}
    return data
```

Then have `get_schema` return `make_plain_datastructure(self.spec.to_dict())`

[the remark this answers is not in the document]
woile: Yeah, this looks much better

Thanks!

woile:
hey @tomchristie I've added the docs, tell me what you think when you have some time. Thanks!

lovelydinosaur:
This is looking awesome to me.

How would you feel about:

* Issuing a PR which *only* has the core changes you'd need in `SchemaGenerator`.
* Maintaining a third party package for the APISpec support.

?

I understand that'd mean some extra work, but it'd be really nice if we could start to show examples of third party packages on top of starlette, rather than building everything in.

woile:
Sure, I'll try to do it this week.

Regarding the documentation, what do you think about having something like awesome -django or awesome-flask but living in the starlette.io documentation. Or are you thinking more like a Django packages.org approach? Do you have any ideas already? Sorry for the wall of questions haha

woile:
Okay updated this part. Tell me what you think. 

Tomorrow I'm gonna work on the package. I'll probably upload it to github but I won't publish to pypi until we have a new version of starlette with this PR merged, so I can pin the version.

If you have some time check [hapi](https://hapijs.com/plugins) way of documenting plugins, I find it really nice.

lovelydinosaur:
> If you have some time check hapi way of documenting plugins, I find it really nice.

Yeah I'd be up for that style. Nice and simple, all in one place.

starlette/schemas.py
```diff
@@ -21,42 +21,66 @@ def render(self, content: typing.Any) -> bytes:
...
             if inspect.isfunction(route.endpoint) or inspect.ismethod(route.endpoint):
-                docstring = route.endpoint.__doc__
                 for method in route.methods or ["GET"]:
                     if method == "HEAD":
                         continue
-                    if route.path not in paths:
-                        paths[route.path] = {}
-                    data = yaml.safe_load(docstring) if docstring else {}
-                    paths[route.path][method.lower()] = data
+                    yield Endpoint(route.path, method.lower(), route.endpoint)
```
lovelydinosaur: Let's just return a list of `Endpoint` instead.

starlette/schemas.py
```diff
@@ -21,42 +21,66 @@ def render(self, content: typing.Any) -> bytes:
         return yaml.dump(content, default_flow_style=False).encode("utf-8")
 
 
+class Endpoint(typing.NamedTuple):
```
lovelydinosaur: I'm a bit wary about calling this `Endpoint`, since the callable is what we're referring to as the "endpoint" elsewhere. Perhaps it should be `EndpointInfo`?

starlette/schemas.py
woile: okay, I'll update it. It feels a bit weird iterating once to collect the endpoints info and then iterating again to generate the schema, do you find it less readable with the `yield`? or am I adding too much unnecessary complexity?

starlette/schemas.py
woile: yes, makes sense, another alternative could be `PathInfo` as well. I'll stick with `EndpointInfo` for now.

starlette/schemas.py
woile: I leave it with the array for now

woile:
Take a look @tomchristie [starlette-apispec](https://github.com/Woile/starlette-apispec) :smile: 

Once we merge this PR, I'll update the depenencies and publish it :tada:

lovelydinosaur:
Released to PyPI 👍

woile:
Released to PyPI :smile: and made a dummy server, everything seems to be working fine :tada: 

[starlette-apispec](https://pypi.org/project/starlette-apispec/)

PR #203, opened 2018-11-09
204 and 1XX response has no content-length

r-m-n:


starlette/responses.py
```diff
@@ -63,6 +63,10 @@ def init_headers(self, headers: typing.Mapping[str, str] = None) -> None:
             populate_content_length = b"content-length" in keys
             populate_content_type = b"content-type" in keys
 
+        # RFC 7230: "a server MUST NOT send a Content-Length header field in any
+        # response with a status code of 1xx (Informational) or 204 (No Content)."
+        populate_content_length = self.status_code >= 200 and self.status_code != 204
```
laixintao:
```suggestion
        populate_content_length = populate_content_length and self.status_code >= 200 and self.status_code != 204
```

laixintao:
Looks good to me 👍

laixintao:
There are some conflicts now... @r-m-n

laixintao:
Any comments on this? @tomchristie 😀

lovelydinosaur:
So I'm not sure about this now. I think #208 removes the motivation for it.

If you don't set a body then you'll be getting the correct behavior now anyways, if you *do* set a body on a `204`/`1xx`, then this change will remove the content length, but still include the body.

I assume we should just close this one now.

PR #208, opened 2018-11-09
Default to empty body on responses

### LINKED ISSUE #201: the raw "Response" should'nt mandatory a content ... no ?
I need to manage myself the etag/304 response ...

to send a 304 if-none-match, I use : 

    Response(status_code=304,content="",headers=rheaders)

But the body content should not be written ... here the content is mandatory : None cause troubble, and empty string is an empty string ...

lovelydinosaur:
I figure that #206 wasn't quite right, because eg. it'd behave slightly incorrectly for cases like `return JSONResponse()`

Closes #201
Closes #206

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -31,13 +31,16 @@ class Response:
 
     def __init__(
         self,
-        content: typing.Any,
+        content: typing.Any = None,
         status_code: int = 200,
         headers: dict = None,
         media_type: str = None,
         background: BackgroundTask = None,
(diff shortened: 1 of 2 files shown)


manatlan:
On GAE/py37 Standard Env...
```python 
      return Response(status_code=304,content=b"")
```
works well with starlette 0.8.1
but not with startlette 0.8.4 ... it continues to break GAE ;-(

I will stay with 0.8.1 for now ... but will dig that problem later

PR #209, opened 2018-11-09
Return proper error response for an empty json body

alexbotello:
This implements the idea I suggested in #135.

If the request is running inside a Starlette application, it will then be raised with `HTTPException` and handled by the exception middleware. A normal ASGI app will return a `PlainTextResponse` with a status of 400. However, there is the minor inconvenience of having to import `HTTPException` inside `Request.json` in order to avoid circular dependence errors.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -6,6 +6,7 @@ from urllib.parse import unquote
 
 from starlette.datastructures import URL, Headers, QueryParams
 from starlette.formparsers import FormParser, MultiPartParser
+from starlette.responses import PlainTextResponse
 from starlette.types import Message, Receive, Scope
 
 try:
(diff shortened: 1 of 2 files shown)


vishes-shell:
@alexbotello thanks for your take on that error.

Actually what we are getting with empty/invalid request body is `JSONDecodeError`. And every application have it's own error responses.

Simplest solution would be to add handler:
```python
app.add_exception_handler(JSONDecodeError, invalid_body_handler)
```

But it's not really readable, since `JSONDecodeError` _could also be raise in decoding json_ in any view and that's _not_ what we actually want to catch.

I would suggest **create exception** like `InvalidBody`, and we then **can add any handler**, that would satisfy every need. And it's also good to EAFP, this way you will handle invalid and empty body.

Something similar to this:

```python
    if not hasattr(self, "_json"):
        body = await self.body()
        try:
            self._json = json.loads(body)
        except JSONDecodeError:
            raise InvalidBody
    ...
```

starlette/requests.py
```diff
@@ -125,6 +126,12 @@ def url_for(self, name: str, **path_params: typing.Any) -> str:
     async def json(self) -> typing.Any:
         if not hasattr(self, "_json"):
             body = await self.body()
+            if body is b"":
+                if "app" in self._scope:
```
lovelydinosaur, line 130: I’m planning to get rid of this approach and just raise HTTPExcpetion in the router/endpoint implementations. The low level interface will document how to catch the exception or override the behaviour.

PR #210, opened 2018-11-10
Minor doc fix for `AsyncioExecutor`

jordaneremieff:
The `executor` argument in the GraphQL doc example needs to be in the `GraphQLApp` args.


lovelydinosaur:
Ace, thanks.

Issue #199, opened 2018-11-08
Testing template reponses


lovelydinosaur:
This is going to require a bit of thinking about.

```python
app = Starlette(template_directory='templates')

@app.route('/')
async def homepage(request):
    template = app.get_template('index.html')
    content = template.render(request=request)
    return HTMLResponse(content)

client = TestClient(app)
response = client.get('/')
assert response.template = 'index.html'
assert 'request' in response.context
```

Two issues here. Firstly we haven't associated the template name or context with the response in any way, so we can't reference them from the template. We could do so, with something like...

```python
return app.render_reponse('index.html', context={'request': request})
```

Or...

```python
return TemplateResponse(template, context={'request': request}))  # I *guess* `template` has the "index.html" name string associated with it somewhere?
```

Secondly, we don't have access to the response instance itself, we've just been sent ASGI messages. We'd need to ensure that `TemplateResponse` included some extra info ("template", "context") in the messages that it sent back, and that `TestClient` was able to additionally annotate the response with that information.

jordaneremieff:
@tomchristie I updated the PR to include one potential pathway for the first part (assuming I understood the issue correctly) - though making the response details available on the `TestClient` appears to be a more difficult challenge.

Is that something that perhaps extensions could be used for? Or just custom message types? I found that if I send the template/context data through the response messages into the adapter that I was still faced with the problem of allowing the mock response to digest that information through the raw request.

PR #211, opened 2018-11-10
Template endpoint

jordaneremieff:
Great work on the template integration!

I was trying it out, and I ended up tailoring an `HTTPEndpoint` to handle the template stuff - figured I'd PR it back into Starlette to see if it would make a good addition. It's fairly minimal at this point.

Below is a usage example:

```python
import uvicorn
from starlette.endpoints import TemplateEndpoint

app = Starlette(debug=True, template_directory="templates")


@app.route("/")
class Home(TemplateEndpoint):

    template_name = "index.html"


@app.route("/hello")
class Hello(TemplateEndpoint):

    template_name = "hello.html"

    def get_context(self, request):
        context = super().get_context(request)
        context["hello"] = "hello world"
        return context


if __name__ == "__main__":
```

If this is something worth including, then I have some questions about if we should also consider providing async versions of the `get_context` and `get` methods on the class, either in place of or in addition to the sync methods. I'm actually using it with async methods in my app, but I think the way I'm using it would be less common than the sync case.

Also, this is based loosely on my familiarity with Django's CBVs, so I'm open to updating the PR to include alternative implementations.


jordaneremieff:
I think I botched something in my local git configuration, closing this and can re-open a clean PR later pending any feedback in https://github.com/encode/starlette/issues/199.

Issue #22, opened 2018-07-12
StaticFile refinements


lovelydinosaur:
* ~`HEAD`~.
* Range responses.
* ~`304` if ETag and Last-Modified indicate no changes against a conditional `GET` request.~
* Consider Manifests / Gzip / Far-future.

laixintao:
I am going to implement range response next :)

PR #212, opened 2018-11-10
Support HTTP 304 for static files.

laixintao:
> 304 if ETag and Last-Modified indicate no changes against a conditional GET request.

from #22

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -1,10 +1,21 @@
 import os
 import stat
+from email.utils import jarsedate_to_datetime
 
 from aiofiles.os import stat as aio_stat
 
+from starlette.requests import Request
 from starlette.responses import FileResponse, PlainTextResponse, Response
 from starlette.types import ASGIInstance, Receive, Scope, Send
+
+NOT_MODIFIED_HEADERS = (
+    "cache-control",
+    "content-location",
+    "date",
+    "etag",
+    "expires",
+    "vary",
+)
 
 
 class StaticFiles:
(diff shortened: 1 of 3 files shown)


starlette/staticfiles.py
```diff
@@ -50,6 +61,26 @@ def __init__(self, scope: Scope, path: str, check_directory: str = None) -> None
...
+            resp_headers["last-modified"]
+        )
+
+    def not_modified_response(self, response: FileResponse) -> Response:
+        headers = {
+            name: value
+            for name, value in response.headers.items()
+            if name in NOT_MODIFIED_HEADERS
+        }
+        return Response(status_code=304, headers=headers)
```
laixintao, line 82: Should I use enum from `status` here?

starlette/staticfiles.py
```diff
@@ -50,6 +61,26 @@ def __init__(self, scope: Scope, path: str, check_directory: str = None) -> None
         if not (stat.S_ISDIR(stat_result.st_mode) or stat.S_ISLNK(stat_result.st_mode)):
             raise RuntimeError("StaticFiles path '%s' is not a directory." % directory)
 
+    def is_not_modified(self, response: FileResponse) -> bool:
+        req_headers = Request(self.scope).headers
```
lovelydinosaur: Use `headers = Headers(scope=self.scope)`

starlette/staticfiles.py
```diff
@@ -66,5 +97,7 @@ def __init__(self, scope: Scope, path: str, check_directory: str = None) -> None
                 response = FileResponse(
                     self.path, stat_result=stat_result, method=self.scope["method"]
                 )
+                if self.is_not_modified(response):
```
lovelydinosaur: This looks like a strange way around to do this. Could we not do something more like:

```python
if not stat.S_ISREG(mode):
    response = PlainTextResponse("Not Found", status_code=404)
elif self.is_not_modified(stat_result):
    response = ...
else:
    response = FileResponse(...)
```

laixintao:
Thanks for review, I will update this PR later.

starlette/staticfiles.py
laixintao: Hi, I just realized that you need `etag` and `if-modified-since` header for checking 304. So I move the `set_stat_header()` logic to a classmethod. What do you think?

In `StaticFile` they got the `etag` and `if-modified-since` in the first place:

https://github.com/evansd/whitenoise/blob/1b5a7f3a0f151ee1449d88da9e4813644e435fd4/whitenoise/responders.py#L36-L38

starlette/staticfiles.py
lovelydinosaur: Nope

starlette/staticfiles.py
```diff
@@ -50,6 +62,25 @@ def __init__(self, scope: Scope, path: str, check_directory: str = None) -> None
...
+        if "if-modified-since" not in req_headers:
+            return False
+        last_req_time = req_headers["if-modified-since"]
+        return parsedate(last_req_time) >= parsedate(last_modified)  # type: ignore
+
+    def not_modified_response(self, stat_headers: dict) -> Response:
+        headers = {
+            name: value
+            for name, value in stat_headers.items()
+            if name in NOT_MODIFIED_HEADERS
```
lovelydinosaur, line 80: Any resources on which headers 304 responses should include and why?

lovelydinosaur (commented):
Looks great! I'd like to understand why we put that particular set of headers in 304 responses, but otherwise I think we're about good to go.

lovelydinosaur:
Let go for it. Nice one!

Issue #182, opened 2018-11-06
Support route convertors


manatlan:
Hi, I've just discovered starlette (seems to be the perfect asgi fw ! great)
Looking into docs/unitests ... but it seems that regex routing is not implemented yet (will look into the code)

Is it possible to declare something like that ?
```python 
@app.route("/api/{name:.*}")
def ...
```

It's the only thing it seems to miss (to quit aiohttp)

lovelydinosaur:
We don't *currently* have any support for that, no.
And yes, it would be (very) good to do so.

I'm not sure exactly what we'd want to adopt here - if we should use something like [flask's URL convertors](http://exploreflask.com/en/latest/views.html), or if we should just use regexs. What existing different options do other frameworks currently use? Which options support type coercion, and which always return strings for the path parameters?

manatlan:
I don't known flask at all (I just know that it was a joke against bottle, at the begining ..)

I'm coming from bottle(wsgi) ( https://bottlepy.org/docs/dev/routing.html ). Bottle does type coercions.

And in asgi world: I use aiohttp ( https://aiohttp.readthedocs.io/en/stable/web_quickstart.html#resources-and-routes ), which just return strings.

But in all cases : you'll need to add a sort of "path/catchall" ... to let the developper parse the path as he wants.
Because, currently ; It's impossible to make an endpoint with an abritratry number of subpaths

Currently i do : (it's not pretty, and it works only for 6 ;-)  
```python
@app.route('/{p1}/{p2}/{p3}/{p4}/{p5}/{p6}')
@app.route('/{p1}/{p2}/{p3}/{p4}/{p5}')
@app.route('/{p1}/{p2}/{p3}/{p4}')
@app.route('/{p1}/{p2}/{p3}')
@app.route('/{p1}/{p2}')
@app.route('/{p1}')
@app.route('/')
async def gotopath(request):
    fullpath="/".join( request.path_params.values() )
```

Dith3r:
> Because, currently ; It's impossible to make an endpoint with an abritratry number of subpaths

It's possible to declare your own regex. 

```
@app.route("/(?P<path>.*)")
```

lovelydinosaur:
@Dith3r That's true actually, yeah. (Tho it won't work with `request.url_for(...)`)

lovelydinosaur:
I've retitled the issue with the feature that we'd actually like to have here. Converters for routes.

So, initially just making sure that we can support this kind of thing...

[...]

PR #214, opened 2018-11-12
Add support for route converters

alexbotello:
Refs #182. Here's what I think route converters could look like in Starlette. If nothing else, it can be used as a base to work against.

Supports:
* `{param:int}`
* `{param:float}`
* `{param:path}`

The `convert_params` interface allows new converters to be added easily.

Example usage:
```python
# Integer
@app.route("/get/{number:int}")
async def func(request):
    number = request.path_params["number"]
    return JSONResponse({"num": number})

response = client.get('/get/25')
number = response.json()["num"]
type(number) == int  # True

# Float
@app.route("/get/{number:float}")
async def func(request):
    number = request.path_params["number"]
    return JSONResponse({"num": number})

response = client.get('/get/25.5')
number = response.json()["num"]
type(number) == float  # True

# Path
@app.route("/get/{param:path}")
async def func(request):
    return PlainTextResponse("Hello, world!")

response = client.get('/get/hello/world/1')
response.text == "Hello, world!"  # True
response.status_code == 200  # True
```

### DIFF
--- a/starlette/converters.py
+++ b/starlette/converters.py
@@ -1,0 +1,32 @@
+import typing
+
+
+class ValidationError(Exception):
+    pass
+
+
+class Converter:
+    def convert(self, value: str) -> typing.Any:
+        raise NotImplementedError()  # pragma: no cover
+
+
+class IntConverter(Converter):
+    def convert(self, value: str) -> int:
+        try:
+            converted_value = int(value)
+        except ValueError:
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Good stuff. Taking a look at this, may merge parts in with work I'm doing here, too. TBD.

PR #215, opened 2018-11-12
Support File Upload via GraphQL

Bomberus:
# Description
This change allows file upload via GraphQL. 
See [https://github.com/jaydenseric/graphql-multipart-request-spec](url)

# How to use
Install graphene-file-upload and import

Create Schema:
```python
from graphene_file_upload.scalars import Upload
class UploadMutation(graphene.Mutation):
    class Arguments:
        file = Upload(required=True)

    success = graphene.Boolean()

    async def mutate(self, info, file):
        # variable file contains binary value

        return UploadMutation(success=True)

class Mutation(graphene.ObjectType):
    uploadMutation = UploadMutation.Field()

schema=graphene.Schema(mutation=Mutation)
```

# Example in Postman
POST /graphql
Content-Type: multipart/form-data

query: mutation ($file: Upload!) { uploadMutation(file: $file) { success } }
file_map: { "file0": "file" }
variables: { "file": ""}
file0: YOURFILE

The idea is to overwrite the variables object with the binary value via file_map

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -52,6 +52,17 @@ class GraphQLApp:
                 data = {"query": text}
             elif "query" in request.query_params:
                 data = request.query_params
+            elif "multipart/form-data" in content_type:
+                form = await request.form()
+                data = {}
+                file_map = json.loads(form["file_map"])
+                data = {"query": form["query"] }
+                variables = json.loads(form["variables"])
+                
+                for key, path in file_map.items():
+                    variables[path] = await form[key].read()
+                    
+                data["variables"] = variables
             else:
                 return PlainTextResponse(
                     "Unsupported Media Type",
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks! Is this a specified part of GraphQL, or a custom implementation case?

Bomberus:
I think it is not officially supported by GraphQL. 

But many popular frameworks, like apollo or graphl-yoga adapting this pattern.
https://github.com/prisma/graphql-yoga/blob/master/examples/file-upload/index.ts
https://github.com/jaydenseric/apollo-upload-client

For the failing jenkins jobs, I think its a windows/linux formatting issue. Will fix that when I am at home.

lovelydinosaur:
Okay, I'd be very happy to see this as a third party package, or even as an example gist that we could link to from the docs.

PR #217, opened 2018-11-13
Fix incorrect decoding of application/x-www-form-urlencoded request body

supriyo-biswas:
The `+` character in a query string (present in `application/x-www-form-urlencoded` request bodies) is always interpreted as a space. However, because `urllib.parse.unquote()` was being used, so the `+` was left as-is.

This PR replaces the call with `urllib.parse.unquote_plus()` which takes care of this issue.

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -3,7 +3,7 @@ import io
 import tempfile
 import typing
 from enum import Enum
-from urllib.parse import unquote
+from urllib.parse import unquote_plus
 
 from starlette.concurrency import run_in_threadpool
 from starlette.datastructures import Headers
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks! It’d be great if we could get this confirmed either in a new test case or by adapting an existing test.

supriyo-biswas:
I'm trying to write a test, however, it seems like the data parameter is not being decoded at all?
https://github.com/supriyo-biswas/starlette/commit/e7664a04d20cfcb2f0942afaaac8e4b01626ab84

Result of running the tests:

```
>       assert response.json() == {"form": {"abc": "123 123 !"}}
E       AssertionError: assert {'form': {'ab...3+123%20%21'}} == {'form': {'abc': '123 123 !'}}
E         Differing items:
E         {'form': {'abc': '123+123%20%21'}} != {'form': {'abc': '123 123 !'}}
E         Full diff:
E         - {'form': {'abc': '123+123%20%21'}}
E         ?                      ^   ^^^^^^
E         + {'form': {'abc': '123 123 !'}}
E         ?                      ^   ^^
```

However, when using starlette normally, I can confirm that the request body is being decoded correctly with my changes.

lovelydinosaur:
You're using a `GET` request in that test, which will put the data in the query parameters.
Try a similar test but with `POST`.

lovelydinosaur:
Fab, thanks!

PR #227, opened 2018-11-21
Remove extraneous print statement in TrustedHostMiddleware

alexbotello:
^


vishes-shell:
As mentioned in [Flag “print” statements in Python code](https://stackoverflow.com/questions/25249586/flag-print-statements-in-python-code) there is [`flake8-print`](https://pypi.org/project/flake8-print/)

lovelydinosaur:
Thanks @vishes-shell - If we start using flake8 elsewhere then we should enable that too.

PR #228, opened 2018-11-21
Clean up all unused imports

alexbotello:
It would be useful if the linting script could cover this too. `flake8` possibly? However, there's the issue where imports used specifically for commented type defs would be flagged as errors by `flake8`.

For example:
In `starlette/applications`:
```python
from starlette.schemas import BaseSchemaGenerator
...
self.schema_generator = None  # type: typing.Optional[BaseSchemaGenerator]
```

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,6 +1,6 @@
 import typing
 
-from starlette.datastructures import URL, URLPath
+from starlette.datastructures import URLPath
 from starlette.exceptions import ExceptionMiddleware
 from starlette.lifespan import LifespanHandler
 from starlette.middleware.base import BaseHTTPMiddleware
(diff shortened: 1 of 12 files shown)


lovelydinosaur:
Great! What did you use to flag these up?

This looks like it does the job...

`autoflake --remove-all-unused-imports --in-place -r starlette`

We could add that to `scripts/lint` - We wouldn't get strict CI failures, but we would have it be applied automatically anytime we run the listing.

alexbotello:
> `autoflake --remove-all-unused-imports --in-place -r starlette`
> 
The problem with this is what I referenced above. Running this command will cause our `mypy` type checks to fail, as it will remove imports used specifically for commented type definitions. 

eg. `# type: typing.List[typing.Optional[Message]]` or `# type: typing.Optional[BaseSchemaGenerator]`

`mypy` will error:
```
starlette/middleware/wsgi.py:71: error: Name 'Message' is not defined
starlette/applications.py:21: error: Name 'BaseSchemaGenerator' is not defined
```
These are the only two cases in the code that are affected, so it wouldn't be a hassle to change. But, it would prevent us from using type defs in this style moving forward

lovelydinosaur:
Okay, that’s a no-go then.

PR #235, opened 2018-11-23
add tests for chunked requests

ericls:
Was doing some research on ASGI and found the `testclient` from `starlette` extremely helpful, but it was missing the functionality of passing a generator as body, which is supported by `requests`. So I added that to `_TestClient` and because there's no existing tests for chunked encoding requests, I also added that.


lovelydinosaur:
Nice work!

PR #243, opened 2018-11-30
Add DatabaseMiddleware

lovelydinosaur:
Example:

```python
import os
import sqlalchemy
from starlette.middleware.database import DatabaseMiddleware


metadata = sqlalchemy.MetaData()

notes = sqlalchemy.Table(
    "notes",
    metadata,
    sqlalchemy.Column("id", sqlalchemy.Integer, primary_key=True),
    sqlalchemy.Column("text", sqlalchemy.String),
    sqlalchemy.Column("completed", sqlalchemy.Boolean),
)

app = Starlette()
app.add_middleware(DatabaseMiddleware, database_url=os.environ['DATABASE_URL'])


@app.route("/notes", methods=["GET"])
async def list_notes(request):
    query = notes.select()
    results = await request.database.fetchall(query)
    content = [
        {"text": result["text"], "completed": result["completed"]} for result in results
    ]
    return JSONResponse(content)


@app.route("/notes", methods=["POST"])
async def add_note(request):
    data = await request.json()
    query = notes.insert().values(text=data["text"], completed=data["completed"])
    await request.database.execute(query)
    return JSONResponse({"text": data["text"], "completed": data["completed"]})
```

Queries:

Supports either SQLAlchemy Core queries, or raw SQL.

Database sessions deal with underlying connections transparently. They'll obtain a connection from the pool when required, and return it as soon as it's no longer held anywhere.

* `request.database.fetchall(query)`
* `request.database.fetchone(query)`
* `request.database.fetchval(query)`
* `request.database.execute(query)`

Transactions:

Available as a decorator, as a context manager, or with a low-level interface.

* `@transaction`
* `async with request.database.transaction()`
* `transaction = request.database.transaction()` / `transaction.start()` / `transaction.commit()` / `transaction.rollback()`

Testing:

Use `rollback_on_shutdown` when instantiating `DatabaseMiddleware` to support test-isolated sessions.

```python
app.add_middleware(
    DatabaseMiddleware,
    database_url=os.environ['DATABASE_URL'],
    rollback_on_shutdown=os.environ['TESTING']
)
```

You'll need to use `TestClient` as a context manager, in order to perform application startup/shutdown.

```python
with TestClient(app) as client:
    # Entering the block performs application startup.
    ...
    # Exiting the block performs application shutdown.
```

If you're using py.test you can create a fixture for `client`.

---

* TODO:

- [ ] sqlite & mysql backends.
- [ ] Document how to integrate with alembic.
- [ ] `executemany`
- [ ] Cursor support.
- [ ] Strict locking around connection acquire in Postgres backend
- [ ] Middleware should strictly ensure closed connections/transactions


woile:
Hey tom fantastic idea. I was wondering, isn't this a bit to tied to postgres? I see only the `asyncpg` dependency, what if you wanted to use `aiomysql` for example? Would it be possible?

And asking from ignorance, would be possible to use nosql databases with this middleware?

I love the idea of having a `request.release_db_connection`!

Midnighter:
> During test sessions, use TestClient context to create a transaction, and rollback on completion.

For your inspiration regarding testing, it's really cool to be able to roll back entire sessions, i.e., something is inserted into the database in a test case (`session.commit`) but rolled back entirely after the test case. I personally do this with fixtures and pytest [but the documentation gives an examples for unittest](https://docs.sqlalchemy.org/en/latest/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites).

jordaneremieff:
Great work. :)

Similar to https://github.com/encode/starlette/pull/243#issuecomment-443418746, I wonder, is this perhaps too tied to asyncpg to be considered a generic "DatabaseMiddleware"? I think it would be a good idea to consider making this more configurable for existing alternatives as well as any potential new alternatives that may come about in the "Starlette as a library" sense.

jordaneremieff:
@tomchristie Why shouldn't a websocket lifetime possess an entire DB connection (I may be confusing terms)? The way I understand http vs. websocket requests in an ASGI context would lead me to believe that it would make most sense for db connections to be started/shutdown the same for either protocol.

woile:
Tom I was wondering, isn't it too much responsibility on starlette's core adding support for relation databases?
Wouldn't this be a good opportunity to start building an ecosystem around it?
To me it feels like if I want batteries included I can go to DRF, and if I want something more minimalistic I can go with starlette, and plug what I need. 
I would like to know your opinion about it, maybe I'm not looking at it from the right angle.

Thanks!

lovelydinosaur:
@Woile - The important thing here is that the database support is properly encapsulated. You can either include the middleware or not, or roll something custom based on the interfaces here. Before starting this I *assumed* that it'd likely be too big a bridge to jump over, but actually I don't think that's the case at all. SQLAlchemy core is a *really* good cutoff point for what we're doing here, and the database drivers themselves are doing all the heavy lifting. All we need to do here is figure out how the top-level API looks (which this pretty much already does), and implement the wrappers around the database drivers themselves.

What I want from starlette is for it to be equally suitable for slimming right down, or for building proper full-scale services with. All of the components we're providing in the core library are completely optional, but it's super important that we provide a comprehensive set of fundamental building blocks.

lovelydinosaur:
Places where we likely *won't* go to in any foreseeable future: Including an ORM. (This isn't that)

lovelydinosaur:
Places where we certainly will go: demonstrating how to properly integrate alembic in order to support database migrations, providing mysql and sqlite implementations (wrapping `aiomysql` and `aiosqlite`)

lovelydinosaur:
Docs: https://www.starlette.io/database/

woile:
Thanks a lot for taking the time to explain, much appreciated!

rajasimon:
Shall we document how to include app path to alembic migrations?

_env.py_
```
import os
import sys

APP_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
sys.path.append(APP_PATH)

import app
config.set_main_option('sqlalchemy.url', str(app.DATABASE_URL))
target_metadata = app.metadata
```

PR #244, opened 2018-12-01
Add context with request when executing GraphQL query

pvanliefland:
Currently, there is no way to access the `request` object within a GraphQL resolver or mutation. I think this is problematic because it makes some middlewares useless in a GraphQL context:

1. If I want to authenticate my users, I will probably have to store the logged in user in the scope
2. The database middleware presented in #243 adds a `@db` property on the request object

I suggest to pass a simple dict context containing the request when executing GraphQL stuff.

What do you think?

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -85,7 +85,9 @@ class GraphQLApp:
         )
         return JSONResponse(response_data, status_code=status_code)
 
-    async def execute(self, request, query, variables=None, operation_name=None):  # type: ignore
+    async def execute(
+        self, request, query, variables=None, operation_name=None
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Yes, yes, and yes.
Looks like it’s just missing a little coverage.

pvanliefland:
Strange... let me check :)

lovelydinosaur:
Great stuff! Welcome to make a pull request adding this to the documentation too, or else I'll get around to it when possible.

pvanliefland:
Nice, I'll make a PR for the documentation ASAP!

PR #245, opened 2018-12-01
Resource middlewares

pvanliefland:
I think it would be great to be able to attach "resources" objects to the scope/request. By resources, I mean either service resources (such as a database connection) or data resources (such as the user corresponding to an auth token).

I really like the way you implemented the db property in #243, but I believe other users would enjoy the flexibility of adding their own services or resources to the scope and to the request (other database engines, multiple databases, authentication...). My code here is "just" a generalisation of the code in your PR, so that we can do stuff like:

```python
@app.route("/last-order")
def last_order(request):
    user = request.resource("user")
    store = request.resource("store")
    if user is None:
        return JSONResponse({})
    return JSONResponse({"customer": user, "lastOrder": store.get_last_order()})
```

Ideally, we should be able to access such resources in a websocket endpoint.

What do you think about the general idea here?

### DIFF
--- a/starlette/middleware/resource.py
+++ b/starlette/middleware/resource.py
@@ -1,0 +1,64 @@
+import functools
+import typing
+
+from starlette.types import ASGIApp, ASGIInstance, Message, Receive, Scope, Send
+
+
+class ResourceMiddleware:
+    def __init__(self, app: ASGIApp, name: str) -> None:
+        self.app = app
+        self.name = name
+
+    def __call__(self, scope: Scope) -> ASGIInstance:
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Good plan! My preference here would likely be to have this as a third-party middleware.

Arbitrary state can be accessed on the request using the scope lookup syntax, such as `request["resources"]`. So long as you use that then there's no need for any modifications to the request instance. (Tho I think it's worth us having specific native interfaces for `request.session`, `request.database`, and `request.user`)

PR #255, opened 2018-12-07
Doc: Fix FileResponse path argument

rajasimon:
I'm not sure what causes this but when I first tested the FileResponse with path argument leading / throws `file doesn't exist error`. 

```
response = FileResponse('/statics/favicor.ico') # Not Working
response = FileResponse('statics/favicon.ico') # Working
```
I'm using Python 3.7.1 and also I tested with Python 3.6.5 ( Linux ).

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -195,6 +195,6 @@ class App:
         self.scope = scope
 
     async def __call__(self, receive, send):
-        response = FileResponse('/statics/favicon.ico')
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks, yup!

`/statics/favicon.ico` is wrong, since it'd refer to the root of the filesystem, whereas `statics/favicon.ico` is a relative path.

PR #257, opened 2018-12-09
Support text JSON in WebSocket endpoint, handle malformed JSON

jordaneremieff:
Currently the endpoint only supports bytes messages and will error when attempting to decode the messages received as text. I've been using `encoding = "text"` to receive JSON messages from a frontend WebSocket connection, e.g.:

```
websocket.send(JSON.stringify({
    "content": "My Content"
}))
```
but I think the JSON encoding should also handle text by default. 

Additionally, I added a try/except block to catch any JSON parser errors and raise a `RuntimeError` in that case.

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -84,12 +84,16 @@ class WebSocketEndpoint:
             return message["bytes"]
 
         elif self.encoding == "json":
-            if "bytes" not in message:
+            try:
+                if "text" in message:
+                    message_json = json.loads(message["text"])
+                elif message["bytes"]:
+                    message_json = json.loads(message["bytes"].decode("utf-8"))
+            except json.decoder.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Okay. The one problem is see is asymmetry. Which of text/bytes are we expecting to write output as?

jeffbuttars:
I've encountered issues with text vs bytes JSON on Websockets as well. I can say that text is much easier to deal with on the client side, usually a browser. 

In other, non-starlette, situations I've ended up defaulting to text with bytes as an option do to how much easier it makes things for clients.

lovelydinosaur:
Okay. In terms of the JSON spec, it's a binary protocol, but practically most implementations end up treating it as a textual one.

Conflicted about if and how we should modify `websocket.send_json()` https://github.com/encode/starlette/blob/master/starlette/websockets.py#L146

PR #261, opened 2018-12-11
routing: fix KeyError with optional group

blueyed:
14650bf caused a regression when using an optional regexp group
(`(?P<foo>…)?/…`).

Not sure about the fix, but at least there is a test for it included.. :)

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -175,6 +175,8 @@ class Route(BaseRoute):
             if match:
                 matched_params = match.groupdict()
                 for key, value in matched_params.items():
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks. We don't document using regexs directly, and I don't think we should support them.
Is there any way we can ensure we're only using the `{name}` and `{name:type}` style matches?

blueyed:
Does `{name}` have support for being optional?

lovelydinosaur:
You can either use two seperate routes, or use `{name:path}`, which may be empty.

PR #262, opened 2018-12-11
Remove _TestClient indirection

blueyed:
pytest can be instructed with `__test__ = False` to ignore the class
during discovery.


lovelydinosaur:
Great, yup!

PR #263, opened 2018-12-11
tests: skip test_database without DATABASE_URL

blueyed:
(instead of causing a collection error)


lovelydinosaur:
Great, thank you!

PR #264, opened 2018-12-11
setup.py: add asyncpg to "full" extras_require

blueyed:


lovelydinosaur:
Yes, we oughta have this. Although I also think we're past the point that `starlette[full]` is useful, so may think about removing it at some point.

blueyed:
I was looking for `testing`, which is good to have I think nevertheless (and should include pytest then for example).

PR #266, opened 2018-12-11
Fix grammar with assertion message

blueyed:
It should probably also not be an `assert` in the first place (which
might be optimized out)?!


lovelydinosaur:
I'd be perfectly happy with it being changed to a `ValueError`

PR #267, opened 2018-12-11
tests: enable branch coverage

blueyed:


lovelydinosaur:
Thanks. I'd happily take PRs that incrementally add further branch coverage.

Also, if we *get* to 100% branch coverage then I'd probably be happy to start enforcing that we don't drop below that.

PR #269, opened 2018-12-13
Add TemplateResponse.

### LINKED ISSUE #199: Testing template reponses

lovelydinosaur:
Adds `TemplateResponse(template, context)`, which expose `.template` and `.context` information back to the test client.

There's an issue here to consider

This PR adds `http.response.template` messages. We probably ought to only send these messages when the server indicates that it's able to deal with them via advertising [an ASGI extension](https://asgi.readthedocs.io/en/latest/extensions.html).

If we don't do that then we'll be sending these messages to servers and likely raising warnings into logs.

However it's not obvious how to do this, since `TemplateResponse` doesn't have the `scope` information available to it - the `__call__` method mirrors an ASGI instance interface, and just takes `recieve`/`send`.

Closes #199.

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -86,6 +86,51 @@ class App:
         await response(receive, send)
 ```
 
+### TemplateResponse
+
+The `TemplateResponse` class return plain text responses generated
+from a template instance, and a dictionary of context to render into the
+template.
+
+A `request` argument must always be included in the context. Responses default
+to `text/html` unless an alternative `media_type` is specified.
+
+```python
+from starlette.responses import TemplateResponse
+from starlette.requests import Request
+
+from jinja2 import Environment, FileSystemLoader
+
+
+env = Environment(loader=FileSystemLoader('templates'))
+
+
+class App:
+    def __init__(self, scope):
+        self.scope = scope
+
+    async def __call__(self, receive, send):
+        template = env.get_template('index.html')
+        context = {
+            'request': Request(self.scope),
+        }
+        response = TemplateResponse(template, context)
+        await response(receive, send)
+```
+
+The advantage with using `TemplateResponse` over `HTMLResponse` is that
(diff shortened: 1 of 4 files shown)


lovelydinosaur:
One other case where it'd be helpful to have the scope available: content negotiated responses.

We'd kinda like to be able to just `return Response(data)` and have a bunch of possible renderers deal with the content rendering, depending on the client's `Accept` header.

lovelydinosaur:
Resolution: Strictly require "request" to be part of the context.

This is important anyway in order to allow `url_for(...)` to function, and better that we raise an error early on in endpoint code if it's not included, rather than raising an error from template code calling into `url_for`.

I think this is probably a good enough compromise here.

lovelydinosaur:
Aside: Adding this to the documentation has been a big reminder of how much of the docs *really* ought to be updated to be from the perspective of documenting framework first, low-level interfaces second.

PR #273, opened 2018-12-14
Add BackgroundTasks for multiple jobs. Support GraphQL background tasks.

### LINKED ISSUE #180: Support for background tasks for `GraphQLApp`
Since Response provided by starlette take a background argument, Any plans to add support for running a list of background tasks after the response has been sent in the case of `GraphQLApp`?

A naive implementation looks like this

```python
class Holder(object):
    def __init__(self):
        self._list = []

    def add(self, function, *args, **kwargs):
        func = functools.partial(function, *args, **kwargs)
        self._list.append(func)

    def clean(self):
        self._list = []

    def run_all_tasks(self):
        for i in self._list:
            i()
```

```python
class EnhancedGraphqlApp(GraphQLApp):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.holder = Holder()
        
    async def execute(self, query, variables=None, operation_name=None):  # type: ignore
        print("Started")
        print(self.holder._list)
        if self.is_async:
            result = await self.schema.execute(
                query,
                variables=variables,
                operation_name=operation_name,
                context={"holder": self.holder},
                executor=self.executor,
                return_promise=True,
            )
        else:
            func = functools.partial(
                self.schema.execute,
                context={"holder": self.holder},
                variables=variables,
                operation_name=operation_name,
            )
            result = await loop.run_in_executor(None, func, query)
        return result

    def run_tasks(self):
        for i in self.holder._list:
            i()
        self.holder.clean()

    async def handle_graphql(self, request: Request) -> Response:
        # existing implementation
        ... 
        tasks = BackgroundTask(self.run_tasks)
        return JSONResponse(response_data, background=tasks, status_code=status_code)
```

And usage in the Schema class would be just like how context is used generally

```python
def hello(name):
    print(f"My name is {name}")

class Query(object):
    hello = graphene.String(name=graphene.String(default_value="stranger"))
    alls = graphene.List(AllModel)

    async def resolve_alls(self, info):
        rresult = await async_print_all_models()
        holder = info.context.get('holder')
        holder.add(hello,"James")
        return [AllModel(name=x.name) for x in rresult]
```

lovelydinosaur:
Closes #180

Adding background tasks from GraphQL:

```python
class Query(graphene.ObjectType):
    ...

    def resolve_some_field(self, info):
        ...
        background = info.context["background"]
        background.add_task(func, *args, **kwargs)
        ...
```

Adding multiple background tasks to a response:

```python
async def some_endpoint(request):
    tasks = BackgroundTasks()
    tasks.add_task(func, *args, **kwargs)
    tasks.add_task(func, *args, **kwargs)
    return JSONResponse(content, background=tasks)
```

### DIFF
--- a/docs/background.md
+++ b/docs/background.md
@@ -6,6 +6,8 @@ the response has been sent.
 
 ### Background Task
 
+Used to add a single background task to a response.
+
 Signature: `BackgroundTask(func, *args, **kwargs)`
 
 ```python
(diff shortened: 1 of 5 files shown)


austincollinpena:
[Related for adding background tasks with Ariadne graphql](https://github.com/encode/starlette/issues/957)

PR #276, opened 2018-12-16
Add security flags to session cookie

HarrySky:
Flags: Secure, HttpOnly and SameSite=lax

SameSite flag can be set to 2 values:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies

I am making it "lax" by default, but maybe it should be configurable somehow?

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -82,8 +82,8 @@ The following arguments are supported:
 * `secret_key` - Should be a random string.
 * `session_cookie` - Defaults to "session".
 * `max_age` - Session expiry time in seconds. Defaults to 2 weeks.
-* `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to "lax".
-* `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to False.
+* `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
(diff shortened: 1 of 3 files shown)


HarrySky:
Oh, I forgot that Secure header can only be used with HTTPS connections, that may be the reason why check failed

HarrySky:
Should I add something like `secure: bool = False` to `__init__` and then add Secure flag if it is `True`?

HarrySky:
@tomchristie can I add `# pragma: no cover` to line 30? I think we can't test that. Or should I change my code somehow?

lovelydinosaur:
You can test with: TestClient(base_url=“https://testserver”)

HarrySky:
@tomchristie added the test, local `./scripts/test` don't fail, but Travis check failed, what should I do? I did not find any CONTRIBUTING.md or something.

HarrySky:
Just run `./scripts/lint` and it changed a lot of files by just removing lines between import, should I commit it?

Example from `git diff`:
```
 import pytest
-
 from starlette import status
 from starlette.websockets import WebSocket, WebSocketDisconnect
```

HarrySky:
Now I get this problem
`ERROR: /home/travis/build/encode/starlette/starlette/middleware/sessions.py Imports are incorrectly sorted.`

But nothing changed, only one space in imports was removed.

UPDATE: Fixed it :)

HarrySky:
@tomchristie should I add or change something else in this pull request?

lovelydinosaur:
Lovely stuff, yup! Should be a new release out sometime this week.

PR #278, opened 2018-12-17
Handle auth error responses at the auth backend level

pvanliefland:
Use case: allow more flexible authentication errors response generation.

Examples:
* JSON error response
* 401 error code

Usage:
```python
class BasicAuth(AuthenticationBackend):
        try:
            some_auth_logic()
        except (AnException) as exc:
            raise AuthenticationError("Auth error message")

        return AuthCredentials(["authenticated"]), SimpleUser(username)

    def error_response(self, request: Request, exc: Exception) -> Response:
        return JSONResponse({"error": str(exc)}, status_code=401)
```

### DIFF
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -4,7 +4,7 @@ import typing
 
 from starlette.exceptions import HTTPException
 from starlette.requests import Request
-from starlette.responses import RedirectResponse, Response
+from starlette.responses import PlainTextResponse, RedirectResponse, Response
 
 
 def has_required_scope(request: Request, scopes: typing.Sequence[str]) -> bool:
@@ -56,6 +56,9 @@ class AuthenticationBackend:
         self, request: Request
     ) -> typing.Optional[typing.Tuple["AuthCredentials", "BaseUser"]]:
         raise NotImplemented()  # pragma: no cover
+
+    def error_response(self, request: Request, exc: Exception) -> Response:
+        return PlainTextResponse(str(exc), status_code=400)
 
 
 class AuthCredentials:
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Haven't quiet figured out for myself if we'd prefer to have this customizable on the middleware class or on the backend.

pvanliefland:
Well it depends... I was under the impression that for now, you can only:
* Add the `AuthenticationMiddleware` once
* Have a single `AuthenticationBackend` for your `AuthenticationMiddleware`

In this case, it makes sense that everything happens within your custom backend - having to subclass AuthenticationMiddleware and write a custom backend feels a bit overkill.

A `on_error` param would also work:

```python
def on_auth_error(request: Request, exc: Exception):
    return JSONResponse({"error": str(exc)}, status_code=401)

app.add_middleware(AuthenticationMiddleware, backend=BasicAuth(), on_error=on_auth_error)
```

What do you think? Let me know what you prefer, I can rework the PR...

lovelydinosaur:
Yeah lets go with that approach - sounds good to me.

pvanliefland:
Which one? The `error_response` method on the backend class, or the `on_error` parameter for the middleware?

lovelydinosaur:
An (optional) `on_error` for the middleware class.

pvanliefland:
See #281

haggen:
I'm starting now with Starlette and this seem counter intuitive to me. Why not cascade the exception back to the exception handler of the application? `401` is response status code after all.

I already have an `exception_handler` for the application that shows a friendly page with recourse to the user and it seems strange to me that I have to re-implement this and pass it along in a different place (middleware options), even though both handlers handle HTTP errors.

Specially confusing because one is async but the other is not and I have been pulling by hair with the "coroutine is not callable" error since I hadn't notice it and was using the same async handler for both the application exceptions and the authentication middleware exceptions.

Does that make sense or am I mixing things up in my head?

Thanks in advance.

PR #280, opened 2018-12-18
Add tests for fixed path handling and docs

tiangolo:
This PR tests and documents how fixed paths are matched when there are parameterized paths that also match.

---

Having two endpoints, one for:

```
/users/{username}
```

and one with a fixed path for:

```
/users/me
```

It is necessary to add the one for `/users/me` first, as paths are matched in order of addition to the router. And a URL with path `/users/me` would match `/users/{username}` first with a `username` of `me` if it was added before.

This PR adds tests for that and documents the behavior.

I also added a small route in the applications docs section to "hint" how it works.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -59,11 +59,7 @@ You should add that route for `/users/me` before the one for `/users/{username}`
 
 ```python
 app = Router([
-    Route('/', endpoint=Homepage, methods=['GET']),
-    Mount('/users', app=Router([
-        Route('/', endpoint=Users, methods=['GET', 'POST']),
-        Route('/users/me', endpoint=UserMe, methods=['GET'])
-        Route('/{username}', endpoint=User, methods=['GET']),
-    ]))
+    Route('/users/me', endpoint=UserMe, methods=['GET']),
+    Route('/{username}', endpoint=User, methods=['GET']),
 ])
 ```
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -43,3 +43,27 @@ app = Router([
 
 The router will respond with "404 Not found" or "405 Method not allowed"
 responses for requests which do not match.
+
+Incoming paths are matched against each `Route` in order.
+
+If you need to have a `Route` with a fixed path that would also match a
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -19,6 +19,16 @@ def user(request):
     return Response(content, media_type="text/plain")
 
 
+def user_me(request):
+    content = "User fixed me"
+    return Response(content, media_type="text/plain")
+
+
+def user_no_match(request):
```
lovelydinosaur: Tests are failing since these two lines are never reached by coverage.

I'd suggest: `raise NotImplementedError()  # pragma: no cover`

lovelydinosaur: Ah I see you're already ahead of me! :)

docs/routing.md
```diff
@@ -43,3 +43,27 @@ app = Router([
...
+You should add that route for `/users/me` before the one for `/users/{username}`:
+
+```python
+app = Router([
+    Route('/', endpoint=Homepage, methods=['GET']),
+    Mount('/users', app=Router([
+        Route('/', endpoint=Users, methods=['GET', 'POST']),
+        Route('/users/me', endpoint=UserMe, methods=['GET'])
+        Route('/{username}', endpoint=User, methods=['GET']),
+    ]))
```
lovelydinosaur: I'd suggest we slim this example down, so that:

* We don't bother with the `Mount`.
* We only include `/users/me` and `/users/{username}`.

tests/test_routing.py
tiangolo: Hehe :smile: 

Still, do you think it's more explicit / better to use a `raise NotImplementedError()` ?

docs/routing.md
tiangolo: Sure thing.

tests/test_routing.py
lovelydinosaur: I'd marginally prefer it, but not *that* fussed either way.

tiangolo:
Thanks! :tada:

PR #285, opened 2018-12-18
Add Third Parties section to docs with: Starlette APISpec, Starlette API, Responder and FastAPI

tiangolo:
This PR adds a section to the docs at the end with third-party packages.

I'm still not sure if "Third Parties" is the best name. Other options that come to mind:

* External Packages
* Plugins (even though 2 of those are actually frameworks, powered by Starlette)
* Third Party Packages (long, but more explicit)

Also, I don't know if I'm being too verbose in the docs.

---

@Woile and @PeRDy, do you have anything to comment there about Starlette APISpec and Starlette API? (I imagine @kennethreitz probably won't care too much about Responder's section).

### DIFF
--- a/docs/thirdparties.md
+++ b/docs/thirdparties.md
@@ -1,0 +1,86 @@
+
+Starlette has a rapidly growing community of developers, building tools that integrate into Starlette, tools that depend on Starlette, etc.
+
+Here are some of those third party packages:
+
+## Plugins
+
+### Starlette APISpec
+
+Link: <a href="https://github.com/Woile/starlette-apispec" target="_blank">https://github.com/Woile/starlette-apispec</a>
+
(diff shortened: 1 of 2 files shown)


perdy:
Thanks for your effort adding other libs. I don't have anything to comment, happy with it.

lovelydinosaur:
Nice work! 😄

tiangolo:
Awesome! Thanks!

woile:
Good job! Looks good.

Should we also add some sections about ORMs and about data serialization (like marshmallow & pydantic)?

lovelydinosaur:
I guess *at the moment* I'd probably consider that as out-of-scope, but I could probably go either way on it.

PR #287, opened 2018-12-21
Fix multi field forms parsing after already having read the body

tiangolo:
This implements tests and the fix for the following scenario:

* The ASGI app uses a `Request` object and calls the `.body()` method (e.g. to check if there is a body in the request).
* Then it calls the `.form()` method.
* And the form had more than one key-value pairs.

:warning: Currently, it would not get the last key-value pair.

---

After this PR, it also gets the last key-value pair.

I added 2 tests that fail with the current code.

The change is just a `yield b""`.

But, without it, the `FormParser`s don't know that the form contents are finished, and don't add the last key-value pair to the results.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -132,6 +132,7 @@ class Request(HTTPConnection):
     async def stream(self) -> typing.AsyncGenerator[bytes, None]:
         if hasattr(self, "_body"):
             yield self._body
+            yield b""
             return
 
         if self._stream_consumed:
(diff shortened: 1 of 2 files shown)


tiangolo:
Hey @tomchristie , sorry to bother, but have you been able to check this?

I guess it might have gone under the radar as it was done during the holidays.

Do you want me to update any of it?

lovelydinosaur:
Nice work, thanks!

tiangolo:
Awesome, thanks! :tada:

PR #292, opened 2018-12-26
Add contributing guidelines

### LINKED ISSUE #290: Contributing guidelines
Hi!

While figuring out a potential solution for #289, I had to setup the repo and install Starlette locally.

Although the process is rather straight-forward, I think it would be good practice to have contribution guidelines to make things clear and facilitate contributor onboarding.

I already went ahead and committed a [draft](https://github.com/encode/starlette/commit/df8805aa1ba26e46d90673a53e7029bc113380b7) to my fork based on the guidelines I wrote for Bocadillo. If it looks fine, I'd be happy to open a PR. :+1:

florimondmanca:
Fixes #290

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,18 @@ By default, tests involving a database are excluded. To include them, set the `S
 export STARLETTE_TEST_DATABASE="postgresql://localhost/starlette"
 ```
 
+## Linting
+
+We use [Black] as a code formatter. To run it along with a few other linting tools, we provide a stand-alone linting script:
+
+```bash
+./scripts/lint
+```
+
+If linting has anything to say about the code, it will format it in-place.
+
+To keep the code style consistent, you should apply linting before committing.
+
 ## Documentation
 
 The documentation is built with [MkDocs], a Markdown-based documentation site generator.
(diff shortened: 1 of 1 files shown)


florimondmanca:
@tomchristie I have added hints about the `test` and `lint` scripts.

An install script would be a good idea to make sure people use a correctly named virtualenv, etc.
That said, since this PR focuses on the `CONTRIBUTING.md` file, I don't think we should add it now. Plus it would probably need some extra work and debugging to make it fault-tolerant enough (looking at you, Windows?).

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+The Starlette team happily welcomes contributions. This document will help you get ready to contribute to Starlette!
+
+> **NOTE**: before writing any code, please first [open an issue] and discuss your ideas with the maintainers.
+
+## Setting up the repository
+
+1. Fork the repo.
+2. Clone your fork on your local computer: `git clone https://github.com/<username>/starlette.git`.
+3. Install Starlette locally and run the tests (see next sections).
+4. Create a branch for your work, e.g. `fix/some-bug`.
```
lovelydinosaur: I guess we should use the full `git checkout -b fix/some-bug` command here.  I don't particularly care about the two part 'fix/<name>' style, so perhaps just use `fix-some-bug` as the example branchname instead?

florimondmanca: Sure! I added a fix for this.

lovelydinosaur:
Fab let's go with this for now.
I'd be happy to accept a PR for a `./scripts/install`, if you fancy it:

* Ensure `python3` meets the minimum required version.
* Create a virtualenv in `./venv`, using -p python3.
* Include the script in the contributing notes.
* Include a note, that the instructions are currently suited to mac and linux, but that we'd happily take pull requests that either updated the instructions appropriately (I don't know windows users need to do in order to run shell scripts, if at all?) or to make the scripts more cross compatible. (If they were all python based for example.)

Issue #286, opened 2018-12-21
Support for AWS Lambda


rajeev:
I would like to add some sort of "driver" for AWS Lambda.

We want to use starlette to develop our API backend on AWS Lambda. At some later date, we might want to move off Lambda (depending on cost, etc). For that reason, we want to build the API in such a way that we can readily move to a full uvicorn+starlette stack with minimal changes.

I've been investigating the starlette for this purpose. I'm not a starlette expert. If one were to build a "driver" for AWS Lambda, how would that work? An outline for a good performant solution would be helpful so that we can implement the corresponding code in the best way possible

rajeev:
Closing this issue as it is irrelevant.

erichaus:
@rajeev I'm using [Mangum](https://github.com/jordaneremieff/mangum) and it's working out well...

PR #296, opened 2018-12-29
Support for AWS Lambda ALB event source

rajeev:
Allows us to write a Starlette app and use it transparently on AWS Lambda. This uses the existing Python 3.7 runtime. 

Also see: #286

### DIFF
--- a/starlette/aws.py
+++ b/starlette/aws.py
@@ -1,0 +1,53 @@
+import asyncio
+import base64
+import urllib.parse
+
+from starlette.types import Message
+from starlette.applications import Starlette
+
+
+class LambdaFunction(Starlette):
+
+    def lambda_handler(self, event, context):
+        connection_scope = {
+            'type': 'http',
+            'http_version': '1.1',
+            'scheme': 'http',
+            'method': event['httpMethod'],
+            'root_path': '',
+            'path': event['path'],
+            'query_string': urllib.parse.urlencode(event['queryStringParameters']),
+            'headers': event['headers'].items(),
+            'x-aws-lambda': {
+                'requestContext': event['requestContext'],
+                'lambdaContext': context
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


rajeev:
This is just a draft. Very basic testing has been done within Lambda. Need to come up with comprehensive test cases. Also need to consider using the Lambda Custom Runtimes to make better use of asyncio concurrency.

lovelydinosaur:
Amazing work. Almost certainly I’d like to see this end up as a third party package, rather than supported in the core library.

rajeev:
All right I will withdraw the Pull Request and setup a new project for that.

rajeev:
@tomchristie I'm not too familiar with the internals of asgi and starlette ... could I count on your suggestions and review from time to time? just so that I can be confident I'm doing the right thing?

lovelydinosaur:
Sure. Can you point me at any basic info for integrating Python with lambda from scratch?

rajeev:
@tomchristie 

Generally writing a Lambda function in Python, basically means providing a synchronous handler that returns a specially formatted Python dict.

This PR targets the simplest way of using asyncio in such a system: synchronous Lambda invocation handler calling a Python async view.

AWS just recently launched Custom Runtimes (which doesn't have much to do with Python per se) https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html

That's what I want to achieve ultimately, a custom runtime to make best of use of Lambda with asyncio concurrency. 

I will start a new package shortly and inform you here once it's setup. It would be really great if you would review it and provide comments.

rajeev:
Repo with the code has been published at https://github.com/alokinsoft/starlette-lambda

lovelydinosaur:
Great! Want to create a pull request adding it to the third party docs?... https://www.starlette.io/third-party-packages/

PR #299, opened 2019-01-02
include license file

carlodri:
as per discussion in #121


lovelydinosaur:
Looks good yup. Let's double check that this has worked as expected when we roll the next release.

PR #300, opened 2019-01-03
Install script

### LINKED ISSUE #298: Development install script
Installing Starlette for development is manual and some constraints or platform-specific cases may not be handled properly.

We'd need an install script to automate the process. After running it, contributors should be able to start working on Starlette right away.

Some guidelines listed by @tomchristie in #292:

- Ensure python3 meets the minimum required version.
- Create a virtualenv in ./venv, using -p python3.
- Include the script in the contributing notes.
- Include a note, that the instructions are currently suited to mac and linux, but that we'd happily take pull requests that either updated the instructions appropriately (I don't know windows users need to do in order to run shell scripts, if at all?) or to make the scripts more cross compatible. (If they were all python based for example.)

florimondmanca:
Fixes #298 

@tomchristie Open to suggestions for cleaner checking of the minimum Python version; I'm quite new to bash scripting and maybe there are secret solutions to comparing strings. :)

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,0 +1,30 @@
+#!/bin/sh -e
+
+MIN_VERSION="(3,6)"
+VERSION_OK=`python -c "import sys; print(sys.version_info[0:2] >= $MIN_VERSION and '1' or '');"`
+
+if [[ -z "$VERSION_OK" ]] ; then
+    PYTHON_VERSION=`python -c "import sys; print('%s.%s' % sys.version_info[0:2]);"`
+    DISP_MIN_VERSION=`python -c "print('%s.%s' % $MIN_VERSION)"`
+    echo "ERROR: Python $PYTHON_VERSION detected, but $DISP_MIN_VERSION+ is required."
+    echo "Please upgrade your Python distribution to install Starlette."
+    exit 1
+fi
+
+REQUIREMENTS="requirements.txt"
+VENV="venv"
+PIP="$VENV/bin/pip"
+
+function execute() {
+    echo "Executing $1…"
+    $1
+}
+
+execute "python -m venv $VENV"
+execute "$PIP install -r $REQUIREMENTS"
+execute "$PIP install -e ."
+
+echo "\nAll set! Activate your virtual environment using the command below."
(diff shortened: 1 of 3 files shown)


scripts/install
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh -e
+
+[[ $1 = "-p" ]] && PYTHON=$2 || PYTHON="python"
```
lovelydinosaur: I'd suggest we don't bother with this - just use "python".
If someone wants to install with some other version other than their system default then they can run virtualenv manually.
It's important that we keep the script simple so that it's as obvious as possible what it does.

scripts/install
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+REQUIREMENTS="requirements.txt"
+VENV="venv"
+PIP="$VENV/bin/pip"
+
+function execute() {
+    echo "Executing $1…"
+    $1
+}
```
lovelydinosaur: I'd replace this with `set -x`

scripts/install
florimondmanca: I hear your concern, but:

- I believe on macOS `python` generally refers to the system's 2.7 version, and I'd bet some Linux distros still do the same.
- It seems many Windows installs expose the Python interpreter as `py`.

If we just hardcode the `python` executable, the script is unusable for a lot of users which IMO defeats its purpose.

If the difficulty of understanding what this lines does is the problem, I can add a thoughtful comment.

lovelydinosaur: Okay.
Thinking about it twice, the default should be `python3` in any case.

florimondmanca: Oops, that's actually what I indicated in `CONTRIBUTING.md` but not in the script. Thanks, that's fixed.

scripts/install
florimondmanca: Didn't know about this! Fixed.

lovelydinosaur:
Thanks! I might make a few tweaks (*) to this, but looks good to pull in now!

(*) Will probably drop the "you can now activate your virtualenv stuff" - users don't *actually* have to do that to run the tests.

PR #302, opened 2019-01-03
Fix leaking Secret values in print and logging statements

### LINKED ISSUE #301: Secret objects leak value in print, logging, and format statements
According to the docs, `Secret` is meant to hide a value from tracebacks, logging, and format statements.

However, since the value is revealed by the `__str__` method, it will still appear in `print`, `logging`, and `format` statements.

```python
In [1]: from starlette.datastructures import Secret

In [2]: s = Secret('foo')

In [3]: print(s)
foo

In [4]: import logging

In [5]: logging.critical(s)
CRITICAL:root:foo

In [6]: format(s)
Out[6]: 'foo'
```

This happening on the latest `master` commit (b94f02e as of this writing).

One possible solution would be to hide the value in `__str__` and expose the value by accessing the `value` attribute.

```python
class Secret:
    """
    Holds a string value that should not be revealed in tracebacks etc.
    You should access the ``value`` attribute at the point it is required.
    """

    def __init__(self, value: str):
        self.value = value

    def __repr__(self) -> str:
        return "%s('**********')" % self.__class__.__name__

    __str__ = __repr__
```

sloria:
`print(Secret("value"))` will no longer print `"value"`.
To get the underlying value, you must access the
`.value` property.

I didn't bother changing the `URL` implementation to
do the same because it would break a lot of public API.

close #301

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -50,25 +50,27 @@ If none of those match, then `config(...)` will raise an error.
 For sensitive keys, the `Secret` class is useful, since it helps minimize
 occasions where the value it holds could leak out into tracebacks or logging.
 
-To get the value of a `Secret` instance, you must explicitly cast it to a string.
+To get the value of a `Secret` instance, you must access the `value` property.
 You should only do this at the point at which the value is used.
 
 ```python
 >>> from myproject import settings
+>>> settings.SECRET_KEY
+Secret('**********')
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Perhaps we should consider this for a future version, as it's more explicit, but let's update DatabaseURL too if we do so.

PR #303, opened 2019-01-03
Fix documentation for Secret and DatabaseURL

sloria:
see https://github.com/encode/starlette/issues/301#issuecomment-451186236


lovelydinosaur:
Thanks!

PR #307, opened 2019-01-07
Type of subprotocol is str, not List[str]

ciscorn:
The ASGI spec says:

- `type`: `websocket.accept`
- `subprotocol`: The subprotocol the server wishes to accept, as a **unicode string**. Optional, defaults to `None`.


lovelydinosaur:
Incorperated into #308 - thanks!

PR #308, opened 2019-01-07
type of subprotocol is str, not List[str]

lovelydinosaur:


lovelydinosaur:
Thanks @ciscorn - Needed to deal with some unrelated issues in order to get the build passing on this.

PR #311, opened 2019-01-10
Improve authentication.requires() to support class based endpoints

steinnes:
I ran into an issue on my API where the OpenAPI docstrings for a class-based endpoint which required authentication weren't appearing in the schema.

Upon investigation I noticed that `starlette.authentication.requires` was returning a wrapper function for the classes, which made the openapi schema generator unable to detect the class and inspect the methods for docstrings.

This was the most simple way I could think of to fix the issue.

### DIFF
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import inspect
 import typing
 
 from starlette.exceptions import HTTPException
@@ -22,8 +23,17 @@ def requires(
     scopes_list = [scopes] if isinstance(scopes, str) else list(scopes)
 
     def decorator(func: typing.Callable) -> typing.Callable:
-        if asyncio.iscoroutinefunction(func):
+        if inspect.isclass(func):
+            class Wrapper(func):
+                async def dispatch(self, request: Request) -> Response:
+                    if not has_required_scope(request, scopes_list):
+                        if redirect is not None:
+                            return RedirectResponse(url=request.url_for(redirect))
+                        raise HTTPException(status_code=status_code)
+                    return await super().dispatch(request)
+            return Wrapper
 
+        if asyncio.iscoroutinefunction(func):
             @functools.wraps(func)
             async def wrapper(request: Request) -> Response:
                 if not has_required_scope(request, scopes_list):
--- a/tests/test_authentication.py
+++ b/tests/test_authentication.py
@@ -9,10 +9,12 @@ from starlette.authentication import (
     SimpleUser,
     requires,
 )
+from starlette.endpoints import HTTPEndpoint
 from starlette.middleware.authentication import AuthenticationMiddleware
 from starlette.requests import Request
 from starlette.responses import JSONResponse
 from starlette.testclient import TestClient
+from starlette.schemas import SchemaGenerator
 
 
 class BasicAuth(AuthenticationBackend):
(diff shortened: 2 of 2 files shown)


lovelydinosaur:
Okay, I'm not sure about this.
An alternative might be that we recommed this style: https://docs.djangoproject.com/en/2.1/topics/class-based-views/intro/#decorating-the-class

lovelydinosaur:
So...

```python
class MyEndpoint(HTTPEndpoint):
    @required("authenticated")
    def dispatch(self, request):
        return super().dispatch(request)
```

Does that work in the way I'd expect?

steinnes:
I think both of those approaches would work, and would alleviate any issues with the route name (https://github.com/encode/starlette/blob/master/starlette/routing.py#L72).

I'm going to try them and and update here.  I think overriding the dispatch method would be enough for me, almost feel silly not to have thought of it :-)

steinnes:
Overloading the dispatch method and wrapping it in `requires` causes the wrapper to get an unexpected positional argument (`self`), so in order for that to work I think we'd need to use something like `method_decorator` (assuming that's a wrapper to "fix" the signature of another wrapper, I haven't looked too deeply into it yet).

steinnes:
This is a deeper rabbit hole than I thought, I'm closing this and solving my own issues in my own project like this:
```python
from starlette.authentication import has_required_scope

class AuthenticatedEndpoint(HTTPEndpoint):
    required_scopes = []
    status_code = 401

    def dispatch(self, request: Request) -> Response:
        if not has_required_scope(request, self.required_scopes):
            raise HTTPException(status_code=self.status_code)
        return super().dispatch(request)
```

lovelydinosaur:
See #312

steinnes:
I can use #312  for sure, thanks @tomchristie 👍

HarrySky:
@tomchristie unrelated to issue, are you planning to release new version to pip soon?

lovelydinosaur:
@HarrySky Yeah I should think so. Need to review what's going in the release notes. You waiting on something in particular?

HarrySky:
@tomchristie No, just waiting for new features, no hurry :smile:

steinnes:
Since we're discussing tangentially related things here, what do you think about allowing `requires` to receive (and pass on) `**kwargs`?  It would allow for more sugary view implementations, a la:

```python
@app.route("/users/{user_id}")
def get_user(request, user_id):
  ...
```

I even tried it out by extending Starlette and implementing my own `route` decorator method, although issues with other parts of starlette where the view method signature is expected to be fixed (such as in the case of `@requires`) has made me wary of relying on it.

```python
class MyAPI(Starlette):
...
    def route(self, *args, **kwargs):
        """ Overload route to pass request path parameters as kwargs """
        real_route_decorator = super().route(*args, **kwargs)

        def inner(fn):
            if inspect.isclass(fn):
                class Wrapper(fn):
                    async def dispatch(self, req):
                        if hasattr(self, 'before_request'):
                            self.before_request(**req.path_params)
                        return await super().dispatch(req)

                return real_route_decorator(Wrapper)

            is_coroutine = asyncio.iscoroutinefunction(fn)

            @wraps(fn)
            async def wrapper(req, *args, **kwargs):
                kwargs.update(req.path_params)
                if is_coroutine:
                    return await fn(req, *args, **kwargs)
                else:
                    return fn(req, *args, **kwargs)
            return real_route_decorator(wrapper)
        return inner
```
(this obviously has the same DNA as the changes to `requires` I proposed in this PR.. explains why that was the "most simple way" I could think of)

lovelydinosaur:
> what do you think about allowing requires to receive (and pass on) **kwargs

Possibly, but it gets a bit complicated having to catch the different cases. ('request' could be a keyword argument here too.)

Actually thinking about it, that'd be a problem for this latest change too.

lovelydinosaur:
@steinnes ---> #314

PR #312, opened 2019-01-10
Support `@requires` as a method decorator

lovelydinosaur:
Allows this...

```python
@app.route("/dashboard")
class Dashboard(HTTPEndpoint):
    @requires("authenticated")
    async def get(self, request):
        ...
```

Or this...

```python
@app.route("/dashboard")
class Dashboard(HTTPEndpoint):
    @requires("authenticated")
    async def dispatch(self, request):
        await super(Dashboard, self).dispatch(request)
```

### DIFF
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -24,8 +24,9 @@ def requires(
 
     def decorator(func: typing.Callable) -> typing.Callable:
         if asyncio.iscoroutinefunction(func):
+
             @functools.wraps(func)
-            async def wrapper(*args) -> Response:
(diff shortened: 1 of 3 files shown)


starlette/authentication.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import inspect
```
steinnes, line 3: Remnant from something else?

lovelydinosaur: Yes, thanks!
I thought our linting picked those up but I guess not.

PR #314, opened 2019-01-10
Flexible requires decorator

lovelydinosaur:
`@requires` can now be applied to any function or method, so long as it includes a `request` argument.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,6 +12,7 @@ ujson
 asyncpg
 
 # Testing
+autoflake
 black
 codecov
 isort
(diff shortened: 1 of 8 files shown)


steinnes:
This is awesome!  I'll test this today with my path params -> kwargs wrapper :-)

steinnes:
.... and now the `inspect` import just removed is necessary again 😅

lovelydinosaur:
Note that `request` *has* to be called exactly that. No `req`, `r`, or any of that nonsense.
I think that's a reasonable constraint tho.

PR #316, opened 2019-01-12
Transfer JSON as text over WebSockets

ciscorn:
I think we usually send JSON data as a text, like `ws.send(JSON.stringify(...))`.


lovelydinosaur:
Okay. It’s less correct, but probably more pragmatic.
May be some finer details to consider here, also we’ll need to figure out how best to deal with this as from the POV of not breaking existing systems.

PR #319, opened 2019-01-14
Ignore database coverage if STARLETTE_TEST_DATABASE is unset

lovelydinosaur:


starlette/requests.py
```diff
@@ -84,7 +84,7 @@ def session(self) -> dict:
         return self._scope["session"]
 
     @property
-    def database(self) -> typing.Any:
+    def database(self) -> typing.Any:  # pragma: no cover
```
blueyed, line 87: Why the "pragma: no cover"?
Shouldn't it be covered by Travis and be good to know in general?

lovelydinosaur: we Exclude the database tests if they’re not setup, but we can’t conditionally exclude this line, so we just do so unconditionally. (Then running the tests without setting up a Postgres DB will skip the DB tests, but still check for full 100% coverage of everything else)

blueyed: Is it really more important to have 100% coverage being reported without a full test setup, than missing something crucial being covered with a full test setup?

lovelydinosaur: The method here is trivial, and there’s no feasible way we’d miss coverage of it, given that the tests enforce coverage onto every *other* aspect of the database code. So on balance here, yes. It’s really valuable that contributors can know that they’ve maintained coverage, and not everyone should have to setup a Postgres database in order to determine that.

PR #320, opened 2019-01-14
Add `request.is_disconnected()`

### LINKED ISSUE #297: Detect/handle closed client connections
It appears like starlette does not cancel inner processing for when a client disconnects.

With the following code it will keep on fetching from the remote, although the
client has disconnected already, e.g. when cancelling a curl call.

```python
@app.route(…)
async def foo(request):
    …
    if content_type.startswith("video/"):
        bytes_total = req.headers["Content-Length"]
        logger.info("Streaming {} bytes...", bytes_total)

        async def generator(bytes_total):
            import aiohttp
            async with aiohttp.ClientSession() as session:
                chunk = 0
                chunk_size = 102400
                bytes_read = 0
                async with session.request(method, str(url), chunked=chunk_size) as resp:
                    async for data, end_of_http_chunk in resp.content.iter_chunks():
                        chunk += 1
                        bytes_read += chunk_size
                        logger.info("chunk {} {}/{}", chunk, bytes_read, bytes_total)
                        yield data
        return StreamingResponse(generator(bytes_total), media_type=content_type)
```

It might be useful if the handler could be notified of this, in case it wants
to handle this, but in general I think a late middleware, responsible for
sending could handle this maybe?!

lovelydinosaur:
Determine if the client has disconnected: `if await request.is_disconnected(): ...`

Closes #297

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -1,3 +1,4 @@
+import asyncio
 import http.cookies
 import json
 import typing
@@ -120,6 +121,7 @@ class Request(HTTPConnection):
         assert scope["type"] == "http"
         self._receive = receive
         self._stream_consumed = False
+        self._is_disconnected = False
 
     @property
     def method(self) -> str:
@@ -148,6 +150,7 @@ class Request(HTTPConnection):
                 if not message.get("more_body", False):
                     break
             elif message["type"] == "http.disconnect":
+                self._is_disconnected = True
                 raise ClientDisconnect()
         yield b""
 
@@ -187,3 +190,15 @@ class Request(HTTPConnection):
             for item in self._form.values():
                 if hasattr(item, "close"):
                     await item.close()  # type: ignore
+
+    async def is_disconnected(self) -> bool:
+        if not self._is_disconnected:
+            try:
+                message = await asyncio.wait_for(self._receive(), timeout=0.0000001)
+            except asyncio.TimeoutError as exc:
+                message = {}
+
+            if message.get("type") == "http.disconnect":
+                self._is_disconnected = True
+
+        return self._is_disconnected
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -129,6 +129,13 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
         }
 
         async def receive() -> Message:
+            nonlocal request_complete, response_complete
+
+            if request_complete:
+                while not response_complete:
+                    await asyncio.sleep(0.0001)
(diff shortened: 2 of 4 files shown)


starlette/testclient.py
```diff
@@ -129,6 +129,13 @@ def send(  # type: ignore
         }
 
         async def receive() -> Message:
+            nonlocal request_complete, response_complete
+
+            if request_complete:
+                while not response_complete:
+                    await asyncio.sleep(0.0001)
```
florimondmanca: Any hint as to why this `while` loop is required since entering this block ensures that `request_complete` is true? :)

florimondmanca:
I like this. The `request.is_disconnected()` API is clean and lets the user define what they want to do with it.

I'm just wondering about calling this function periodically and potentially very frequently. Consider the streaming response use case, i.e. calling `request.is_disconnected()` to break the generator when the client has disconnected:

```python
@app.route('/foo')
async def foo(request):
    async def chunks(num):
        for i in range(num):
            if await request.is_disconnected():
               break
            yield f"chunk #{i}"
    return StreamingResponse(chunks(1_000_000))
```

Is there any performance issue to be expected in this context? I suppose the 0.1µs timeout was chosen to avoid any overhead.

starlette/testclient.py
florimondmanca: Ignore this comment, I didn't see `request_complete` vs `*response*_complete`. 🤦‍♂️

starlette/requests.py
```diff
@@ -187,3 +190,15 @@ def receive(self) -> Receive:
             for item in self._form.values():
                 if hasattr(item, "close"):
                     await item.close()  # type: ignore
+
+    async def is_disconnected(self) -> bool:
+        if not self._is_disconnected:
+            try:
+                message = await asyncio.wait_for(self._receive(), timeout=0.0000001)
```
blueyed, line 197: In https://github.com/encode/starlette/issues/297#issuecomment-454024709 `timeout=0` was used.
What's the difference?

And what about using just self._receive.disconnected?  (https://github.com/encode/starlette/issues/297#issuecomment-451089517)

lovelydinosaur: > In #297 (comment) timeout=0 was used. What's the difference?

Turns out that doesn't actually work (according to the tests). Needs to be non-zero.

> And what about using just self._receive.disconnected?

That's an implementation detail that'll only happen to work with uvicorn, in current versions. The API contract between the ASGI server and the client is just "recieve is an awaitable that returns messages".

PR #322, opened 2019-01-15
Customisable debug generator

florimondmanca:
This PR generalises the `DebugGenerator`: instead of using hard-coded HTML templates and CSS, they are now defined as class attributes. The debug generator class is also made customisable in the same fashion on `ServerErrorMiddleware`.

The idea is that, from an ASGI toolkit perspective, the `ServerErrorMiddleware` and the `DebugGenerator` are useful (although undocumented) components that help with implementing traceback responses in debug mode.

Let me know if this fits Starlette's scope at all.

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -18,7 +18,7 @@ STYLES = """\
 
 TEMPLATE = """
     <style type='text/css'>{style}</style>
-    <title>Starlette Debugger</title>
+    <title>{title}</title>
     <h1>500 Server Error</h1>
     <h2>{error}</h2>
     <div class='traceback-container'>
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Okay, so...

Reviewing this PR, my first thought was "I'd rather see any configuration be on the middleware class itself". Which then led to looking again at "why do we even have two seperate classes here?".

Upshot is: I'm not sure yet exactly what set of customizations I think are worth exposing, but in either case I think that there's a necessary precursor to *any* of them, which is:

Drop the current style of:

```python
debug_gen = DebugGenerator(exc)
debug_gen.generate_html()
```

And instead just refactor it down to:

```python
self.generate_html(exc)
```

(Note that we don't want to store state on the class inside `generate_html` or anything it calls into, but there's no need to do so.)

With that in mind I'm going to close this off for now, pending that refactor. (You're welcome to take it on if you'd like?)

Once that's complete we'll be in a better position to assess if and how we address style customization on the middleware class.

PR #323, opened 2019-01-15
Merge DebugGenerator into ServerErrorMiddleware

florimondmanca:
Following up on #322: refactored the debug generation code to all happen in `ServerErrorMiddleware`.

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -37,36 +37,6 @@ FRAME_TEMPLATE = """
 """
 
 
-class DebugGenerator:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Great stuff! That's a nice improvement right there.

Next up to consider: Do we really need a seperate "FRAME_TEMPLATE"? Could we pass `traceback` to the template, and iterate through the frames in `TEMPLATE` itself? I may take a bit of a look at this myself too.

lovelydinosaur:
Ack sorry, forgetting we're not in `jinja` land here.

PR #326, opened 2019-01-18
Set Max-Age on Session Cookies

pvanliefland:
As of now, no `Max-Age` or `Expires` value is set when sending the session cookie.

I think it makes sense to use `max_age` both for signing session data and for the cookie max age.

WDYT?


lovelydinosaur:
Great stuff, yup!

Issue #325, opened 2019-01-17
SchemaGenerator should consider route 'methods' for endpoint


DrPyser:
In `BaseSchemaGenerator.get_endpoints`, when the endpoint being inspected is an endpoint class, an `EndpointInfo` is generated for any method of that class matching an http method name, even if the route also specifies the set allowed methods. 
Example:
```
app.route("/endpoint", methods=["POST"])
class Endpoint(HTTPEndpoint):
    def post(self, req):
        """
        ...
        """
        ...
    def get(self):
        """
        ...
        """
        ...
```
The schema generated will include both methods for the `/endpoint` path, even if only `POST` is allowed for the route. I don't think this is valid behavior.

Consider my situation, where I want to create two different routes using the same endpoint class, but with different allowed methods:
```
@app.route("/users", methods=["POST"])
@app.route("/users/{username}", methods=["DELETE"])
class Users(endpoints.HTTPEndpoint):
    def delete(self, req, username):
        """
        ...
        """
        ...
    def post(self, req):
        """
        ...
        """
        ...
```
I don't know if this would be considered good practice or appropriate usage of starlette. If so, the generated schema will erroneously include entries for both methods on both paths(`/users` and `/users/{username}`).

I think the fix simply involves checking the `route.methods` attribute to generate `EndpointInfo` in `BaseSchemaGenerator.get_endpoints`:
https://github.com/encode/starlette/blob/2f51a38266473a555b1a25633896c150090a1c4f/starlette/schemas.py#L61-L67
``` 
                allowed_methods = route.methods if route.methods is not None else \
                  ["get", "post", "put", "patch", "delete", "options"]
                for method in allowed_methods:
                    if not hasattr(route.endpoint, method):
                        continue
                    func = getattr(route.endpoint, method)
                    endpoints_info.append(
                        EndpointInfo(route.path, method.lower(), func)
                    )
```

Thanks for the great work!

lovelydinosaur:
Seems valid yup. Would you be up for creating a pull request addressing this issue?

DrPyser:
Sure.

lovelydinosaur:
Closing this as stale.

Not sure if it's still an issue or not. If it is, it'd need someone to be invested enough to deal with it.

PR #328, opened 2019-01-19
Check route.methods to generate schema for class endpoints

DrPyser:
See #325 .


lovelydinosaur:
Needs scripts/lint to be run.

Issue #187, opened 2018-11-07
Support for python 3.5.


manatlan:
Hi ...

It seems that starlette is py>=3.6 .... so it doesn't work with python3.5, and so raspberry/raspbian (py3.5.3) ;-(
Is it just the f-string which block that ? or it's anything else ?

If it's just fstring : I can propose patchs ...

Thanks in advance

extraymond:
Will this be possible by writing a instance with __aiter__ and __anext__?

I'm not sure if functionally wised they are the same as async generator though.

lovelydinosaur:
@extraymond Yes I believe so. That'd be the fix.

lovelydinosaur:
If someone wanted to help progress this then the sensible thing to do would be:

* Create a pull request that adds python 3.5 to the travis builds.
* Either comment out any code that's a syntax error in 3.5, or add test skip markers, until all the remaining tests are passing.

That'll then isolate for us *exactly* what code would need alteration in order to work with 3.5, and we could take next steps from there.

extraymond:
@tomchristie  Thx for the tips!

After fiddling around with pypy3.5, I found the following todo-items for pypy compatible.

**1. yield syntax inside starlette.request or such**

1. Replace yield with __aiter__, __anext__ in a  class or use [async_generator package](https://pypi.org/project/async_generator/) to maintain minimal changes.
2. If using async_generator, output type is changing from typing.AsyncGenerator to the async_generator one. Since starlette.request is used in all majority of starlette, this breaks a lot of type checking stuff.

I'm not sure whether its good idea to use class based async_generator, or just use the package to maintain minimal code changes. Would love to know what's your idea on solving this.

**2. typing.Namedtuple behavior in 3.5**

For **starlette.schemas.EndpointInfo**, it should be declared with 

```python
EndpointInfo = typing.NamedTuple('EndpointInfo',
    [
     ('path', str),
     ('http_method', str),
     ('func', typing.Callable)
    ])
```

lovelydinosaur:
> Would love to know what's your idea on solving this.

I wouldn't want the additional dependency there.

(So: Implement it as a class)

lovelydinosaur:
> Replace yield with aiter, aexit in a class

Okay, I guess the next thing would be giving a brief example of how that might look (even if it's not completely implemented, and is just a code snippet on this thread, that'd still be useful)

extraymond:
Example for starlette.request.Request.stream

```python
class async_stream:

[...]

PR #334, opened 2019-01-21
f-strings everywhere

taoufik07:
except https://github.com/encode/starlette/blob/master/starlette/middleware/sessions.py#L58 and https://github.com/encode/starlette/blob/master/starlette/middleware/sessions.py#L68

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -28,16 +28,16 @@ class URL:
                     break
 
             if host_header is not None:
-                url = "%s://%s%s" % (scheme, host_header, path)
+                url = f"{scheme}://{host_header}{path}"
             elif server is None:
                 url = path
             else:
                 host, port = server
                 default_port = {"http": 80, "https": 443, "ws": 80, "wss": 443}[scheme]
                 if port == default_port:
-                    url = "%s://%s%s" % (scheme, host, path)
+                    url = f"{scheme}://{host}{path}"
                 else:
-                    url = "%s://%s:%s%s" % (scheme, host, port, path)
+                    url = f"{scheme}://{host}:{port}{path}"
 
             if query_string:
                 url += "?" + query_string.decode()
@@ -115,8 +115,8 @@ class URL:
             if username is not None:
                 userpass = username
                 if password is not None:
-                    userpass += ":%s" % password
-                netloc = "%s@%s" % (userpass, netloc)
+                    userpass += f":{password}"
+                netloc = f"{userpass}@{netloc}"
 
             kwargs["netloc"] = netloc
 
@@ -133,7 +133,7 @@ class URL:
         url = str(self)
         if self.password:
             url = str(self.replace(password="********"))
-        return "%s(%s)" % (self.__class__.__name__, repr(url))
+        return f"{self.__class__.__name__}({repr(url)})"
 
 
 class DatabaseURL(URL):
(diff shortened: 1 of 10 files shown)


starlette/middleware/wsgi.py
```diff
@@ -43,7 +43,7 @@ def build_environ(scope: Scope, body: bytes) -> dict:
         elif name == "content-type":
             corrected_name = "CONTENT_TYPE"
         else:
-            corrected_name = "HTTP_%s" % name.upper().replace("-", "_")
+            corrected_name = f"HTTP_{name.upper().replace("-", "_")}"
```
lovelydinosaur: This isn't valid syntax.  (See test failure... https://travis-ci.org/encode/starlette/jobs/482660913)

starlette/datastructures.py
```diff
@@ -115,8 +115,8 @@ def replace(self, **kwargs: typing.Any) -> "URL":
             if username is not None:
```
lovelydinosaur, line 115: Missing `netloc += ":%d" % port`

starlette/routing.py
```diff
@@ -103,16 +103,16 @@ def compile_path(
...
+        ), f"Unknown path convertor '{convertor_type}'"
         convertor = CONVERTOR_TYPES[convertor_type]
 
         path_regex += path[idx : match.start()]
-        path_regex += "(?P<%s>%s)" % (param_name, convertor.regex)
+        path_regex += f"(?P<{param_name}>{convertor.regex})"
 
         path_format += path[idx : match.start()]
-        path_format += "{%s}" % param_name
+        path_format += f"{{{param_name}}}"
```
lovelydinosaur: This ends up being non-obvious. I'd probably rather we use `"{" + param_name + "}"` or `netloc += ":%d" % port`

starlette/middleware/wsgi.py
```diff
@@ -43,7 +43,7 @@ def build_environ(scope: Scope, body: bytes) -> dict:
         elif name == "content-type":
             corrected_name = "CONTENT_TYPE"
         else:
-            corrected_name = "HTTP_%s" % name.upper().replace("-", "_")
+            corrected_name = f"HTTP_{name.upper().replace('-', '_')}"
```
lovelydinosaur: Let's perhaps pull `.upper().replace('-', '_')` out of the f-string entirely. So we have:

`f"HTTP_{name}".upper().replace("-", "_")`

I think that ends up being cleaner. (I don't really want us to use expressions inside f-strings if we can avoid it.)

lovelydinosaur:
Does also mean we'd have to pass on https://github.com/encode/starlette/issues/187

lovelydinosaur:
Which, yeah, I've currently said we should probably do anyway, but I *do* still like the idea of being able to run starlette on a raspberry pi easily, since it's a great target for it.

em92:
> Does also mean we'd have to pass on #187

> I do still like the idea of being able to run starlette on a raspberry pi easily

I suggest to create separate python3.5 branch or repository for that. There you can revoke those async generators without decreasing readability or adding extra dependencies.

em92:
> it's too difficult to keep in sync.

Merging from master and rebasing to exclude/modify incompatible commits for python 3.5 does not seem to be difficult. I didn't try in practice myself, but I have seen project, that was doing this. As example: https://github.com/brugal/wolfcamql/ merges commits from https://github.com/ioquake/ioq3 (See commit messages containing "ioquake3 patches")

lovelydinosaur:
Let's have this.
Great stuff, thanks! :)

PR #335, opened 2019-01-22
Fix typo in scope check for 'websocket'

tchan09:
Noticed this typo when tying `AuthenticationMiddleware` with websockets.

### DIFF
--- a/starlette/middleware/authentication.py
+++ b/starlette/middleware/authentication.py
@@ -7,7 +7,7 @@ from starlette.authentication import (
     AuthenticationError,
     UnauthenticatedUser,
 )
-from starlette.requests import Request
+from starlette.requests import HTTPConnection
 from starlette.responses import PlainTextResponse, Response
 from starlette.types import ASGIApp, ASGIInstance, Receive, Scope, Send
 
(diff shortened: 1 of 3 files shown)


tchan09:
Actually, hold up. I think I need to patch `Request` as well. Let me test the full flow first.

starlette/middleware/authentication.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         )  # type: typing.Callable[[Request, AuthenticationError], Response]
 
     def __call__(self, scope: Scope) -> ASGIInstance:
-        if scope["type"] in ["http", "websockets"]:
+        if scope["type"] in ["http", "websocket"]:
```
lovelydinosaur, line 31: Yup.

starlette/requests.py
```diff
@@ -118,7 +118,7 @@ def url_for(self, name: str, **path_params: typing.Any) -> str:
 class Request(HTTPConnection):
     def __init__(self, scope: Scope, receive: Receive = empty_receive):
         super().__init__(scope)
-        assert scope["type"] == "http"
+        assert scope["type"] in ("http", "websocket")
```
lovelydinosaur: We don't want this. Instead we should change `AuthenticationMiddleware.asgi` to use `conn = HTTPConnection(scope)` and `auth_result = await self.backend.authenticate(conn)`.

Signatures in `starlette/authentication.py` should also change to use the `HTTPConnection`base class instead of `Request`.

lovelydinosaur: That way the authentication will be suitable for either HTTP requests or WebSocket connections.

tchan09: Okay I see your point about using the parent `HTTPConnection` class and I've made that change.

What are your thoughts about the default error handler? It calls `PlainTextResponse` and that seems to be a http specific response (it inherits from `Response` and has message type `http.response.start`).

Is there an error response that can work for both http and websocket? I reckon for a websocket connection, the socket should be closed at this point.

Issue #172, opened 2018-11-01
Improvements to `SchemaGenerator`


lovelydinosaur:
* Include mounted paths.
* Ensure that OpenAPI path parameters in doctsrings match up with the router path parameters.
* Automatically add any path parameters that are not already included?

There's also a stack of really nice third party work that we can do here, eg packaging up Swagger UI or Redoc or API Star into API docs apps.

Contributor work towards any of the above would be very much welcome! 😄

tiangolo:
First, Starlette source code is a pleasure to read. 
It's been a long time since I enjoyed reading an already established project's source code like this one. Awesome  :clap: :cake: :taco: 

---

Here are my plans.

I imagine you would prefer me to build it as a separate project to not bloat Starlette, to keep it as pure ASGI as possible and to avoid putting very interdependent stuff together here.

But if you think it would be acceptable/desirable to have it integrated into Starlette, my first option would be, of course, to create a huge PR / set of PRs implementing it as part of Starlette.

---

Here's what I want to build.

First, my main point/concern is to achieve a specific set of features, that overlap a lot with what you did in APIStar. 

I'm actually starting from the development experience I want to have, and then finding, mixing and creating the tools that would provide it.

It want it to have this:

* Support for declaring OpenAPI 3, with all/most of its attributes. Not necessarily implementing all the logic, but allowing the schema to be declared. Actually, my plan is to build it around OpenAPI, using its keywords, etc.
* Very simple and intuitive code API.
* Great editor support. Autocomplete and type checking everywhere. For example, I want the endpoint route decorator to have explicit default declaration of `name=None, description=None, **kwargs`, etc. instead of just `**kwargs`. So that when I'm creating a route, I can get autocompletion and suggestions of the parameters I need to put instead of having to go and find the docs for that function to see what I can pass (part of having simple and intuitive API).
* Have code / class / type based declarations. Not strings with a subformat (as happens with YAML in docstrings). I want to have very much like what APIStar was doing (I think it was brilliant).
* Have data type declarations for path parameters, query parameters, headers, and body. For that, I want to use [Pydantic](https://pydantic-docs.helpmanual.io/), as it is based on the same Python type hints. So, whenever I get a value passed, it has the correct type inference, completion, etc. And Pydantic already has validation and serialization.
[...]

PR #336, opened 2019-01-22
Improvements to schema generation

JoseKilo:
Three changes here:

-  Include mounted paths in schemas (part of #172) 
- Allow regular docstrings before schema definitions 
- Don't include empty definitions in the schemas

Second one is to support something like this:

```python
@app.route("/regular-docstring")
def regular_docstring(request):
    """
    This a regular docstring example (not included in schema)

    ---

    responses:
      200:
        description: This is included in the schema.
    """
    pass  # pragma: no cover
```

Which splits regular docstring from the schema definition using the yaml start marker: `---`.

Third one is to avoid errors in swagger-ui when an endpoint doesn't have any docstrings.

Happy to split / amend if needed

### DIFF
--- a/starlette/schemas.py
+++ b/starlette/schemas.py
@@ -81,14 +81,6 @@ class BaseSchemaGenerator:
 
         return endpoints_info
 
-    def _extract_yaml(self, docstring: str) -> str:
-        """
-        We support having regular docstrings before the schema definition.
-
-        Here we return just the schema part from a docstring.
-        """
-        return docstring.split("---")[-1]
-
     def parse_docstring(self, func_or_method: typing.Callable) -> typing.Optional[dict]:
         """
         Given a function, parse the docstring as YAML and return a dictionary of info.
@@ -97,7 +89,10 @@ class BaseSchemaGenerator:
         if not docstring:
             return None
 
-        docstring = self._extract_yaml(docstring)
+        # We support having regular docstrings before the schema
+        # definition. Here we return just the schema part from
+        # the docstring.
+        docstring = docstring.split("---")[-1]
 
         parsed = yaml.safe_load(docstring)
 
--- a/starlette/schemas.py
+++ b/starlette/schemas.py
@@ -81,13 +81,13 @@ class BaseSchemaGenerator:
 
         return endpoints_info
 
-    def parse_docstring(self, func_or_method: typing.Callable) -> typing.Optional[dict]:
+    def parse_docstring(self, func_or_method: typing.Callable) -> dict:
         """
         Given a function, parse the docstring as YAML and return a dictionary of info.
         """
         docstring = func_or_method.__doc__
         if not docstring:
-            return None
+            return {}
 
         # We support having regular docstrings before the schema
         # definition. Here we return just the schema part from
@@ -99,7 +99,7 @@ class BaseSchemaGenerator:
         if not isinstance(parsed, dict):
             # A regular docstring (not yaml formatted) can return
             # a simple string here, which wouldn't follow the schema.
-            return None
+            return {}
 
         return parsed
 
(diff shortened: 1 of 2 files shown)


starlette/schemas.py
```diff
@@ -1,8 +1,9 @@
 import inspect
 import typing
+from dataclasses import dataclass
```
lovelydinosaur: I'd rather not introduce this here.

JoseKilo: removed

starlette/schemas.py
```diff
@@ -65,14 +77,40 @@ def get_endpoints(
...
         return endpoints_info
 
-    def parse_docstring(self, func_or_method: typing.Callable) -> dict:
+    def _extract_yaml(self, docstring: str) -> str:
+        """
+        We support having regular docstrings before the schema definition.
+
+        Here we return just the schema part from a docstring.
+        """
+        yaml_docs = docstring.split("---")
```
lovelydinosaur: Can we just do `docstring.split("---")[-1]` ?

tests/test_schemas.py
```diff
@@ -1,73 +1,174 @@
...
+
 from starlette.applications import Starlette
 from starlette.endpoints import HTTPEndpoint
 from starlette.schemas import OpenAPIResponse, SchemaGenerator
 from starlette.testclient import TestClient
 
-app = Starlette()
-app.schema_generator = SchemaGenerator(
-    {"openapi": "3.0.0", "info": {"title": "Example API", "version": "1.0"}}
-)
```
lovelydinosaur: It'd be nice if we could keep a smaller change-footprint. It's hard to accurately review, with a big diff like this.

lovelydinosaur:
I guess there's conflicts here with #328, perhaps? (Not necessarily a blocker, but worth mentioning)

tests/test_schemas.py
JoseKilo: I know, I was trying to keep it minimal. But adding more routes to the same 2 tests looked more difficult to maintain. Happy to revert the test refactor.

starlette/schemas.py
JoseKilo: Very True, changed.

JoseKilo:
@tomchristie I've reverted the tests refactor, so that the diff is easier to review.

rjwittams:
This would be useful - the schema generator doesn’t really work at all when mounts are used right now

starlette/schemas.py
```diff
@@ -65,14 +77,35 @@ def get_endpoints(
                     endpoints_info.append(
                         EndpointInfo(route.path, method.lower(), func)
                     )
+
         return endpoints_info
 
-    def parse_docstring(self, func_or_method: typing.Callable) -> dict:
+    def _extract_yaml(self, docstring: str) -> str:
```
lovelydinosaur: Let's not have the extra indirection here.

starlette/schemas.py
```diff
@@ -65,14 +77,35 @@ def get_endpoints(
...
+        """
+        We support having regular docstrings before the schema definition.
+
+        Here we return just the schema part from a docstring.
+        """
+        return docstring.split("---")[-1]
+
+    def parse_docstring(self, func_or_method: typing.Callable) -> typing.Optional[dict]:
```
lovelydinosaur: Would it make sense to return an empty dictionary in cases where we couldn't extract any info?

lovelydinosaur (changes requested):
Looking good. A couple of very minor things to look at.

starlette/schemas.py
JoseKilo: Yes, that looks cleaner. Another question is whether we want to exclude those cases from the final result (`if not parsed:` below). I think it makes sense.

lovelydinosaur: Actually that makes sense yup. Perhaps we'll also be able to move towards dropping the `include_in_schema` argument in routing?

JoseKilo: Yes, maybe it's not longer necessary. I haven't used it myself, but maybe there are still cases when you want to explicitly disable some endpoints. But then you wouldn't use yaml syntax in the docstrings ...

JoseKilo: I guess in that case you could have something that looks like yaml but it isn't yaml, for example:

```python
@app.route("/foo")
def foo(request):
    """
    This endpoint does the following:
        - Receive input
        - Some magic
        - Return an output
    """
    pass  # pragma: no cover
```

alimcmaster1:
thanks very much @JoseKilo @tomchristie was about to raise this as an issue. But will just bump up to latest release thanks!

Issue #333, opened 2019-01-21
QueryParams unexpected bahavior


taoufik07:
Hey, in QueryParams when calling the `.keys()` it returns the keys from the `self._list` which can return duplicate keys if a key have multiple values. Same for `.values()` and `.items()`, it returns all the values, while the expected behavior is to return for each key the last value.

```python
>>> q = QueryParams(query_string="a=1&a=2")
>>> q.keys() # ['a', 'a'] the expected ['a']
>>> q.values() # [1, 2] the expected [2]
>>> q.items() # (a, 1) (a, 2) the expected (a, 2)
```
I can easily fix those, but I suggest to create a MultiValueDict that can be subclassed by other classes. 

So what do you think, should I keep it that way and fix the behavior or add a MultiDict ?

** In the QueryParams implementation, the items are stored in both a dict and a list, is it more efficient than django's `Mapping[str, List]` or just because back then there were more important things to do.

lovelydinosaur:
Let's treat this as two different issues.

1. Yup, perfectly happy to have the single-value variants instead.
2. Perhaps we'll have a base class too, but let's discuss that separately, later.

taoufik07:
okie dokie :D

taoufik07:
@tomchristie you want `.keys()` to return a `generator`, a `dict_keys`  or a `list` ?

lovelydinosaur:
I'd suggest just starting with a list, as it currently is.
QueryParams have a limited size, so that'll be just fine.
I'd be okay with considering later pull requests that use dictkeys, but it'd depend on the additional complexity it adds.

PR #337, opened 2019-01-22
Fix QueryParams behavior

taoufik07:
Now QueryParams behaves like a dict  
ref #333

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -282,7 +282,7 @@ class QueryParams(typing.Mapping[str, str]):
         return iter(self.keys())
 
     def __len__(self) -> int:
-        return len(self._list)
+        return len(self._dict)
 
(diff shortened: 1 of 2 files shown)


starlette/datastructures.py
```diff
@@ -252,26 +252,25 @@ def __init__(
         elif scope is not None:
             _items = parse_qsl(scope["query_string"].decode("latin-1"))
 
-        self._dict = {k: v for k, v in reversed(_items)}
+        self._dict = {k: v for k, v in _items}
```
lovelydinosaur, line 255: Note that `QueryParams([('a', 'first'), ('a', 'second')`])` will have different behavior here.

lovelydinosaur: Sorry, ignore this.

lovelydinosaur:
Thanks! Built a little on top of this with #338

PR #338, opened 2019-01-22
Ensure query params behavior matches standard mappings.

### LINKED ISSUE #333: QueryParams unexpected bahavior

lovelydinosaur:
Closes #333
Superseeds #337

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -242,7 +242,7 @@ class QueryParams(typing.Mapping[str, str]):
             assert query_string is None, "Cannot set both 'params' and 'query_string'"
             assert scope is None, "Cannot set both 'params' and 'scope'"
             if isinstance(params, QueryParams):
-                _items = list(params.itemlists())
+                _items = list(params.multi_items())
             else:
                 _items = list(params.items())
         elif items is not None:
@@ -270,7 +270,7 @@ class QueryParams(typing.Mapping[str, str]):
     def items(self) -> typing.List[typing.Tuple[str, str]]:  # type: ignore
         return list(self._dict.items())
 
-    def itemlists(self) -> typing.List[typing.Tuple[str, str]]:
+    def multi_items(self) -> typing.List[typing.Tuple[str, str]]:
         return list(self._list)
 
     def get(self, key: typing.Any, default: typing.Any = None) -> typing.Any:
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Question: Do we consider this to be a straight-up bugfix, or an API behavior change?

taoufik07:
A bugfix, cause that's the behavior that would anyone expect

lovelydinosaur:
Right then. Possible follow ups:

1. I'd be perfectly happy to see `keys`, `values` and `items` return "keyview"/"valuesview"/"itemsview" or whatever the heck those are called. It'd need some work on the type annotations to get right.
2. Base class - I don't particularly care. I'd be happy to see a pull request for it, and judge it on merit, but keep in mind that the general cases would not be str: str as we have now, and might need to include both mutable and immutable types?

taoufik07:
**Update**: Tweaking the QueryParams is not a good idea, because removing an item gonna be expensive since we need to remove it from `_dict` and remove each occurrence in `_list`.
So I think we should go with Mapping[Any, List[Any]].

lovelydinosaur:
Okay, we now have `FormData`. Both `FormData` and `QueryParams` could inherit from a shared `ImmutableMultiDict` base class if we wanted. That'd be the first step.

We'd then want to also add a (mutable) `MultiDict`, but we should address that as a separate issue.

lovelydinosaur:
`FormData` and `QueryParams` themselves should remain immutable. If a user wants a mutable data structure containing that info, then make a mutable copy eg. `query_params = MultiDict(request.query_params)`

taoufik07:
Alright !

PR #341, opened 2019-01-22
Use `urlsplit`, not `urlparse`.

### LINKED ISSUE #340: Params dropped in test client
https://github.com/encode/starlette/blob/10b61f6cc28ec233ad3265b2822ae713dc26a68d/starlette/testclient.py#L70

params needs to be added back to path here, or urls like http://server/path;params will have the ;params chopped off.

lovelydinosaur:
Closes #340


lovelydinosaur:
New learning for today. You almost never really want `urlparse`.

PR #342, opened 2019-01-22
[WIP] MultiDict and ImmutableMixin

taoufik07:
it's a **WIP**, I need to add some methods, tests and implement ImmutableMixin.

@tomchristie I need to know what do you think before moving further.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -325,6 +325,63 @@ class MultiDict(dict):
         return super().pop(key)
 
 
+class ImmutableMultiDict:
+    """
+    An immutable multidict.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self._dict = MultiDict(*args, **kwargs)
+
+    def __getitem__(self, key) -> typing.Any:
+        return self._dict[key]
+
+    def __copy__(self) -> typing.Mapping:
+        return self.__class__(self)
+
+    def __deepcopy__(self, memo) -> typing.Mapping:
+        copy_ = self.__class__(copy.deepcopy(self._dict, memo))
+        return copy_
+
+    def get(self, key, default=None) -> typing.Any:
+        return self._dict.get()
+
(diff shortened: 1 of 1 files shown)


taoufik07:
I've got another [implementation](https://gist.github.com/taoufik07/30ce9b76b8f6d291bafcffb8cf079baf) where I've used a `_dict` attribute instead of subclassing `dict`

lovelydinosaur:
See #343.

A mutable `MultiDict` class, should subclass the existing `ImmutableMultiDict`, and add the extra MutableMapping API behavior for `__setitem__`, `__delitem__`, `pop`, `popitem`, `clear`, `update`, and `setdefault`.

Additionally, for multi-value updates we probably want `setlist(key, values)`, `appendlist(self, key, value)` and `setlistdefault(self, key, default_list=None)`.

taoufik07:
in #343 removing an item gonna be expensive, cause we need to remove every occurrence in `_list`

taoufik07:
Finally got some time, check this https://github.com/taoufik07/starlette/tree/MultiDict-2

lovelydinosaur:
> in #343 removing an item gonna be expensive, cause we need to remove every occurrence in _list

I'm not concerned about that. It's a rare operation, typically small list sizes, and only an `O(n)` operation.

lovelydinosaur:
Closing this off now, given #343.

Still up for a `MultiDict` implementation that builds on top of the existing `ImmutableMultiDict` class.

taoufik07:
> I'm not concerned about that. It's a rare operation, typically small list sizes, and only an `O(n)` operation.

Okay !

taoufik07:
@tomchristie I did it your way (I guess xD) https://github.com/taoufik07/starlette/tree/multidict-3

lovelydinosaur:
Opened the PR here, with comments: https://github.com/encode/starlette/pull/345

PR #345, opened 2019-01-24
MultiDict

lovelydinosaur:


starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
...
+
+    def __delitem__(self, key: typing.Any) -> None:
+        self._list = list(filter(lambda item: item[0] != key, self._list))
+        del self._dict[key]
+
+    def pop(self, key: typing.Any) -> typing.Any:
+        list_ = list(filter(lambda item: item[0] == key, self._list))
+        poped = self._dict.pop(key)
+        self.setlist(key, [v for _, v in list_[:-1]])
+        return poped
```
lovelydinosaur: I think the behavior of the standard `Mapping` operations should match `dict`.
So, `pop` should remove the item entirely. (Rather than remove one value from the list)

Also we're missing the `default` argument.

```python
def pop(self, key: typing.Any, default: typing.Any=None) -> typing.Any:
    self._list = [(existing_key, value) for (existing_key, value) in self._list if existing_key != key]
    return self._dict.pop(key, default)
```

Or...

```python
def pop(self, key: typing.Any, default: typing.Any=None) -> typing.Any:
    self._list = list(filter(lambda item: item[0] != key, self._list))
    return self._dict.pop(key, default)
```

starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
...
+
+    def popitem(self, key: typing.Any) -> typing.Tuple:
+        return key, self.pop(key)
```
lovelydinosaur: `popitem` doesn't take any arguments. I guess this kinda thing...

```python
key, value = self._dict.popitem()
self._list = list(filter(lambda item: item[0] != key, self._list))
return (key, value)
```

starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
...
+        self._list = [*filter_obj, *((key, value) for value in values)]
+        self._dict[key] = self._list[-1][1]
+        # Case list is empty
+
+    def setlistdefault(
+        self, key: typing.Any, default_list: typing.List = None
+    ) -> typing.List:
+        if key not in self:
+            if default_list is None:
+                default_list = [None]  # or just []
```
lovelydinosaur: Just `[]` please. 😄

starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
         return f"{self.__class__.__name__}({repr(items)})"
 
 
+class MultiDict(ImmutableMultiDict):
+    def __setitem__(self, key: typing.Any, value: typing.Any) -> None:
+        self.setlist(key, [value])
+        self._dict[key] = value
```
lovelydinosaur: Just the `setlist()` call should be enough. No need to set `._dict` again.

starlette/datastructures.py
lovelydinosaur: Also let's just call it `default`

starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
...
+
+        return self[key]
+
+    def setlist(self, key: typing.Any, values: typing.List) -> None:
+        filter_obj = filter(lambda item: item[0] != key, self._list)
+        if not values:
+            values.append(None)
+        self._list = [*filter_obj, *((key, value) for value in values)]
+        self._dict[key] = self._list[-1][1]
+        # Case list is empty
```
lovelydinosaur:
```python
items = [(key, value) for value in values]
self._list = list(filter(lambda item: item[0] != key, self._list)) + items
self._dict[key] = values[-1]
```

starlette/datastructures.py
```diff
@@ -290,6 +291,95 @@ def __repr__(self) -> str:
...
+
+    def update(
+        self,
+        values: typing.Union[
+            "MultiDict",
+            typing.Mapping,
+            typing.List[typing.Tuple[typing.Any, typing.Any]],
+        ] = None,
+        **kwargs: typing.Any,
+    ) -> None:
```
lovelydinosaur, line 347: I think `update` should just overwrite any existing keys, rather than appending to them.

starlette/datastructures.py
taoufik07: values is a tuple (key, value).

starlette/datastructures.py
taoufik07: if an empty was provided should we remove the item from dict or set the value in dict to None or raise an error ?

starlette/datastructures.py
taoufik07: Sorry, I didn't get it, could you provide some example

starlette/datastructures.py
lovelydinosaur: Actually, let's just omit this method entirely. I can't think of obvious enough use cases. It'd be reasonable to push users towards `if key not in q: q.setlist(values)`, which I think is more obvious.

starlette/datastructures.py
lovelydinosaur: Your test case gives:

```python
q = MultiDict([("a", "123"), ("b", "456")])
q.update({"a": "789"})
assert q.getlist("a") == ["123", "789"]
```

But I'd like to see:

```python
q = MultiDict([("a", "123"), ("b", "456")])
q.update({"a": "789"})
assert q.getlist("a") == ["789"]
q == MultiDict([("a", "789"), ("b", "456")])
```

starlette/datastructures.py
taoufik07: if values was empty, what should we do ?

```python
q = MultiDict([("a", "123")])
q.setlist("a", []) # ?
q.setlist("b", []) # ?
```

starlette/datastructures.py
taoufik07: Thanks, and if multiple values were provided, we do smt like :

```python
q = MultiDict([("a", "123"),  ("b", "456")])
q.update([("a", "ABC"), ("a", "DEF")])
assert q.getlist("a") == ["ABC", "DEF"] # ?
```

starlette/datastructures.py
lovelydinosaur: Delete it from the dict.

taoufik07: Alright !

starlette/datastructures.py
lovelydinosaur: Yup 👍

starlette/datastructures.py
lovelydinosaur: I guess it'd actually be:

```python
if not values:
    del self[key]
else:
    items = [(key, value) for value in values]
    self._list = list(filter(lambda item: item[0] != key, self._list)) + items
    self._dict[key] = values[-1]
```

starlette/datastructures.py
lovelydinosaur: I guess the implementation will need to do something like:

```python
items = ...
mapping = dict(items)
self._list = list(filter(lambda item: item[0] not in mapping, self._list)) + items
self._dict.update(mapping)
```

taoufik07: close to what I did
```python
keys = [k for k, _ in itertools.chain(items_, kwargs.items())]
        self._list = [*((k, v) for k, v in self._list if k not in keys), *items_, *kwargs.items()]
        self._dict.update(self._list)
```

taoufik07:
Sorry for the delay, I had to take care of some urgent matters :)

starlette/datastructures.py
```diff
@@ -290,6 +291,83 @@ def __repr__(self) -> str:
...
+        self._list = [(k, v) for k, v in self._list if k != key]
+        return self._dict.pop(key, default)
+
+    def popitem(self) -> typing.Tuple:
+        key, value = self._dict.popitem()
+        self._list = [(k, v) for k, v in self._list if k != key]
+        return key, value
+
+    def poplist(self, key: typing.Any) -> typing.List:
+        poped_list = [v for k, v in self._list if k == key]
```
lovelydinosaur: Let's call this "values"

taoufik07: Done Captain ! Any other changes ?

lovelydinosaur:
Great! I might make a few minor finesses on this, but I think it's good to come in as it stands. Thanks!

PR #354, opened 2019-01-25
Use correct name for yaml.dump

HenrikOssipoff:
This commit changes the incorrectly named `yaml.dumps` to `yaml.dump` in the docs.


lovelydinosaur:
Thanks!

PR #356, opened 2019-01-25
Update link to https

taoufik07:
I wanted to use a BSD license, so I've copied pasted it from starlette (since it was in my tree browser :3) and forgot to update the copyright holder. I've discovered it when I clicked unintentionally on the license file and I discovered that you're not using https.

Lesson learned :D


lovelydinosaur:
👍 Thanks

taoufik07:
You need to update all the other projects

PR #357, opened 2019-01-26
Add missing argument to database docs example

HenrikOssipoff:
This commit adds a missing required argument in the docs example of using the DatabaseMiddleware.

I'm not entirely sure if it's best to set it to false as I've done here, but I assume that's the use-case most people are looking for.

I've also slightly clarified a comment in the test part of the same docs that seemed incomplete.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -155,7 +155,8 @@ if TESTING:
   DATABASE_URL = DATABASE_URL.replace(database='test_' + DATABASE_URL.database)
 
 
-# Use 'rollback_on_shutdown' during testing, to ensure we have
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks, good catch! Actually we should default it to False on the middleware class, rather than updating the docs like this.

HenrikOssipoff:
That would've been my first assumption as well, but I figured there was a reason for this. I'll change the PR :)

lovelydinosaur:
Ace, thanks!

HenrikOssipoff:
Sorry about the wait - it now defaults to false. Commits squashed.

lovelydinosaur:
👍

PR #363, opened 2019-01-28
Routing for Hosts and Subdomains

### LINKED ISSUE #360: Support virtual hosts
It would be really nice to have a virtual hosts support (like with tornado) : 

> We support virtual hosts with the `add_handlers` method, which takes in a host regular expression as the first argument:
> 
> ```
> application.add_handlers(r"www\.myhost\.com", [
>     (r"/article/([0-9]+)", ArticleHandler),
> ])
> ```
> If there’s no match for the current request’s host, then default_host parameter value is matched against host regular expressions.

I've seen some references to a `base_url` variable in the code, but it doesn't seem possible to actively set it.

The point is to have only one instance of `uvicorn` with one instance of `Starlette` but being able to handle multiple domains or subdomains.

lovelydinosaur:
* Adds `app.host(host, app)`, which mirrors the existing `app.mount(path, app)`.
* Adds a `Host(...)` routing class, which mirrors the existing `Mount` routing class.

Closes #360

Examples:

Website and API, with separate routing rules:

```python
app = Starlette()
site = Router()  # Use eg. `@site.route()` to configure this.
api = Router()  # Use eg. `@api.route()` to configure this.

app.host('www.example.org', site)
app.host('api.example.org', api)
```

A custom subdomain:

```python
app = Starlette()

site = Router()  # Use eg. `@site.route()` to configure this.
subdomains = Router()  # Use eg. `@subdomains.route()` to configure this.

app.host('www.example.org', site)
app.host('{subdomain}.example.org', api)
```

Features:

* Subdomain components are available in URL reversing, eg. `request.url_for("users_list", subdomain=...)`
* Namespaces route naming is supported. eg. `app.host('api.example.org', api, name="api')` will then require reverse lookups like eg. `request.url_for("api:users_list")`
* Matched subdomain parts are available on the existing `request.path_params`.

Considerations:

* The name of `request.path_params` isn't exactly correct anymore. Should we move to `request.route_params` or something?
* For testing purposes users will need an easy way to mock up the host for requests, rather than eg. 127.0.0.1. Do we want that available in Starlette or in Uvicorn? Or do we just point folks at how to edit their `hosts` file?
* ~We might like subdomain parameters to be optional. If they're not included then `request.url_for` should default to current domain. (Assuming it matches the pattern?)~ Actually probably forget about that - we don't expect it with submounted paths, we shouldn't expect it here.
* We need to document this.
* I'm also becoming less convinced with the benefits of the decorator style routing, vs. composing route instances.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -67,6 +67,9 @@ class Starlette:
     def mount(self, path: str, app: ASGIApp, name: str = None) -> None:
         self.router.mount(path, app=app, name=name)
 
+    def host(self, host: str, app: ASGIApp, name: str = None) -> None:
+        self.router.host(host, app=app, name=name)
+
     def add_middleware(self, middleware_class: type, **kwargs: typing.Any) -> None:
         self.error_middleware.app = middleware_class(
             self.error_middleware.app, **kwargs
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -157,26 +157,35 @@ class DatabaseURL(URL):
 
 class URLPath(str):
     """
-    A URL path string that also holds an associated protocol.
+    A URL path string that may also hold an associated protocol and/or host.
     Used by the routing to return `url_path_for` matches.
     """
 
-    def __new__(cls, path: str, protocol: str) -> str:
-        assert protocol in ("http", "websocket")
+    def __new__(cls, path: str, protocol: str = "", host: str = "") -> str:
+        assert protocol in ("http", "websocket", "")
         return str.__new__(cls, path)  # type: ignore
 
-    def __init__(self, path: str, protocol: str) -> None:
+    def __init__(self, path: str, protocol: str = "", host: str = "") -> None:
         self.protocol = protocol
+        self.host = host
 
     def make_absolute_url(self, base_url: typing.Union[str, URL]) -> str:
         if isinstance(base_url, str):
             base_url = URL(base_url)
-        scheme = {
-            "http": {True: "https", False: "http"},
-            "websocket": {True: "wss", False: "ws"},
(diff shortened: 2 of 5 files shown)


lovelydinosaur:
Let's get this in. We can expand the documentation in due course.

sigg3:
Hi guys,

Hate to bump this, but this is still not in the docs so I've been trying to figure it out by looking at the patch and the added tests. Still having a minor issue. I have this code in my app.py:

`
""" The App """
app = Starlette(
                debug=True,
                middleware=middleware,
                on_startup=[initialize_korona],
                on_shutdown=[clean_up_korona]
)

""" Routing"""
site_main = Router(
     routes=[
        Route('/', hjem, name="homepage"),
        Route('/egen', utvalg_egen),
        Route('/om', om_tjenesten),
        Route('/k/{kom}', kommune),
        Mount('/css', StaticFiles(directory="static"), name="css")
        ]
)

site_subdomains = Router(
    routes=[
        Route('/', subdomain_parser, name="sub"),
        Mount('/css', StaticFiles(directory="static"), name="css")
        ]
)

 """Configure flow"""
app.host('www.localhost', site_main)
app.host('din.localhost', site_main)
app.host('korona.localhost', site_main)
app.host('{subdomain}.localhost', site_subdomains)`

Which seems to work alright for www, din and korona (to my great relief), but as for the site_subdomains I get the dreaded "'NoneType' object is not callable" error since the subdomain_parser function is not a basic ASGI function (scope, send, receive) but only has a request parameter..

Is there a reason why site_subdomains would be treated differently, when I have the app.host() in front of it too?

What am I doing wrong here? I feel like I'm very close. Full [app.py is here]https://github.com/sigg3/koronakom/blob/subdomains/app.py).

PR #364, opened 2019-01-29
Initial support for MySQL database

rajeev:
This is a copy of starlette's PostgreSQL driver, with modifications for MySQL. Includes changes to documentation, a mysql module within starlette.database and a minimal change to the Database middleware to accept MySQL DB URLs.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -1,7 +1,10 @@
 Starlette includes optional database support. There is currently only a driver
-for Postgres databases, but MySQL and SQLite support is planned.
+for Postgres and MySQL databases, but SQLite support is planned.
 
-Enabling the built-in database support requires `sqlalchemy`, and an appropriate database driver. Currently this means `asyncpg` is a requirement.
+Enabling the built-in database support requires `sqlalchemy`, and an appropriate database driver. 
+
+PostgreSQL: requires `asyncpg`
+MySQL: requires `aiomysql`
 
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Great work! Have you tested this locally? What OS are you running, and what tooling did you use? What extra requirements did you install?

I'm building on top of this in #366, and trying to run the database tests against both mysql and postgres.

rajeev:
@tomchristie Im building this as I go as we need MySQL support in our project, and it works on my local (Mac OS). Extra requirement: aiomysql (as described in the docs). Working on improving it for some use cases and then update the PR.

rajeev:
And yes, it works on local.

lovelydinosaur:
Pulled into #366

lovelydinosaur:
Okay, there's a bunch of stuff in #366 that 

* Coercing results back to python types (eg. so you get True/False for boolean columns, not 1/0.)
* Fix for `fetchrow` being `fetchone`
* Update the test cases to also run MySQL.

There's also one currently *unfixed* aspect, which is using SAVEPOINTS for nested transactions.

The biggest reason we need that is in order to properly support how test cases should interact with the database. (We need to: 1. Create a test database. 2. Start a transaction for each test case, rollback the transaction at the end of each test case. 3. Drop the test database) Any transactions that occur within the application then end up needing to be savepoints.

PR #366, opened 2019-01-29
MySQL driver support

lovelydinosaur:


lovelydinosaur:
Parameters for query compilation differ here between mysql and postgresql.
Hyperlinking to different implementations for later reference:

* https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/connection.py#L72
* https://github.com/aio-libs/aiomysql/blob/master/aiomysql/sa/connection.py#L146
* https://github.com/CanopyTax/asyncpgsa/blob/master/asyncpgsa/connection.py#L63

lovelydinosaur:
Remaining issue here: result sets for mysql are currently returning plain tuples, rather than using `_result_columns` from the compiled query to determine how to map key names onto the tuple results.

Options here are:

* Delegate the work to `aiomysql.sa` rather than doing it ourselves. We could also be doing the same with `asyncpgsa`, rather than having our own `compile(query...)`.
* Have `compile` hand back the `_result_columns` info.

lovelydinosaur:
To test:

`export STARLETTE_TEST_DATABASES=mysql+pymysql://<username>:<password>@localhost:3306/starlette_test`

Note that `pymysql` is required for testing here, since we need a sync driver when executing `metadata.create_all()`

You can test multiple database types, by setting `STARLETTE_TEST_DATABASES` as a comma-seperated list of database URLs.

lovelydinosaur:
Currently coercing data types using `.python_type` - We should look into using `result_processor` to do so in a more robust/complete way.

First test is now passing, but the other two are failing - looks like the session rollback isn't working, and the database is already populated at the start of those tests.

rajeev:
@tomchristie I just saw this PR now and looks like you're building on top of the patch I provided. I'm not sure how to continue development on this ... should I create a new PR based on this branch? Should I simply wait for MySQL support? Please let me know ...

lovelydinosaur:
I'm going to continue taking a look at this right now - the test aren't passing just yet, tho some progress towards it. Probably sensible just to wait for a bit, but otherwise you could fork the repo, and create a branch off `rajeev-mysql`.

lovelydinosaur:
Not completely happy with the implementation yet, but this gets us over the line.

lovelydinosaur:
* ~We should drop the `pymysql` visibility down into the test case setup itself, rather than having to use it in the `STARLETTE_TEST_DATABASES` envvar.~ Done.
* We should change the interfaces so that `fetchone` and `fetchall` return `Record` instances. The postgres case should then use the same type coercion.
* The transaction and connection state managment is a bit grungy.

rajeev:
@tomchristie Just wondering, how come I don't show up in the contributors graph? After all, I did spend the time to hack the PG driver into support for MySQL ...

PR #371, opened 2019-01-31
Add content_type attribute to UploadFile

roy-ht:
Thanks for your awesome project!

I want to look "Content-Type" header in multipart form request in some cases.

* Add `content_type` to UploadFile class
* If no content_type found, `UloadFile.content_type == None`

ref: https://tools.ietf.org/html/rfc7578#section-4.4

I don't know it's good to be `None` as the default behavior. For example, [Flask's `FileStorage.mimetype`returns empty string as default](https://github.com/pallets/werkzeug/blob/c8144ed2bcab242c80b9c5bbfe0a6f2124af0315/werkzeug/http.py#L371), but [`FileStorage.content_type` returns `None` as default](https://github.com/pallets/werkzeug/blob/c8144ed2bcab242c80b9c5bbfe0a6f2124af0315/werkzeug/datastructures.py#L2704).

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -422,8 +422,9 @@ class UploadFile:
     An uploaded file included as part of the request data.
     """
 
-    def __init__(self, filename: str, file: typing.IO = None) -> None:
+    def __init__(self, filename: str, content_type: str, file: typing.IO = None) -> None:
         self.filename = filename
+        self.content_type = content_type
         if file is None:
             file = tempfile.SpooledTemporaryFile()
         self.file = file
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -423,7 +423,7 @@ class UploadFile:
     """
 
     def __init__(
-        self, filename: str, content_type: str, file: typing.IO = None
+        self, filename: str, file: typing.IO = None, content_type: str = ""
     ) -> None:
         self.filename = filename
         self.content_type = content_type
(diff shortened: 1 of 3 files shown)


starlette/datastructures.py
```diff
@@ -422,8 +422,11 @@ class UploadFile:
     An uploaded file included as part of the request data.
     """
 
-    def __init__(self, filename: str, file: typing.IO = None) -> None:
+    def __init__(
+        self, filename: str, content_type: str, file: typing.IO = None
```
lovelydinosaur: Let's have `content_type` as the last parameter here, and have it default to `""`.

starlette/formparsers.py
```diff
@@ -200,11 +200,12 @@ def on_end(self) -> None:
                 elif message_type == MultiPartMessage.HEADERS_FINISHED:
                     headers = Headers(raw=raw_headers)
                     content_disposition = headers.get("Content-Disposition")
+                    content_type = headers.get("Content-Type")
```
lovelydinosaur: Let's use `headers.get("Content-Type", "")`

lovelydinosaur (changes requested):
I like this, yes!
Let's change the default to empty string, rather than `None`. After that I think it'll be good to go.

Nice work. 😄

roy-ht:
@tomchristie Done 😃

lovelydinosaur (approved):

lovelydinosaur:
Ace!

PR #373, opened 2019-01-31
Invoke mounted apps to handle lifespan message

cfhamlet:
Submounting other applications is very powerful.

Right now, only the root app process the lifespan message. So the event handlers registered on the root app can be called when the server startup/shutdown.  It is proper and reasonable. But sometime it is useful to invoke the submounted apps to process lifespan message in the meantime. 

I write a ``MountedAppLifespanHandler`` to invoke the submounted apps when the lifespan is coming. But I do not think it is proper to replace the recent ``LifespanHandler`` for the following compatible and design reasons:
1. It change the recent startup/shutdown behavior.
1. The ``__init__`` method is not compatible, need a ``ASGIApp``  as the first param.
2. I think it is optional not required.

There is no direct way to initialize my handler, because the lifespan handler creation is hard coded:

```
class Router:
    ......
        assert scope["type"] in ("http", "websocket", "lifespan")

        if scope["type"] == "lifespan":
            return LifespanHandler(scope)
    ......
```

I need to initialize my handler like this:

```
class Router:
    ......
        assert scope["type"] in ("http", "websocket", "lifespan")

        if scope["type"] == "lifespan":
            return MountedAppLifespanHandler(self, scope)
    ......
```

So, maybe my handler can be a future feature when the framework support custom lifespan handler initialization.

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -565,7 +565,7 @@ class MountedAppLifespanHandler:
                 self.instance = instance
                 self.startup_event = asyncio.Event()
                 self.shutdown_event = asyncio.Event()
-                self.receive_queue = asyncio.Queue()
+                self.receive_queue = asyncio.Queue()  # type: asyncio.Queue
 
             async def receive(self) -> Message:
                 return await self.receive_queue.get()
@@ -608,10 +608,8 @@ class MountedAppLifespanHandler:
 
         class MountedAppLifespan:
             def __init__(self, app: ASGIApp, scope: Scope) -> None:
-                self.lifespans = []
-                if not hasattr(app, "routes"):
-                    return
-                for route in app.routes:
+                self.lifespans = []  # type: typing.List[Lifespan]
+                for route in getattr(app, "routes", []):
                     try:
                         if not isinstance(route, Route):
                             self.lifespans.append(Lifespan(route(scope)))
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
> Submounting other applications is very powerful.
>
> Right now, only the root app process the lifespan message. So the event handlers registered on the root app can be called when the server startup/shutdown. It is proper and reasonable. But sometime it is useful to invoke the submounted apps to process lifespan message in the meantime.

Yes, yes, and yes.

We should work towards making sure that submounted applications receive lifespan messages. I'm not sure how best to tackle that yet. I don't think this pull request will be the way forward, but I'll try to take a look at how we should be doing it, in a built-in way.

dgrahn:
Is there a reason this issue is lingering?

dgrahn:
Here's a workaround.

```python
app.mount('/', service_app)

for func in service_app.router.lifespan.startup_handlers:
    app.add_event_handler('startup', func)

for func in service_app.router.lifespan.shutdown_handlers:
    app.add_event_handler('shutdown', func)
```

cfhamlet:
> Here's a workaround.
> 
> ```python
> app.mount('/', service_app)
> 
> for func in service_app.router.lifespan.startup_handlers:
>     app.add_event_handler('startup', func)
> 
> for func in service_app.router.lifespan.shutdown_handlers:
>     app.add_event_handler('shutdown', func)
> ```

My PR is just an immature solution. I'm not sure is there official solution right now. This issue should close when there is a proper way to solve this situation.

dgrahn:
Maybe something like this?

```python
app.mount(path, sub_app, trigger_events=True)
```

PR #375, opened 2019-02-02
Disable graphiql by default

jordaneremieff:
This allows for GraphIQL to be enabled/disabled (default: disabled) using an additional keyword argument. It is a good security practice to disable this console on production systems.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -17,11 +17,13 @@ class Query(graphene.ObjectType):
 
 
 app = Starlette()
-app.add_route('/', GraphQLApp(schema=graphene.Schema(query=Query)))
+app.add_route('/', GraphQLApp(schema=graphene.Schema(query=Query), graphiql=True))
 ```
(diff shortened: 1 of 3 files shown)


tiangolo:
My 2 opinionated cents:

I like the idea of being able to turn it off, but I'm not sure about having it off by default.

For example, GitHub has it on, with production data: https://developer.github.com/v4/explorer/

jordaneremieff:
I think having it default to off is more appropriate because it is considered a development tool first as well as it being [recommended](https://graphql.org/learn/serving-over-http/#graphiql) to disable it in production.  This would also be consistent with the various other Graphene integrations for popular frameworks.

lovelydinosaur (changes requested):
Thanks @erm! I'm happy enough with us defaulting to having it on, but the configuration switch would make sense.

lovelydinosaur:
👍

PR #377, opened 2019-02-05
Update Starlette API section in third-party docs

perdy:
Just updated Starlette API section within Third Party docs to reflect the additional functionalities and a minimal example of the usage.


lovelydinosaur:
Thanks!

Aside: It might not be a bad idea to call the application class `StarletteAPI` here.

PR #381, opened 2019-02-10
Fix URL.replace when port  is not provided

taoufik07:
When the port is not provided in the `URL.replace`, it uses None.
ref https://github.com/kennethreitz/responder/issues/284

### DIFF
--- a/starlette/middleware/httpsredirect.py
+++ b/starlette/middleware/httpsredirect.py
@@ -11,7 +11,7 @@ class HTTPSRedirectMiddleware:
         if scope["type"] in ("http", "websocket") and scope["scheme"] in ("http", "ws"):
             redirect_scheme = {"http": "https", "ws": "wss"}[scope["scheme"]]
             url = URL(scope=scope)
-            url = url.replace(scheme=redirect_scheme, netloc=url.hostname)
+            url = url.replace(scheme=redirect_scheme, netloc=url.netloc)
             return RedirectResponse(url, status_code=301)
 
         return self.app(scope)


taoufik07:
My bad, forgot the tests !!

em92:
And you left `print`

em92:
Btw, to fix bug in responder it is required to unrestrict 0.10.x version in responder. Or also apply patch to 0.9.x
https://github.com/kennethreitz/responder/blob/69eb843604c43b85ca796721f40c3547fa5bf424/setup.py#L25

taoufik07:
I'm aware of that and I think responder should update to .10, but it's up to @kennethreitz

lovelydinosaur:
It's unclear to me what this fixes.
Want to add a test case so that's more clear?
Ideally if you could also aim for a more minimal diff, that doesn't change the structure around so much - that'd also make it easier to clearly review.

lovelydinosaur (changes requested):
Thanks! Needs a test case to make it obvious what behavioral change is intended here. Ideally a more minimal diff, too.

lovelydinosaur:
> I'm aware of that and I think responder should update to .10, but it's up to @kennethreitz

I'd suggest just be proactive and issue the PR on responder making the version update.

taoufik07:
> I'd suggest just be proactive and issue the PR on responder making the version update.

Once we fix this issue, I will PR

taoufik07:
Sorry, we only need to pass `url.netloc` in https middleware instead of `url.hostname`.
https://github.com/encode/starlette/blob/208b4aa2078a0f58289ffdd9a6100f80ba7ffaf6/starlette/middleware/httpsredirect.py#L14

And I was saying to my self why yesterday I wanted to update the https middleware, now I remember xD.

lovelydinosaur:
> Sorry, we only need to pass `url.netloc` in https middleware instead of `url.hostname`.

I think the most robust behavior would be to use `url.hostname` if the port is `None`, `80`, or `443`, and `url.netloc` if there's an explicit non-standard port.

*Aside*: Even so - it's not definitive that this is *neccessarily* related to the reported issue on responder.

lovelydinosaur:
What do you think to this?...

```python
        if scope["type"] in ("http", "websocket") and scope["scheme"] in ("http", "ws"):
            redirect_scheme = {"http": "https", "ws": "wss"}[scope["scheme"]]
            netloc = url.hostname if url.port in (None, 80, 443) else url.netloc
            url = URL(scope=scope)
            url = url.replace(scheme=redirect_scheme, netloc=netloc)
            return RedirectResponse(url, status_code=301)
```

taoufik07:
looks great

taoufik07:
but aren't netloc and hostname gonna be the same if the port is in (None, 80, 443) ?

lovelydinosaur:
Superseeded by #382 
Released as 0.10.2

lovelydinosaur:
Still not obvious that this is neccessarily related to the issue as raised on responder, tho.

taoufik07:
when `localhost:5042` is redirected by the HTTPSMiddleware it uses url.hostname `localhost` and with this fix it now uses the netloc if the port is not standard

em92:
> Released as 0.10.2

Version bump commit not in master branch. https://github.com/encode/starlette/blob/master/starlette/__init__.py

lovelydinosaur:
@em92 Thanks. Resolved.

PR #384, opened 2019-02-11
Request.ujson() for parity with UJSONResponse

iced:
Proper solution should be some setting - e.g. `app = Starlette("app", json="ujson")` - will you accept PR with this approach?

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -13,7 +13,12 @@ try:
 except ImportError:  # pragma: nocover
     parse_options_header = None  # type: ignore
 
+try:
+    import ujson
+except ImportError:  # pragma: nocover
+    ujson = None  # type: ignore
 
+    
 class ClientDisconnect(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
If anything I’d probably rather just drop `UJSONResponse` and document how to subclass `JSONResponse` for users that really need it.

It handles edge cases in subtly different ways to the standard library, and is rarely going to be the actual bottleneck point in production codebases, so I think it’s a poor trade-off for us to recommend.

lovelydinosaur:
Thanks for making the contribution, tho let’s pass on it given the above.

iced:
No problem, I'll just monkey patch it for my projects. For some of my cases ujson gives ~20% extra rps, so it's kinda forced decision :)

lovelydinosaur:
I'd probably suggest

```python
body = await request.body()
data = ujson.loads(body)
```

Rather than monkeypatching, but either ways.

(Or have a helper function that gives you `data = await load_ujson(request)`)

PR #387, opened 2019-02-13
Add python version restriction

em92:
Reference: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires


lovelydinosaur:
Yup great, thanks!

PR #390, opened 2019-02-14
Move to 'databases' instead of built-in 'DatabaseMiddleware'

lovelydinosaur:
Move to recommending the stand-alone package https://github.com/encode/databases rather than the built-in `DatabaseMiddleware`.

You no longer need to pass the `request` around everywhere you want database access.

This moves `DatabaseMiddleware` into a pending deprecation state, but does not remove it. It'll currently still continue to work just fine. We'll only drop it once we make a proper version bump.

Not much to change if you're moving to using it...

**1. Create a `database` instance, rather than installing a middleware.**

Remove:

`app.add_middleware(DatabaseMiddleware, database_url=DATABASE_URL)`

Add:

```python

@app.on_event("startup")
async def startup():
    await database.connect()

@app.on_event("shutdown")
async def shutdown():
    await database.disconnect()
```

**2. Use `database.<...>` instead of `request.database.<...>`**

* `request.database.fetchone(...)` -> `database.fetch_one()`
* `request.database.fetchall(...)` -> `database.fetch_all()`
* `request.database.execute(...)` -> `database.execute()`
* New: `database.execute_many()`
* New: `database.iterate()`
* Not yet ported: `request.database.fetchval(...)` (Use `fetchone()['coilumn_name']`)

**3. Use `@database.transaction()` instead of Starlette's built-in `@transaction` decorator.**

Remove:

```python
from starlette.databases import transaction

@transaction
async def my_endpoint(request):
   ...
```

Add:

```python
@database.transaction()
async def my_endpoint(request):
   ...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
+test.db
 .coverage
 .pytest_cache/
 .mypy_cache/
--- a/docs/database.md
+++ b/docs/database.md
@@ -1,23 +1,31 @@
-Starlette includes optional database support. There is currently only a driver
-for Postgres and MySQL databases, but SQLite support is planned.
+Starlette is not strictly tied to any particular database implementation.
 
-Enabling the built-in database support requires `sqlalchemy`, and an appropriate database driver. 
+You can use it with an asynchronous ORM, such as [GINO](https://python-gino.readthedocs.io/en/latest/),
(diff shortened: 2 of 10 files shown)


cancan101:
Using this new approach, is it possible to use the sql alchemy orm functionality, assuming the route is synchronous? e.g. Some form of `db.session.query`?

PR #392, opened 2019-02-16
Add note about routes order

tiangolo:
Add explicit note about routes order: `/users/me` before `/users/{user_id}`.

---

This is equivalent to this new (more verbose) section in FastAPI: https://fastapi.tiangolo.com/tutorial/path-params/#order-matters

I tried to make it short as I see the docs style in Starlette is concise, but let me know if you prefer it more verbose (I can copy/adapt more from that section in FastAPI).

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -53,6 +53,12 @@ You can use any of the following to add handled routes to the application:
 * `@app.route(path)` - Add an HTTP route, decorator style.
 * `@app.websocket_route(path)` - Add a WebSocket route, decorator style.
 
+#### Order matters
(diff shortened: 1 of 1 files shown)


tiangolo:
I just realized I had added a note about it in this PR: https://github.com/encode/starlette/pull/280/files.

But that note is in the docs for Routing: https://www.starlette.io/routing/

The note in this PR is a shorter version in the Applications section: https://www.starlette.io/applications/

---

If you think the existing documentation in Routing is enough and this PR is not needed, feel free to just close it.

PR #393, opened 2019-02-16
Update FastAPI example to make style/difference explicit

tiangolo:
Update FastAPI example to make style difference explicit.

The current example looks more similar to Flask (the use of Python types is not explicit).

This update makes it more similar to APIStar (with explicit Python types).

And it shows more clearly the difference with other frameworks/styles.


lovelydinosaur (approved):

tiangolo:
Thanks!

PR #395, opened 2019-02-17
Proposal: remove code from third party section

woile:
Hey folks, here a proposal to make third party section simpler. This doesn't need to be merge, just up for discussion, what do you think?

reason: code should not be present here, it's hard to maintain, a short synopsis is enough.
Page is also smaller, more visibility to other packages.

What we want to give people is just a reference, and then they can go to that page an explore the app.

Some examples:

- http://flask.pocoo.org/extensions/
- https://hapijs.com/plugins
- https://djangopackages.org/categories/apps/
- https://github.com/falconry/falcon/wiki/Complementary-Packages

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -186,7 +186,7 @@ import pytest
 from starlette.config import environ
 from starlette.testclient import TestClient
 from sqlalchemy import create_engine
-from sqlalchemy_utils import database_exists, create_database
+from sqlalchemy_utils import database_exists, create_database, drop_database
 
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Basically in favour from me, yup. It was okay when there’s only a couple of examples, but it’s worth slimming down now.

tiangolo:
I think it's @tomchristie's call.

---

But still, cool for me. I agree. :+1:

PR #399, opened 2019-02-18
fix: jinja annotations when jinja not installed

woile:
+ jinja annotations are now strings.
+ minor fix in docs
+ lint

Error when jinja not installed:

```
venv/lib/python3.6/site-packages/starlette/applications.py:4: in <module>
    from starlette.exceptions import ExceptionMiddleware
venv/lib/python3.6/site-packages/starlette/exceptions.py:7: in <module>
    from starlette.responses import PlainTextResponse, Response
venv/lib/python3.6/site-packages/starlette/responses.py:289: in <module>
    from starlette.templating import _TemplateResponse as TemplateResponse
venv/lib/python3.6/site-packages/starlette/templating.py:44: in <module>
    class Jinja2Templates:
venv/lib/python3.6/site-packages/starlette/templating.py:54: in Jinja2Templates
    def get_env(self, directory: str) -> jinja2.Environment:
E   AttributeError: 'NoneType' object has no attribute 'Environment'
```

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -281,7 +281,7 @@ from alembic.config import Config
 
 @pytest.fixture(scope="session", autouse=True)
 def create_test_database():
-    url = str(app.DATABASE_URL)
+    url = str(app.database._url)
     engine = create_engine(url)
     assert not database_exists(url), 'Test database already exists. Aborting tests.'
     create_database(url)             # Create the test database.
--- a/starlette/database/mysql.py
+++ b/starlette/database/mysql.py
@@ -4,11 +4,11 @@ import typing
 import uuid
 from types import TracebackType
 
+import aiomysql
 from sqlalchemy.dialects.mysql import pymysql
 from sqlalchemy.engine.interfaces import Dialect
 from sqlalchemy.sql import ClauseElement
 
-import aiomysql
 from starlette.database.core import (
     DatabaseBackend,
     DatabaseSession,
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -51,7 +51,7 @@ class Jinja2Templates:
     def __init__(self, directory: str) -> None:
         self.env = self.get_env(directory)
 
-    def get_env(self, directory: str) -> jinja2.Environment:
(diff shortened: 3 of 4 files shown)


docs/database.md
```diff
@@ -281,7 +281,7 @@ from alembic.config import Config
 
 @pytest.fixture(scope="session", autouse=True)
 def create_test_database():
-    url = str(app.DATABASE_URL)
+    url = str(app.database._url)
```
lovelydinosaur, line 284: `app.DATABASE_URL` is already correct in the docs here, but we should add `import app` above.
(It's *not* the application instance, it's the module)

lovelydinosaur:
Thanks! One tweak to make, but I'll go fix that up.

docs/database.md
woile: Please verify tom. My issue was that the next assert:
`assert not database_exists(url)` was failing because the value of `app.DATABASE_URL` was the dev value, not the TEST_DATABASE_VALUE.

According to the docs, the `database` gets generated this way:
```
if TESTING:
    database = databases.Database(TEST_DATABASE_URL, force_rollback=True)
else:
```

so the assert can only check against `database._url`. Or at least that's what I understood.

Maybe I've missed something hehe

woile: I mean, not matter what I did, even if `TESTING` is `True`, the assert would always fail.

lovelydinosaur: Right, then `app.TEST_DATABASE_URL`

woile: But isn't the idea to check that you are not referencing the "prod/dev" database? if you point it to `TEST_DATABASE_URL` it won't fail.

Jitsusama:
@tomchristie; any idea on when you might cut a release with this fix? I'm stuck with either having to add jinja2 as a dependency in order to run 0.10.4, or version pin starlette to 0.10.3 in my Pipenv because of this.

lovelydinosaur:
@Jitsusama Sorted

Jitsusama:
Thanks @tomchristie!

Issue #396, opened 2019-02-18
Roadmap


lovelydinosaur:
I've been thinking about how best to mature Starlette as it becomes more and more capable for large-scale projects. **The key thing I want the framework to focus on is low-complexity of the underlying stack.**

Rather than have a monolithic application object, where we put all the functionality on, I'd like to see us focus on independent components. I think this is really important, because it makes it *so much* easier to understand the stack that you're building on top of. Components are independently testable, can be understood in isolation, and can be swapped out or adapted more easily.

One example here is we currently have `template_dirs` on the application instance. That makes it less clear what interactions the application instance might have with template dirs under the hood. Instead we should keep all the templating properly independent from other parts of the stack.

A related aspect of this is routing. Right now we promote the decorator style, which is convenient. I think we should move towards promoting a more explicit style made up of the actual `BaseRoute` instances that the `Router` class uses. The reason here is that it explicitly surfaces information about how the underlying stack function. If you want to implement an alternative to `Route`, `WebSocketRoute`, `Mount`, `Host`, it's already more clear how you'd do that. Even the call stack that gets passed through is being made more explicit. "Look at this class implementation if you want to understand what's happening here".

Here's a sketch of how a more explicit style would fit together:

**settings.py**

```python
DEBUG = config(...)
DATABASE_URL = config(...)
```

**resources.py**

```python
cache = Cache(...)
database = Database(...)
templates = Jinja2Templates(...)
```

**routing.py**

```python
routes = [
    Route(...),
    Route(...),
    WebSocketRoute(...),
    Mount(...)
]
```

**application.py**

```python
middleware = [
    ...
]

app = Starlette(routes=routes, middleware=...)
```

taoufik07:
I've always liked how django is organized and when working with other frameworks I try to follow django's structure.

We can add a `views.py` or `controller.py` where the endpoints will live, also add some cli to init the project.

tiangolo:
Thanks for sharing the plans and roadmap!

---

A small opinion that might or not be relevant about routing:

In the same spirit of having components of related parts together, I like the decorator style of routing, because it puts in the same code region (the same section in a file) two very related parts: the path/route and the function that handles it.

Having routes in a separated section that reference functions imported from other files (more Django style than Flask style) makes the used classes more explicit, but it also makes the relationship between the function that handles a specific path and the path itself less explicit.

I guess this is some kind of tradeoff of two types of explicitness, between classes used and path/function relationship.

[...]

PR #401, opened 2019-02-18
Lifespan route instance

lovelydinosaur:
First pass towards having `Lifespan` just be another routed component.


lovelydinosaur:
Refs #396

PR #403, opened 2019-02-19
Apply linting to setup.py

JoseKilo:
I thought this would make sense. Instead of changing a smaller bit on my other PR.


lovelydinosaur:
👍 Good call.

PR #404, opened 2019-02-19
Request state

### LINKED ISSUE #379: Generic interface for storing per-request state.
From this: https://github.com/encode/starlette/issues/374#issuecomment-461684385

I think we should probably support a public interface for storing arbitary per-request state, within the ASGI scope.

That could either be:

1. `request['whatever'] = ...` - Expose the ASGI scope, and let the user handle it directly if they really need to.
2. `request.scope['whatever'] = ...` - Same as before, but let's move it into a different bit of public API.
3. `request.state.whatever = ...` - Store arbitrary state, and always keep it isolated from the rest of the namespace in the ASGI scope. (Implementation wise, store it all in a scope['state'] dictionary)

With (1) and (2) we might need to think a bit about mutability and it's implications. Eg. If the scope is modified, do we update the `request.url`, `request.headers` or whatever else it might affect? Probably we just put a disclaimer on "request.scope" gives you access to the underlying ASGI scope - make sure you've got a bit of an idea what you're doing.

Incidentally, I think I prefer (2) over the "request is dict-like" interface that we currently have.

lovelydinosaur:
Closes #379


tiangolo:
:clap: :clap: :tada: :cake: :rocket:

PR #406, opened 2019-02-20
Add SQLAlchemy ORM example to databases, using request.state

tiangolo:
Add SQLAlchemy ORM example to databases, using `request.state`.

Based on the equivalent in FastAPI: https://fastapi.tiangolo.com/tutorial/sql-databases/

---

I tried to simplify it as much as possible, limit the verbosity of the docs, and keep the style of the rest of the Starlette docs (e.g. without type hints).

I assumed you would prefer it in the same `database.md`, as a short version/note at the end.

But if you prefer, I can change it to a `database-async.md` or similar, in a separate menu section.

Also, if you want I can raise the verbosity of the content/docs.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -385,7 +385,7 @@ def read_user(request):
 
 
 @app.middleware("http")
-async def close_db(request, call_next):
+async def db_session_middleware(request, call_next):
     request.state.db = Session()
     response = await call_next(request)
     request.state.db.close()
--- a/docs/database.md
+++ b/docs/database.md
@@ -294,3 +294,120 @@ def create_test_database():
 
 [sqlalchemy-core]: https://docs.sqlalchemy.org/en/latest/core/
 [alembic]: https://alembic.sqlalchemy.org/en/latest/
+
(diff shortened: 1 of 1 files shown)


docs/database.md
```diff
@@ -294,3 +294,120 @@ def create_test_database():
 
 [sqlalchemy-core]: https://docs.sqlalchemy.org/en/latest/core/
 [alembic]: https://alembic.sqlalchemy.org/en/latest/
+
+
+## Integrating `SQLAlchemy` ORM
```
woile, line 299: My 2 cents:
The existing example is already using sqlalchemy (but in an async way with `databases`), this looks more like a migration example, maybe the title could be renamed to `Migrating from Flask`. What do you think?

tiangolo: The existing example uses SQLAlchemy core with async. This uses SQLAlchemy ORM with async.

The focus is on being able to use the ORM too.

tiangolo:
Cool, thanks for the feedback.

PR #409, opened 2019-02-23
Update Config docs, removing defunct DatabaseURL references

### LINKED ISSUE #408: Config/Settings dump?
What's the best way to dump all of the settings using starlette.config? I want to create an endpoint to make it easy to review the settings being used but I don't want the Secrets to be viewable, e.g. things like:

    JWT_SECRET = config("JWT_SECRET", cast=starlette.datastructures.Secret)

We are currently using:

```
@router.get("/settings", tags=["Info"])
def get_settings() -> dict:
    dctconfig = dict(settings.config)
    return dctconfig
```
But it shows the the secrets (instead of Secret('**********')) along with everything else.

wshayes:
Closes #408 
1. Removed DatabaseURL Class which was removed in v0.11 and changed to URL instead
2. Added example for creating a settings API endpoint to show the settings while maintaining the Secret obfuscation for passwords, etc (example how to do this was shared by @tomchristie in #409)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -9,13 +9,13 @@ that is not committed to source control.
 ```python
 from starlette.applications import Starlette
 from starlette.config import Config
-from starlette.datastructures import CommaSeparatedStrings, DatabaseURL, Secret
+from starlette.datastructures import CommaSeparatedStrings, Secret
 
 # Config will be read from environment variables and/or ".env" files.
 config = Config(".env")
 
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks! I'd rather not have the endpoint example, but the other fixes are great!
Any chance you'd be able to update the pull request?

lovelydinosaur:
Ta!

PR #411, opened 2019-02-24
Fix lower case header value preflight request

jgelens:
Some browers send lower case values for the `Access-Control-Request-Headers`. This diff makes sure everything is handled case insensitive.

screenshot 2019-02-24 at 10 11 14


lovelydinosaur:
Nice one. Yup!

PR #415, opened 2019-02-26
Staticfiles packages

lovelydinosaur:
Adds support for `packages` to `StaticFiles`. Eg.

Useful for bundling up reusable components. (Eg. in the same way that Django's INSTALLED_APPS does)

```python
routes= [
    ...
    Mount('/static', app=StaticFiles(directory='static', packages=['bootstrap4']), name="static")
]
```


alimcmaster1:
+1 handy feature! thanks!

PR #417, opened 2019-03-02
Doc hint about enabling code reload support

ahtik:
As currently starlette(debug=True) does not enable reload for code changes, maybe it's useful to point out the corresponding reload feature from the uvicorn and gunicorn packages. As I understand, uvicorn.run() itself doesn't allow this, so one must rely on command line (https://github.com/kennethreitz/responder/issues/239#issuecomment-440260673).

This PR suggests only uvicorn flag hint. gunicorn --reload arg with UvicornWorker has been a bit unstable for me (just the reload part).

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ $ uvicorn example:App
 INFO: Started server process [11509]
 INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
 ```
-Run uvicorn with `--debug` to enable auto-reloading on code changes and display error tracebacks.
+Run uvicorn with `--reload` to enable auto-reloading on code changes.
 
 ## Modularity
 
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ $ uvicorn example:App
 INFO: Started server process [11509]
 INFO: Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
 ```
+Run uvicorn with `--debug` to enable auto-reloading on code changes and display error tracebacks.
 
 ## Modularity
 


lovelydinosaur:
Yeah it's a bit confusing currently.
`uvicorn --debug` does two things:

* Adds a DebugMiddleware.
* Enables auto-reload.

I really think we ought to *drop* `--debug` as a flag there and just use `--reload`. (The debug middleware is then just an *application* level concern - ie. Starlette(debug=True), rather than being duplicated in both places.)

I'd like to leave this PR open at the moment as I don't want to introduce a docs change stating "Use --debug" when the clearer thing we could do would be move towards a "use --reload", but I also don't want to close this, as we should keep it on our radar until we can switch it to recommending --reload.

ahtik:
Not 100% sure if worth mentioning but still, this current `--debug` behavior is actually pretty close to what Flask&Werkzeug is doing, which additionally provides use_reloader=True and use_debugger=True defaults for --debug flag (https://github.com/pallets/flask/blob/master/flask/app.py#L935. Reloader itself, https://github.com/pallets/werkzeug/blob/master/werkzeug/_reloader.py#L319).

But yeah, I do realize Starlette is not having a dependency on uvicorn even for development mode, so it's a bit different use-case. Then again there are already quite a few Starlette-based things like fastapi, responder etc who could possibly benefit from a similar knobs, so maybe this --reload could be still allowed for API access, not just the CLI (ref: https://github.com/kennethreitz/responder/issues/239#issuecomment-440266419)

lovelydinosaur:
Thanks!

lovelydinosaur:
Also note that `uvicorn.run(app, reload=True)` *does* now work, too. 👍

colinkcurtis:
> Also note that `uvicorn.run(app, reload=True)` _does_ now work

I am using uvicorn==0.6.1, however for me when using:

`if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8000, reload=True)`

I do not have the --reload behavior that I get when running uvicorn from the CLI

EDIT: I just found your comments, here: https://github.com/kennethreitz/responder/issues/326#issuecomment-469632317

So, it looks like I'll need to pin uvicorn==0.5.* to utilize this functionality? Thanks...

lovelydinosaur:
reloading can only work from the command line or with string imports. (When the app isn't already in memory)

> So, it looks like I'll need to pin uvicorn==0.5.* to utilize this functionality?

No. I was recommending that responder pinned to the latest median version at the time. (They still ought to, tho it's 0.6.* now)

colinkcurtis:
Ok - thank you.

PR #418, opened 2019-03-04
Add documentation for UploadFile class in the Request section

tiangolo:
Add documentation for `UploadFile` class in the Request section, the class received when getting uploaded files in a request.

This is extracted and adapted from FastAPI that uses/inherits the same Starlette class https://fastapi.tiangolo.com/tutorial/request-files/.

If it's too verbose, I can simplify it and eliminate explanations, etc.

### DIFF
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -121,9 +121,7 @@ state with `disconnected = await request.is_disconnected()`.
 Request files are normally sent as multipart form data (`multipart/form-data`).
 
 When you call `await request.form()` you receive a `starlette.datastructures.FormData` which is an immutable
-multidict, containing both file uploads and text input.
-
-When one of the fields is a file upload, you will get a `starlette.datastructures.UploadFile`.
+multidict, containing both file uploads and text input. File upload items are represented as instances of `starlette.datastructures.UploadFile`.
 
 `UploadFile` has the following attributes:
 
@@ -132,27 +130,39 @@ When one of the fields is a file upload, you will get a `starlette.datastructure
 * `file`: A <a href="https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile" target="_blank">`SpooledTemporaryFile`</a> (a <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> object). This is the actual Python file that you can pass directly to other functions or libraries that expect a "file-like" object.
 
 
-`UploadFile` has the following `async` methods. They all call the corresponding file methods underneath (using the internal `SpooledTemporaryFile`) in a threadpool.
+`UploadFile` has the following `async` methods. They all call the corresponding file methods underneath (using the internal `SpooledTemporaryFile`).
 
-* `write(data)`: Writes `data` (`str` or `bytes`) to the file.
-* `read(size)`: Reads `size` (`int`) bytes/characters of the file.
-* `seek(offset)`: Goes to the byte position `offset` (`int`) in the file.
+* `async write(data)`: Writes `data` (`str` or `bytes`) to the file.
+* `async read(size)`: Reads `size` (`int`) bytes/characters of the file.
+* `async seek(offset)`: Goes to the byte position `offset` (`int`) in the file.
     * E.g., `await myfile.seek(0)` would go to the start of the file.
-    * This is especially useful if you run `await myfile.read()` once and then need to read the contents again.
-* `close()`: Closes the file.
+* `async close()`: Closes the file.
 
 As all these methods are `async` methods, you need to "await" them.
 
-For example, inside of an `async` route you can get the contents with:
+For example:
 
-```Python
-contents = await myfile.read()
-```
+```python
+from starlette.requests import Request
+from starlette.responses import Response
 
-If you are inside of a normal `def` route, you can access the `UploadFile.file` directly, for example:
 
-```Python
-contents = myfile.file.read()
+class App:
+    def __init__(self, scope):
+        assert scope["type"] == "http"
+        self.scope = scope
+
+    async def __call__(self, receive, send):
+        request = Request(self.scope, receive)
+        form = await request.form()
+        myfile = form["file"]
+        filename = myfile.filename
+        contents_bytes = await myfile.read()
+        contents_str = contents_bytes.decode("utf-8")
+        response = Response(
+            f"The file {filename} contains: {contents_str}", media_type="text/plain"
+        )
+        await response(receive, send)
 ```
 
 #### Other state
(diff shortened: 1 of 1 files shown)


docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
+
+```Python
+contents = await myfile.read()
+```
+
+If you are inside of a normal `def` route, you can access the `UploadFile.file` directly, for example:
+
+```Python
+contents = myfile.file.read()
+```
```
lovelydinosaur, line 149: Let's not bother with this section.

docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
+    * This is especially useful if you run `await myfile.read()` once and then need to read the contents again.
+* `close()`: Closes the file.
+
+As all these methods are `async` methods, you need to "await" them.
+
+For example, inside of an `async` route you can get the contents with:
+
+```Python
+contents = await myfile.read()
+```
```
lovelydinosaur: I think an example that included the `data = await request.form()` would be marginally clearer.
Also let's use lowercase `python` on backticks for consistency with other places in the docs.

docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
+* `file`: A <a href="https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile" target="_blank">`SpooledTemporaryFile`</a> (a <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> object). This is the actual Python file that you can pass directly to other functions or libraries that expect a "file-like" object.
+
+
+`UploadFile` has the following `async` methods. They all call the corresponding file methods underneath (using the internal `SpooledTemporaryFile`) in a threadpool.
+
+* `write(data)`: Writes `data` (`str` or `bytes`) to the file.
+* `read(size)`: Reads `size` (`int`) bytes/characters of the file.
+* `seek(offset)`: Goes to the byte position `offset` (`int`) in the file.
+    * E.g., `await myfile.seek(0)` would go to the start of the file.
+    * This is especially useful if you run `await myfile.read()` once and then need to read the contents again.
```
lovelydinosaur: I don't think we need the second bullet point here.

docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
+`UploadFile` has the following attributes:
+
+* `filename`: A `str` with the original file name that was uploaded (e.g. `myimage.jpg`).
+* `content_type`: A `str` with the content type (MIME type / media type) (e.g. `image/jpeg`).
+* `file`: A <a href="https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile" target="_blank">`SpooledTemporaryFile`</a> (a <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> object). This is the actual Python file that you can pass directly to other functions or libraries that expect a "file-like" object.
+
+
+`UploadFile` has the following `async` methods. They all call the corresponding file methods underneath (using the internal `SpooledTemporaryFile`) in a threadpool.
+
+* `write(data)`: Writes `data` (`str` or `bytes`) to the file.
```
lovelydinosaur: I guess we could use a style like `async write(data)`?

docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
+When one of the fields is a file upload, you will get a `starlette.datastructures.UploadFile`.
+
+`UploadFile` has the following attributes:
+
+* `filename`: A `str` with the original file name that was uploaded (e.g. `myimage.jpg`).
+* `content_type`: A `str` with the content type (MIME type / media type) (e.g. `image/jpeg`).
+* `file`: A <a href="https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile" target="_blank">`SpooledTemporaryFile`</a> (a <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> object). This is the actual Python file that you can pass directly to other functions or libraries that expect a "file-like" object.
+
+
+`UploadFile` has the following `async` methods. They all call the corresponding file methods underneath (using the internal `SpooledTemporaryFile`) in a threadpool.
```
lovelydinosaur: I don't think we need to mention the threadpool - at this level of docs we can consider that as an implementation detail.

docs/requests.md
```diff
@@ -116,6 +116,45 @@ In some cases such as long-polling, or streaming responses you might need to
...
 state with `disconnected = await request.is_disconnected()`.
 
+#### Request Files
+
+Request files are normally sent as multipart form data (`multipart/form-data`).
+
+When you call `await request.form()` you receive a `starlette.datastructures.FormData` which is an immutable
+multidict, containing both file uploads and text input.
+
+When one of the fields is a file upload, you will get a `starlette.datastructures.UploadFile`.
```
lovelydinosaur: We could shorten this line and the preceeding one, into a single para. Perhaps. *"containing both file uploads and text input. File upload items are represented as instances of `starlette.datastructures.UploadFile`."*?

lovelydinosaur (changes requested):
Great stuff! Some thoughts here on how we might slim it down a little.

docs/requests.md
tiangolo: Sure! Removed.

docs/requests.md
tiangolo: Perfect. I used the same from above, updated to include this.

docs/requests.md
tiangolo: Sure. Removed. :fire: 

Do you want to keep the:

> E.g., `await myfile.seek(0)` would go to the start of the file.

docs/requests.md
tiangolo: Done :heavy_check_mark: .

docs/requests.md
tiangolo: Sure. Updated it. Let me know if you prefer to delete the details about the <a href="https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile" target="_blank">`SpooledTemporaryFile`</a>.

docs/requests.md
tiangolo: Sure thing.

docs/requests.md
lovelydinosaur: I think we can keep this example really short...

```python
form = await request.form()
filename = form["upload_file"].filename
contents = await form["upload_file"].read()
```

Don't really need to include a complete endpoint - just a snippet similar to the above would be fine.

Thanks so much for this!

tiangolo: Get it.

Done! :heavy_check_mark:

> Thanks so much for this!

That's the least I can do :blush: . Thanks for Starlette! :rocket:

lovelydinosaur:
Looks great! Let's get it in.

tiangolo:
Awesome! Thanks 🎉

PR #422, opened 2019-03-06
reqs: minor cleanup

blueyed:
- codecov is only required on CI/Travis, install it there then.
- databases is a requirement for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,5 @@ script:
     - scripts/test
 
 after_script:
+    - pip install codecov
     - codecov
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 # Optionals
 aiofiles
-databases[sqlite]
 graphene
 itsdangerous
 jinja2
@@ -12,7 +11,7 @@ ujson
 # Testing
 autoflake
 black
-codecov
+databases[sqlite]
 isort
 mypy
 pytest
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Optionals
 aiofiles
-databases[sqlite]
+asyncpg
```
lovelydinosaur: Is `asyncpg` intentional here?

requirements.txt
```diff
@@ -12,11 +12,12 @@ ujson
 # Testing
 autoflake
 black
-codecov
+databases[sqlite]
 isort
 mypy
 pytest
 pytest-cov
+sqlalchemy
```
lovelydinosaur: This is a requirement of `databases` anyways - we don't need to include it additionally, right?

.travis.yml
```diff
@@ -16,4 +16,5 @@ script:
     - scripts/test
 
 after_script:
+    - pip install codecov
```
lovelydinosaur, line 19: Fair enough.

requirements.txt
blueyed: Oh, moved it there since it was in setup.py, but obsolete since e396f07.

One good reason for havings reqs in one place I guess.

requirements.txt
blueyed: ok

blueyed: fixed.

lovelydinosaur:
Ta!

PR #423, opened 2019-03-06
WIP: add tox config, revisit coverage/pytest config

blueyed:
Main motivation: have QA failure reported in a separate/single job on Travis.

But tox allows for easier isolated running of tests in general.

### DIFF
--- a/tests/.ignore_lifespan
+++ b/tests/.ignore_lifespan
@@ -1,3 +1,0 @@
-[coverage:run]
-omit =
-  starlette/middleware/lifespan.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,24 @@
-sudo: required
 dist: xenial
 language: python
 
 cache: pip
 
-python:
-    - "3.6"
-    - "3.7"
-    - "3.8-dev"
+jobs:
+    include:
+        - env: TOXENV=py36-coverage
+          python: "3.6"
+        - env: TOXENV=py37-coverage
+          python: "3.7"
+        - env: TOXENV=py38-coverage
+          python: "3.8-dev"
+        - env: TOXENV=mypy,autoflake,black
+          python: "3.7"
 
 install:
     - pip install -U -r requirements.txt
 
 script:
-    - scripts/test
+    - tox
 
 after_script:
     - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: python
 
 cache: pip
 
+env:
+    global:
+        - PYTEST_ADDOPTS=--cov-report=xml
+
 jobs:
     include:
         - env: TOXENV=py36-coverage
@@ -21,4 +25,9 @@ script:
     - tox
 
 after_script:
-    - codecov
+    - |
+      # Upload coverage.
+      if [[ -f coverage.xml ]]; then
+          pip install codecov
+          codecov -f coverage.xml
+      fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ jobs:
           python: "3.7"
 
 install:
-    - pip install -U -r requirements.txt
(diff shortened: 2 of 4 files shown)


blueyed:
What do  you think about moving requirements into `extras_require`?
I think extras for "testing", "optional", "qa", and "docs" might make sense.  "full" could be removed to "optional" then.

tests/.ignore_lifespan
```diff
@@ -1,3 +0,0 @@
-[coverage:run]
-omit =
-  starlette/middleware/lifespan.py
```
blueyed, line 3: I found it very confusing to have this instead of `.coveragerc` - moved it to setup.cfg.

blueyed:
Coverage is unaffected: https://codecov.io/github/encode/starlette/commit/790c566714ba34d7fdcec3454dadd8c89bfc9335

lovelydinosaur:
> Main motivation: have QA failure reported in a separate/single job on Travis.

Sorry - I'm not sure what you mean here.

blueyed:
> Sorry - I'm not sure what you mean here.

Currently QA (mypy, …) is just run at the end of every job, but
a) it is only necessary once
b) you cannot see from build jobs, that failure was just because of this

Therefore it is better to have a single (Travis) job that runs those - then the test jobs are ok, but the QA one is not.

lovelydinosaur:
* I'd rather not move to using tox. We don't have lots of language-version issues to worry about here, and I'm perfectly happy for development to mainly test against a fixed version, with multiple versions tested in CI only.
* I'd be happy to see an alternate travis configuration that only ran the linting in one separate case, but I'd rather see a more minimal diff towards that.
* The ignore lifespan hack is something we'll be able to drop anyways so let's leave that for a moment - we just needed a temporary backport of some stuff until `responder` pinned it's Starlette dependencies.

blueyed:
> * I'd rather not move to using tox. We don't have lots of language-version issues to worry about here, and I'm perfectly happy for development to mainly test against a fixed version, with multiple versions tested in CI only.

It does not mean to move there, but just allows everybody to easily run tests as on Travis - it is easier to grasp than the custom test runner you currently have (but which I've left alone).
tox is rather popular, especially for pytest based projects.


> * The ignore lifespan hack is something we'll be able to drop anyways so let's leave that for a moment - we just needed a temporary backport of some stuff until `responder` pinned it's Starlette dependencies.

It makes sense to have coverage configuration in general - it was just the oddly named file that confused me.
It is better to have the config there, instead of `--cov…` options hidden in some script.

lovelydinosaur:
> It makes sense to have coverage configuration in general

Happy to consider it, but maybe let's keep it as a separate PR.

Keep in mind that I *also* need to balance these requests against wanting to keep a consistent style across each of `uvicorn`, `starlette`, `databases`, `typesystem` - So I have a higher barrier to what I'll consider for changes there. I'm quite keen on the lightweight style for tests at the moment. Not 100% against moving to tox, but a change with the most minimal footprint possible is more likely to convince me to switch.

lovelydinosaur:
> I think extras for "testing", "optional", "qa", and "docs" might make sense. "full" could be removed to "optional" then.

Possibly yes. Something like `"dev"`, `"docs"`, `"optionals"` might be okay.

blueyed:
> > It makes sense to have coverage configuration in general
> 
> Happy to consider it, but maybe let's keep it as a separate PR.

Where should it go? .coveragerc, setup.cfg (tox.ini is out of the question for now)?

blueyed:
> Keep in mind that I _also_ need to balance these requests against wanting to keep a consistent style across each of `uvicorn`, `starlette`, `databases`, `typesystem` - So I have a higher barrier to what I'll consider for changes there.

Maybe https://github.com/zsimic/setupmeta/ is worth considering then to ease this in general?!

PR #424, opened 2019-03-06
setup.py: use extras_require more

blueyed:
Ref: https://github.com/encode/starlette/pull/423#issuecomment-470277919


.travis.yml
```diff
@@ -10,7 +10,7 @@ python:
     - "3.8-dev"
 
 install:
-    - pip install -U -r requirements.txt
+    - pip install -U -e '.[full]'
```
lovelydinosaur, line 13: Is it possible for us to keep the dev + docs stuff out of setup.py, but within the `requirements.txt` maybe?

Either way around I don't think we should install them with `pip install starlette[full]`, since they'e not in any way requirements for the starlette package itself.

PR #428, opened 2019-03-07
doc: fixes for Databases chapter

blueyed:


docs/database.md
```diff
@@ -198,20 +198,20 @@ import app
...
-  yield                            # Run the tests.
-  drop_database(url)               # Drop the test database.
+    """
+    Create a clean database on every test case.
+    For safety, we should abort if a database already exists.
+
+    We use the `sqlalchemy_utils` package here for a few helpers in consistently
+    creating and dropping the database.
+    """
+    url = str(app.TEST_DATABASE_URL)
```
blueyed, line 208: Also below - if this makes sense.  Feel free to amend/edit yourself from here.

blueyed: Pushed.

docs/database.md
```diff
@@ -282,8 +282,7 @@ import app
 
 @pytest.fixture(scope="session", autouse=True)
 def create_test_database():
-    url = str(app.DATABASE_URL)
-    engine = create_engine(url)
+    url = str(app.TEST_DATABASE_URL)
     assert not database_exists(url), 'Test database already exists. Aborting tests.'
     create_database(url)             # Create the test database.
     config = Config("alembic.ini")   # Run the migrations.
```
blueyed, line 288: I also need `config.set_main_option("sqlalchemy.url", url)` here, since this is removed from the ini, and does not appear to come via the env.

```python
    @pytest.fixture(scope="session", autouse=True)
    def create_test_database():
        from alembic import command
        from alembic.config import Config

        url = str(settings.TEST_DATABASE_URL)
        # assert not database_exists(url), 'Test database already exists. Aborting tests.'
        create_database(url)             # Create the test database.
        config = Config("alembic.ini")   # Run the migrations.
>       command.upgrade(config, "head")

tests/conftest.py:40:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.venv/lib/python3.7/site-packages/alembic/command.py:276: in upgrade
    script.run_env()
.venv/lib/python3.7/site-packages/alembic/script/base.py:475: in run_env
    util.load_python_file(self.dir, "env.py")
.venv/lib/python3.7/site-packages/alembic/util/pyfiles.py:90: in load_python_file
    module = load_module_py(module_id, path)
.venv/lib/python3.7/site-packages/alembic/util/compat.py:156: in load_module_py
    spec.loader.exec_module(module)
migrations/env.py:83: in <module>
    run_migrations_online()
migrations/env.py:68: in run_migrations_online
    poolclass=pool.NullPool,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

configuration = {'here': '…/project', 'script_location': 'migrations'}, prefix = 'sqlalchemy.', kwargs = {'poolclass': <class 'sqlalchemy.pool.impl.NullPool'>}
options = {'_coerce_config': True, 'poolclass': <class 'sqlalchemy.pool.impl.NullPool'>}

    def engine_from_config(configuration, prefix="sqlalchemy.", **kwargs):
        …

        options = dict(
            (key[len(prefix) :], configuration[key])
            for key in configuration
            if key.startswith(prefix)
        )
        options["_coerce_config"] = True
        options.update(kwargs)
>       url = options.pop("url")
E       KeyError: 'url'

.venv/lib/python3.7/site-packages/sqlalchemy/engine/__init__.py:497: KeyError
```

lovelydinosaur: Okay, is that possibly because `config.set_main_option('sqlalchemy.url', str(app.DATABASE_URL))` in the env should instead be:

```python
if app.TESTING:
    config.set_main_option('sqlalchemy.url', str(app.TEST_DATABASE_URL))
else:
    config.set_main_option('sqlalchemy.url', str(app.DATABASE_URL))
```

lovelydinosaur: Or does that still not resolve the issue?

blueyed: Yes, or `app.database._url`.
(I got the KeyError because I've had the actual `config.set_main_option('sqlalchemy.url', …` commented, likely when I've noticed that it used the real db)

Maybe the following pattern in `settings` is better?
```python
TESTING = config("TESTING", cast=bool, default=False)

DATABASE_URL = config("DATABASE_URL", cast=databases.DatabaseURL)
if TESTING:
    DATABASE_URL = DATABASE_URL.replace(database="test_" + DATABASE_URL.database)
```
I.e. do not have a separate setting (`TEST_DATABASE_URL`) in the first place?
But having it on the other hand allows to override this separately (via env), and makes it clearer that it is the test db really.

btw: the doc uses `DATABASE_URL=sqlite:///test.db`, which is a bit confusing, since you have "test.db" and "test_test.db" then.  It could use "app.sqlite" there maybe instead? (also changing the ext).

lovelydinosaur: > Yes, or app.database._url.

We should change that to be a public API `database.url`, and then update the docs.

Also the test suite can then do `assert database.url == TEST_DATABASE_URL` before it runs anything.

lovelydinosaur: > sqlite:///test.db

Yeah `database.sqlite` or something would be better.

lovelydinosaur:
@blueyed Okay, so `database.url` is now public API in the latest release of `databases`.

It returns a `DatabaseURL` instance, so I guess the relvant bits in the docs will want this...

```python
config.set_main_option('sqlalchemy.url', str(database.url))
```

docs/database.md
```diff
@@ -258,7 +258,7 @@ config = context.config
 
 # Configure Alembic to use our DATABASE_URL and our table definitions...
 import app
-config.set_main_option('sqlalchemy.url', str(app.DATABASE_URL))
+config.set_main_option('sqlalchemy.url', str(app.database._url))
```
lovelydinosaur, line 261: `str(app.database.url)`

PR #434, opened 2019-03-11
Populate SCRIPT_NAME in WSGIMiddleware

### LINKED ISSUE #430: Path prefix for PATH_INFO with WSGIMiddleware
When mounting an app (e.g. Django) through `WSGIMiddleware` it seems to be required to also adjust `PATH_INFO` accordingly.

With the following, Django would e.g. redirect to "/admin/" (without the outer "/dj" prefix).

```python
djapp = django.core.wsgi.get_wsgi_application()
app.mount("/dj", app=WSGIMiddleware(djapp))
```

I've got this working through an additional "path_prefix" argument for WSGIMiddleware, but I think it makes sense to either prepend the mount path in `build_environ` always (using `"PATH_INFO": scope["root_path"] + scope["path"]`, and making it visible to the mounted app though), or handle e.g. "Location" response headers by prepending it there.
The latter could likely result in redirect-loops, and does not help with Django POSTing to "/admin/login" then.

I've not investigated too much into this yet, what do you think?

lovelydinosaur:
Closes #430

### DIFF
--- a/starlette/middleware/wsgi.py
+++ b/starlette/middleware/wsgi.py
@@ -13,7 +13,7 @@ def build_environ(scope: Scope, body: bytes) -> dict:
     """
     environ = {
         "REQUEST_METHOD": scope["method"],
-        "SCRIPT_NAME": "",
+        "SCRIPT_NAME": scope.get("root_path", ""),
         "PATH_INFO": scope["path"],
         "QUERY_STRING": scope["query_string"].decode("ascii"),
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
See https://asgi.readthedocs.io/en/latest/specs/www.html#wsgi-compatibility

blueyed:
Thanks!

PR #435, opened 2019-03-12
ASGI 3

lovelydinosaur:
Refs https://github.com/django/asgiref/pull/78 and https://github.com/django/asgiref/pull/80

In short:

* ASGI 2 defines the interface as: `a callable taking 'scope' that returns an async callable taking 'receive' and 'send' arguments'
* ASGI 3 defines the interface as `an async callable taking 'scope', 'receive' and 'send' arguments'.

So previously...

```python
def hello_world(scope):
    async def inner(receive, send):
        ...
    return inner
```

Would become more simply, this:

```python
async def hello_world(scope, receive, send):
    ...
```

It's a big change to have to make throughout the stack, and about as late in the day as we'd still be able to get away with it, but it'll be worth it for the simpler interface, clearer tracebacks, and easier-to-write applications and middleware classes.

See also https://github.com/encode/uvicorn/pull/325


lovelydinosaur:
Would need a version bump to 0.12.0.

lovelydinosaur:
@florimondmanca @tiangolo Particularly relevant to you folks.

(Are FastAPI and Bocadillo pinned against the current 0.11.x release cycle?)

I'll prob. coordinate with responder folks to make the upgrade there, once this is in.

florimondmanca:
Big change indeed! But agreed that the simpler interface is worth it. Bocadillo is safe as it has Starlette pinned to `starlette>=0.11, <0.12`. Thanks for the heads up!

mbeacom (approved):
🙌

tiangolo:
Sorry for the delay and thanks for the ping.

Yep, FastAPI is pinned too to `11.x`. :heavy_check_mark:

lovelydinosaur:
Okay, we'd best just push on with this.

Might roll a `-beta` prerelease. We probably don't want a general release until either or both of hypercorn, daphne have support for ASGI 3.

PR #441, opened 2019-03-19
[RFC] Add URL.path_query_fragment

blueyed:
This is helpful for getting the part after scheme, user, host etc.,
e.g. for local redirects.

While it is easy to wrap this with an own helper I believe it is used often in general.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -95,6 +95,17 @@ class URL:
     def is_secure(self) -> bool:
         return self.scheme in ("https", "wss")
 
+    @property
+    def path_query_fragment(self) -> str:
+        r = self.path
+        query = self.query
+        if query:
+            r += '?' + query
+        fragment = self.fragment
+        if fragment:
+            r += '#' + fragment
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
The query fragment is a client-side thing - it never gets sent to the server.
We *probably* don't want to include it because of that.

blueyed:
Yes, if it is not sent/known then it will also not be included.
But URL can also be used to build client-side URLs, e.g. links to anchors.

Would you take a `path_query` property then?

blueyed:
@tomchristie 
What about `path_query`?

It is kind of tedious to do this manually for redirect responses etc.

lovelydinosaur:
I think a `full_path` property would be absolutely fine.

(And I guess we *may as well* include the fragment in there too if one exists, tho we may want a code comment or two in `fragment` to make it clearer to users that it doesn't normally end up server-side)

blueyed:
I've figured that using `replace` works better in my use case anyway.

Issue #433, opened 2019-03-11
url_for for one endpoint with combinations of routes 


em92:
```python
from starlette.applications import Starlette
from starlette.responses import PlainTextResponse


app = Starlette()


@app.route("/")
@app.route("/{page:int}")
@app.route("/cat/{cat}")
@app.route("/cat/{cat}/{page:int}")
def http_method(request):
    cat = request.path_params.get("cat", "")
    page = request.path_params.get("page", 0)
    return PlainTextResponse("cat: {}, page: {}".format(cat, page))


print(app.url_path_for("http_method", cat=None, page=1))  # expected /1
print(app.url_path_for("http_method", cat="name", page=1))  # expected /cat/name/1
print(app.url_path_for("http_method", cat="name", page=None))  # expected /cat/name
```

First print does not match with expected, second print matches, third one throws exception

em92:
Can be fixed with:
```diff
diff --git a/starlette/routing.py b/starlette/routing.py
index 15786b1..8118032 100644
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -190,6 +190,7 @@ class Route(BaseRoute):
         return Match.NONE, {}
 
     def url_path_for(self, name: str, **path_params: str) -> URLPath:
+        path_params = {k: v for k, v in path_params.items() if v is not None}
         seen_params = set(path_params.keys())
         expected_params = set(self.param_convertors.keys())
 ```

If it is acceptable, I suggest to add PR with test cases yourself @tomchristie. I am quite busy at work.

**Comment by em92** [comment]
Temporary solution, before PR is merged:

```python
from starlette.routing import Route

def fixed_url_path_for(self, name: str, **path_params: str):
    path_params = {k: v for k, v in path_params.items() if v is not None}
    return self.bugged_url_path_for(name, **path_params)


Route.bugged_url_path_for = Route.url_path_for
Route.url_path_for = fixed_url_path_for
```

lovelydinosaur:
I'm not convinced that we should change the behavior...

The following do not work:

```python
print(app.url_path_for("http_method", cat=None, page=1))  # expected /1
print(app.url_path_for("http_method", cat="name", page=1))  # expected /cat/name/1
print(app.url_path_for("http_method", cat="name", page=None))  # expected /cat/name
```

But these do...


```python
print(app.url_path_for("http_method", page=1))  # expected /1
print(app.url_path_for("http_method", cat="name", page=1))  # expected /cat/name/1
[...]

PR #444, opened 2019-03-22
 Take out path_params items with None value in url_path_for

em92:
Fixes https://github.com/encode/starlette/issues/433


lovelydinosaur:
Closing as per #433

Issue #440, opened 2019-03-19
pytest event loop is already running


ryananguiano:
Because the test client calls `loop.run_until_complete(connection(receive, send))`, I cannot use anything that modifies the event loop in a pytest fixture without getting `RuntimeError: This event loop is already running`.

I would like to use a package like aresponses to mock aiohttp requests like this:
```
@pytest.fixture
def mock_endpoint(aresponses):
    aresponses.add('mock.com', '/test/', 'get', 'this is my mock response')
```

After messing with it for a couple hours, the only way I was able to get successful tests was to wrap the application in middleware:
```
@pytest.fixture
def app():
    class App:
        def __init__(self, app):
            self.app = app

        def __call__(self, scope):
            return functools.partial(self.asgi, scope=scope)

        async def asgi(self, receive, send, scope):
            async with ResponsesMockServer() as aresponses:
                aresponses.add('mock.com', '/test/', 'get', 'this is my mock response')
                inner = self.app(scope)
                await inner(receive, send)
    return App(application)


@pytest.fixture
def test_client(app):
    with TestClient(app) as test_client:
        yield test_client

```

I am going to modify this to allow myself to dynamically pass responses through the `@pytest.mark` decorator, but this workflow is not really convenient at all.

Am I missing something here? Is there a better way to set up my test clients, or should I just keep going down this route?

Thanks

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/440">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/440/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/440/pledge.svg">
</picture>
</a>

ryananguiano:
As a follow up, I was able to get to a much cleaner workflow with this code:

```python
import functools
import pytest
from aresponses import ResponsesMockServer


@pytest.fixture
def mocked_responses_app(request):
    mocked_responses = []
    for marker in request.node.iter_markers('mocked_responses'):
        if marker and marker.args:
            if len(marker.args) == 1 and isinstance(marker.args[0], (list, tuple)):
                responses = marker.args[0]
            else:
                responses = marker.args
[...]

PR #447, opened 2019-03-23
Draft for the async test client (#440)

gvbgduh:


gvbgduh:
Based on `requests_async`. ASGI 2 is not supported for now. Works together with `@pytest.mark.asyncio` that runs the event loop.

lovelydinosaur:
So, a suggestion:

I think for this incoming iteration we should maybe stop using `async with TestClient()` for the lifespan handling. There's no particular reason we need to be bundling Lifespan handling up in the same class as the HTTP client. Perhaps a standalone `async with AppLifespan(app):` would be better? (Perhaps also with an API for non-context-manager style usage)

gvbgduh:
Yes, that sounds reasonable, it'll give much better granularity as having the lifespan task in the loop might not always be convenient. 
On the other hand, having the lifespan handled helps to take care of the app events (eg. `startup` or `shutdown`), so the app can be more closely replicated within the test case. I'm very happy to follow this.

Having said that, I still might need a bit of clarity as I have some architectural confusion.
It probably should be an additional test client (like AsyncTestClient or anything else?). Mostly for backwards compatibility with existing test cases. And if so `requests_async.ASGISession` takes care of everything for this already, but is there need for an alias within the `starlette`?
Or, more generally, should `requests_async` be brought into dependencies?

This might also impact `AppLifespan` implementation as it seems it can inherit and extend `requests_async.ASGISession` quite well, so the same module [`requests_async.asgi`] might be a good place. But I might be missing bits of the whole picture. 

Therefore, some advice would be much appreciated and I can target my PR more precisely.

jordic:
Hey just to share with yours, we just developed this small asgi test client, 
https://github.com/vinissimus/async-asgi-testclient
(Only for understanding better the main spec, but we are using it)

lovelydinosaur:
Right now let's punt on this.
Yes we *will* want to point at `requests_async` or something as an async test client, but I'm also confident that pulling the lifespan stuff out of the rest of the test client is *also* going to be the right thing for us to do, so let's keep those bits seperate.

PR #449, opened 2019-03-25
Adding a `%` to quote_plus's safe in RedirectResponse 

archived001:
Fix the bug when `%` is quoted

```
>>> quote_plus(str("http://localhost:8000/?var=dsfs%20sd"), safe=":/#?&=@[]!$&'()*+,;")
'http://localhost:8000/?var=dsfs%2520sd'
```

https://github.com/huge-success/sanic/issues/1356


lovelydinosaur:
Gotcha, thanks!

PR #455, opened 2019-03-28
Add py.typed to package

### LINKED ISSUE #452: py.typed missing in published artifacts
I didn’t check for earlier versions, but at least 0.11.4 on PyPI does not include `py.typed`. I assume this is an oversight, given it is mentioned in `setup.py`?

https://github.com/encode/starlette/blob/77b84a08c1e4de0db64a197b58ac363a26c51d4f/setup.py#L49

uranusjr:
This package would be installed along with the package (since it is listed in `package_data`), to let mypy pick up type hints in the package source.

Quoting [Mypy’s documentation](https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages) on how to create a PEP 561 compatible package:

> [P]ackages that supply type information via type comments or annotations in the code should put a py.typed in their package directory.

The `package_data` entry is already added in #253.

Fix #452.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,4 +72,5 @@ setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
     ],
+    zip_safe=False,
 )


lovelydinosaur:
> [P]ackages that supply type information via type comments or annotations in the code should put a py.typed in their package directory.

😕 Not sure I’m up for uglifying things like that.

uranusjr:
Well that’s what the standard dictates 🤷‍♂️

Without that there’d be no way for users (read: me) to use the annotations, and have to either disable it when calling any Starlette functions, or duplicating the work. Both are unfortunate IMO given how well type hinting is already set up in Starlette.

I guess there are more beautiful ways to distribute the annotations, e.g. Build them into as a stub-only package (like how TypeScript stubs are distributed on NPM), or hack `setup.py` to generate that file at build-time. All are unnecessarily complicating the situation though, IMO.

lovelydinosaur:
I don’t get why on earth it’s required? We have our annotations in the source code, right there. Why does mypy dictate that we introduce an oddly named module in order to make use of them?
I think I’m not getting something here, because it seems like an unreasonable requirement for mypy to make.

lovelydinosaur:
Gimme a little while to get used to the idea 😑

lovelydinosaur:
‪I don’t get it. Can’t mypy do: “if a third party package has type annotations, then check them. If it doesn’t, don’t.”‬?

florimondmanca:
I am confused too. I’ve using mypy to check typing in most of my recent projects, a lot of which use libraries that provide type annotations like Starlette. I believe I am able to have mypy read those even though the package does not have a py.typed file — eg showing unknown attribute access on Starlette’s Request.

Edit: it seems that mypy requires this file because it can’t otherwise know at a glance which packages should be type checked and which shouldn’t without causing “spurious errors”. See the rationale in PEP 561: https://www.python.org/dev/peps/pep-0561/

I’d agree it is not terribly elegant. I don’t see why not comply with the spec though.

woile:
I started reading about this, and to my surprise, I think we are also missing this:

```
zip_safe=False,  # https://mypy.readthedocs.io/en/latest/installed_packages.html
```

[example](https://github.com/samuelcolvin/pydantic/blob/master/setup.py#L74)

lovelydinosaur:
Let’s start by considering this as any other issue: what exactly is the behaviour with mypy without py.typed? What is the expected behaviour? What’s a minimal replicable example?

florimondmanca:
Here you go @tomchristie :

- Current behaviour: if code submitted to mypy for type checking contains Starlette imports, mypy cannot find type annotations for the imported modules.
- Expected behaviour: it should be able to find those.
- Minimal replicable example (copied from Starlette's quickstart app):

```python
# app.py
import uvicorn

app = Starlette(debug=True)

@app.route('/')
async def homepage(request):
    return JSONResponse({'hello': 'world'})

if __name__ == '__main__':
```

```bash
$ mypy app.py
app.py:1: error: Cannot find module named 'starlette.applications'
app.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
app.py:2: error: Cannot find module named 'starlette.responses'
app.py:3: error: Cannot find module named 'uvicorn'
```

(So it seems uvicorn has the same issue.)

>  I believe I am able to have mypy read those even though the package does not have a py.typed file — eg showing unknown attribute access on Starlette’s Request.

I realise that in Bocadillo I have a `mypy.ini` file that sets `ignore_unknown_imports = True`. As a result, the above errors aren't shown and the autocompletion I mentioned in this earlier answer doesn't come from mypy, but from another part of the VSCode Python extension.

lovelydinosaur:
Right, well. Let's get `zip_safe=False` into this PR as well.

I don't know that I'm wild about it still, will have to take a call sometime tho.

uranusjr:
`zip_safe` flag added

starlette/py.typed
```diff
@@ -0,0 +1 @@
+
```
blueyed: just a nitpick: doesn't need an empty line, does it?

uranusjr: Whatever makes you happy

matthewspeck:
Been looking forward to this. Any chance it'll be available in a release soon?

lovelydinosaur:
Yup, soon.

PR #458, opened 2019-04-01
Implement custom convertor registration

### LINKED ISSUE #231: Add user-defined converters to routing
To solve application task in https://github.com/encode/starlette/issues/230 I use my own defined convertor:
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import starlette.convertors as conv

class UnderscoreConvertor(conv.Convertor):
    regex = ".*"

    def convert(self, value: str) -> str:
        return "_" + value

    def to_string(self, value: str) -> str:
        return value[1:]

conv.CONVERTOR_TYPES['underscore'] = UnderscoreConvertor()



import uvicorn

app = Starlette()

@app.route("/test/{username:underscore}")
def http_test(request):
    return PlainTextResponse(username + "\n")

if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=8080)
```

So feature request is allowing user-defined converters in more acceptable way. Something like:
`starlette.convertors.add("underscore", UnderscoreConverter())`

uranusjr:
A very low-effort implementation to resolve #231.


tests/test_converters.py
```diff
@@ -0,0 +1,71 @@
...
+    def convert(self, value: str) -> bytes:
+        return bytes.fromhex(value)
+
+    def to_string(self, value: bytes) -> str:
+        return value.hex()
+
+
+@pytest.fixture(autouse=True)
+def reset_convertors(request):
+    request.addfinalizer(reset)
```
lovelydinosaur: What is `request` and `addfinalizer` here?

starlette/convertors.py
```diff
@@ -61,9 +61,20 @@ def to_string(self, value: typing.Any) -> str:
...
     "str": StringConvertor(),
     "path": PathConvertor(),
     "int": IntegerConvertor(),
     "float": FloatConvertor(),
 }
+
+CONVERTOR_TYPES = dict(_DEFAULT_CONVERTOR_TYPES)
+
+
+def add(key: str, convertor: Convertor) -> None:
```
lovelydinosaur: Probably `register_url_convertor` or something like that.

starlette/convertors.py
```diff
@@ -61,9 +61,20 @@ def to_string(self, value: typing.Any) -> str:
...
+    CONVERTOR_TYPES[key] = convertor
+
+
+def reset() -> None:
```
lovelydinosaur: I'm not sure `reset` should be part of the public API - we really only need it for the purposes of our own tests, right? Perhaps we should just have `URL_CONVERTORS` and `register_url_convertor`?

lovelydinosaur:
Yeah great call!
Really we ought to expand out https://www.starlette.io/routing/ too, but happy to pick that up at a later date if you like.

starlette/convertors.py
uranusjr: Yes it’s only for tests, I guess I can just move this to `test_convertors.py` (and I just realised I spelled that filename wrong).

tests/test_converters.py
uranusjr: `request` is a magic Pytest argument to access the “test context”. `request.addfinalizer` is a common pattern to [register clean up code](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code); this fixture essentially calls `reset` after every test to ensure the environment is clean.

florimondmanca: My two cents: can this also be achieved with a `yield` fixture?

```python
@pytest.fixture(autouse=True)
def reset_converters_after_test():
    yield
    reset()
```

uranusjr: I believe so, yes.

lovelydinosaur: Good call @florimondmanca  - I'd prefer that, since "request" is *really* ambiguous.

steinnes:
Can I help in some way to push this forward?  I've really missed being able to register my own convertors :-)

lovelydinosaur:
I just need a gear change from pushing on `httpcore` to get back onto starlette, uvicorn, databases etc. for a bit. Oughta be taking a look through next week.

PR #459, opened 2019-04-01
Don't block the event loop in WebSocketTestSession

patrys:
It's typical for event-loop-based servers to try to do a blocking receive in a while loop. `Queue.get()` is blocking in a synchronous way and it does not yield control back to the asyncio executor. Let's explicitly yield control until the queue is no longer empty.

Warning: without the fix in place the test will hang indefinitely.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -297,6 +297,8 @@ class WebSocketTestSession:
             self._send_queue.put(exc)
 
     async def _asgi_receive(self) -> Message:
+        while self._receive_queue.empty():
+            await asyncio.sleep(0)
         return self._receive_queue.get()
 
     async def _asgi_send(self, message: Message) -> None:
--- a/tests/test_testclient.py
+++ b/tests/test_testclient.py
@@ -1,8 +1,10 @@
+import asyncio
 import pytest
 
 from starlette.applications import Starlette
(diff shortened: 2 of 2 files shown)


lovelydinosaur:
Great, let's have this!

Given https://github.com/encode/requests-async we may end up moving to fully async test clients instead - but I see no reason not to have this in place until then.

lovelydinosaur:
Push 0.12.0.b2 to pypi - You'll need to install with `pip install --pre` since it's labelled as a beta pre-release (ASGI 3).

Ideally we'd wait for Daphne and Hypercorn releases before we drop the beta label. eg. https://github.com/django/daphne/pull/255

martin82:
This merge still causes 100% CPU use when receiving due to the busy-waiting loop. Maybe it could be a blocking self._receive_queue.get call in a separate executor instead?
I solved this for myself by replacing the _asgi_receive function body with 
`return await self._loop.run_in_executor(None, self._receive_queue.get)`
but maybe it should be a pre-allocated single thread executor instead of the default.

patrys:
It does use 100% CPU but it does not block the event loop which was a bigger issue since under certain scenarios it would lock the test suite up and hang forever. I believe @tomchristie plans to make the test client asynchronous using `http3` so we won't have to have a busy loop at all.

PR #461, opened 2019-04-02
Make StreamingResponse compatible with normal iterators

tiangolo:
Make `StreamingResponse` compatible with normal iterators.

Continuing from #457, https://github.com/encode/starlette/pull/457#issuecomment-478965259

---

I also removed the mention to `iterator_to_async` from the docs, only mention that it supports both async and normal generators.

> I'd also be interested in if there's any nice ways around we could use something like this and drop `aiofiles` [...]

Let's see what do you think about this one and then I'll take a stab at that :smile:

### DIFF
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -37,7 +37,7 @@ def _interceptable_next(iterator: Iterator) -> Any:
         raise _StopSyncIteration
 
 
-async def iterator_to_async(iterator: Iterator) -> AsyncGenerator:
+async def iterate_in_threadpool(iterator: Iterator) -> AsyncGenerator:
     while True:
         try:
             result = await run_in_threadpool(_interceptable_next, iterator)
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import typing
+from typing import Any, AsyncGenerator, Iterator
 
 try:
     import contextvars  # Python 3.7+ only.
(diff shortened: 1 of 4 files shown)


starlette/concurrency.py
```diff
@@ -22,3 +23,24 @@
...
+
+def _interceptable_next(iterator: Iterator) -> Any:
+    try:
+        result = next(iterator)
+        return result
+    except StopIteration:
+        raise _StopSyncIteration
+
+
+async def iterator_to_async(iterator: Iterator) -> AsyncGenerator:
```
lovelydinosaur: How about `iterate_in_threadpool` to mirror `run_in_threadpool`.

tiangolo: Great idea. I struggled a lot to name it and wasn't quite convinced either.

lovelydinosaur:
Do we strictly need the `_StopSyncIteration` indirection?

lovelydinosaur:
Can we not catch StopIteration directly?

tiangolo:
`StopIteration` is used internally by coroutines, it was used before to yield values, here's a short summary in the PEP: https://www.python.org/dev/peps/pep-0492/#why-stopasynciteration

~I can think of an alternative using classes, based on this example: https://www.python.org/dev/peps/pep-0492/#example-2, would you prefer that?~

tiangolo:
Because `StopIteration` was used to yield values, catching it directly in `iterate_in_threadpool` is not possible, it is converted to a `RuntimeError`.

Actually, now that I think of it, using a class would still need the indirection, as we need an exception that is propagated up/out even when run inside of `run_in_threadpool`.

lovelydinosaur:
👍 Released as prerelease version 0.12.0b3

tiangolo:
Awesome! Thanks! :tada: :rocket:

PR #466, opened 2019-04-05
Allow to edit headers in request object

em92:
It would allow to make `TrustedProxiesMiddleware` middleware like below
```python
import uvicorn
from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware

app = Starlette()


@app.route("/")
def root(request: Request):
    return PlainTextResponse(request.client.host)


class TrustedProxiesMiddleware(BaseHTTPMiddleware):
        if "x-forwarded-for" in request.headers:
            if request.client.host != "127.0.0.1":
                del request.headers["x-forwarded-for"]
                if request.headers.get("x-forwarded-proto"):
                    del request.headers["x-forwarded-proto"]


app.add_middleware(ProxyHeadersMiddleware)
app.add_middleware(TrustedProxiesMiddleware)

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0")
```

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -4,7 +4,14 @@ import json
 import typing
 from collections.abc import Mapping
 
-from starlette.datastructures import URL, Address, FormData, Headers, QueryParams
+from starlette.datastructures import (
+    URL,
+    Address,
+    FormData,
+    Headers,
+    MutableHeaders,
+    QueryParams,
+)
 from starlette.formparsers import FormParser, MultiPartParser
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Headers on `Request` are *deliberatly* immutable, since it doesn't *really* make sense to pitch up the application logic as "this header wasn't there, never existed".

The headers should accuratly reflect what actually came in over the wire. Middleware can store state or deal with the headers appropriately without needing to mutate them

PR #467, opened 2019-04-08
Corrects usage of CommaSeparatedStrings in docs

em92:
Previous example will give list of one string: `127.0.0.1, localhost`, while it is expected list of two strings `127.0.0.1` and `localhost`


lovelydinosaur:
Gothca, thanks.
Looks like an issue with the .env parsing rather than CommaSeparatedStrings, since...

```python
>>> CommaSeparatedStrings('"127.0.0.1", "localhost"')
CommaSeparatedStrings(['127.0.0.1', 'localhost'])
```

Works correctly, at that level.

PR #471, opened 2019-04-09
doc: fix middleware

blueyed:
It is still broken with regard to `dispatch/dispatch_func` (since 3e2e343) I guess.

Either the `__init__` here should call the parent's constructor, or use `dispatch_func` directly?!

Ref: https://github.com/blueyed/starlette/blob/c2a126ec5797cae28440454c0bf23b8348a630c1/starlette/middleware/base.py#L17


lovelydinosaur:
Thanks!

blueyed:
What about the remaining issue?

> It is still broken with regard to `dispatch/dispatch_func` (since [3e2e343](https://github.com/encode/starlette/commit/3e2e3430d1a08e885325c96205330b13f426fb4c)) I guess.
> 
> Either the `__init__` here should call the parent's constructor, or use `dispatch_func` directly?!

PR #480, opened 2019-04-11
Implement file handling independent of aiofiles, add support for file-like objects

tiangolo:
This PR removes `aiofiles` as a dependency, handling files directly.

It also adds support for file-like objects in `FileResponse` apart from file paths.

This is related to  https://github.com/encode/starlette/pull/461 and @tomchristie's comment at https://github.com/encode/starlette/pull/457#issuecomment-478965259

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,6 @@ For a more complete example, see [encode/starlette-example](https://github.com/e
 Starlette does not have any hard dependencies, but the following are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
-* [`aiofiles`][aiofiles] - Required if you want to use `FileResponse` or `StaticFiles`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
 * [`python-multipart`][python-multipart] - Required if you want to support form parsing, with `request.form()`.
 * [`itsdangerous`][itsdangerous] - Required for `SessionMiddleware` support.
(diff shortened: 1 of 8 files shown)


florimondmanca:
Good stuff!

Any reason why changing `path` to `file_or_path` instead of adding a new `file` parameter, and making sure only one of them is given? Would preserve the current API on `FileResponse`.

tiangolo:
Good point. Many libraries receive a file or path single parameter, this would be equivalent to that. E.g. Pillow: https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.open

But having just an additional parameter would work the same.

@tomchristie what do you prefer?

dhirschfeld:
An argument which can take either a file-like object or a path is quite a common idiom in Python and is, IMHO quite a user-friendly api.

*  https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas-read-csv
* https://docs.scipy.org/doc/numpy/reference/generated/numpy.loadtxt.html#numpy-loadtxt
* https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.savefig.html#matplotlib-pyplot-savefig

Exposing two different arguments to a user and forcing them to choose depending on the type of object they have doesn't strike me as very *pythonic* - Python is duck-typed so the user shouldn't have to care about the type of argument they pass - the class should handle those details internally (as is done in this PR)

Kludex:
Hola @tiangolo :wave: 

Closing this due to recent `anyio` integration and `aiofiles` removal on #1157.

tiangolo:
Nice! Thanks @Kludex for the cleanup. 🙇

PR #488, opened 2019-04-22
Send lifespan.startup.failed if exception occurs during startup

### LINKED ISSUE #486: Error in app startup event is not logged and does not prevent startup
Related #431 #177  and mentioned here https://github.com/encode/uvicorn/issues/333#issuecomment-484885300

Scenario:
I have a 'startup' event handler:
```python
@app.on_event('startup')
async def startup():
    initialise()
```
and I am running under uvicorn for local dev just via the:
```python
if __name__ == '__main__':
```

If there is an uncaught exception in the startup event handler, the exception will be swallowed (not logged) and uvicorn will continue to startup.

The only anomaly seen in the logs is that uvicorn will report:
```
ASGI 'lifespan' protocol appears unsupported
```

I would like to see:
- the exception and traceback logged
- process aborted, since we haven't fully started up yet

florimondmanca:
Fixes #486.

Note: requires https://github.com/encode/uvicorn/pull/350 to work. With the current state of uvicorn (0.7.0), the `ASGI 'lifespan' protocol appears unsupported.` message is still displayed and startup does not abort.

With:

```python
# app.py

app = Starlette()

@app.on_event("startup")
async def startup():
    raise ValueError("Nope!")

```

`$ uvicorn app:app` results in:

```
INFO: Started server process [71364]
INFO: Waiting for application startup.
ERROR: Traceback (most recent call last):
  File "./starlette/routing.py", line 474, in __call__
    await self.startup()
  File "./starlette/routing.py", line 459, in startup
    await handler()
  File "./app.py", line 8, in startup
    raise ValueError("Nope")
ValueError: Nope

ERROR: Application startup failed. Exiting
```

i.e. the error traceback is logged and startup aborts :tada:.

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -1,6 +1,7 @@
 import asyncio
 import inspect
 import re
+import traceback
 import typing
 from enum import Enum
 
@@ -469,8 +470,14 @@ class Lifespan(BaseRoute):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         message = await receive()
         assert message["type"] == "lifespan.startup"
-        await self.startup()
-        await send({"type": "lifespan.startup.complete"})
+        try:
+            await self.startup()
+        except BaseException as exc:
+            msg = traceback.format_exc()
+            await send({"type": "lifespan.startup.failed", "message": msg})
+            raise exc from None
+        else:
+            await send({"type": "lifespan.startup.complete"})
 
         message = await receive()
         assert message["type"] == "lifespan.shutdown"
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -470,15 +470,15 @@ class Lifespan(BaseRoute):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         message = await receive()
         assert message["type"] == "lifespan.startup"
+
         try:
             await self.startup()
-        except BaseException as exc:
+        except BaseException:
             msg = traceback.format_exc()
             await send({"type": "lifespan.startup.failed", "message": msg})
-            raise exc from None
-        else:
-            await send({"type": "lifespan.startup.complete"})
+            raise
 
+        await send({"type": "lifespan.startup.complete"})
         message = await receive()
         assert message["type"] == "lifespan.shutdown"
         await self.shutdown()
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -463,7 +463,10 @@ class TestClient(requests.Session):
     async def wait_startup(self) -> None:
         await self.receive_queue.put({"type": "lifespan.startup"})
         message = await self.send_queue.get()
-        assert message["type"] in ("lifespan.startup.complete", "lifespan.startup.failed")
+        assert message["type"] in (
+            "lifespan.startup.complete",
+            "lifespan.startup.failed",
+        )
 
     async def wait_shutdown(self) -> None:
         await self.receive_queue.put({"type": "lifespan.shutdown"})
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -465,7 +465,7 @@ class TestClient(requests.Session):
         message = await self.send_queue.get()
         if message is None:
             self.task.result()
-        assert message["type"] == "lifespan.startup.complete"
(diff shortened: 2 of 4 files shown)


starlette/testclient.py
```diff
@@ -463,9 +463,7 @@ def __exit__(self, *args: typing.Any) -> None:
     async def wait_startup(self) -> None:
         await self.receive_queue.put({"type": "lifespan.startup"})
         message = await self.send_queue.get()
-        if message is None:
-            self.task.result()
-        assert message["type"] == "lifespan.startup.complete"
+        assert message["type"] in ("lifespan.startup.complete", "lifespan.startup.failed")
```
lovelydinosaur: In the failure case I *think* we should probably wait for another message (which will be `None` as the sentinel value for "all done on this messaging queue now) and then call `self.task.result()` (which will raise an exception)

Tho we'd want to make sure that we're verifying this with a test case that ensures that startup failures bubble out correctly.

lovelydinosaur:
Great work! I think there's maybe a tweak needed on the test client, as described above.
Also, run `scripts/lint` to fix the formatting errors.

starlette/testclient.py
florimondmanca: Okay, I think I've got it fixed. The tests already verify that the `RuntimeError` raised in the startup callback bubbles out of the app, so I think we're good. Let me know!

starlette/routing.py
```diff
@@ -469,8 +470,14 @@ def decorator(func: typing.Callable) -> typing.Callable:
...
         message = await receive()
         assert message["type"] == "lifespan.startup"
-        await self.startup()
-        await send({"type": "lifespan.startup.complete"})
+        try:
+            await self.startup()
+        except BaseException as exc:
+            msg = traceback.format_exc()
+            await send({"type": "lifespan.startup.failed", "message": msg})
+            raise exc from None
```
blueyed: Why is `from None` here used?

starlette/routing.py
```diff
@@ -469,8 +470,14 @@ def decorator(func: typing.Callable) -> typing.Callable:
...
+        else:
+            await send({"type": "lifespan.startup.complete"})
```
blueyed: The `else` is not really needed here.

blueyed (commented):
Nice.
Could use a test I think

starlette/routing.py
florimondmanca: Hm, personal code style preference. It has no functional value; I’ll remove it. :-)

florimondmanca: This and your other comment should be fixed now.

florimondmanca:
You’re right @blueyed! Thought I did add one, but only on the uvicorn PR. Will do. Thanks for reviewing! 👍

blueyed:
Great!
I wonder how/if https://github.com/encode/starlette/pull/505 is affected / should handle this btw.

florimondmanca:
Hi @tomchristie, is there any chance we can get this and https://github.com/encode/uvicorn/pull/350 merged/released soon? It seems users keep encountering the « swallowed exception » issue.

Getting this out here — if you need help with maintenance and/or reviews I’d be happy to give a helping hand.

jkseppan:
Just to confirm, performing the following commands helped me debug why my FastAPI application was not starting:

```
pipenv install -e 'git+https://github.com/florimondmanca/starlette.git@fix/startup-error-handling#egg=starlette'
pipenv install -e 'git+https://github.com/florimondmanca/uvicorn.git@lifespan-startup-failed#egg=uvicorn'
```

The relevant lines in `Pipfile` are 

```
fastapi = {extras = ["all"],version = "==0.33.0"}
starlette = {editable = true,git = "https://github.com/florimondmanca/starlette.git",ref = "fix/startup-error-handling"}
uvicorn = {editable = true,git = "https://github.com/florimondmanca/uvicorn.git",ref = "lifespan-startup-failed"}
```

and exceptions from the code run on the startup event are propagated both with `uvicorn main:app --reload` and `gunicorn main:app --bind 0.0.0.0:80 --workers 2 --worker-class uvicorn.workers.UvicornH11Worker`. I have not so far noticed any downside from using these versions of starlette and uvicorn.

florimondmanca:
Thanks for getting this in, @tomchristie!

PR #491, opened 2019-04-24
TestClient: rewind body always

blueyed:
This is necessary for:

1. middleware creates a StreamingResponse
2. `app.exception_handler(500)` returns a `PlainTextResponse`
3. the middleware's `body_stream` raises the exception with `task.result()` in the end
4. StreamingResponse does not send the final `"more_body": False`

Therefore the test client has an empty content without the seek.

An alternative/additional fix might be to send the `more_body=False`
always in a `finally` clause from the middleware, but making the
TestClient adhere to how this gets handled outside of tests appears to
be good anyway.

TODO:

- [ ] test?


Kludex:
Is there a real code example that will benefit from this?

Kludex:
Closed as it's clearly stale. 😗

PR #492, opened 2019-04-26
Route: inspect signature for `func(request) -> response`

blueyed:
Without this trying to add a route for an app created via the following
would be treated wrongly:

    from channels.http import AsgiHandler
    from asgiref.compatibility import double_to_single_callable

    application = double_to_single_callable(AsgiHandler)

`double_to_single_callable` returns a function:
https://github.com/django/asgiref/blob/e31a26463d002399b3301c92c570de9c9f4a9d1d/asgiref/compatibility.py#L27-L36

Not really sure what I am doing here, and if maybe even `is_double_callable` should/could be used?!

I guess this reflects https://github.com/django/asgiref/commit/fae4ef0#diff-89559d52e4cc62e36fe3d457cea26194L43 somehow.
/cc @andrewgodwin

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -149,8 +149,9 @@ class Route(BaseRoute):
         self.name = get_name(endpoint) if name is None else name
         self.include_in_schema = include_in_schema
 
-        if inspect.isfunction(endpoint) or inspect.ismethod(endpoint):
-            # Endpoint is function or method. Treat it as `func(request) -> response`.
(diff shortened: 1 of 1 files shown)


blueyed:
Also only style issues in CI - I assume black was updated maybe?

Kludex:
I didn't understand the issue, and no one commented for more than two years. For that reason, I'm closing the issue.

In case we should take this in consideration, please let's discuss over our Gitter server or open an issue. :)

Issue #497, opened 2019-05-01
Store single request object in scope


blueyed:
I've thought about doing so with regard to the templating, so that it does not have to be passed in via `context`, but https://github.com/encode/starlette/issues/495 looks like it would also help fixing state in this regard.

For templating it would be about saving the request there, and this could then be used instead of creating new objects then.

lovelydinosaur:
Storing a request inside the scope isn't ever a useful thing to do, because the scope itself gives us everything that we need in order to instantiate a request instance.

If we have the scope available then we can always just do `request = Request(scope=scope)`

A good way to think of this is that the scope *is the state itself*, and that a Request instance is *an interface onto that state*.

blueyed:
I see, and it makes sense.

But aside from added overhead (creating an instance several times per request), it requires to store global state for a request then in the scope (for https://github.com/encode/starlette/issues/495).

I've seen it more like `scope` is the initial state, and then it gets turned into a `Request` for http, which knows about `body`, while other connection types use different classes/interfaces.

blueyed:
@tomchristie 
Any feedback on my previous comment?

I've noticed this to be a problem again with `request.state`, where it would create new State objects for "new" Request objects (created via middleware and app for the same scope).

I am using a lazy property on State (via descriptor), and it would be good to have it only being created once really then.  The workaround is to store/cache it on the `request.state` dict additionally.

basepi:
Can we revisit this discussion? @dmontagu hits the nail on the head -- if we want to be able to create the `Request` from the scope, we need to propagate cached state like `request._body`. Otherwise, you hit issues like #495. In our Elastic APM middleware I've had to preemptively make sure we use the workaround in https://github.com/encode/starlette/issues/495#issuecomment-513138055 during the pre-request stage so that everyone can access the body who needs to.

Issue #493, opened 2019-04-26
Question about Request.Body()


ricardomomm:
Hi dears developers, how can I retrieve the request.body inside an exception_handler? 

My scenario:

I have an exception_handler hook that receives a request and an exception, I'm also using FastApi which tries to parse the body to a JSON, if it does not work it raises an HTTPException. So I'm trying to build a hook to log invalid JSON requests, this exception is being correctly captured by my exception_handler but when I'm trying to retrieve the body I can't it hangs, it simple doesn't work. It does make sense as after reading for the first time it is being cached inside the Request object but the Request object is being recreated before calling the exception_handler so I'm losing my cache and I'm able to ready from the stream again.

https://github.com/encode/starlette/blob/d23bfd0d8ff68d535d0283aa4099e5055da88bb9/starlette/exceptions.py#L79

Not sure if it is supposed to be like this by design, but I think my scenario is not that fancy, it should be something possible, what do you guys think? Any workaround?

My code:

```@app.exception_handler(HTTPException)  
async def httpexception_logger(request: Request, exc: HTTPException):
    body = await request.body()
    text: str = bytes.decode(body)
    logger.error("Exception at %s, status code: %s, body: %s: exception %s", request.url, exception.status_code, text, exc)
    ... other stuff...
```

lovelydinosaur:
What’s the simplest possible code sample someone could use to replicate the issue? (Ideally just using Starlette itself if possible)

lovelydinosaur:
I expect there’s an issue with the request body already having been streamed, but we ought to be able to handle things more gracefully.

ricardomomm:
Hi @tomchristie , I was able to reproduce with the code below:

```python
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.exceptions import HTTPException
from json.decoder import JSONDecodeError
import uvicorn

# Force an exception sending an invalid json to "/"
# it will trigger an exception and it will be handled
# by log_exceptions and the assert will trigger an
# exception because request.body() cannot be read
# anymore (stream already consumed)

app = Starlette(debug=True)

@app.exception_handler(JSONDecodeError)
async def log_exceptions(request, exc):
    body: bytes = await request.body() # hangs forever
    body_text: str = bytes.decode(body)
    assert body_text, "I could not retrieve the body"
    return JSONResponse({ "error": "Unexpected error happened!" })


@app.route('/', methods=["POST"])
async def homepage(request):
    body_bytes = await request.body()
[...]

Issue #489, opened 2019-04-23
Questions about middlewares


hyperknot:
I'd like to do a very simple task, that is add a middleware which runs before the endpoints and adds a few properties to the request object. Very similar to how authentication middleware does it, but I'd like to make it more flexible.

My aim is to have something like this:
- request.user_.user_data
- request.user_.org_data
- request.user_.allowed_regions
for example

Now, my problem is the following:
- BaseHTTPMiddleware in the docs starts with `response = await call_next(request)` and it adds a Header **after** the request has been processed. How can I use BaseHTTPMiddleware to add processing **before** an endpoint runs?
- Can I add custom properties to the request object in a Middleware?
- In case I cannot do it with BaseHTTPMiddleware I'm confused about the pure ASGI 2 vs. 3. middlewares. All the examples in starlette 0.11.4 are on ASGI 2, but ProxyHeadersMiddleware from uvicorn github is ASGI 3, do I understand it right?
- I get a `INFO: ASGI 'lifespan' protocol appears unsupported.` if I try to add something based on the skeleton of ProxyHeadersMiddleware. Will it disappear if I upgrade to 0.12.beta?

em92:
> BaseHTTPMiddleware in the docs starts with response = await call_next(request) and it adds a Header after the request has been **processed**. How can I use BaseHTTPMiddleware to add processing before an endpoint runs?

Processed by what?

> Can I add custom properties to the request object in a Middleware?

Yes. To add custom props see: https://www.starlette.io/requests/#other-state

> In case I cannot do it with BaseHTTPMiddleware I'm confused about the pure ASGI 2 vs. 3. middlewares. All the examples in starlette 0.11.4 are on ASGI 2, but ProxyHeadersMiddleware from uvicorn github is ASGI 3, do I understand it right?

ProxyHeadersMiddleware from uvicorn 0.6.1 is ASGI 2.

> I get a INFO: ASGI 'lifespan' protocol appears unsupported.

Make sure, that "startup" lifespan event does not throw exceptions.

hyperknot:
> Processed by what?

By the view / endpoint function. What is not clear is how to write middlewares in B.. (for before) and A.. (for after) states in the following diagram.

HTTP Request -> B1 -> B2 -> B3 -> view function -> A1 -> A2 -> A3 -> HTTP Response 

[...]

PR #498, opened 2019-05-01
[WIP/RFC] Store request in scope

blueyed:
Ref: https://github.com/encode/starlette/issues/497


blueyed:
This is useful for templating already, but also when using Starlette with other frameworks/apps, e.g. django-channels, and you want to have e.g. shared template processors (which could then use Starlette's or Django's request).

gwdekker:
This would also help with https://github.com/encode/starlette/issues/493 (and maybe https://github.com/encode/starlette/issues/489 ?)

lovelydinosaur:
Closing as per #497.

Issue #495, opened 2019-04-30
Consume request body in middleware is problematic


yihuang:
```
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import PlainTextResponse
from starlette.middleware.base import BaseHTTPMiddleware


class SampleMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next):
        _ = await request.form()
        return await call_next(request)


app = Starlette()


@app.route('/test', methods=['POST'])
async def test(request):
    _ = await request.form()  # blocked, because middleware already consumed request body
    return PlainTextResponse('Hello, world!')


app.add_middleware(SampleMiddleware)
```

```
$ uvicorn test:app --reload
```

```
$ curl -d "a=1" http://127.0.0.1:8000/test
# request is blocked
```

blueyed:
Where is it blocking?

Is there a second request instance involved? (otherwise it should just return, shouldn't it?)
https://github.com/encode/starlette/blob/7eb43757307d4702ee6a1f2739388242c703e47e/starlette/requests.py#L178-L193

blueyed:
Looks like it: https://github.com/encode/starlette/blob/7eb43757307d4702ee6a1f2739388242c703e47e/starlette/middleware/base.py#L26 and https://github.com/encode/starlette/blob/7eb43757307d4702ee6a1f2739388242c703e47e/starlette/routing.py#L38.

Can you write a test for starlette itself?

(just for reference: https://github.com/encode/starlette/pull/498)

lovelydinosaur:
"Consume request body in middleware is problematic"

Indeed. Consuming request data in middleware *is* problematic.
Not just to Starlette, but *generally*, everywhere.

On the whole you should avoid doing so if at all possible.

There's some work we could do to make it work better, but there will *always* be cases where it can't work (eg. if you stream the request data, then it's just not going to be available anymore later down the line).

wyfo:
Coming from FastAPI issue referenced above.

@tomchristie I don't understand the issue about consuming the request in a middleware, could you explain this point ? In fact, I have the need (which is current where I work) to log every requests received by my production server. Is there a better place than a middleware to do it and avoid duplicating code in every endpoint ? (I would like something like this https://github.com/Rhumbix/django-request-logging)

For now, I found this workaround (but that's not very pretty):
```
async def set_body(request: Request, body: bytes):
    async def receive() -> Message:
        return {"type": "http.request", "body": body}

[...]

PR #500, opened 2019-05-02
#493 and #495 cache body content

ricardomomm:
Hi dear developers, not sure if it is 100% compliant with your code base, but I'll be more than welcome to refactor it if not. This is one of the tests that I did locally in order to be able to consume the Request.Body more than one time when recreating a Request object from the Scope, this fix is caching the body content inside the scope rather than a local variable.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -139,8 +139,8 @@ class Request(HTTPConnection):
         return self._receive
 
     async def stream(self) -> typing.AsyncGenerator[bytes, None]:
-        if hasattr(self, "_body"):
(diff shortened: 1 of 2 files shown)


ricardomomm:
Just linking #493 and #495

lovelydinosaur:
Thanks!

So we *may* do something like this at some point, but we'll want to have a more comprehensive look at it - eg. what should the behaviour look like when consuming the request body as a stream.

At the moment I'd like us to punt on this, but there's likely somethere here we'll come back to.

PR #502, opened 2019-05-04
WIP/RFC: templating: do not require request in context

blueyed:


starlette/templating.py
```diff
@@ -38,6 +40,13 @@ def __init__(
                     "context": self.context,
                 }
             )
+
+        context = self.context.copy()
+        context.setdefault("request", request)
+
+        # TODO async?!
+        self.body = self.template.render(context).encode()
```
alex-oleshkevich, line 54: I believe it should be async as template rendering involves IO (read template from disk).

blueyed: Yeah.. but that requires to use a different method / explicit setup with Jinja2 - but I agree that it should default to being async.

blueyed: This is not in scope for this PR though I'd say, just something I've noticed / noted there.

Kludex: https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Template.render_async

It's available now.

starlette/templating.py
```diff
@@ -30,11 +31,12 @@ def __init__(
...
         self.template = template
         self.context = context
-        content = template.render(context)
-        super().__init__(content, status_code, headers, media_type, background)
+        super().__init__(None, status_code, headers, media_type, background)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         request = self.context.get("request", {})
+        if not request:
+            request = Request(scope)
```
Kludex, line 39:
```suggestion
        request = self.context.get("request", Request(scope))
```

Kludex (changes requested):
@blueyed Thanks for the PR! :tada:  

Can we have a use-case for this?

Also, can you check my comments? We should render asynchronously. 

Sorry for taking so long to review! :sweat_smile:

Kludex:
Closed as stale.

PR #504, opened 2019-05-08
Websocket routes can have additional keyword arguments passed

meadsteve:
I've been using starlette and I've wanted to have a `WebSocketEndpoint` which took some extra arguments into the constructor. This PR extends `WebSocketRoute` to take `**kwargs` which are passed to the constructor. I tried to make this match how the middleware work.


meadsteve:
I've instead implemented this with an anonymous class

PR #505, opened 2019-05-09
TestClient: do not throw the same exception multiple times

blueyed:
Without this patch it would throw the same exception again during shutdown
(when exiting the context manager).

This leads to pytest reporting it twice (during "call" as failure, and
as error for "teardown").

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -88,6 +88,8 @@ class _WrapASGI2:
 
 
 class _ASGIAdapter(requests.adapters.HTTPAdapter):
+    raised_server_exception: typing.Union[None, BaseException] = None
+
     def __init__(self, app: ASGI3App, raise_server_exceptions: bool = True) -> None:
         self.app = app
         self.raise_server_exceptions = raise_server_exceptions
@@ -231,10 +233,27 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
             loop = asyncio.new_event_loop()
             asyncio.set_event_loop(loop)
 
+        class WrapForServerErrors:
+            def __init__(self, adapter: _ASGIAdapter) -> None:
+                self.adapter = adapter
+
+            async def __call__(
+                self, scope: Scope, receive: Receive, send: Send
+            ) -> None:
+                try:
+                    await self.adapter.app(scope, receive, send)
+                except BaseException as exc:
+                    if self.adapter.raise_server_exceptions:
+                        if exc is not self.adapter.raised_server_exception:
+                            raise exc from None
+
+        app = WrapForServerErrors(self)
+
         try:
-            loop.run_until_complete(self.app(scope, receive, send))
+            loop.run_until_complete(app(scope, receive, send))
         except BaseException as exc:
             if self.raise_server_exceptions:
+                self.raised_server_exception = exc
                 raise exc from None
 
         if self.raise_server_exceptions:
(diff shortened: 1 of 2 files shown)


blueyed:
Now with test.. \o/
(requires a http middleware also)

blueyed:
Only mypy is failing.

blueyed:
Any input on this?
I am still seeing problems in this regard, where I basically would have to swallow the raised exception several times in `TestClient.__exit__`, before the lifespan end can be awaited / handled.

Should the lifespan app also be wrapped in a similar way?

I have not investigated, but it might be that the exception gets raised in `__exit__` for every middleware instance being used (I have ~5 there).

blueyed:
Should TestClient have its own error_middleware?

starlette/testclient.py
```diff
@@ -451,7 +472,16 @@ def __enter__(self) -> requests.Session:
...
     def __exit__(self, *args: typing.Any) -> None:
         loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.wait_shutdown())
+        shutdown = asyncio.ensure_future(self.wait_shutdown())
+        while True:
+            try:
+                loop.run_until_complete(shutdown)
+            except BaseException as exc:
+                if exc is self.adapter.handled_server_exception:
+                    continue
```
blueyed: Not covered / triggered with py38.

starlette/testclient.py
```diff
@@ -451,7 +472,16 @@ def __enter__(self) -> requests.Session:
...
+                raise
+            else:
+                break
```
blueyed, line 485: The `break` is reported as not covered in py38, but is - likely some optimization not handled by (current) coverage.py

blueyed: I consider this to be quite a hack anyway - but at least this is finally getting not in my way anymore - and has tests for the different variants.

tests/test_testclient.py
```diff
@@ -113,3 +113,55 @@ def app(scope):
...
+            pytest.fail("should_only_fail_once")
+
+        with TestClient(app, raise_server_exceptions=True) as client:
+            with pytest.raises(pytest.fail.Exception):
+                client.get("/error")
+    else:
+
+        @app.route("/error")
+        async def error(request):
+            pytest.fail("should_not_raise")
```
unmade, line 161: Could it be that exception raised by `pytest.fail` handles differently by pytest itself?

If I replace all `pytest.fail` with `raise Exception` and rollback changes in `testclient.py` the test passes fine.

blueyed: I think it is rather due to it being not of type Exception, but BaseException.

blueyed: IIRC BaseException/Exception is handled differently also in asyncio already in some place(s).

unmade: > I think it is rather due to it being not of type Exception, but BaseException.

Indeed

blueyed: Might be related to this, but do not remember really: https://github.com/python/cpython/blob/443fe5a52a3d6a101795380227ced38b4b5e0a8b/Lib/asyncio/base_events.py#L161-L168

lovelydinosaur:
Going to close this as stale.

Broader rule for the moment: We shouldn't make any changes to the TestClient until after https://github.com/encode/starlette/pull/1376

Issue #295, opened 2018-12-28
HTTPException with headers - pull request proposal


tiangolo:
I would like to be able to raise `HTTPException` with custom headers.

I'm willing to implement it and send a PR if this is something acceptable for Starlette.

Currently, that's not possible. As the headers would have to be set in an exception handler, and it wouldn't have a way to know the associated realms of a specific exception raised.

---

Use case:

I would like to be able to raise an authentication error 401 Unauthorized, including `WWW-Authenticate`, with a specific type (like `Basic`) and possibly, with a specific set of realms.

---

What do you think? Would that be acceptable/desirable in Starlette?

lovelydinosaur:
I'd *probably* prefer not to actually.

The use case for exceptions-rather-than-responses is when you need to strictly decouple the exception from the kind of response it generates.

For example, generic 404 exceptions that the developer *might* want to render as an HTML page, or as a JSON response depending on if they're developing a website or API.

Having built-in execeptions that mandate the specific of the response such as headers starts to defeat the point, and just makes it less clear if the developer should use an exception or a response.

tiangolo:
Get it. Thanks for the explanation.

I guess it also doesn't make as much sense here in Starlette, as all the responses are handled in the route endpoint, or in middleware directly, but not in between.

...I just prefer to keep as much as possible of the features/changes "upstream" (in Starlette) when it makes sense for the project, before adding custom stuff to FastAPI. So I prefer to propose it here first :slightly_smiling_face: 

---

I'll handle it outside of Starlette, but just as future reference, here's my case in FastAPI:

As I'm having injected dependencies (in this specific scenario for security/auth), they can return a value from the request that gets injected in the endpoint function, or they can raise an exception, that aborts all the rest of the process.

But I can't return a response directly from the dependency that would somehow override the response handled by the endpoint function (and it doesn't seem very natural in that case). So I want to be able to raise an exception from the security dependency that results in an HTTP 401 error response with a standard `WWW-Authenticate` header, etc.

PR #507, opened 2019-05-11
Add custom headers to HTTPException

florimondmanca:
Hi there,

Dealing with HTTP errors sometimes requires setting custom headers in the response. Ideally these should be defined when raising the `HTTPException`, which is what this PR proposes to allow.

FYI, FastAPI uses a custom `HTTPException` that adds just this: https://github.com/tiangolo/fastapi/blob/master/fastapi/exceptions.py

### DIFF
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -9,11 +9,16 @@ from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 
 class HTTPException(Exception):
-    def __init__(self, status_code: int, detail: str = None) -> None:
+    def __init__(
+        self, status_code: int, detail: str = None, headers: dict = None
+    ) -> None:
         if detail is None:
             detail = http.HTTPStatus(status_code).phrase
+        if headers is None:
+            headers = {}
         self.status_code = status_code
         self.detail = detail
+        self.headers = headers
 
 
 class ExceptionMiddleware:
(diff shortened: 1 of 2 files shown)


tiangolo:
Just for reference, it was proposed (but rejected) here: https://github.com/encode/starlette/issues/295

florimondmanca:
Ah, I see, thanks!

Tom wrote:

> Having built-in execeptions that mandate the specific of the response such as headers starts to defeat the point, and just makes it less clear if the developer should use an exception or a response.

I think it's a fair point. There's also the fact that allowing to add headers to the exception puts the responsibility on the person who wrote the error handler (e.g. the framework user themselves) not to forget to send exception headers. See:

```python
app = Starlette()

@app.exception_handler(401)
async def unauthorized(request, exc):
    return JSONResponse(
        content={"detail": str(exc)},
        status_code=exc.status_code,
        headers=exc.headers
    )
```

Error handlers written up to now didn't send headers anyway, so that's not a problem — except when people start sending headers in `HTTPException` and they'll wonder why they don't get through, only to realise that the ancient handler they use don't pass them through.

As a safety net, maybe Starlette (or any other framework that would use `HTTPException`) can add `exc.headers` to the response if the error handler didn't do so — but then you'd ask whether this was done intentionally or not.

So, the alternative would be to use custom exceptions, I assume?

```python
app = Starlette()

class BasicUnauthorized(HTTPException):
    pass

@app.exception_handler(BasicUnauthorized)
async def basic_unauthorized(request, exc):
    return JSONResponse(
        content={"detail": str(exc)},
        status_code=exc.status_code,
        headers={"WWW-Authenticate": "Basic"}
    )
```

florimondmanca:
Alright, I think I'll close this for now. :-)

PR #508, opened 2019-05-12
docs: Add missing import

jodal:
The first example on the GraphQL page has complete imports. Then a number of examples without imports follows. This last example is complete, with the exception of this one missing import.


lovelydinosaur:
Thanks!

PR #512, opened 2019-05-13
flake8 fixes

blueyed:


lovelydinosaur:
Legit. 👍

blueyed:
Would you take a flake8 config?  https://github.com/blueyed/starlette/blob/ef49e73e1043fdf7ce409a190ab632cfd568953e/setup.cfg#L26-L32

lovelydinosaur:
Bit hard to say without seeing the pull request, but something along those lines would be good.
Whatever the minimal tooling we'd need in order to drop unused import, variables is prob. something I would go for.

blueyed:
Well, I won't create a PR for `scripts/*`, only the config.

Issue #510, opened 2019-05-13
'access-control-allow-origin' header seems to "leak" between requests


steinnes:
We've encountered cases where credentialed requests (ie. with cookies), followed by non-credentialed requests will get an `Access-Control-Allow-Origin` header in the response, containing the `Origin` from the previous credentialed request.

Example:

```python
import uvicorn

from starlette.applications import Starlette
from starlette.responses import PlainTextResponse
from starlette.middleware.cors import CORSMiddleware


app = Starlette()
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_headers=["*"],
    allow_methods=["*"],
)


@app.route("/")
def homepage(request):
    return PlainTextResponse("Homepage", status_code=200)


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=5100)
```

1. curl without a cookie, get `access-control-allow-origin: *` back 
```
$ curl -v -H 'Origin: https://someplace.org' http://127.0.0.1:5100
* Rebuilt URL to: http://127.0.0.1:5100/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 5100 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:5100
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: https://someplace.org
>
< HTTP/1.1 200 OK
< date: Mon, 13 May 2019 10:32:58 GMT
< server: uvicorn
< content-length: 8
< content-type: text/plain; charset=utf-8
< access-control-allow-origin: *
```

2. curl with a cookie, get the Origin header back as `access-control-allow-origin`:
```
$ curl -v -H 'Origin: https://someplace.org' --cookie "foo=bar"  http://127.0.0.1:5100
* Rebuilt URL to: http://127.0.0.1:5100/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 5100 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:5100
> User-Agent: curl/7.54.0
> Accept: */*
> Cookie: foo=bar
> Origin: https://someplace.org
>
< HTTP/1.1 200 OK
< date: Mon, 13 May 2019 10:34:10 GMT
< server: uvicorn
< content-length: 8
< content-type: text/plain; charset=utf-8
< access-control-allow-origin: https://someplace.org
```

[...]

PR #513, opened 2019-05-13
CORS: preflight_response: cookie is not used here

blueyed:
Ref: https://github.com/encode/starlette/issues/510#issuecomment-491836700

It gets only called for "OPTIONS", but strictly speaking we should maybe check that there is no cookie before?!  https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Simple_requests

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -98,7 +98,6 @@ class CORSMiddleware:
         requested_origin = request_headers["origin"]
         requested_method = request_headers["access-control-request-method"]
         requested_headers = request_headers.get("access-control-request-headers")
-        requested_cookie = "cookie" in request_headers
 
         headers = dict(self.preflight_headers)
         failures = []


steinnes:
I'm not an expert on how the standards bodies expect OPTIONS preflight requests and responses to behave, but it does seem that they should not contain (or set) cookies.  From the perspective of `"be liberal in what you accept, but strict in what you send"` (not sure if that's the verbatim quote) I'd say this is fine?

lovelydinosaur:
Sorry - I don't understand the question.
There's an unused variable here, so we ought to remove it right?
Any idea how it's been left dangling in the first place?

blueyed:
> There's an unused variable here, so we ought to remove it right?

Yes.

> Any idea how it's been left dangling in the first place?

Have not checked - but that's the question mainly: is it required to handle/check cookies there?

Dustyposa:
@blueyed Have solved this problem? 
```python
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_headers=["*"],
    allow_methods=["*"],
)
```
I used this config, but still can only recevie `OPTION` request.  
With from front's request still had Cross-domain problem

PR #523, opened 2019-05-21
Update starlette.png, added starlette.svg - raw version of image, upd…

DamianSkrzypczak:
…ated  tags and added framework title thus logo doesnt contain it


DamianSkrzypczak:
I've created logo proposition for project some time ago and according to @tomchristie's advice - I'm making PR for it.  

**THIS IS ONLY CONCEPT** - I don't assume it will be final version or even that it will pass this PR.
Coloring is matching website but of course, as everything, it can be changed. 

I'm waiting for any suggestions or comments - if its good, if its bad, all feedback is welcome!!! :smile:  

This is how it would look on web-page:
image

DamianSkrzypczak:
**Worth mentioning:** 
I didn't change logo URL to one that shows my proposition in README.md. 
Instead, it still pointing at master. I can of course change if necessary for time of PR.

DamianSkrzypczak:
+0 ? :smile: (refreshing PR)

lovelydinosaur:
It's nice work, but I think we'll probably pass on it vs. our existing logo.

lovelydinosaur:
Thank you so much for taking the time tho.

Issue #293, opened 2018-12-27
Security and password utils


jordaneremieff:
Given that Starlette implements a basic authentication interface at this point, are there any plans to also provide some common password hashing functionality/security utils?

For example, both [Werkzeug](http://werkzeug.pocoo.org/docs/0.14/utils/#module-werkzeug.security) and [Django](https://docs.djangoproject.com/en/2.1/topics/auth/passwords/) offer some tools in this regard. I think security is something easy to get wrong and password authentication is a common enough problem to warrant shipping some utils with Starlette as well.

Alternatively, we may want to document recommended approaches/best practices to provide some guidance to those implementing the authentication interface or a third-party extension.

lovelydinosaur:
We’ll want to provide login/logout endpoints, yes.

jordaneremieff:
@tomchristie would this also include our own password salting/hashing/validation methods (and potentially support for multiple hashing algorithms), or more simply to provide the building blocks for implementations to use their own security methods with the interface?

I've been digging into this a bit and may be able to put forward a PR at some point to get the ball rolling depending on how much should be included here.

lovelydinosaur:
We’d do something similar to Django’s hashing strategy. Allow it to be configurable, but use PBKDF2 by default since the stdlib includes support for that.

Chaostheorie:
Has there been any recent developments regarding this feature set? I'm at the moment implementing [secure](https://pypi.org/project/secure/0.1.8/) for one of my applications. It may be a viable option for secure cookie based authentication that may support login/ logout enpoints.

lovelydinosaur:
Let's close this off for now.

Right now we need to get the maintenance load into balance before we consider if we'd want any more feature work to be in-scope for Starlette or not.

PR #529, opened 2019-05-26
Add password hashing classes.

alex-oleshkevich:
This adds password hashers to the framework.

refs #293

Included types: `BCryptPasswordHasher`, `Argon2PasswordHasher`, `PBKDF2PasswordHasher`,
`PlainPasswordHasher`.

Also provides aggregate `PasswordChecker` to manage mutiple password hashers.

## Interface
```python
class PasswordHasher:
    async def check(self, password: str, encoded: str) -> bool: ...

    async def make(self, password: str, salt: str = None, **kwargs) -> str: ...
```

## Example

```python
from starlette.passwords import (
    PlainPasswordHasher, PBKDF2PasswordHasher, PasswordChecker
)
hashers = [PlainPasswordHasher(), PBKDF2PasswordHasher()]
checker = PasswordChecker(hashers)

await checker.make('password_to_hash')
await checker.check('plain_password', 'hashed_password')

print(checker.requires_update) # True if password needs to be rehashed
```

upd:
renamed `InsecurePasswordHasher` to `PlainPasswordHasher`

### DIFF
--- a/docs/passwords.md
+++ b/docs/passwords.md
@@ -19,9 +19,9 @@ that the hashed password needs to be updated (eg. algorithm has changed).
 
 ```python
 from starlette.passwords import (
-    PlainPasswordHasher, PBKDF2PasswordHasher, PasswordChecker
+    InsecurePasswordHasher, PBKDF2PasswordHasher, PasswordChecker
 )
-hashers = [PlainPasswordHasher(), PBKDF2PasswordHasher()]
+hashers = [InsecurePasswordHasher(), PBKDF2PasswordHasher()]
 checker = PasswordChecker(hashers)
 
 await checker.make('password_to_hash')
@@ -32,7 +32,7 @@ print(checker.requires_update) # True if password needs to be rehashed
 
 ## Included hashers
 
-### PlainPasswordHasher
+### InsecurePasswordHasher
 
 This class does not encode the password and returns it "as is".  
 Warning: this hasher does not offer any security and is provided exclusively 
--- a/docs/passwords.md
+++ b/docs/passwords.md
@@ -1,5 +1,5 @@
 Starlette includes several password hashing classes that let you
-to make and check passwords.
+create and check passwords.
 
 
 ## PasswordChecker
--- a/docs/passwords.md
+++ b/docs/passwords.md
@@ -5,7 +5,7 @@ create and check passwords.
 ## PasswordChecker
 
 Among these classes is a special class `PasswordChecker`.
-It is intended to be a common interface to a multiple underlying hashers.
+It is intended to be a common interface to multiple underlying hashers.
 
 `PasswordChecker#check` will iterate over all given hashers asking each to verify
 the given password. If none of them succeeds then it returns `False`.
--- a/docs/passwords.md
+++ b/docs/passwords.md
@@ -34,7 +34,7 @@ print(checker.requires_update) # True if password needs to be rehashed
 
 ### InsecurePasswordHasher
 
-This class does not encode password and returns it "as is".
+This class does not encode the password and returns it "as is".
 
 ### PBKDF2PasswordHasher
 
--- a/docs/passwords.md
+++ b/docs/passwords.md
@@ -1,0 +1,70 @@
+Starlette includes several password hashing classes that let you
+to make and check passwords.
+
+
+## PasswordChecker
+
+Among these classes is a special class `PasswordChecker`.
+It is intended to be a common interface to a multiple underlying hashers.
+
+`PasswordChecker#check` will iterate over all given hashes asking each to verify
+the given password. If none of them succeed then returns `False`.
+
+Note, that `PasswordChecker#make` will use only the first item from 
+the `hashers` list to encode a plain password.
+
+This class also provides `requries_update` property which indicates
+that the hashed password needs to be updated (eg. algorithm has changed).
+
+
+```python
+from starlette.passwords import (
+    InsecurePasswordHasher, PBKDF2PasswordHasher, PasswordChecker
+)
+hashers = [InsecurePasswordHasher(), PBKDF2PasswordHasher()]
+checker = PasswordChecker(hashers)
+
+await checker.make('password_to_hash')
+await checker.check('plain_password', 'hashed_password')
+
+print(checker.requires_update) # True if password needs to be rehashed
+```
+
+## Included hashers
+
+### InsecurePasswordHasher
+
+This class does not encode password and returns them "as is".
+
+### PBKDF2PasswordHasher
+
+PBKDF2 password hashing algorithm.
+Does not have any dependencies. 
+
+### BCryptPasswordHasher
+
+Uses bcrypt algorithm as a hashing backend. 
+Requires `bcrypt` library to be installed.
+
+### Argon2PasswordHasher
+
+Uses argon2 algorithm as a hashing backend. 
+Requires `argon2` library to be installed.
+
+
+## Custom password hasher
+
+When you need to make your own password hasher 
+then create a new class and implement `starlette.passwords.PasswordHasher` interface:
+
+```python
+from typing import Any
+from starlette.passwords import PasswordHasher
+
+class MyPasswordHasher(PasswordHasher):
+    async def make(self, password: str, salt: str = None, **kwargs: Any) -> str:
+        ...
+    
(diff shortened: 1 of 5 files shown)


docs/passwords.md
```diff
@@ -0,0 +1,70 @@
+Starlette includes several password hashing classes that let you
+to make and check passwords.
```
blueyed:
```suggestion
create and check passwords.
```

docs/passwords.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## PasswordChecker
+
+Among these classes is a special class `PasswordChecker`.
+It is intended to be a common interface to a multiple underlying hashers.
```
blueyed:
```suggestion
It is intended to be a common interface to multiple underlying hashers.
```

docs/passwords.md
```diff
@@ -0,0 +1,70 @@
...
+await checker.check('plain_password', 'hashed_password')
+
+print(checker.requires_update) # True if password needs to be rehashed
+```
+
+## Included hashers
+
+### InsecurePasswordHasher
+
+This class does not encode password and returns it "as is".
```
blueyed:
```suggestion
This class does not encode the password and returns it "as is".
```

docs/passwords.md
```diff
@@ -0,0 +1,70 @@
...
+### Argon2PasswordHasher
+
+Uses argon2 algorithm as a hashing backend. 
+Requires `argon2` library to be installed.
+
+
+## Custom password hasher
+
+When you need to make your own password hasher 
+then create a new class and implement `starlette.passwords.PasswordHasher` interface:
```
blueyed: better:

> You can create your own password hasher by creating a new class and implementing the `starlette.passwords.PasswordHasher` interface:

starlette/passwords.py
```diff
@@ -0,0 +1,191 @@
...
+    import bcrypt
+except ImportError:  # pragma: nocover
+    bcrypt = None
+
+ALNUM_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+
+class PasswordHasher(abc.ABC):
+    algorithm: str = ""
+    rounds: int = 150000
```
blueyed: Is that a sensible value?

blueyed (commented):
Just a quick/incomplete review.

tiangolo:
Just a question, wouldn't it make sense to just integrate/document how to use Passlib instead?: https://passlib.readthedocs.io/en/stable/

It could be kept as an optional dependency, and it provides a lot of tools already, including algorithms compatible with Django and other tools.

alex-oleshkevich:
> Just a question, wouldn't it make sense to just integrate/document how to use Passlib instead?: https://passlib.readthedocs.io/en/stable/
> 
> It could be kept as an optional dependency, and it provides a lot of tools already, including algorithms compatible with Django and other tools.

I was thinking about it. There isn't any activity since summer 2017. 

Also, passlib isn't async, password hashing sometimes is a CPU bound task and may block the thread.
Classes from this PR are asynchronous.

starlette/passwords.py
alex-oleshkevich: do you mean to set default to null and override this on per-class basis?

tiangolo:
> I was thinking about it. There isn't any activity since summer 2017.

> Also, passlib isn't async, password hashing sometimes is a CPU bound task and may block the thread.
Classes from this PR are asynchronous.

@alex-oleshkevich Get it. Thanks for the clarification.

lovelydinosaur:
Thanks so much for this @alex-oleshkevich  - Apologies that I've not gotten around to reviewing just yet. I'll aim to take a proper look sometime soon.

rafalp:
Perhaps we could replace the `PlainPasswordHasher` with `SimplePasswordHasher` or `MD5PasswordHasher` that would do single MD5 pass?

I think documentation should also warn devs about this hasher not offering any security at all, and being provided exclusively for making password operations faster in tests suite?

alex-oleshkevich:
> Perhaps we could replace the `PlainPasswordHasher` with `SimplePasswordHasher` or `MD5PasswordHasher` that would do single MD5 pass?
> 
> I think documentation should also warn devs about this hasher not offering any security at all, and being provided exclusively for making password operations faster in tests suite?

Added a warning.
Not sure about the name. May be `InsecurePasswordHasher` is a better alternative?

docs/passwords.md
```diff
@@ -0,0 +1,71 @@
...
+
+print(checker.requires_update) # True if password needs to be rehashed
+```
+
+## Included hashers
+
+### PlainPasswordHasher
+
+This class does not encode the password and returns it "as is".  
+Warning: this hasher does not offer any security and is provided exclusively
```
rafalp:
```suggestion

> **Warning**: this hasher does not offer any security and is provided exclusively 
```

rafalp:
Yeah, `PlainPasswordHasher` is misnomer - it doesn't really hash anything.

rafalp:
I still don't think its good to have a class named `Hasher` that actually **doesn't** hash - this name is a lie to developer. ;)

IMHO this class should be dropped or replaced `MD5PasswordHasher` that will actually hash password, albeit with simple and fast algorithm.

dmontagu:
> Also, passlib isn't async, password hashing sometimes is a CPU bound task and may block the thread.
Classes from this PR are asynchronous.

~~I noticed you are using `run_in_threadpool` -- given the GIL, doesn't this still block the event loop? I would think you'd need to use an analogous `run_in_processpool` to ensure you aren't blocking the event loop.~~

~~Maybe the password-hashing functions release the GIL?~~

Edit: I ran a bunch of password checks concurrently and noticed similar speed ups whether using threadpool or processpool, so this is a non-issue (I assume the GIL is being released).

These are great!

I think it might be a good idea to add a section/note to the documentation discussing the tradeoffs of different hashers (or linking to a discussion that does), OR just providing a single recommendation that will fit most applications for users that just aren't going to dig in deeply on this topic.

alex-oleshkevich:
> I think it might be a good idea to add a section/note to the documentation discussing the tradeoffs of different hashers (or linking to a discussion that does), OR just providing a single recommendation that will fit most applications for users that just aren't going to dig in deeply on this topic.

I think that that topic is out of scope of framework.

dmontagu:
@alex-oleshkevich 
The introduction page for Starlette explicitly says:
> Starlette is designed to be used either as a complete framework, or as an ASGI toolkit. You can use any of its components independently.

If a user-friendly interface to something as security-sensitive as password hashing is going to be baked into the framework (which doesn't otherwise have any warnings against usage by non-advanced users), I personally think there should be an obligation to make sure it isn't readily misused/misunderstood. Putting responsibility fully on the user for even *being aware* of the potential security issues (such as MD5 hashing insecurity) seems like it would inevitably result in someone getting burned, especially given the current growth rate of Starlette/ASGI.

The [Django docs on passwords](https://docs.djangoproject.com/en/2.2/topics/auth/passwords/) discuss (in relative depth) the tradeoffs of the various hashers. I think content like this goes a long way towards drawing increased adoption of the framework, but if you want to elide a discussion of this entirely, I think it would merit either some mechanism by which a default hasher is used (Django has sane defaults, as described near the top of those docs). Or, if there is no default, then at least some sort of warning against making use of this feature unless you know what you are doing.

rafalp:
I've been thinking - wouldn't it make more sense to split this work into a separate repository? That way it would be possible to improve the library with PR's and releases without needing to release new Starlette version. It would also make the library more readily available to those not using Starlette, and keep Starlette itself focused on being async microframework/http server library.

alex-oleshkevich:
> I've been thinking - wouldn't it make more sense to split this work into a separate repository? That way it would be possible to improve the library with PR's and releases without needing to release new Starlette version. It would also make the library more readily available to those not using Starlette, and keep Starlette itself focused on being async microframework/http server library.

Yes, that is a valid idea. May be I will proceed with it.

Kludex:
Given the fact that we're aiming to have a minimal set of features on [encode](https://github.com/encode) projects, and considering this is a 3 years old PR, I'll take the lead and close it.

Thanks everybody for the discussion, and thanks @alex-oleshkevich for the interest! :tada:

PR #530, opened 2019-05-27
Allow callable functions to be passed into the config's casting

BoyangHuang:
in `starlette/config.py`, the `cast` parameter has an expected type `type`. Below is an example of a simple use case where passing a function would also be useful. 

```python
def cast_to_datetime(date_time_string: str) -> datetime:
    return datetime.strptime(date_time_string, "%Y-%m-%d %H:%M:%S")

config = Config()

time = config("MY_TIME", cast=cast_to_datetime, default="2017-01-12 14:12:06") 

```

This works like a charm, but mypy will complain that `cast` can only be `type`. This contribution is meant to allow callables to be passed into `cast` as well.

### DIFF
--- a/starlette/config.py
+++ b/starlette/config.py
@@ -54,12 +54,12 @@ class Config:
             self.file_values = self._read_file(env_file)
 
     def __call__(
-        self, key: str, cast: type = None, default: typing.Any = undefined
+        self, key: str, cast: typing.Callable = None, default: typing.Any = undefined,
     ) -> typing.Any:
         return self.get(key, cast, default)
 
     def get(
-        self, key: str, cast: type = None, default: typing.Any = undefined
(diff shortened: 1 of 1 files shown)


starlette/config.py
```diff
@@ -54,12 +54,18 @@ def __init__(
             self.file_values = self._read_file(env_file)
 
     def __call__(
-        self, key: str, cast: type = None, default: typing.Any = undefined
+        self,
+        key: str,
+        cast: typing.Union[type, typing.Callable] = None,
```
lovelydinosaur: I guess this could *just* be `cast: Callable` right? (Since types will be callable)

BoyangHuang: @tomchristie Right you are. Updated the PR to reflect that fact.

lovelydinosaur:
Thanks!

PR #531, opened 2019-05-28
WebSocketTestSession uses current event loop 

### LINKED ISSUE #487: Should WebSocketTestSession use current event loop?
Hi there! 

First of all, thanks for the beautiful framework!

I'm in the middle of migrating one of my apps from Sanic to FastAPI and I'm a little bit stuck with testing web socket endpoints. 

In the web socket handler I use redis connection pool (aioredis), which is created on the app startup (thanks to according event). However, in test whenever I try to access redis I keep getting this error:
```
RuntimeError: Task ...  attached to a different loop
```

I think the problem can be in this line:
https://github.com/encode/starlette/blob/master/starlette/testclient.py#L266

Here is what happening: to run events in test we use TestClient as context manager. Event handlers runs in current event loop, but WebSocketTestSession runs app in new event loop

So, my question is there specific reason to create new event loop in the WebSocketTestSession? It seems that we can just use current loop.

To illustrate the problem I created **pull request**:
https://github.com/unmade/starlette/pull/1/files

P.S: There seems to be incorrect return value in type annotation. I can fix it also:
https://github.com/encode/starlette/blob/master/starlette/testclient.py#L444
https://github.com/encode/starlette/blob/master/starlette/testclient.py#L426

unmade:
Resolve #487

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -264,6 +264,8 @@ class WebSocketTestSession:
         self.scope = scope
         self.accepted_subprotocol = None
         self._loop = asyncio.get_event_loop()
+        if self._loop.is_running():
+            self._loop = asyncio.new_event_loop()
         self._receive_queue = queue.Queue()  # type: queue.Queue
         self._send_queue = queue.Queue()  # type: queue.Queue
         self._thread = threading.Thread(target=self._run)
@@ -423,7 +425,7 @@ class TestClient(requests.Session):
 
     def websocket_connect(
         self, url: str, subprotocols: typing.Sequence[str] = None, **kwargs: typing.Any
-    ) -> typing.Any:
+    ) -> WebSocketTestSession:
         url = urljoin("ws://testserver", url)
         headers = kwargs.get("headers", {})
         headers.setdefault("connection", "upgrade")
@@ -441,7 +443,7 @@ class TestClient(requests.Session):
 
         return session
 
-    def __enter__(self) -> requests.Session:
+    def __enter__(self) -> "TestClient":
         loop = asyncio.get_event_loop()
         self.send_queue = asyncio.Queue()  # type: asyncio.Queue
         self.receive_queue = asyncio.Queue()  # type: asyncio.Queue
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -263,7 +263,9 @@ def __init__(self, app: ASGI3App, scope: Scope) -> None:
         self.app = app
         self.scope = scope
         self.accepted_subprotocol = None
-        self._loop = asyncio.new_event_loop()
+        self._loop = asyncio.get_event_loop()
+        if self._loop.is_running():
+            self._loop = asyncio.new_event_loop()
```
blueyed, line 268: I think this should use the same method as the TestClient (https://github.com/blueyed/starlette/blob/8fd8152b1536dc42d1d122791f0ee4cdeec0aaa1/starlette/testclient.py#L232-L236) - and therefore probably get factored out into a function/method of its own.

unmade: I was initially going to do so, but they have some differences.
In `TestClient` exception is raised on calling `get_event_loop` because there is no current event loop at all. 
In `WebSocketTestSession` no exception is raised on calling `get_event_loop`. However if current event loop is running there will be RuntimeError in [_run](https://github.com/unmade/starlette/blob/487-testclient-websocket-same-loop/starlette/testclient.py#L289) method. This `_run` method runs in thread, so I can't do fallback here. 

Condition `if self._loop.is_running` is tested by [here](https://github.com/unmade/starlette/blob/487-testclient-websocket-same-loop/tests/test_testclient.py#L132-L134) where you can see nested calls to endpoints each of which creates new `WebSocketTestSession`

blueyed: Ah, it is the other way around..
(reminded me of https://github.com/encode/starlette/pull/505 btw - maybe you can provide some insight there / help with it)

unmade: Sure, I replied you in #505

songzhi: Please merge it!!!😭,otherwise I can only edit files by myself,it feels insecure😂. In my code,I  edit the _run method as this:
``` python
if loop.is_running():
    loop.call_soon(self.app, scope, receive, send)
else:
```
It works,maybe there is more appropriate way to deal with this.

unmade:
Closing, see https://github.com/encode/starlette/issues/487#issuecomment-593375719

PR #532, opened 2019-05-29
Do not use requests.packages.urllib3, but urllib3 directly

blueyed:
The dynamic nature of `requests.packages` makes it harder for tools to
follow it [1].

1: https://github.com/kennethreitz/requests/blob/6f81e78dfe0b7419bde97cae594febfffab8e431/requests/packages.py

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -7,6 +7,7 @@ import queue
 import threading
 import types
 import typing
+import urllib3
 from urllib.parse import unquote, urljoin, urlsplit
 
 import requests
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Reasonable idea yup. I'm going to pass, since `requests` isn't likely to change that around at this point, and I'd rather that we end up raising an import error for "requests" rather than "urllib3", if it's not installed.

blueyed:
Ok, too bad - since it makes the code impossible to follow automatically.

FWIW, I was about to submit a PR for requests to issue a DeprecationWarning there.

Anyway, not used too much in starlette on the other hand.

Issue #534, opened 2019-06-06
get_directories in staticfiles.py has bad asserts


vinhquevu:
I don't think you can use F-Strings in assert statements.  I get "invalid syntax" if I try.

The line: 
`directory = os.path.normpath(os.path.join(spec.origin, "..", "statics"))`
Always assumes there is a folder `statics`?  I believe the standard is `static` and not `statics` and the user is not given the option to circumvent the step.  Does it make more sense to only include this step if `directory is None`.

The directory argument does not get the package path prepended, which I feel is implied in the documentation.

The `assert os.path.isdir(directory)` will fail every time since it will return the `__init__.py` of a package.

dmontagu:
* I have no problem using f-strings in assert statements in python 3.7.3.
    * Tangentially related, it looks like the f is missing from the f-string asserts on [this line](https://github.com/encode/starlette/blob/master/starlette/staticfiles.py#L73) and [this line](https://github.com/encode/starlette/blob/master/starlette/staticfiles.py#L77).
* I personally would agree that it feels a little weird to look for `"statics"` vs. `"static"`. The examples from the staticfiles docs all use the folder name `"static"`, rather than `"statics"` (though they aren't using the packages option to specify the folder). But it's not a big deal to me. Maybe there's a reason for it? (E.g., it could be a breaking change at this point?)
* The docs [here](https://www.starlette.io/staticfiles/) say
> directory - A string denoting a directory path.

I personally don't think this implies any package's path would be prepended -- if it was, what package's path would you even expect it to be?
* I take it you are referring to [this line](https://github.com/encode/starlette/blob/master/starlette/staticfiles.py#L75) when you say the `os.path.isdir(directory)` will fail every time. But I don't think that's right -- the `".."` argument to `os.path.join` ensures you get the parent folder. (I verified that line does not cause any errors for me, and correctly gets the `"statics"` directory's full path.)

vinhquevu:
The f-string failure I thought was strange and could be an issue in my python env, I'm using a debian build, 3.6.

The docs read this:
> The packages option can be used to include "static" directories contained within a python package. The Python "bootstrap4" package is an example of this.
> 
> ```
> from starlette.routing import Router, Mount
> from starlette.staticfiles import StaticFiles
> 
> app = Router(routes=[
>      Mount('/static', app=StaticFiles(directory='static', packages=['bootstrap4']), name="static"),
> ])
> ```
[...]

PR #535, opened 2019-06-07
Update staticfiles.py

dmontagu:
Add the f for the f strings.

Noticed while investigating #534

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -70,11 +70,11 @@ class StaticFiles:
             assert spec is not None, f"Package {package!r} could not be found."
             assert (
                 spec.origin is not None
-            ), "Directory 'statics' in package {package!r} could not be found."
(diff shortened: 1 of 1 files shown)


dmontagu:
Issue #534 also pointed out the issue of calling the static folder "statics" vs. "static" when using the `packages` option. Is there a reason for this? (I would understand if at this point the goal was just to prevent breaking changes.)

dmontagu:
Segmentation fault for python 3.8-dev? That's weird.

dmontagu:
@tomchristie perhaps related to https://github.com/python-pillow/Pillow/issues/3884 ? They suggest deleting the travis build caches.

lovelydinosaur:
Cache deleted. Build restarted.

blueyed (approved):

lovelydinosaur:
Thanks for the bump @blueyed! 3.8 build passed after I restarted it.

PR #541, opened 2019-06-13
remove unused code

keakon:


starlette/staticfiles.py
```diff
@@ -145,7 +145,6 @@ def get_path(self, scope: Scope) -> str:
     async def lookup_path(
         self, path: str
     ) -> typing.Tuple[str, typing.Optional[os.stat_result]]:
-        stat_result = None
```
lovelydinosaur, line 148: Yup.

starlette/staticfiles.py
```diff
@@ -27,7 +27,7 @@ class NotModifiedResponse(Response):
     )
 
     def __init__(self, headers: Headers):
-        return super().__init__(
```
lovelydinosaur, line 30: Yup.

starlette/routing.py
```diff
@@ -161,9 +161,9 @@ def __init__(
         if methods is None:
             self.methods = None
         else:
-            self.methods = set([method.upper() for method in methods])
+            self.methods = set(method.upper() for method in methods)
             if "GET" in self.methods:
-                self.methods |= set(["HEAD"])
```
lovelydinosaur, line 167: Yeah, okay - prefer your version of these two lines.

starlette/middleware/wsgi.py
```diff
@@ -93,7 +94,7 @@ def __init__(self, app: typing.Callable, scope: Scope) -> None:
                     self.exc_info[1], self.exc_info[2]
                 )
         finally:
-            if not sender.done():
```
lovelydinosaur, line 96: I don't see any value in the two lines changed here.

starlette/routing.py
```diff
@@ -231,8 +231,6 @@ def __init__(
             # Endpoint is a class. Treat it as ASGI.
             self.app = endpoint
 
-        regex = "^" + path + "$"
-        regex = re.sub("{([a-zA-Z_][a-zA-Z0-9_]*)}", r"(?P<\1>[^/]+)", regex)
```
lovelydinosaur, line 236: Any idea where these lines have been left over from? They don't make sense right now, but might be a good idea to understand how they came to be here?

starlette/middleware/wsgi.py
keakon: `self.loop.create_task()` may raise an exception, then `sender` will be referenced before assignment.

starlette/routing.py
keakon: This [commit](https://github.com/encode/starlette/commit/de71b8594807dae907eba636f01dc13fdb355f9d#diff-34fba745b50527bfb4245d02afd59246L122) removed the pattern in `Route`, but forgot `WebSocketRoute`.

starlette/middleware/wsgi.py
lovelydinosaur: Okay, then let's move `sender = ` to just outside the `try ... finally` block

blueyed:
@keakon 
Ping.
Rebasing should fix the CI failure also.

lovelydinosaur:
Looks good - thanks!

PR #543, opened 2019-06-13
Update response() calls in docs for ASGI 3

rafalp:
I've noticed that code examples in docs are missing `scope` argument in `response()` call, producing `TypeError: __call__() missing 1 required positional argument: 'send'` error when ran.

This PR fixes those.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -104,7 +104,7 @@ class App:
 
     async def __call__(self, receive, send):
         response = PlainTextResponse('Hello, world!')
-        await response(receive, send)
+        await response(self.scope, receive, send)
 ```
 
 Run the `App` application in `example.py`:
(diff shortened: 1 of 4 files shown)


rafalp:
On second thought, wouldn't it make more sense to move `scope` from `__init__` to `__call__` so it can be introspected per request? Like this:

```python
class App:
        ...
```

Or maybe simplify examples further with reducing code to single function?

```python
    ...
```

lovelydinosaur:
I'd switch 'em out to standard function calls now yup.
(We couldn't do that prior to ASGI 3)
You may also want to make sure to merge the latest master to resolve the build issues.

rafalp:
@tomchristie changed those to plain functions and rebased on current master.

lovelydinosaur:
Really great work! Thank you!

Issue #30, opened 2018-07-18
Include more information in `DebugMiddleware`.


lovelydinosaur:
Right now we just display a raw `format_trackback()`.

It'd make sense to have a proper page with lines of context around each file in the stack, and local variables at each point. (As with Django or Werkzeug's debugger pages)

supriyo-biswas:
What about integrating the werkzeug debugger itself? This feature was there previously in APIStar.

marcosschroh:
Ok, perfect!

marcosschroh:
Related to #179

Kludex:
I guess the local variables are still missing, correct?

lovelydinosaur:
Closing as per #1386

mireq:
@supriyo-biswas i tried to integrate werkzeug.  It's simple hack and should work without problems. Package is published on pypi - [starlette-werkzeug-debugger](https://pypi.org/project/starlette-werkzeug-debugger/) and it's implemented as [middleware](https://github.com/mireq/starlette-werkzeug-debugger/blob/78487667362b6fa97f8b30638e401f90db973eaa/starlette_werkzeug_debugger/__init__.py).

PR #544, opened 2019-06-13
feat: ServerErrorMiddleware improved

marcosschroh:
* Frames have been reverted. 
* Stylng improved. 
* Collapse added on every frame added. 
* Context around error with line added
* Default limit is 7, means that there are 3 lines shown at the top and 3 lines at the bottom. ServerErrorMiddleware accepts limit in the constructor. 

Screen Shot 2019-06-14 at 01 08 42

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import traceback
 import typing
 
@@ -8,6 +9,9 @@ from starlette.responses import HTMLResponse, PlainTextResponse, Response
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 STYLES = """
+p {
+    color: #211c1c;
+}
 .traceback-container {
     border: 1px solid #038BB8;
 }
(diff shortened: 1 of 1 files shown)


marcosschroh:
related with https://github.com/encode/starlette/issues/30

lovelydinosaur:
Ohhh... this looks really nice!
Not sure if "limit"is a super-obvious name here or not - could be okay, but are there any other more descriptive options we might use? What does Werkzeug's middleware use for naming there (if it has an equivelent)?

lovelydinosaur:
Also - I think you'll need to merge master to resolve the existing build issues.

marcosschroh:
> Ohhh... this looks really nice!
> Not sure if "limit"is a super-obvious name here or not - could be okay, but are there any other more descriptive options we might use? What does Werkzeug's middleware use for naming there (if it has an equivelent)?

Yes, it was hard to fine a descriptive name. At first I used `total_context` but then I chose `limit`. Maybe `context_limit`? I think Werkzeug is using limit as well https://github.com/pallets/werkzeug/blob/master/src/werkzeug/debug/repr.py#L183

lovelydinosaur:
How about we just don’t include that as a configurable-on-init at all.
Less dials and levers is a good thing.

marcosschroh:
> How about we just don’t include that as a configurable-on-init at all.
> Less dials and levers is a good thing.

Perfect! The `__init__` method is the same as before.

lovelydinosaur:
What else do we want from #30 at this point?
I guess a fixed width font for the code might be a good idea?

marcosschroh:
Ok. Do you want to have a kind of kind of same font styling as https://user-images.githubusercontent.com/647359/48124781-14093980-e275-11e8-8fa6-d000db2e8c45.png ?

We can fix the `title` and `subtitle` to match the styling and also add the text `most recent call first`  

Later, we can  add the `locals`.

What do you think?

lovelydinosaur:
> Do you want to have a kind of kind of same font styling as

Hellz yes! 😅

> Later, we can add the locals.

👍👍👍

PR #547, opened 2019-06-14
Drop 3.8-dev from build matrix

lovelydinosaur:
We've currently got failing builds on Python 3.8, which appears to be a compilation issue with "typed-ast" (A dependency of mypy)

Going to pin it to a slightly earlier version here to see if it resolves the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ cache: pip
 python:
     - "3.6"
     - "3.7"
-    - "3.8-dev"
 
 install:
     - pip install -U -r requirements.txt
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
There's an unresolved issue with the builds on 3.8-dev - seems to be that the `typed-ast` pacakge (required for mypy) isn't compiling.

Best thing for the moment is going to be just drop 3.8-dev from the matrix until we've got a resolution here. Anyone fancy trying to dig into what might be happening here? Are there any other projects with seeing a similar issue?

blueyed:
For reference: previous build failure: https://travis-ci.org/encode/starlette/jobs/543799592 (using Python 3.8.0b1+).
See #575 for a follow-up.

Issue #551, opened 2019-06-15
request.state is not persisted between middlewares.


didip:
I am not finding my `request.state.xyz = 1` on the next middleware execution. Bug?

lovelydinosaur:
Good point yup - it's not being initialized based on what's in the scope.

We should tweak things here, so that it has `__getattr__`, `__setattr__`, and `__delattr__` methods, each of which actually get/set/delete keys within the scope['state'] dictionary.

lovelydinosaur:
As a side effect, that'd also resolve #545

didip:
Super fast response, thank you!

Being able to pass around several singleton objects will help us break the app structure better so that everything is not defined in one giant Python file.

lovelydinosaur:
If you'd like to take a crack at this, would happily accept pull requests...

I think we'd want to change things around slightly so eg...

See https://github.com/encode/starlette/blob/master/starlette/requests.py#L109

Would probably become something like this?...

```python
    @property
    def state(self) -> State:
        if not hasattr(self, '_state'):
            # Ensure 'state' has an empty dict if it's not already populated.
            self._scope.setdefault('state', {})
            # Create a state instance with a reference to the dict in which it should store info
            self._state = State(self._scope['state'])
        return self._state
```

```python
class State:
    def __init__(self, state_dict):
        self.state_dict = state_dict

    # __getattr__, __setattr__, __delattr__ methods which update keys in the dict.
```

didip:
Yup, was just looking at that precise code block. Give me a minute.

didip:
Hm... the new test I wrote still fails.

## Code

```
class HTTPConnection(Mapping):
    """
    A base class for incoming HTTP connections, that is used to provide
    any functionality that is common to both `Request` and `WebSocket`.
    """

    def __init__(self, scope: Scope, receive: Receive = None) -> None:
        assert scope["type"] in ("http", "websocket")
        self._scope = scope

        # Ensure 'state' has an empty dict if it's not already populated.
        self._scope.setdefault('state', {})

    @property
    def state(self) -> State:
        if not hasattr(self, '_state'):
            # Create a state instance with a reference to the dict in which it should store info
            self._state = State(self._scope['state'])
        return self._state

```

## Test Code

```
def test_state_data_across_multiple_middlewares():
    expected_value = 'yes'

    class aMiddleware(BaseHTTPMiddleware):
        async def dispatch(self, request, call_next):
            request.state.show_me = expected_value
[...]

PR #552, opened 2019-06-16
Begin work in making sure `request.state` is preserved between middlewares.

didip:
One test still fails, and `request.state` is still not preserved but I'd like to share to you the code. I am not clear why the `scope` always empty on every new `Request`.

See issue #551 

cc @tomchristie

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -19,8 +19,8 @@ class ClientDisconnect(Exception):
 
 
 class State(object):
-    def __init__(self, state: typing.Dict={}):
-        super(State, self).__setattr__('_state', state)
+    def __init__(self, state: typing.Dict = {}):
+        super(State, self).__setattr__("_state", state)
 
     def __setattr__(self, key: typing.Any, value: typing.Any) -> None:
         self._state[key] = value
@@ -29,10 +29,12 @@ class State(object):
         if key in self._state:
             return self._state[key]
         else:
-            raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
+            raise AttributeError(
+                "'{}' object has no attribute '{}'".format(self.__class__.__name__, key)
+            )
 
     def __delattr__(self, key: typing.Any) -> None:
-        if key == '_state':
+        if key == "_state":
             super(State, self).__delattr__(key)
         else:
             del self._state[key]
@@ -49,7 +51,7 @@ class HTTPConnection(Mapping):
         self._scope = scope
 
         # Ensure 'state' has an empty dict if it's not already populated.
-        self._scope.setdefault('state', {})
+        self._scope.setdefault("state", {})
 
     def __getitem__(self, key: str) -> str:
         return self._scope[key]
@@ -127,9 +129,9 @@ class HTTPConnection(Mapping):
 
     @property
     def state(self) -> State:
-        if not hasattr(self, '_state'):
+        if not hasattr(self, "_state"):
             # Create a state instance with a reference to the dict in which it should store info
-            self._state = State(self._scope['state'])
+            self._state = State(self._scope["state"])
         return self._state
 
     def url_for(self, name: str, **path_params: typing.Any) -> str:
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -19,25 +19,19 @@ class ClientDisconnect(Exception):
 
 
 class State(object):
-    def __init__(self, state={}):
-        self._state = state
+    def __init__(self, state: typing.Dict={}):
+        super(State, self).__setattr__('_state', state)
 
-    def __setattr__(self, key, value: typing.Any) -> None:
-        if key == '_state':
-            super(State, self).__setattr__(key, value)
+    def __setattr__(self, key: typing.Any, value: typing.Any) -> None:
+        self._state[key] = value
+
+    def __getattr__(self, key: typing.Any) -> typing.Any:
+        if key in self._state:
+            return self._state[key]
         else:
-            self._state[key] = value
+            raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
 
-    def __getattr__(self, key) -> typing.Any:
-        if key == '_state':
-            super(State, self).__getattr__(key)
-        else:
-            if key in self._state:
-                return self._state[key]
-            else:
-                raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
-
-    def __delattr__(self, key) -> None:
+    def __delattr__(self, key: typing.Any) -> None:
         if key == '_state':
             super(State, self).__delattr__(key)
         else:
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -19,7 +19,8 @@ class ClientDisconnect(Exception):
 
 
 class State:
-    pass
+    def __init__(self, state_dict={}):
+        self.__dict__.update(state_dict)
 
 
 class HTTPConnection(Mapping):
@@ -31,6 +32,9 @@ class HTTPConnection(Mapping):
     def __init__(self, scope: Scope, receive: Receive = None) -> None:
         assert scope["type"] in ("http", "websocket")
         self._scope = scope
+
+        # Ensure 'state' has an empty dict if it's not already populated.
+        self._scope.setdefault('state', {})
 
     def __getitem__(self, key: str) -> str:
         return self._scope[key]
@@ -108,9 +112,10 @@ class HTTPConnection(Mapping):
 
     @property
     def state(self) -> State:
-        if "state" not in self._scope:
-            self._scope["state"] = State()
-        return self._scope["state"]
+        if not hasattr(self, '_state'):
+            # Create a state instance with a reference to the dict in which it should store info
+            self._state = State(self._scope['state'])
+        return self._state
 
     def url_for(self, name: str, **path_params: typing.Any) -> str:
         router = self._scope["router"]
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -18,9 +18,30 @@ class ClientDisconnect(Exception):
     pass
 
 
-class State:
-    def __init__(self, state_dict={}):
-        self.__dict__.update(state_dict)
+class State(object):
+    def __init__(self, state={}):
+        self._state = state
+
+    def __setattr__(self, key, value: typing.Any) -> None:
+        if key == '_state':
+            super(State, self).__setattr__(key, value)
+        else:
+            self._state[key] = value
+
+    def __getattr__(self, key) -> typing.Any:
+        if key == '_state':
+            super(State, self).__getattr__(key)
+        else:
+            if key in self._state:
+                return self._state[key]
+            else:
+                raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
+
+    def __delattr__(self, key) -> None:
+        if key == '_state':
+            super(State, self).__delattr__(key)
+        else:
+            del self._state[key]
 
 
 class HTTPConnection(Mapping):
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -34,9 +34,7 @@ class State(object):
             )
 
     def __delattr__(self, key: typing.Any) -> None:
-        if key == "_state":
-            super(State, self).__delattr__(key)
-        else:
+        if hasattr(self, "_state"):
             del self._state[key]
 
 
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -34,8 +34,7 @@ class State(object):
             )
 
     def __delattr__(self, key: typing.Any) -> None:
-        if hasattr(self, "_state"):
-            del self._state[key]
+        del self._state[key]
 
 
 class HTTPConnection(Mapping):
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -6,41 +6,47 @@ from starlette.responses import PlainTextResponse
 from starlette.testclient import TestClient
 
 
+class CustomMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request, call_next):
+        response = await call_next(request)
+        response.headers["Custom-Header"] = "Example"
+        return response
+
+
+app = Starlette()
+app.add_middleware(CustomMiddleware)
+
+
+@app.route("/")
+def homepage(request):
+    return PlainTextResponse("Homepage")
+
+
+@app.route("/exc")
+def exc(request):
+    raise Exception()
+
+
+@app.route("/no-response")
+class NoResponse:
+    def __init__(self, scope, receive, send):
+        pass
+
+    def __await__(self):
+        return self.dispatch().__await__()
+
+    async def dispatch(self):
+        pass
+
+
+@app.websocket_route("/ws")
+async def websocket_endpoint(session):
+    await session.accept()
+    await session.send_text("Hello, world!")
+    await session.close()
+
+
 def test_custom_middleware():
-    class CustomMiddleware(BaseHTTPMiddleware):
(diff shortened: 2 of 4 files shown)


starlette/requests.py
```diff
@@ -32,6 +33,9 @@ def __init__(self, scope: Scope, receive: Receive = None) -> None:
         assert scope["type"] in ("http", "websocket")
         self._scope = scope
 
+        # Ensure 'state' has an empty dict if it's not already populated.
+        self._scope.setdefault('state', {})
```
lovelydinosaur: Let's *only* do this if/when the `.state` property is first accessed.

starlette/requests.py
```diff
@@ -19,7 +19,8 @@ class ClientDisconnect(Exception):
 
 
 class State:
-    pass
+    def __init__(self, state_dict={}):
+        self.__dict__.update(state_dict)
```
lovelydinosaur: This isn't right - you'll instead want to save a private `_state_dict` attribute on the instance, and read/modify it in `__getattr__`/`__setattr__`/`__delattr__`.

didip: When I did that previously, I got infinite recursive calls. I will debug again later to figure out why.

That said, a little off-tangent here...

I found out that I can pass around singleton objects by just assigning them to app object. Example:

app.settings[‘conf’] = conf

Then I can call it with:

request.app.settings[‘conf’].

This is similar to how Tornado does it.

didip:
```
class State(object):
    def __init__(self, state={}):
        self._state = state

    def __setattr__(self, key, item: typing.Any) -> None:
        self._state[key] = item

    def __getattr__(self, key) -> typing.Any:
        return self._state[key]

    def __delattr__(self, key) -> None:
        del self._state[key]
```

## Error
```
starlette/requests.py:23: in __init__
    self._state = state
starlette/requests.py:26: in __setattr__
    self._state[key] = item
starlette/requests.py:29: in __getattr__
    return self._state[key]
starlette/requests.py:29: in __getattr__
    return self._state[key]
E   RecursionError: maximum recursion depth exceeded while calling a Python object
```

didip: I got it working using private `dict`:

```
class State(object):
    def __init__(self, state={}):
        self._state = state

    def __setattr__(self, key, value: typing.Any) -> None:
        if key == '_state':
            super(State, self).__setattr__(key, value)
        else:
            self._state[key] = value

    def __getattr__(self, key) -> typing.Any:
        if key == '_state':
            super(State, self).__getattr__(key)
        else:
            if key in self._state:
                return self._state[key]
            else:
                raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))

    def __delattr__(self, key) -> None:
        if key == '_state':
            super(State, self).__delattr__(key)
        else:
            del self._state[key]
```

I see now why you want to do this. Every state object has access to the same dict.

All tests passed, including the new ones I wrote. Not sure how the test coverage missed that 1 line...

didip: Code coverage still stubbornly skips this line: `super(State, self).__getattr__(key)`

Even though I wrote several asserts to test the existence of keys:

```
def test_request_state_object():
    scope = {'state': {'old': 'foo'}}

    s = State(scope['state'])
    assert s._state == scope['state']
    assert getattr(s, '_state') == scope['state']

    s.new = 'value'                   # test __setattr__
    assert s.new == 'value'           # test __getattr__
    assert s._state['new'] == 'value' # test if inner _state dict is updated.

    del s.new # test __delattr__

    try:
        assert s.new == 'value' # will bombed with AttributeError
    except AttributeError as e:
        assert str(e) == "'State' object has no attribute 'new'"

    del s._state # Deleting _state dict should not bombed.
```

starlette/requests.py
```diff
@@ -18,8 +18,30 @@ class ClientDisconnect(Exception):
...
+
+    def __setattr__(self, key, value: typing.Any) -> None:
+        if key == '_state':
+            super(State, self).__setattr__(key, value)
+        else:
+            self._state[key] = value
+
+    def __getattr__(self, key) -> typing.Any:
+        if key == '_state':
+            super(State, self).__getattr__(key)
```
lovelydinosaur: This is the single line missing coverage ATM

didip: Done. 100% coverage! 💯💥

didip:
Strange, why would it fail? It completed 100%.

lovelydinosaur:
Needs linting: https://travis-ci.org/encode/starlette/jobs/546425652#L401

Run `scripts/lint` to apply `black` and force a consistent code style everywhere.

starlette/requests.py
```diff
@@ -18,8 +18,24 @@ class ClientDisconnect(Exception):
...
+
+    def __getattr__(self, key: typing.Any) -> typing.Any:
+        if key in self._state:
+            return self._state[key]
+        else:
+            raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__.__name__, key))
+
+    def __delattr__(self, key: typing.Any) -> None:
+        if key == '_state':
+            super(State, self).__delattr__(key)
```
lovelydinosaur: Actually we don't need to care about this case.

tests/middleware/test_base.py
```diff
@@ -6,47 +6,46 @@
...
-        response = await call_next(request)
-        response.headers["Custom-Header"] = "Example"
-        return response
-
+def test_custom_middleware():
+    class CustomMiddleware(BaseHTTPMiddleware):
+        async def dispatch(self, request, call_next):
+            response = await call_next(request)
+            response.headers["Custom-Header"] = "Example"
+            return response
```
lovelydinosaur: It'd be better to leave all these bits as they were - it's super important that we always try to keep the footprint of a pull request absoultely minimal, and keep it scoped to one thing only.

starlette/requests.py
```diff
@@ -108,9 +127,10 @@ def user(self) -> typing.Any:
 
     @property
     def state(self) -> State:
-        if "state" not in self._scope:
-            self._scope["state"] = State()
-        return self._scope["state"]
+        if not hasattr(self, '_state'):
+            # Create a state instance with a reference to the dict in which it should store info
```
lovelydinosaur, line 130: Let's move the `self._scope.setdefault('state', {})` to here.

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
...
+    assert s._state['new'] == 'value' # test if inner _state dict is updated.
+
+    del s.new # test __delattr__
+
+    try:
+        assert s.new == 'value' # will bombed with AttributeError
+    except AttributeError as e:
+        assert str(e) == "'State' object has no attribute 'new'"
+
+    del s._state # Deleting _state dict should not bombed.
```
lovelydinosaur: We won't need this line, once we've dropped the `__delattr__` case for `_state`.

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
...
+    assert getattr(s, '_state') == scope['state']
+
+    s.new = 'value'                   # test __setattr__
+    assert s.new == 'value'           # test __getattr__
+    assert s._state['new'] == 'value' # test if inner _state dict is updated.
+
+    del s.new # test __delattr__
+
+    try:
+        assert s.new == 'value' # will bombed with AttributeError
```
lovelydinosaur: Just something like:

```python
with pytest.assertRaises(AttributeError):
    request.state.new
```

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
     assert disconnected_after_response
 
 
+def test_request_state_object():
```
lovelydinosaur, line 248: I think it'd probably be marginally nicer if we tested this by instantiating a request object, and accessing `request.state` on it.

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
...
+    scope = {'state': {'old': 'foo'}}
+
+    s = State(scope['state'])
+    assert s._state == scope['state']
+    assert getattr(s, '_state') == scope['state']
+
+    s.new = 'value'                   # test __setattr__
+    assert s.new == 'value'           # test __getattr__
+    assert s._state['new'] == 'value' # test if inner _state dict is updated.
```
lovelydinosaur: Let's not test against private API.

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
...
 
 
+def test_request_state_object():
+    scope = {'state': {'old': 'foo'}}
+
+    s = State(scope['state'])
+    assert s._state == scope['state']
+    assert getattr(s, '_state') == scope['state']
+
+    s.new = 'value'                   # test __setattr__
```
lovelydinosaur: We can probably drop the "test __setattr__" style comments.

tests/test_requests.py
```diff
@@ -240,11 +240,32 @@ def test_request_is_disconnected():
...
+
+    del s._state # Deleting _state dict should not bombed.
+
+
 def test_request_state():
     async def app(scope, receive, send):
         request = Request(scope, receive)
         request.state.example = 123
-        response = JSONResponse({"state.example": request["state"].example})
+        response = JSONResponse({"state.example": request.state.example})
```
lovelydinosaur, line 271: Yes, much nicer!

tests/test_websockets.py
```diff
@@ -281,5 +281,5 @@ def test_websocket_scope_interface():
         send=mock_send,
     )
     assert websocket["type"] == "websocket"
-    assert dict(websocket) == {"type": "websocket", "path": "/abc/", "headers": []}
-    assert len(websocket) == 3
+    assert dict(websocket) == {"type": "websocket", "path": "/abc/", "headers": [], 'state': {}}
```
lovelydinosaur: This won't have changed once we *only* set `state` to `{}` if `.state` is accessed.

tests/test_requests.py
```diff
@@ -174,8 +174,8 @@ def test_request_scope_interface():
     """
     request = Request({"type": "http", "method": "GET", "path": "/abc/"})
     assert request["method"] == "GET"
-    assert dict(request) == {"type": "http", "method": "GET", "path": "/abc/"}
-    assert len(request) == 3
+    assert dict(request) == {"type": "http", "method": "GET", "path": "/abc/", "state": {}}
+    assert len(request) == 4
```
lovelydinosaur, line 183: This won't have changed once we only set state to {} if .state is accessed.

tests/middleware/test_base.py
```diff
@@ -84,3 +83,41 @@ def homepage(request):
     response = client.get("/homepage")
     assert response.text == "Homepage"
     assert response.headers["Custom"] == "Example"
+
+def test_state_data_across_multiple_middlewares():
```
lovelydinosaur, line 89: I'd suggest that we could simplify this down to a simpler style of test that:

* Has a single middleware, which sets some state on the request.
* Has a view, which returns `JSONResponse(dict(request.state))` 

That'd be sufficient for ensuring that `.state` is persisted across two different request instances.

lovelydinosaur (changes requested):
Okay, this is *great* stuff!
I've added several review comments, for bits of cleanup. Let me knwo if anything's not clear there.
(Equally, I'm happy to pick this up and finish off the last bits if needed)

starlette/requests.py
didip: If I move that here, my multi middleware test failed.

tests/test_requests.py
didip: The pytest that is part of this project doesn't seem to have `assertRaises`.

tests/middleware/test_base.py
didip: When I moved `"state"` init inside `@property def state():`, my test bombed in a meaningful way:

The middle state var called `bar` was not persisted.

Stacktrace:

```
_____________________________________________________________________________ test_state_data_across_multiple_middlewares ______________________________________________________________________________

    def test_state_data_across_multiple_middlewares():
        expected_value1 = "foo"
        expected_value2 = "bar"
    
        class aMiddleware(BaseHTTPMiddleware):
                request.state.foo = expected_value1
                return response
    
        class bMiddleware(BaseHTTPMiddleware):
                request.state.bar = expected_value2
                response.headers["X-State-Foo"] = request.state.foo
                return response
    
        class cMiddleware(BaseHTTPMiddleware):
                response.headers["X-State-Bar"] = request.state.bar
                return response
    
        app = Starlette()
        app.add_middleware(aMiddleware)
        app.add_middleware(bMiddleware)
        app.add_middleware(cMiddleware)
    
        @app.route("/")
        def homepage(request):
            return PlainTextResponse("OK")
    
        client = TestClient(app)
>       response = client.get("/")

tests/middleware/test_base.py:122: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.7/site-packages/requests/sessions.py:546: in get
starlette/testclient.py:421: in request
    json=json,
venv/lib/python3.7/site-packages/requests/sessions.py:533: in request
venv/lib/python3.7/site-packages/requests/sessions.py:646: in send
starlette/testclient.py:238: in send
    raise exc from None
starlette/testclient.py:235: in send
/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py:584: in run_until_complete
    return future.result()
starlette/applications.py:133: in __call__
    await self.error_middleware(scope, receive, send)
starlette/middleware/errors.py:122: in __call__
    raise exc from None
starlette/middleware/errors.py:100: in __call__
starlette/middleware/base.py:25: in __call__
    response = await self.dispatch_func(request, self.call_next)
tests/middleware/test_base.py:109: in dispatch
    response.headers["X-State-Bar"] = request.state.bar
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <starlette.requests.State object at 0x10ab52e48>, key = 'bar'

    def __getattr__(self, key: typing.Any) -> typing.Any:
        if key in self._state:
            return self._state[key]
        else:
            raise AttributeError(
>               "'{}' object has no attribute '{}'".format(self.__class__.__name__, key)
            )
E           AttributeError: 'State' object has no attribute 'bar'

starlette/requests.py:33: AttributeError
```

starlette/requests.py
```diff
@@ -18,8 +18,24 @@ class ClientDisconnect(Exception):
...
+    def __getattr__(self, key: typing.Any) -> typing.Any:
+        if key in self._state:
+            return self._state[key]
+        else:
+            raise AttributeError(
+                "'{}' object has no attribute '{}'".format(self.__class__.__name__, key)
+            )
+
+    def __delattr__(self, key: typing.Any) -> None:
+        if hasattr(self, "_state"):
```
lovelydinosaur: I don't think we need the `if` condition here.

lovelydinosaur:
Great stuff! Followed up and completed this work in #557

didip:
Awesome! Thank you for helping to get it merged.

PR #555, opened 2019-06-17
Include `StaticFiles` in template docs

jordaneremieff:
Include `StaticFiles` app import in template docs example.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -7,12 +7,13 @@ what you want to use by default.
 ```python
 from starlette.applications import Starlette
 from starlette.templating import Jinja2Templates
+from starlette.staticfiles import StaticFiles
 
 
 templates = Jinja2Templates(directory='templates')
 
 app = Starlette(debug=True)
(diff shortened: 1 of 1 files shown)


docs/templates.md
```diff
@@ -7,20 +7,21 @@ what you want to use by default.
...
+app.mount('/static', StaticFiles(directory='static'), name='static')
 
 
 @app.route('/')
 async def homepage(request):
     return templates.TemplateResponse('index.html', {'request': request})
 ```
 
-Note that the incoming `request` instance must be included as part of the
+Note that the [StaticFiles](https://www.starlette.io/staticfiles/) application must be mounted and configured to serve the templates, and the incoming `request` instance must be included as part of the
```
lovelydinosaur: I think that may be a bit confusing there. The static files app isn't the thing that serves the templates. (Tho it may serve any media that's hyperlinked in the templates with `url_for`, such as images)

lovelydinosaur:
Thanks!

PR #556, opened 2019-06-17
Allow None for JSONResponse

sirex:
Previously `JSONResponse(None)` rendered empty string, but it should be `'null'` instead.

This change fixes it.


lovelydinosaur:
Yup!

lovelydinosaur:
Nice one!

PR #557, opened 2019-06-19
Request state and scope

### LINKED ISSUE #545: mypy errors when accessing Request.state
From https://www.starlette.io/requests/:
> Other state
>
> If you want to store additional information on the request you can do so using request.state.
>
> For example:
>
> request.state.time_started = time.time()

This causes mypy errors since any accessed/set attributes aren't present in the `State` class. As far as I can tell, there isn't a good way to tell mypy to expect certain attributes to potentially be set.

I could imagine solutions that somehow make `Request.state` generic, and those would probably be the most type-safe, but that seems like it would take a lot of work for little gain, and could make the API a little uglier.

As a workaround, I propose adding a `def __getattr__` in an `if typing.TYPE_CHECKING:` block within the class definition. This will silence the mypy errors without changing anything about how the class behaves during runtime.

One counter-argument to this approach would be that this would prevent state attribute accesses from being called out by mypy even if you want the warnings. But then you'd have to add `# type: ignore` to disable it, and would lose any ability to check the continued validity going forward. Given the suggested usage pattern, I just don't see much value to keeping these errors (though I could be convinced otherwise).

Alternatively, the state could be replaced by a `Dict[str, Any]` (or something that behaves more like one anyway), or given one as an attribute.

### LINKED ISSUE #551: request.state is not persisted between middlewares.

lovelydinosaur:
Follow up to #552

Adds `request.scope` and `request.state`.
Middleware now passes scope in-place.

Closes #545 
Closes #551
Closes #552

At a later point we'll also want to start moving towards deprecating the request-as-dict interface, in favor of `request.scope`.

### DIFF
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -6,41 +6,47 @@ from starlette.responses import PlainTextResponse
 from starlette.testclient import TestClient
 
 
+class CustomMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request, call_next):
+        response = await call_next(request)
+        response.headers["Custom-Header"] = "Example"
+        return response
+
+
+app = Starlette()
+app.add_middleware(CustomMiddleware)
+
+
+@app.route("/")
+def homepage(request):
+    return PlainTextResponse("Homepage")
+
+
+@app.route("/exc")
+def exc(request):
+    raise Exception()
+
+
+@app.route("/no-response")
+class NoResponse:
+    def __init__(self, scope, receive, send):
+        pass
+
+    def __await__(self):
+        return self.dispatch().__await__()
+
+    async def dispatch(self):
+        pass
+
+
+@app.websocket_route("/ws")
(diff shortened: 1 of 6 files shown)


tests/middleware/test_base.py
```diff
@@ -84,3 +84,46 @@ def homepage(request):
...
+
+
+def test_state_data_across_multiple_middlewares():
+    expected_value1 = "foo"
+    expected_value2 = "bar"
+
+    class aMiddleware(BaseHTTPMiddleware):
+        async def dispatch(self, request, call_next):
+            request.state.foo = expected_value1
+            print("a", dict(request))
```
didip, line 96: Don't you want to remove these print statements?

lovelydinosaur: Yes. Yes, I do. Thanks!

PR #559, opened 2019-06-20
Docs: fix WebSocket example

JoseKilo:
This example was updated from method to function here: https://github.com/encode/starlette/commit/9e962e79f24d3b09d5e832ad0a4bbf3e5ec917e1

Right now it raises:
```
TypeError: app() missing 1 required positional argument: 'send'
```


lovelydinosaur:
Thanks!

PR #561, opened 2019-06-24
Update framework example in readme to use ASGI 3

rafalp:
I've missed the code snippet in readme when doing #543


lovelydinosaur:
Thanks!

PR #563, opened 2019-06-24
Make 307 the default redirect status code

### LINKED ISSUE #464: Use "308 Permanent Redirect" for redirect slashes behavior.
Hi,
I stumbled upon a quirk in starlette that is not properly documented. It seems like all of my HTTP request to a route without trailing slash are being redirected to route with trailing slashes. Say I am hitting `http://hostname/mountpoint/api` it will be redirected (302) to `http://hostname/mountpoint/api/`. This messed up your api calls; if you call POST to `http://hostname/mountpoint/api` it will be redirected to GET `http://hostname/mountpoint/api/`.

I dig on the source and see this redirect_slashes flag and setting it to False fix this. I feel this behavior (the auto redirection) should be documented.

tiangolo:
:recycle: Make 307 the default redirect status code respecting HTTP method, and use 308 for HTTPS redirects (equivalently for 301).

This should fix #464.

---

I'm leaving `307` as the default, which is the equivalent of the current default `302`.

And I'm setting explicit `308` in the places where a `301` was used (HTTPS redirect middleware).

This also means that when using `redirect_slashes` it would use a `307`, as the equivalent of the currently used `302`. Although the discussion in that issue talks about only `308`, not `307`.

### DIFF
--- a/starlette/middleware/httpsredirect.py
+++ b/starlette/middleware/httpsredirect.py
@@ -13,7 +13,7 @@ class HTTPSRedirectMiddleware:
             redirect_scheme = {"http": "https", "ws": "wss"}[url.scheme]
             netloc = url.hostname if url.port in (80, 443) else url.netloc
             url = url.replace(scheme=redirect_scheme, netloc=netloc)
-            response = RedirectResponse(url, status_code=301)
+            response = RedirectResponse(url, status_code=308)
             await response(scope, receive, send)
         else:
             await self.app(scope, receive, send)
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
Yes please!
Thanks, @tiangolo.

tiangolo:
Awesome, thanks!

blueyed:
Thanks for this!

Strictly speaking it should be done for `request.scope['http_version'] >= '1.1'` (pseudo code) only though - i.e. `HTTP/1.0` does not support it (according to Wikipedia at least (https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection)).
Not a big issue probably (since you can still handle it yourself if necessary, but noticed this with an older test suite).

lovelydinosaur:
Thanks @blueyed - interesting catch!
I think we'll probably just omit that in the interests of pragmatism.

PR #565, opened 2019-06-26
handle kwargs if args is None

sander-su:
Make it possible to use starlette authentication decorators with fastapi

When using fastap (which is build on top of starlette) request is in kwargs and args is None when using the @ requires decorator. However args[0] is evaluated when the kwargs.get() is invoked and this results in a IndexError: tuple index out of range. 
Changing the evaluation of kwargs and args

### DIFF
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -42,8 +42,8 @@ def requires(
                 *args: typing.Any, **kwargs: typing.Any
             ) -> None:
                 websocket = kwargs.get("websocket")
-                if websocket == None:
-                    websocket =  args[idx]
+                if websocket is None:
+                    websocket = args[idx]
 
                 assert isinstance(websocket, WebSocket)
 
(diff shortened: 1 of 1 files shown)


sander-su:
@tomchristie

lovelydinosaur:
Thanks for this!

I've been a bit hesistent mostly because I'm not sure the one-size-fits-all decorater approach is really what we want to be doing. The "request" might be an arg might be a kwarg, let's just try to guess is really wooly behavior. (That's not anything critical of your PR, just that I'm a bit uncertain that we've got this right in the first place)

I think we might want the mull this over for a bit (and/or have a more in depth discussion with the FastAPI/Bocadillo teams about auth/permissions, in order to figure out good general patterns for us to adopt here)

sander-su:
Thanks for the response

For me the reason to have this on starlette and not a more high level framework is that we only need those frameworks is you are doing REST (in our case). If you have a graphql app then startlette is sufficient. 

Another approach could be to have some common middleware and common functions and frameworks should define the decorator and routes required.

I'm working on a oidc and jwt lib (which use oic) for which I think I have a first version soon.

Kludex:
I'm closing this as stale, and as it doesn't look there is interest on getting this merged.

PR #566, opened 2019-06-27
Use `http3` for TestClient.

lovelydinosaur:
We'll want to migrate to using `http3` for the TestClient, rather than `requests`, since it'll allow us to provide both a sync client, and an async client. We'll also be able to recommend using it as a async client for making HTTP requests.

Still a chunk of work needed to get this done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@
 
 **Documentation**: [https://www.starlette.io/](https://www.starlette.io/)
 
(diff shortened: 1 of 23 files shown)


blueyed:
Would be great if this helped with https://github.com/encode/starlette/pull/505 - maybe consider taking the new tests from there.

lovelydinosaur:
Closing this off as outdated.

PR #573, opened 2019-07-09
Tweak redirects and exception debugger

lovelydinosaur:
Default redirects actually should be 303 (which is what we want for eg. form redirects)
HTTPS middleware and Router slash redirects should explicitly be 308 (Permanent redirect, preserving method)

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -206,7 +206,7 @@ class ServerErrorMiddleware:
         }
         return FRAME_TEMPLATE.format(**values)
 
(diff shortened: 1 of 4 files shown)


blueyed:
> slash redirects should explicitly be 308 (Permanent redirect, preserving method)

JFI: note that this causes problems during development, where paths are likely to change etc and then the browser not re-trying (a common problem with Django when using permanent redirects there).

lovelydinosaur:
Okay. Happy to roll with 307 there instead.

lovelydinosaur:
I would have *thought* it shouldn’t make a difference either way unless cache headers are being added as well?

PR #574, opened 2019-07-13
Authentication Compatibility for Bocadillo

skrtu:
They use req, starlette uses request ¯\_(ツ)_/¯

Changes the authentication package to look for the "req" parameter instead of only for "websocket" or "request". This allows it to work with Bocadillo.

I'm on Windows :( so I didn't run tests yet. Will add them.

### DIFF
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -27,7 +27,7 @@ def requires(
         type = None
         sig = inspect.signature(func)
         for idx, parameter in enumerate(sig.parameters.values()):
-            if parameter.name == "request" or parameter.name == "websocket":
+            if parameter.name == "request" or parameter.name == "websocket" or parameter.name == "req":
                 type = parameter.name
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks! I'm going to be a "no" on this.
If anything I think it shows that our approach for a general purpose decorator that can be used against any other framework's interface just isn't a sensible thing for us to do. Going to rethink this in future versions.

florimondmanca:
Hey @skrtu,

It is true that the requires() decorator provided by Starlette is tied to Starlette’s own view system, hence not compatible with Bocadillo (the name of the request argument is not even the biggest problem to using the decorator — the view signature is completely different).

I agree with Tom’s decision to close this as Starlette should not have to change to fit the specific needs of downstream dependencies.

This might be worth an issue on Bocadillo repo, what do you think? :-)

PR #577, opened 2019-07-14
Request path regex

unittolabs:
Hey, guys!

Currently, I making the starlette middleware for the data dog tracing.
I faced the problem that I can't get the route path (regex) for using as an endpoint in statistics views.

For example:
We have user pages `/users/{username}`. When somebody visits the page we want to see the trace, but trace endpoint not should be exact (as `/users/user1` or `/users/user2`) but should be a template string as we set in route (`/users/{username}`). So I proxying with this value to the scope and then to the request object (near `path_params`). I think it will be useful.

Also, I add the status code checking functions from the DRF, and two more RFCs =)
Hope to merge it and release a new path for the framework, then I will continue with my TraceMiddleware.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,129 @@
-*.pyc
-test.db
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -1,7 +1,129 @@
...
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+
+# Custom project settings
+test.db
 starlette.egg-info/
-venv/
```
blueyed, line 7: Please consider using a global gitignore file yourself.

lovelydinosaur:
There's too many unrelated changes included in this pull request for us to be able to consider it in its current state.

I'm not not immediately sure if we do or don't want `path_re`. We need to think very carefully about anything we choose to introduce into the ASGI scope, as we're moving away from general ASGI compatibility anytime we do that.

I *think* it's probably more likely that we'd consider switching the current `endpoint` information, to instead store be a more general `route` or `routes`, so that we're saving the Route instance. (Or possibly, list of all Mounts and Routes that get matched.)

That way we'd be able to provide *any* kind of routing information we need, including `request.endpoint` or whatever else.

Happy to consider a more restricted pull request, that *only* deals with that aspect, and doesn't include unrelated changes. 😀

hadrien:
> Currently, I making the starlette middleware for the data dog tracing.

@unitto1 I am super interested in that. Will it be open source?

blueyed:
JFI: it reminded me of https://github.com/encode/starlette/pull/261.

unittolabs:
@hadrien
Yes, but currently we don't have a required functional in starlette. Hope guys will give me answers in #581 , then I will share the results.

PR #580, opened 2019-07-16
change: JSONResponse accepts JSONEncoder class for serialization

ProvoK:
Hello there!

### Rationale

As of now, it's not possible to pass to JSONResponse a custom JSONEncoder.

Working "hack" is the following.
```python
class MyEncoder(json.JSONEncoder):
    ...

class CustomJSONResponse(JSONResponse):
    def __init__(self, *args, encoder: MyEncoder, **kwargs):
        self.encoder = encoder
        super().__init__(*args, **kwargs)
        return json.dumps(
            ...
            cls=self.encoder,
        ).encode("utf-8")
```
To notice: it's mandatory to assign the encoder **before** calling the `super().__init__` because as of now the content is dumped at init time. Afaik, this is not a good practice in terms of inheritance.

### Why this PR

To permit to easily pass a JSONEncoder class at creation time
```python
JSONEncoder(data=myDataclassInstance, encoder=MyEncoder)
```

The changes are working but still in "wip", it's probably missing something about the docs and maybe more test cases.

PS: About the failing test case, I thing it's caused by [this commit](https://github.com/encode/starlette/commit/059b9894946c85366b777b8d3d724e046cd2e875). I could fix that in another PR if necessary

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -39,7 +39,7 @@ class Response:
         media_type: str = None,
         background: BackgroundTask = None,
     ) -> None:
-        self.body = self.render(content)
+        self.content = content
         self.status_code = status_code
         if media_type is not None:
             self.media_type = media_type
@@ -79,6 +79,10 @@ class Response:
             raw_headers.append((b"content-type", content_type.encode("latin-1")))
 
         self.raw_headers = raw_headers
+
+    @property
+    def body(self):
+        return self.render(self.content)
 
     @property
     def headers(self) -> MutableHeaders:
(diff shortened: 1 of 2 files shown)


ProvoK:
Furthermore: an even better approach could be one similar to Flask, in which you can assign to the App object a JSON encoder class ([ref](https://flask.palletsprojects.com/en/1.0.x/api/#flask.Flask.json_encoder)).

Ofc, this change is not that trivial

lovelydinosaur:
Thanks! At least right now I'd probably rather than we encourage folks to subclass `JSONResponse` and override `.render()`.

ProvoK:
Hey, sorry but I did not understand the current situation for this PR.

Are you suggesting to keep the JSONResponse as is and encourage subclassing in user's codebase? (and then, close this PR?)
PS: Probably the changes to the Response class are still a good improvement for the codebase...as of now the content is dumped at init time, making the class harder to extend

Kludex:
> Are you suggesting to keep the JSONResponse as is and encourage subclassing in user's codebase?

That's what I understood as well. PR welcome to enhance the documentation! :tada: 

> PS: Probably the changes to the Response class are still a good improvement for the codebase...as of now the content is dumped at init time, making the class harder to extend

Feel free to open as issue or go on our Gitter server, so we can discuss it.

Sorry for taking so long! I agree with @tomchristie, so I'll be closing this PR.

PR #581, opened 2019-07-16
Re-open request path regex

unittolabs:
Hey, guys! It's re-opening of #577

I re-pushed only relative to the changes.
Hope it will be merged =)

@tomchristie @blueyed

### DIFF
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -52,6 +52,13 @@ Query parameters are exposed as an immutable multi-dict.
 
 For example: `request.query_params['search']`
 
+#### Path regex
+
+Regular expresion of triggered route. Represented as a string.
+
+For example: `GET /users/unitto` => `request.path_re`
+will return something like a `/users/{username}` 
+
 #### Path Parameters
 
 Router path parameters are exposed as a dictionary interface.
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -82,6 +82,10 @@ class HTTPConnection(Mapping):
         return self.scope.get("path_params", {})
 
     @property
+    def path_re(self) -> str:
+        return self.scope.get("path_re", "")
+
+    @property
     def cookies(self) -> typing.Dict[str, str]:
         if not hasattr(self, "_cookies"):
             cookies = {}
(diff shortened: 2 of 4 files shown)


unittolabs:
seems one of **non** related test is down?

lovelydinosaur:
@unitto1 Yup - now resolved. If you merge master that bit should be sorted.

lovelydinosaur:
Okay, so...

I don't want "path_re" to be part of the scope - it's too overly specific.

However, I *would* probably be okay something like a `"routes"` piece of state that holds the complete list of Routes *and* any Mounts that we matched against.

We can then figure out exactly what interface we'd like the request to present on top of that state.

unittolabs:
@tomchristie mm... you propose to use state instead of scope? or place the `routes` in the scope?

currently, we can get the whole list of routes by `request.app.routes`, but the problem is in detection currently used one. the request contains only the callable of the endpoint and iteration about all of the routes to find out the current again seems to overhead.
so I see only one way - it's to modify `matches`  and save the required info when matches.full

lovelydinosaur:
> place the routes in the scope?

This.

> but the problem is in detection currently used one

Yup, I'm refering to detecting the currently used one. It's just that it's not necceassrily just a single `Route`. It may have been eg. `[Mount, Route]` or `[Mount, Mount, Route]`.

unittolabs:
@tomchristie yes, I lose the moment about mounting.

ok, will try to do this week.

should I place it in `request._scope['route']`?

unittolabs:
One thing is a little confusing me. As I understood the scope object should be a simple object because It used in uvicorn (https://www.uvicorn.org/#http-scope).
In this case, I think it should be a simple object with simple values, in our case it hashmap with different types of strings or iterables.
Placing here a Route or Mount will look oddly.

The second part is about a place where I can implement this.
Currently, it implementing in Route.matches \ WebSocketRoute.matches but I can extend it in a single place - Router.__call__
Does it a good idea?

lovelydinosaur:
I'm going to close off this and #1101 for now.

I do think that at some point we probably want to expose the `request.route` and make associated info, such as the regex available. Eg. `request.route.regex`.

However at the moment we're already putting a bunch of stuff that's outside the ASGI spec into the scope. I don't want to add any more until we've properly audited that, and also don't want any new functionality until we've got the existing tickets all well under control.

It's also very possible that there's actually more design questions lurking under the surface here once you start properly unpicking this issue (eg. what public API does/should `Route` expose? What about mounting?)

For now let's just file this under "we don't provide this".

PR #582, opened 2019-07-16
Setting a finite size to SpooledTemporaryFile.

### LINKED ISSUE #579: UploadFile hogs RAM
Good day,

We are using FastAPI to access large (1GB) files through starlette UploadFile and we have noticed that all the file is loaded into RAM.

Looking at the code I see that starlette uses SpooledTemporaryFile, however it initializes it [1] with default parameters, namely `max_size=0`. The thing is that SpooledTemporaryFile.write() methods calls `self._check()` to switch to disk IO, however with `max_size=0` it never does it [2] and hence the whole file is loaded to RAM.

I suggest to starlette's UploadFile to:
* Set positive `max_size` by default
* Accept user-defined value from constructor

What do you think?

[1] https://github.com/encode/starlette/blob/594b95c1da8c4ed2aff5629ab5ce128a18b6931b/starlette/datastructures.py#L428

[2] https://github.com/python/cpython/blob/master/Lib/tempfile.py#L650

haizaar:
Fixes #579

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -418,6 +418,7 @@ class UploadFile:
     """
     An uploaded file included as part of the request data.
     """
+    spool_max_size = 1024 * 1024
 
     def __init__(
         self, filename: str, file: typing.IO = None, content_type: str = ""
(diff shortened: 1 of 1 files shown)


haizaar:
Also exposed the default value as a class attribute to ease "configuration" through monkey-patching till it's properly configurable.

/CC https://github.com/tiangolo/fastapi/issues/390

lovelydinosaur:
Released as 0.12.4

Thanks!

PR #588, opened 2019-07-26
Fix memory leaks that may be caused by default parameters in BackgroundTasks

abersheeran:
Fix memory leaks that may be caused by default parameters in `BackgroundTasks`.

### DIFF
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -21,8 +21,8 @@ class BackgroundTask:
 
 
 class BackgroundTasks(BackgroundTask):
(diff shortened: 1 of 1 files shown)


erewok (approved):

erewok:
This PR seems fine to me. Does anyone know how to kick off a build in order to get the checks to pass so that it can be merged?

Kludex:
This was fixed on https://github.com/encode/starlette/pull/1173

Thanks @abersheeran, and sorry for taking so long! :sweat_smile:

PR #562, opened 2019-06-24
allow use of characters out of latin-1 in multipart form values

jaceksnet:
basic fix for issue #425

PR #589, opened 2019-07-28
Default multipart encoding should be utf-8 instead of latin-1

ciscorn:
I found #562 is not sufficient for many cases because major web browsers never/cannot append such charset information to request headers.

I've confirmed other frameworks (werkzeug, aiohttp and sanic) use utf-8 as the default encoding of multipart headers, instead of latin-1.

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -160,7 +160,7 @@ class MultiPartParser:
     async def parse(self) -> FormData:
         # Parse the Content-Type header to get the multipart boundary.
         content_type, params = parse_options_header(self.headers["Content-Type"])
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Looks really sensible, yup.

I think we should probably also tweak `_user_safe_decode` so that if a decode error is raised, then we fallback to `latin-1` (Which can't raise a decode error, since it has specified characters for all 256 code points in each byte.)

(Right now it's only falling back to `latin-1` if called with an unknown codec.)

ciscorn:
@tomchristie
I've modified `_user_safe_decode` to handle UnicodeDecodeError.

PR #593, opened 2019-08-04
Fix ignore inline comment.

### LINKED ISSUE #585: starlette should ignore "inline" comment in env file
Hi,
when using https://www.starlette.io/config/ , if we put 
```
VAR = 'value'   # this var is use for ....
```
we get error related to the expected role of the var.

For example in my case 
```
  File "app.py", line 292, in <module>
    Mount('/files', StaticFiles(directory=settings('JOPLIN_RESOURCES'))),
  File "[...]/lib/python3.7/site-packages/starlette/staticfiles.py", line 55, in __init__
    raise RuntimeError(f"Directory '{directory}' does not exist")
RuntimeError: Directory '/home/foxmask/.config/joplin-desktop/resources/'   # path to get the joplin `resources` to be able to display them' does not exist
```
or if that was for a HTTP port :
```
  File "app.py", line 303, in <module>
    uvicorn.run(main_app, host='0.0.0.0', port=settings('HTTP_PORT', cast=int, default=8001))
  File "[...]lib/python3.7/site-packages/starlette/config.py", line 59, in __call__
    return self.get(key, cast, default)
  File "[...]lib/python3.7/site-packages/starlette/config.py", line 69, in get
    return self._perform_cast(key, value, cast)
  File "[...]lib/python3.7/site-packages/starlette/config.py", line 103, in _perform_cast
    f"Config '{key}' has value '{value}'. Not a valid {cast.__name__}."
ValueError: Config 'HTTP_PORT' has value '8001   # listen port number'. Not a valid int.
```

alaminopu:
Closes #585

### DIFF
--- a/starlette/config.py
+++ b/starlette/config.py
@@ -78,6 +78,8 @@ class Config:
                 line = line.strip()
                 if "=" in line and not line.startswith("#"):
                     key, value = line.split("=", 1)
+                    if "#" in value:
+                        value = value.split("#", 1)[0]
                     key = key.strip()
                     value = value.strip().strip("\"'")
                     file_values[key] = value
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -12,6 +12,7 @@ def test_config(tmpdir, monkeypatch):
         file.write("# Do not commit to source control\n")
         file.write("DATABASE_URL=postgres://user:pass@localhost/dbname\n")
         file.write("REQUEST_HOSTNAME=example.com\n")
+        file.write("REQUEST_PORT=8000 # Port \n")
         file.write("SECRET_KEY=12345\n")
         file.write("BOOL_AS_INT=0\n")
         file.write("\n")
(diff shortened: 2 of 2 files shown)


blueyed:
For (clickable, linked) reference: #585

starlette/config.py
```diff
@@ -78,6 +78,8 @@ def _read_file(self, file_name: str) -> typing.Dict[str, str]:
                 line = line.strip()
                 if "=" in line and not line.startswith("#"):
                     key, value = line.split("=", 1)
+                    if "#" in value:
+                        value = value.split("#", 1)[0]
```
blueyed, line 82: This will fail for `SOME_URL=http://example.com#fragment`.

lovelydinosaur:
Updated the description and title to correctly reference the corresponding issue.

starlette/config.py
lovelydinosaur: Indeed, we'd need a smarter approach here.

alaminopu: @blueyed Thanks for pointing out the issue might occur. I will think about a smarter approach here.

lovelydinosaur:
Let's treat this as out of scope, at least for now. See https://github.com/encode/starlette/issues/585#issuecomment-521507083

Happy to accept documentation pull requests, tho.

Issue #599, opened 2019-08-16
Debugger content should be escaped


abersheeran:
![TIM截图20190817000318](https://user-images.githubusercontent.com/25364828/63181405-7f90fa80-c082-11e9-84a2-8e98d05930a3.png)
![TIM截图20190817000415](https://user-images.githubusercontent.com/25364828/63181441-91729d80-c082-11e9-837a-db66b91035f5.png)

lovelydinosaur:
Closed via #601

PR #601, opened 2019-08-16
Fix html escape

abersheeran:
#599 

By the way, apologize for the last Pull request that was not tested locally.
This time I tested it correctly in the local.

TIM截图20190817004316


lovelydinosaur:
Great stuff - yup!

Are there any other places in the debug output that we need to be escaping that we're not already doing so?

abersheeran:
I have checked other places, maybe not.

PR #602, opened 2019-08-16
Remove "self" argument from websockets example

jlaine:
Fix the websockets example by removing the bogus "self" argument from the app
callable.


jlaine:
I just realised this is actually a duplicate of #559

Issue #608, opened 2019-08-21
"url_for" signature prevents use of "name" in path arguments


dansan:
I'm implementing an API in Fastapi, using `APIRouter`s for different resources.
When I want to get the URL of an item in a resource, I use the `starlette.requests.Request` objects `url_for()` method.
When I have a function with a path `/resource/{name}` I cannot use `url_for()` to get the URL, because it will result in:
```
TypeError: url_for() got multiple values for argument 'name'
```
Reproduce with:
```
def func(name, **path_params):
    pass
func("foo", name="bar")
```

As a workaround I have renamed the path variable in my API, but `name` was really fitting, and what I wanted to use, and not at all an uncommon variable name.
From what I see, `name` in `url_for()` seems to be the name of the referenced method in `APIRouter`. Changing the signature of `url_for()` to use `router_method` instead of `name`, would make a collision with the key word arguments in `path_params` less likely. Result would be:
```
def url_path_for(self, router_method: str, **path_params: str) -> URLPath:
```
An inspection on the used arguments could be used to log a deprecation warning until the next release, when `url_path_for()` is used with a named `name=...` instead of positional.

If this proposal is acceptable, I'd create a PR.

lovelydinosaur:
I’d suggest we change it to an unnamed argument. We could do this by changing the argument signature to *args, **kwargs, and ensuring that len(args) == 1

dansan:
PR: https://github.com/encode/starlette/pull/611

dansan:
Rebased on top of current master.

chbndrhnns:
I was hit by that issue just today and would like to see it merged.

tzmara:
Had a problem here too and took me around 30mins to find out why. Looked if there is an issue: Yeah. It is. Since 2019.

PR #611, opened 2019-08-22
allow 'name' in key word arguments of url_for() and url_path_for() (#608)

dansan:
Patch replaces
```
def url_path_for(self, name: str, **path_params: str) -> URLPath:
```
with
```
def url_path_for(self, *args: str, **kwargs: str) -> URLPath:
    assert len(args) == 1, "Exactly one positional argument required: the routes name."
    name = args[0]
```
(also for `url_for()`)
Not entirely sure about the assertion text...

Without the above change, these existing (now modified) tests fail:
* tests/test_routing.py::test_url_path_for
* tests/test_routing.py::test_url_for
* tests/test_routing.py::test_reverse_mount_urls

And this new one:
* tests/test_requests.py::test_request_url_for


dansan:
Were multiple commits, because I had forgotten coverage and black. I squashed them together, thus the force-push.

tests/test_routing.py
```diff
@@ -15,7 +15,7 @@ def users(request):
 
 
 def user(request):
-    content = "User " + request.path_params["username"]
+    content = "User " + request.path_params["name"]
```
lovelydinosaur: Let's leave this as `username` throughout - since it makes more sense in this context.

starlette/routing.py
```diff
@@ -184,15 +184,18 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
                     return Match.FULL, child_scope
         return Match.NONE, {}
 
-    def url_path_for(self, name: str, **path_params: str) -> URLPath:
-        seen_params = set(path_params.keys())
+    def url_path_for(self, *args: str, **kwargs: str) -> URLPath:
+        assert (
+            len(args) == 1
+        ), "Exactly one positional argument required: the routes name."
```
lovelydinosaur: How about we use...

```python
assert len(args) > 0, "Missing route name as the first argument."
assert len(args) < 2, "Invalid positional argument passed."
```

That'd be a bit more verbose, codewise, but it'd be nice since it's a user-facing error that's reasonably likely to be seen, and it'd be useful if we can be as helpful as possible.

lovelydinosaur (changes requested):
Lovely stuff - couple of minor changes suggested - otherwise looks ace!

dansan:
I reverted the `username` → `name` changes, added a decorator to check the `*args` and added a test for the assertions.
I didn't use the decorator in `templating.py`, because I'm unsure about importing issues/policies.
(Wow - Travis is really clobbering me ;) Enforcing good standards and more automated QA... I think I want that at my work too...

starlette/routing.py
```diff
@@ -28,6 +29,16 @@ class Match(Enum):
     FULL = 2
 
 
+def assert_args_is_only_route(func: typing.Callable) -> typing.Callable:
+    @functools.wraps(func)
+    def wrapper(self: BaseRoute, *args: str, **kwargs: str) -> URLPath:
+        assert len(args) > 0, "Missing route name as the first argument."
+        assert len(args) < 2, "Invalid positional argument passed."
+        return func(self, *args, **kwargs)
```
blueyed: Should raise TypeError instead.
("assert" might be a no-op)

blueyed: The decorator could be renamed to "verify_args_is_only_route" then.

starlette/templating.py
```diff
@@ -54,9 +54,11 @@ def __init__(self, directory: str) -> None:
 
     def get_env(self, directory: str) -> "jinja2.Environment":
         @jinja2.contextfunction
-        def url_for(context: dict, name: str, **path_params: typing.Any) -> str:
+        def url_for(context: dict, *args: str, **kwargs: typing.Any) -> str:
+            assert len(args) > 0, "Missing route name as the first argument."
+            assert len(args) < 2, "Invalid positional argument passed."
```
blueyed: Also TypeError here instead.

tests/test_requests.py
```diff
@@ -285,6 +286,24 @@ def test_request_cookies():
     assert response.text == "Hello, cookies!"
 
 
+def test_request_url_for():
+    app = Starlette()
+
+    @app.route("/users/{name}")
+    async def func_users(request):  # pragma: no cover
+        ...
```
blueyed: Instead of "pragma: no cover" consider "raise NotImplementedError()", which could be automatically ignored then (not sure if that was taken already though.. :/)

starlette/routing.py
dansan: Done in 6bf9b0e.

tests/test_requests.py
dansan: Didn't know about that one… had to adapt the coverage config (`tests/.ignore_lifespan`).

dansan:
Hi @blueyed ,
thanks for the review.
I'm unsure about the procedure here. Github was nagging about the "Changes requested" by @tomchristie , so I clicked "request re-review" to get rid of it ^.^ although you're doing it now. Not intended as a rejection of your review, just to make github happy.
I made the changes you requested.

nwalsh1995:
Bumping this PR since I am affected by this same problem and would love to see this land!

dansan:
I updated the code to be mergable to master.

lovelydinosaur:
Great stuff.

I'd probably prefer if we drop the `verify_args_is_only_route` decorator. The method on the `Router` class should make a proper type check.
The `Route`/`Mount`/`WebSocketRoute` cases can just perform a dumber `assert len(args) == 1` check.

dansan:
I replaced the decorator with the asserts and the type check: f4f4cd1f7843e0e6f02fdc4b382facc671accea6

fhscholl:
Any possibility this is going to get merged?

Kludex:
We still have this "bug". Would you mind rebasing it? I know 2.5 years is a lot... But would you mind rebasing it @dansan ?

If not, I'd like to close the PR and creating an issue about this.

dansan:
@Kludex Sorry, had no time before the holidays to get into this.
Patch is rebased.
Thanks for waking this old issue up.

dansan:
@tomchristie could you approve the workflow to run?

aminalaee:
@dansan Can you resolve the conflict please?

dansan:
Rebased on master. Sorry for the delay.

starlette/templating.py
```diff
@@ -65,7 +65,12 @@ def _create_env(
...
         @pass_context
-        def url_for(context: dict, name: str, **path_params: typing.Any) -> str:
+        def url_for(context: dict, *args: str, **path_params: typing.Any) -> str:
+            if len(args) < 1:
+                raise TypeError("Missing route name as the second argument.")
+            elif len(args) > 1:
+                raise TypeError("Invalid positional argument passed.")
+            name = args[0]
             request = context["request"]
             return request.url_for(name, **path_params)
```
aminalaee: Can you help me understand this change? Won't this work?
```suggestion
        def url_for(context: dict, *args: str, **path_params: typing.Any) -> str:
            request = context["request"]
            return request.url_for(*args, **path_params)
```

tests/test_templates.py
```diff
@@ -29,3 +29,13 @@ def test_template_response_requires_request(tmpdir):
     templates = Jinja2Templates(str(tmpdir))
     with pytest.raises(ValueError):
         templates.TemplateResponse("", {})
+
+
+def test_template_env_url_for_args(tmpdir):
```
aminalaee: And maybe we won't need this with that change?

aminalaee (commented):
Thank you for the update.

starlette/templating.py
dansan: It'd work. But we couldn't be sure that the caller of the function uses it as intended.
In the original signature it's obvious, that the first argument must the the routes name, as its the name of the argument. But by removing the variable name (replacing  with `*args`) the caller now has no indication what to do.
Those `TypeErrors` are our *communication* to future starlette users.

aminalaee: Yeah but you can (or already did) the validation in `request.url_for`, right? Isn't this duplicated?

dansan: Yes, you are right. The check can be removed and the one in `request.url_for()` will catch it.
But I fear, that it'll be *confusing* to get the exception message `"url_for() takes exactly one positional argument"` from `request.url_for()` when using the `templating....url_for()` with wrong arguments, as *that* requires exactly 2 positional arguments.
I am unsure what is the best way forward: reduced code vs explicit error messages. As in the long run the maintainers will have to deal with this decision, I'll defer it to them. If you are one of them (or simply confident enough), I'll do the change, no more discussion, just write "do it" :)

dansan: @aminalaee ping. Please decide. I'll follow your decision.

aminalaee: I think I'd use the `url_for` validation and don't do it in the templating. I was waiting for other feedback.

dansan: Done (afefc16).

tests/test_templates.py
dansan: Removed (afefc16).

dansan:
Branch is up2date with encode/starlette→master.

dansan:
Hum.. .I tried linking the PR to the existing issue (encode/starlette#608) by editing the commit message according to what is documented in https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue, but it's not working. The button for the manual way is not enabled for me.

... Ah - but after a while this PR is now linked in the issue. Maybe Github just needs a while to process this...

dansan:
@tomchristie could you approve the workflow to start? Branch could be merged into master if the pipeline is successful. Thank you.

starlette/routing.py
```diff
@@ -239,7 +239,9 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
                     return Match.FULL, child_scope
         return Match.NONE, {}
 
-    def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
+    def url_path_for(self, *args: str, **path_params: typing.Any) -> URLPath:
+        assert len(args) == 1, "url_path_for() takes exactly one positional argument"
```
lovelydinosaur: Let's raise `TypeError` on incorrect usages, rather than `assert`.

starlette/routing.py
```diff
@@ -308,7 +310,9 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
                 return Match.FULL, child_scope
         return Match.NONE, {}
 
-    def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
+    def url_path_for(self, *args: str, **path_params: typing.Any) -> URLPath:
+        assert len(args) == 1, "url_path_for() takes exactly one positional argument"
```
lovelydinosaur: Same here.

starlette/routing.py
```diff
@@ -381,7 +385,9 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
...
```
lovelydinosaur: Here.

starlette/routing.py
```diff
@@ -451,7 +457,9 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
...
```
lovelydinosaur: And here.

lovelydinosaur (changes requested):
This is a nice improvement, isn't it? Good work. 👍

One set of review changes to be addressed.

starlette/routing.py
dansan: Fixed in 1892edd.

starlette/routing.py
dansan: Fixed in 1892edd.

starlette/routing.py
dansan: Fixed in 1892edd.

starlette/routing.py
dansan: Fixed in 1892edd.

Kludex:
I've set the milestone to 0.2x.x (the release after 0.20.0).

adriangb:
Can we somehow type this (maybe via overloads?) as `def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath`? That will tell type checkers that `__name` is a positional only param (even on 3.7). [Ref](https://peps.python.org/pep-0484/#positional-only-arguments)

adriangb:
My main issue with this is that using `*args` makes it harder for users to understand what the correct parameters are.

I'd think we'd want `(/, name: str, **path_params: Any) -> URLPath` or `(name: str, path_params: Mapping[str, Any]) -> URLPath` as our end state. The former will be possible in 2023 when 3.7 is EOL and the latter is possible now. But they are both breaking changes. So maybe as part of this PR we can choose our end goal/state and add a deprecation warning to start nudging us in that direction?

Kludex:
Is it bad to replace `name` by `__name`? @adriangb

adriangb:
> Is it bad to replace `name` by `__name`? @adriangb

It means that anyone calling `url_for(name="foo")` where `"foo"` is the route name would experience a breaking change. But the same thing would happen with the current patch of `name = args[0]`. I think the only difference with the `*args` version is that the `__name` version (and subsequently the `(`/, name: str, ...)` version once Python 3.7 is our minimum version) put the intent in the function signature instead of burying it in the implementation.

Kludex:
Can we deprecate the parameter "name"? Maybe warning when the caller uses "name=".

See:
```python
import inspect


def potato(a: int, b: str) -> str:
    print(inspect.stack()[-1].code_context)
    return b


potato(1, b=2)
```

alex-oleshkevich:
> Can we deprecate the parameter "name"? Maybe warning when the caller uses "name=".

This is a good solution to have before 1.0.
After we can have something like `def url_for(self, name_:str, **path_params: Any) -> URL` and clearly reserve `name_` for internal use by documenting it.

Kludex:
> > Is it bad to replace `name` by `__name`? @adriangb
> 
> It means that anyone calling `url_for(name="foo")` where `"foo"` is the route name would experience a breaking change. But the same thing would happen with the current patch of `name = args[0]`. I think the only difference with the `*args` version is that the `__name` version (and subsequently the `(`/, name: str, ...)` version once Python 3.7 is our minimum version) put the intent in the function signature instead of burying it in the implementation.

Let's take a decision here.

Options:
1. Go with this PR, and further revert + `/` on the signature.
2. Go with a less possible name like `__name`, and further addition of `/` to the signature.
3. ?

abersheeran:
This PR may need to be considered in conjunction with #1385 .

There are several solutions to solve these two problems at the same time:

- `def url_for(self, name: str, path_params: dict[str, Any], query_params: dict[str, Any]) -> str`
- `def url_for(self, *args: str, **kwargs: str) -> URL`: Add query params to URL object.
- `def url_for(self, *args: str, **kwargs: str) -> str`: Unknown parameters in kwargs as query params. (like flask)
- `def url_for(self, __name: str, **path_params: dict[str, Any], __query_params: dict[str, Any]) -> str`
- `def url_for(self, __name: str, **path_params: dict[str, Any]) -> URL`

Kludex:
I believe we prefer this one: `def url_for(self, __name: str, **path_params: dict[str, Any]) -> URL`.

alex-oleshkevich:
Leading underscores are less discoverable. When you type `n` IDEs suggest you `name` parameter. When the parameter starts with `_` then the autocompleting list if empty or not helpful.

What if we do like this `def url_for(self, name_: str, **path_params: dict[str, Any]) -> URL`. This is a far more common pattern I see in the codebases. And I don't remember if I ever seen `name_` query parameter.

Kludex:
The goal is to be less discoverable. :thinking: 

The idea is to move to `def url_for(self, name: str, /, **path_params: dict[str, Any]) -> URL` when Python 3.7 support is no longer needed.

Kludex:
I'm not sure if my last comment was that clear.

The goal of having `__name` is to signalize that we want it to be only a positional argument, so we can move to `name: str, /, **path_params)` later on.

alex-oleshkevich:
There is no difference between `__name` or `name_`. But from DX perspective `name_` is more clear.

Kludex:
> There is no difference between `__name` or `name_`. But from DX perspective `name_` is more clear.

I see the goal here being exactly the opposite... Why would you want the IDE to show `name__`? People would get tempted to use the keyword and write `url_for(name__="name")`. I want to avoid people from writing the parameter so in a future release it can be removed. :thinking:

Kludex:
Is it possible to have a different signature depending on the Python version? Maybe we should already do this...

alex-oleshkevich:
> Is it possible to have a different signature depending on the Python version? Maybe we should already do this...

I am aware that it could be an unexpected side effect when devs will update the project's python version.
Maybe we should enforce it at `mypy` level at this state?

Kludex:
> > Is it possible to have a different signature depending on the Python version? Maybe we should already do this...
> 
> I am aware that it could be an unexpected side effect when devs will update the project's python version. Maybe we should enforce it at `mypy` level at this state?

You mean this?

```python
import typing


if typing.TYPE_CHECKING:
    def potato(__name: str, /) -> str:
        """Potato."""
        return __name
else:
    def potato(__name: str) -> str:
        """Potato."""
        return __name

potato(__name="potato")  # ERROR HERE
```

It's not possible due to syntax error. :cry:

Kludex:
Anyone subscribed here is against the `__name` proposal with further addition of positional-only argument on `__name` (when we drop support for Python 3.7)?

Kludex:
- Closed by https://github.com/encode/starlette/pull/2050

PR #615, opened 2019-08-24
[WIP] Allow using multipart/form-data to upload files with graphql

jareks:
Currently sending multipart/form-data to Starlette's GraphQL results in 415 "unsupported media type" error.
This PR adds basic support for multipart/form-data which can be used for uploading files with GraphQL.

I am not sure if this is good way to implement this functionality and would  greatly appreciate any comments. For what it's worth, this solution works fine in my pet project.

Regarding tests, I am happy to look into them once I get confirmation this PR goes in right direction.

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -7,6 +7,7 @@ from starlette.concurrency import run_in_threadpool
 from starlette.requests import Request
 from starlette.responses import HTMLResponse, JSONResponse, PlainTextResponse, Response
 from starlette.types import Receive, Scope, Send
+from starlette.datastructures import UploadFile, MultiDict
 
 try:
     import graphene
@@ -54,6 +55,7 @@ class GraphQLApp:
         await response(scope, receive, send)
 
     async def handle_graphql(self, request: Request) -> Response:
+        files = MultiDict()
         if request.method in ("GET", "HEAD"):
             if "text/html" in request.headers.get("Accept", ""):
                 if not self.graphiql:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Before proceeding with this - does the GraphQL spec support multipart uploads.
Could you link to any related or relevant docs so that I've got something to check the implementation against. Thanks! 😃

jareks:
Here is the relay modern example is was starting with: https://github.com/facebook/relay/issues/1844#issuecomment-316893590
Please note this uses `uploadables` parameter of fetchQuery as described in Relay's network doc: https://relay.dev/docs/en/network-layer

I see second option: https://github.com/jaydenseric/graphql-multipart-request-spec
This spec seems to be a bit more complicated and I didn't need extra mapping functionality.

In summary: this PR adds very simple method to allow uploading files. Let me know if you think this is useful :)

anilwarbhe:
> Currently sending multipart/form-data to Starlette's GraphQL results in 415 "unsupported media type" error.
> This PR adds basic support for multipart/form-data which can be used for uploading files with GraphQL.
> 
> I am not sure if this is good way to implement this functionality and would greatly appreciate any comments. For what it's worth, this solution works fine in my pet project.
> 
> Regarding tests, I am happy to look into them once I get confirmation this PR goes in right direction.

I am getting the same error 415. Need help on how to achieve file upload in starlette graphql.

Regards

JayH5:

Issue #612, opened 2019-08-22
Application State


elyobo:
It's now possible to store [arbitrary state](https://www.starlette.io/requests/#other-state) on requests (via https://github.com/encode/starlette/pull/404); it would be great to be able to store arbitrary application state on the application and have it available in requests as well.

It seems common to use globals to configure things (e.g. [have a `settings.py` somewhere that's imported all over a project](https://www.starlette.io/config/#a-full-example)), but this seems far less sane (especially for tests, where you might want to verify various different settings) than providing configuration to your application and accessing it from there.

https://github.com/encode/starlette/issues/64 provides a useful point at which state could be attached, potentially, if only there was somewhere to put it like the request state.

What I have done for the moment is to use middleware to wrap up the things that I want to share with all requests and inject that in to the request state instead; it's an acceptable workaround for me, but if I was adding state in the lifecycle hooks I'd have to add the middleware there as well and it starts to get pretty messy.

example

```python
from myproject import settings
@app.middleware('http')                                                      
async def provide_request_settings(request: Request, call_next):
  """Injects settings in to all requests"""                                  
  request.state.settings = settings
  return await call_next(request)                                         
```

octopoulpe:
Hi. Very personal opinion coming here.

I think your middleware is exactly what needs to be done. I do not think it is a workaround as global variables/states should be made local as soon as you go down in your application scope and business modules.

That way you can manage your own state lifecycle in your main application file (with big responsibilities), and treat these variables as stateless local variables in your business methods/modules.

I also think that lifecycle hooks should be centralized in your main file, by calling methods of your submodules explicitly (with control over arguments and order), rather than using @on_event everywhere in your submodules. That way the workflow is explicit and anybody that reads your application main file is able to understand it.

Also, as you say, an external file imported everywhere with a shared state IS a workaround.

elyobo:
I think we largely agree.

> I think your middleware is exactly what needs to be done.

Yes, it's totally fine for this trivial example.

> Also, as you say, an external file imported everywhere with a shared state IS a workaround.

It is, but it's not a good one, and one with major problems for testing - I prefer to keep things as contained as possible and pass in all dependencies instead of reaching out and grabbing globals all over the place.

> I also think that lifecycle hooks should be centralized in your main file, by calling methods of your submodules explicitly (with control over arguments and order), rather than using @on_event everywhere in your submodules.

Absolutely agreed also; but the lack of application state makes this worse.

[...]

PR #620, opened 2019-08-27
Document the `request.app` property.

elyobo:
* Mention it in the Request documentation
* Mention it in the Application documentation in the context of accessing the application state

See #612.

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -80,10 +80,14 @@ exceptions that occur within the application:
 ### Storing state on the app instance
 
 You can store arbitrary extra state on the application instance, using the
-generic `app.state` attribute.  This can then be accessed in middleware or endpoints via `request.app`.
+generic `app.state` attribute.
 
 For example:
 
 ```python
 app.state.ADMIN_EMAIL = 'admin@example.org'
 ```
+
+### Acessing the app instance
+
+Where a `request` is available (i.e. endpoints and middleware), the app is available on `request.app`.  For other situations it can be imported from wherever it's instantiated.
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -80,7 +80,7 @@ exceptions that occur within the application:
 ### Storing state on the app instance
 
 You can store arbitrary extra state on the application instance, using the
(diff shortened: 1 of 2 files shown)


docs/applications.md
```diff
@@ -80,7 +80,7 @@ exceptions that occur within the application:
 ### Storing state on the app instance
 
 You can store arbitrary extra state on the application instance, using the
-generic `app.state` attribute.
+generic `app.state` attribute.  This can then be accessed in middleware or endpoints via `request.app`.
```
lovelydinosaur: Rather than having `request.app` mentioned specifically wrt. `request.app.state`, I'd rather we had a one or two sentence section "Accessing the app instance", which just notes that you can either import the application instance, or that it is also available as `request.app`. (Makes a bit more sense to me that way around). Make sense?

lovelydinosaur (changes requested):
Good call - One minor comment against this.

docs/applications.md
elyobo: Yeah makes sense to have a separate section to describe how the app instance can be documented.

I can't think of any reason to prefer importing a global vs using a local variable if it's available (if you're going to do that there's really no need for the state on the application in the first place, you can use any global state), so think that emphasising the use of `request.app` over importing is worthwhile.

dmontagu: It could make sense if you want to access the `app.state` from a place where you don't have access to a request (e.g., in a repeating task that runs periodically).

(It wouldn't *necessarily* require importing, e.g. if the task was configured in the same file as the app was created, but either way would involve accessing the app instance without being able to read it off of a request.)

elyobo:
Yes, true.

docs/applications.md
elyobo: Updated version pushed, but mentions the use of `request.app` first.  Feel free to edit directly if you prefer different wording @tomchristie.

lovelydinosaur (approved):

lovelydinosaur:
Looks good, yup! Thank you!

elyobo:
Great, thanks!

PR #623, opened 2019-08-29
don't return "error" in graphql response when no error occurs

### LINKED ISSUE #622: GraphQL response should not include error key if no error occured
The [GraphQL Spec](https://graphql.github.io/graphql-spec/June2018/#sec-Errors) states that:
> If no errors were encountered during the requested operation, the errors entry should not be present in the result.

Currently, if no errors are encountered, starlette will return `{"data": {...}, "errors": null}`. 
This is only a small thing, but enough to break some clients.

I have a PR for this incoming.

joshainglis:
Fixes #622

Signed-off-by: Josha Inglis <joshainglis@gmail.com>

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -107,7 +107,9 @@ class GraphQLApp:
             if result.errors
             else None
         )
-        response_data = {"data": result.data, "errors": error_data}
+        response_data = {"data": result.data}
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Great, thank you!

Issue #625, opened 2019-08-30
Multiple directories in StaticFiles


taoufik07:
The feature request is to be able to serve different directories under the same route.

The `StaticFiles` app only allows serving from one directory,  so we can't have have multiple apps under the same route.

### Suggestions :

1- Add an `add_directory` method to `StaticFiles` app, to allow adding multiple directories to the the same `StaticFiles` app and it can be used like so :
```python
static_app = StaticFiles(directory="static")
static_app.add_directory("another_static_dir")
```
We can also add `add_directories` which accepts a list of directories
* I've made a PR for the `add_directory` #626

2- Allow passing a list of directories in `init`, I know it's not a good idea to change the api :/, and to avoid that one way could be adding a kwarg `directories: List[str]`, I'm also aware it's not a good idea :3 (even worse) to expose two ways of doing things `directory` and `directories`but  the way of implementing it should be discussed, ofc if accepted.

lovelydinosaur:
Good summary on this issue, yup!
I'm not really sure what we want the API for this to look like, bit conflicted.

taoufik07:
We can now add the `add_directory`, and later discuss how to pass multiple directories in the constructor, if we wanna allow it and change the existing API ofc.

as-dg:
Any news on this?

Kludex:
@aminalaee you've been creating some PRs around `StaticFiles`. Do you think this makes sense?

aminalaee:
@Kludex yeah. It does make sense. There is no conclusion about the API for it.

Honestly I don't think this is very nice:

```python
static_app = StaticFiles(directory="static")
static_app.add_directory("another_static_dir")
```

We did handle multiple directories for third party packages:

```python
StaticFiles(packages=[('bootstrap4', 'static')])
```

But for local directories I think we can work around it.
Maybe later on we can change the signature to handle `directories` too.

Kludex:
Could we accept both `str` and `List[str]` on `directory`? Is that bad? Or should we create `directories` and make it mutual exclusive with `directory`?

lovelydinosaur:
Closing with the same rationale as https://github.com/encode/starlette/issues/1214

There's plenty decent enough ways users *can* do this if they're really keen. Eg. subclass `StaticFiles` and override [the `get_directories` method](https://github.com/encode/starlette/blob/1ea624bf10c2dfd3a4e2f23888fa4df383e4b6b1/starlette/staticfiles.py#L55).

[...]

PR #626, opened 2019-08-30
Allow adding directories to StatiFiles app

taoufik07:
#625 Implementing the `add_directory`


Kludex:
Closing this as we want to decide the right API for `directories`.

Thanks for the PR @taoufik07 ! :)

Sorry for taking so long to act here!

PR #629, opened 2019-09-01
Add support for server push

jlaine:
This adds support for HTTP/2 and HTTP/3 server push by adding a
`Request.send_push_promise` method, which signals to push-enabled
servers that a push should be sent.

### DIFF
--- a/docs/server-push.md
+++ b/docs/server-push.md
@@ -1,0 +1,34 @@
+
+Starlette includes support for HTTP/2 and HTTP/3 server push, making it
+possible to push resources to the client to speed up page load times.
+
+### `Request.send_push_promise`
+
+Used to initiate a server push for a resource. If server push is not available
+this method does nothing.
+
+Signature: `send_push_promise(path)`
+
+* `path` - A string denoting the path of the resource.
+
+```python
+from starlette.applications import Starlette
+from starlette.responses import HTMLResponse
+from starlette.staticfiles import StaticFiles
+
+app = Starlette()
+
+
+@app.route("/")
+async def homepage(request):
+    """
+    Homepage which uses server push to deliver the stylesheet.
+    """
+    await request.send_push_promise("/static/style.css")
+    return HTMLResponse(
+        '<html><head><link rel="stylesheet" href="/static/style.css"/></head></html>'
+    )
+
+
+app.mount("/static", StaticFiles(directory="static"))
+```
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -6,12 +6,21 @@ from collections.abc import Mapping
 
 from starlette.datastructures import URL, Address, FormData, Headers, QueryParams, State
 from starlette.formparsers import FormParser, MultiPartParser
-from starlette.types import Message, Receive, Scope
+from starlette.types import Message, Receive, Scope, Send
 
 try:
     from multipart.multipart import parse_options_header
 except ImportError:  # pragma: nocover
     parse_options_header = None  # type: ignore
+
+
+SERVER_PUSH_HEADERS_TO_COPY = {
+    "accept",
+    "accept-encoding",
+    "accept-language",
+    "cache-control",
+    "user-agent",
+}
 
 
 class ClientDisconnect(Exception):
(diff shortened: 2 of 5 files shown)


jlaine:
I'll obviously be adding tests + docs, but I'd like to know if the API would be acceptable?

A simple homepage which uses server push to load a stylesheet would look like:

```python
@app.route("/")
async def homepage(request):
    """
    Simple homepage.
    """
    await request.send_push_promise("/static/style.css")
```

jlaine:
@pgjones I read your blog post https://pgjones.dev/blog/push-promise-2019/ which mentions copying over `Accept` headers and friends, where does Quart do this?

pgjones:
It does it [here](https://gitlab.com/pgjones/quart/blob/master/quart/wrappers/request.py#L276).

starlette/requests.py
```diff
@@ -121,11 +121,18 @@ def url_for(self, name: str, **path_params: typing.Any) -> str:
     raise RuntimeError("Receive channel has not been made available")
 
 
+async def empty_send(message: Message) -> None:
+    raise RuntimeError("Send channel has not been made available")
```
pgjones, line 134: @jlaine I'm not sure it should error - a failed server push only affects performance.

jlaine: I mimicked the `request.body()` test. What we are catching here is a programming error internal to `starlette` isn't it?

jlaine:
> It does it [here](https://gitlab.com/pgjones/quart/blob/master/quart/wrappers/request.py#L276).

Thank you Philip! I can certainly understand the rationale for `"accept", "accept-encoding", "accept-language"` and probably `"cache-control"`, but why `"user-agent"`?

pgjones:
I see it as making the push promise exist as if was made by the client that made the request. With this it seemed sensible to have the same user agent.

jlaine:
okidoki, @pgjones I used the exact same set of headers as quart, hope that's OK?

pgjones:
Yep, I think that is a good set.

docs/server-push.md
```diff
@@ -0,0 +1,34 @@
+
+Starlette includes support for HTTP/2 and HTTP/3 server push, making it
+possible to push resources to the client to speed up page load times.
```
lovelydinosaur, line 3: We probably want to note that `hypercorn` is currently the only ASGI server to support server push.

starlette/requests.py
```diff
@@ -121,11 +130,18 @@ def url_for(self, name: str, **path_params: typing.Any) -> str:
...
+
+
 class Request(HTTPConnection):
-    def __init__(self, scope: Scope, receive: Receive = empty_receive):
+    def __init__(
+        self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send
+    ):
```
lovelydinosaur, line 140: I guess we *might* want to think about what we'd need to do to move this up to the top `HTTPConnection` instead, since both `Request` and `WebSocket` now accept `scope`, `receive`, `send`.

We *could* consider having public `def send()` and `def receive()` methods on the HTTPConnection (eg. the WebSocket implementations that add in state transition checks would then call into `super()` for the raw sending.)

Any thoughts on that?

jlaine: My policy is usually "keep it private until you absolutely need to make it public"  :)

docs/server-push.md
jlaine: That's not *strictly* true, aioquic's demo server also supports server push.

Also if you're open to the idea I'd be happy to submit a PR to uvicorn

starlette/requests.py
lovelydinosaur: Yes - that's entirely fair here.

docs/server-push.md
lovelydinosaur: > Also if you're open to the idea I'd be happy to submit a PR to uvicorn

Absolutely, yup!

lovelydinosaur:
Released as 0.12.9 - Thanks!

jlaine:
Awesome thanks so much @tomchristie !

Cool: https://www.starlette.io/server-push/

lovelydinosaur:
Not something that we need to add, but worth mentioning here...

In a mature framework, one nice way to deal with this for serving HTML pages and associated media would be to have a `static_url` template tag/context function that can *automatically* deal with sending resources as push promises (based on some conditional around the if HTTP session  appears to have served the resource yet or not).

We might want to add something like that to Starlette one day.

PR #630, opened 2019-09-02
Add "UploadFile.copy(dest)" method

sscherfke:
This method allows users to conveniently and efficiently persist an uploaded file without reading all of its contents into memory.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -1,3 +1,4 @@
+import shutil
 import tempfile
 import typing
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


Kludex:
I don't see a reason for adding this method instead of just copying it yourself - the way you wish so.

In case you still think this is necessary, let's discuss over our Gitter server. :)

gnat:
Newbie perspective- agreed, but it's not obvious to handle UploadFile inside `run_in_threadpool`. May be more of a documentation issue here more than anything, since moving / copying files uploaded is a very common usecase.

Example, creator of FastAPI not using `run_in_threadpool`: https://github.com/tiangolo/fastapi/issues/362#issuecomment-584104025

PR #635, opened 2019-09-11
TestClient: provide "raw_path" in scope

blueyed:


tests/test_testclient.py
```diff
@@ -87,10 +87,44 @@ def test_testclient_asgi3():
...
+    @app.route("/.*")
+    def scope_route_fallback(request):
+        return JSONResponse({
+            "path": request.scope["path"],
+            "raw_path": str(request.scope["raw_path"]),
+        })
+
+    client = TestClient(app)
+
+    # _ASGIAdapter unquotes "path", so this goes to the fallback route.
```
blueyed, line 115: How do you feel about not using "unquote" for the provided path?
(i.e. there is not really a reason for the test client to behave like a "bad webserver" that would unquote already)

lovelydinosaur:
Going to pick #805 over this one, since I prefer the simpler test case.
Thanks as ever, @blueyed!

blueyed:
> Going to pick #805 over this one, since I prefer the simpler test case.

IIRC this is testing various corner cases here also, i.e. "more tests" - please consider at least checking if it works the same way with #805 then.

PR #636, opened 2019-09-11
tests: minor: use "test_" prefix

blueyed:
The test was run previously, but this is clearer.

### DIFF
--- a/tests/test_testclient.py
+++ b/tests/test_testclient.py
@@ -45,7 +45,7 @@ def test_use_testclient_in_endpoint():
     assert response.json() == {"mock": "example"}
 
 
-def testclient_as_contextmanager():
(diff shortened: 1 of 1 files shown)


tests/test_testclient.py
```diff
@@ -45,7 +45,7 @@ def test_use_testclient_in_endpoint():
     assert response.json() == {"mock": "example"}
 
 
-def testclient_as_contextmanager():
```
lovelydinosaur, line 48: Thanks! Let's go for `test_use_testclient_as_contextmanager`

blueyed: Amended.

lovelydinosaur:
Ta!

PR #640, opened 2019-09-17
Support for default json encoder when using JSONResponse

gnulnx:
This is something I have wanted in a framework for a while now.  Basically the ability to give a default encoder to the JSONResponse object so that we can set json.dumps(content, default=encoder).  

For example based on this PR we can do the following.

```
from starlette.encoders import utf8encoder
...

@app.route('/', methods=['GET'])
async def homepage(request):
    return JSONResponse({'request': request.__dict__}, json_encoder=utf8encoder)
```

Or alternatively if you are just developing and don't care about the encoding.

```
@app.route('/', methods=['GET'])
async def homepage(request):
    return JSONResponse({'request': request.__dict__}, json_encoder=str)
```

The way devs typically handle this now is to wrap json.dumps with json.loads and their favorite encoder.

```
request_dict = json.loads(json.dumps(content, default=utf8encoder))
return JSONResponse({'request': request_dict})
```

The addition of this PR would eliminate the unnecessary json.loads() step by allowing us to pass the encoder to the final json.dumps call inside JSONResponse.render().

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -38,7 +38,7 @@ class Response:
         headers: dict = None,
         media_type: str = None,
         background: BackgroundTask = None,
-        json_encoder: object = None
+        json_encoder: typing.Callable = None
     ) -> None:
         self.json_encoder = json_encoder
         self.body = self.render(content)
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -38,7 +38,9 @@ class Response:
         headers: dict = None,
         media_type: str = None,
         background: BackgroundTask = None,
+        json_encoder: object = None
     ) -> None:
+        self.json_encoder = json_encoder
         self.body = self.render(content)
         self.status_code = status_code
         if media_type is not None:
@@ -149,6 +151,7 @@ class JSONResponse(Response):
             allow_nan=False,
             indent=None,
             separators=(",", ":"),
+            default=self.json_encoder
         ).encode("utf-8")
 
 
--- a/starlette/encoders.py
+++ b/starlette/encoders.py
@@ -1,0 +1,3 @@
+def utf8encoder(val):
+    if isinstance(val, bytes):
+        return val.decode('utf-8')


gnulnx:
If this feature looks promising I'll code up some unit tests and write some docs for it...

starlette/responses.py
```diff
@@ -38,7 +38,9 @@ def __init__(
         headers: dict = None,
         media_type: str = None,
         background: BackgroundTask = None,
+        json_encoder: typing.Callable = None
     ) -> None:
+        self.json_encoder = json_encoder
```
dmontagu, line 43: It feels a little awkward to me to have the option to pass a `json_encoder` to *all* response class `__init__`s, since json encoding is irrelevant in many cases (unlike the other `__init__` arguments).

Is there a reason you wouldn't want to accomplish this pattern by subclassing JSONResponse to override the encoder? (Similar to how UJSONResponse currently works, except overriding the encoding rather than the dumps call)

The current JSONResponse could be modified to hold the encoder in a ClassVar -- that would make it a little easier to override. (Though even without that the modification necessary is pretty small.)

gnulnx: My rational for not subclassing is that would mean a subclass for each encoder type.  Since JSONResponse.render() is just a wrapper over json.dumps it feels slightly better to me to be able to simply pass in the encoder just like I would json.dumps().  


Let me rework this a bit and contain it to JSONResponse and see how it feels.

dmontagu: That makes sense to me, but a quick sanity check -- do you need to support lots of different encoder types? That seems surprising to me, but I might be overlooking something.

gnulnx: OK Here is another approach that confines this to JSONResponse

```
class JSONResponse(Response):
    encoder = None

        return json.dumps(
            content,
            ensure_ascii=False,
            allow_nan=False,
            indent=None,
            separators=(",", ":"),
            default=self.encoder
        ).encode("utf-8")
```

We now have to also add 'self' to our encoder method

```
def utf8encoder(self, val):
    if isinstance(val, bytes):
        return val.decode('utf-8')
```

And then we can use it like this...

```
JSONResponse.encoder = utf8encoder
return JSONResponse({'request': request.__dict__})
```

Things I like about this approach: 

- Functionality is there
- It's contained to JSONResponse

Things I don't like with this approach.

- The encoder has to have 'self' making it no longer a generic encoder that works with vanilla json.dumps.  Would be nice to have a library of encoders that work with both.
- The encoder could be set at module scope reducing the calling signature for each Response call.

I think the approach that will feel the most naturally as a developer using the framework will be to override the JSONResponse.__init__.  That way it could be fully contained in JSONResponse, but since Response.__init__ has all named parameters this means overriding the entire signature which doesn't seem like a great approach either... 

@dmontagu What's your opinion on the above?  I'd certainly rather have the first approach outlined than not have the ability to pass the encoder at all.  The second approach unless I'm missing something obvious doesn't feel solid either.

gnulnx: Note changes are only in comments for now.  If we settle on an approach I'll update PR with code changes.

gnulnx: > That makes sense to me, but a quick sanity check -- do you need to support lots of different encoder types? That seems surprising to me, but I might be overlooking something.

I mostly only use str or the utf8encoder method I submitted, but I like to keep it generic for others.  json.dumps allows us to specify any encoder so I figure why not extend that to JSONResponse?

dmontagu: @gnulnx I think there might be a problem here -- because you are using a class variable on JSONRespnose, when you do:
```python
JSONResponse.encoder = utf8encoder
return JSONResponse({'request': request.__dict__})
```
you are actually setting the encoder for *all* JSONResponse instances.

I am not totally sure but, even if you were careful to never use JSONResponse without setting the encoder (which is already an ugly restriction), I could imagine undesired behavior happening if you are changing the encoder type for JSONResponse inside a `run_in_threadpool` call -- it seems to me it's theoretically possible the encoder type could change between where you set it and where the actual dumping happens due to thread safety issues.

The proposed change would work fine though if the approach was to use subclasses of JSONResponse though.

dmontagu: I think if you want to have the ability to specify a custom encoder without subclassing, the only reasonable way to do this would be to override the `__init__` and add an `encoder` kwarg:

```python
class JSONResponse(Response):
    def __init__(
        self,
        content: typing.Any = None,
        status_code: int = 200,
        headers: dict = None,
        media_type: str = None,
        background: BackgroundTask = None,
        encoder: Any  # not sure what the type should be
    ):
        super().__init__(content, status_code, headers, media_type, background)
        self.encoder = encoder
```

But before you invest much more effort/thought into this, I think it would be good to get confirmation from @tomchristie that this is something starlette would be willing to support (as opposed to preferring the use of subclassing, and keeping the `JSONResponse.__init__` unmodified).

dmontagu: I didn't read your post carefully enough -- the adding of self to the encoder seems terrible, but I think there should be a workaround for it... Let me see

dmontagu: Using `default=type(self).encoder` drops the need for adding the self argument.

But it's still kind of gross, and either way this encoder-overriding approach will have minor performance downsides.

I personally think both the approach that overrides the `__init__` and the subclassing approach are cleaner/preferable.

gnulnx: @dmontagu  I agree that overriding the init method seems cleanest.  Whether that is Response or JSONResponse is I guess really up to @tomchristie.  My initial PR did it in the Response so that we didn't have to duplicate the init signature in both Response and JSONResponse.

Also you are correct about the class variable setting it for all JSONResponse objects.  This could be desirable in many cases.  For instance most projects probably do want to decode as utf8...so maybe you import JSONResponse and then immediately set the encoder to utf8encoder.  Now JSONResponse alway's utf8 encodes for the entire module.  

Personally though, I think it's better to pass it in each time and override init assuming Tom is down with adding this functionality.

Kludex:
I'll take the lead on this one and say that creating a subclass is preferred.

I don't see any reactions on this issue over those years, so I'll close the issue. We can revisit this in the future if someone wants.

Sorry for taking so long to act here.

Issue #644, opened 2019-09-20
GZipMiddleware (unzipping encoded requests)


RobinDavid:
Dear maintainers,

I am looking for the appropriate way to unzip clients requests which are gzip encoded. Looking at the ``GZipMiddleware`` it seems that it only zip responses but do not care about requests.

Thus I tried to do it at the dispatch level without success. I did something like:

```python
@app.middleware("http")
async def dispatch(request, call_next):
    plain_body = gzip.decompress(request.body())
    request._body = plain_body  # I don't think it is the appropriate way
    response = await call_next(request)  # hangs here
    return response
```
I am not extremely familiar with the inner-working of starlette so I don't know how to "push back in the queue" the decoded the request with the decoded body. If found no middlewares doing similar things.

Anyone can pinpoint me the appropriate way to do it ?

Thank you in advance for your help :wink:

dmontagu:
Your approach is close, but it won't work until you reach the endpoint call, because a new `Request` instance is created at each step in the middleware.

You could probably make it work via an endpoint decorator with a little elbow grease -- the request you'd receive at that point would work. (But that would mean decorators everywhere; subideal).

@tomchristie has in other issues pointed out downsides to consuming the request stream in middleware.

I think the *right* way to do this would actually be to override the `body` method on the Starlette request to look for an appropriate header, and un-gzip if appropriate. Currently this isn't especially convenient, but it is possible. You need to do the following:

1. Subclass `starlette.requests.Request`, and override the `body` method with the behavior you want.
2. Create a function that is very similar to `starlette.routing.request_response`, except that it uses your custom subclass of `Request`, rather than the starlette-builtin `Request`.
3. Create a subclass of `starlette.routing.Route`, and modify the `__init__` to set `self.app` using the custom `request_response` from step 2. This could be done via:
```python
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if inspect.isfunction(endpoint) or inspect.ismethod(endpoint):
            self.app = custom_request_response(endpoint)
```

(You could also just set the `request._body` inside your `custom_request_response` -- if you did that, you wouldn't need to subclass `Request`, but I suppose it is a private API...)

dmontagu:
I actually used a similar approach while helping with [this question about converting msgpack to json in middleware](https://github.com/tiangolo/fastapi/issues/521) (it's fastapi, but the same ideas are used).

[...]

PR #645, opened 2019-09-20
Make it easier to override route/request behaviors

dmontagu:
Right now, I don't think there is a good way to override `Request` behavior in starlette. This capability may be desirable for various reasons, such as decompressing requests (https://github.com/encode/starlette/issues/644).

This pull request would make it substantially easier to override `Request` methods if necessary for custom body parsing, etc.

(@tomchristie I would understand if you want to reject this pull request in the interest of keeping starlette simpler, but given how few lines this PR changes, I figured it would be easier to discuss with the concrete implementation as a reference.)

------

### Usage example (assuming it is merged)

The modifications to `Route` and `Router` are included to make it easier to use a custom `Request` subclass in all your routes.

As an example, say you have a custom request class `CustomRequest` that looks like the following:

```python
class CustomRequest(Request):
        await super().body()
        if self.headers.get("Content-Encoding") == "gzip":
            self._body = gzip.decompress(self._body)
        return self._body
```

Then you could create custom subclasses of `Route` and `Router` that will make use of this request type in as little code as:

```python
class CustomRoute(Route):
    request_type = CustomRequest

class CustomRouter(Router):
    route_type = CustomRoute 
```

and use them via:

```python
router = CustomRouter()

@router.route(...)
def endpoint(request):
    ...
```

Overriding the `Starlette` object itself to use a custom route type is just a minor change to the `__init__`:

```python
class CustomStarlette(Starlette):
    def __init__(
        self, debug: bool = False, routes: typing.List[BaseRoute] = None
    ) -> None:
        super().__init__(debug, routes)
        self.router = CustomRouter(routes)
```

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -28,7 +28,9 @@ class Match(Enum):
     FULL = 2
 
 
-def request_response(func: typing.Callable) -> ASGIApp:
+def request_response(
+    func: typing.Callable, request_type: typing.Type[Request] = Request
+) -> ASGIApp:
     """
     Takes a function or coroutine `func(request) -> response`,
     and returns an ASGI application.
@@ -36,7 +38,7 @@ def request_response(func: typing.Callable) -> ASGIApp:
     is_coroutine = asyncio.iscoroutinefunction(func)
 
     async def app(scope: Scope, receive: Receive, send: Send) -> None:
-        request = Request(scope, receive=receive, send=send)
+        request = request_type(scope, receive=receive, send=send)
         if is_coroutine:
             response = await func(request)
         else:
@@ -135,6 +137,8 @@ class BaseRoute:
 
 
 class Route(BaseRoute):
+    request_type: typing.ClassVar[typing.Type[Request]] = Request
+
     def __init__(
         self,
         path: str,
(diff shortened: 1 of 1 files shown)


prostomarkeloff:
LGTM. This replacement it's needed.

dmontagu:
I've thought about this a lot recently (there has been a lot of related activity in fastapi); and I'm actually now less convinced that this is a useful addition to starlette. In particular, I think the `Request` class as it stands doesn't really need extension, and other frameworks (e.g., fastapi) should take responsibility for implementing any other desired functionality (e.g., alternate approaches to consuming the body).

It seems to me more natural that, if you are using starlette directly, whatever logic you want to use in a subclass of `Request` could be applied in the endpoint function; the inability to override methods is mostly problematic when the `Request` is being handled by (non-starlette) framework calls that are expecting a `Request` satisfying specific assumptions. But I think dealing with that should probably be the responsibility of the (non-starlette) framework.

------

That said, I still think there is a conspicuous gap in starlette's ability to add middleware-like functionality that accesses the request body. Obviously this may be problematic in some cases (and the use of request-body-accessing middleware has accordingly been discouraged in other issues), but there appear to be a relatively long list of use cases where it seems more-or-less okay (mostly related to logging/exception-handling of non-streaming json requests).

See the list of linked issues on this PR https://github.com/tiangolo/fastapi/pull/589#issue-324506826 for more detail.

lovelydinosaur:
> It seems to me more natural that, if you are using starlette directly, whatever logic you want to use in a subclass of Request could be applied in the endpoint function

Yeah - that sounds right to me, too.

> That said, I still think there is a conspicuous gap in starlette's ability to add middleware-like functionality that accesses the request body.

Yeah that's a fair call - don't recall off the top of my head if we've got an open issue for that, but it's been cropping up a few times, so def worth treating as a priority.

Issue #651, opened 2019-10-01
Poor performance for large (~100MB) payloads


dmontagu:
I'm noticing especially slow handling of large request bodies when running in uvicorn, and I'm trying to get to the bottom of it.

**If this kind of performance is expected for for payloads of this size for any reason, please let me know.**

------

The script below posts a large payload (~100M in size) to a `starlette.applications.Starlette` endpoint, which just returns a success response. Running via the **starlette `TestClient`**, I get a response in **~0.65 seconds**; running via **uvicorn** it takes **~17.5 seconds (or ~27x slower)**.

(I'll note that this discrepancy becomes much smaller as the size of the payload decreases -- I think it was about 3-5x for 10MB, and not really significant below 1MB.)

I was able to get speeds comparable to the `TestClient` runs speed using a flask implementation. I also get similar slowdowns when running via gunicorn with a uvicorn worker (I haven't tested other servers; not sure if there are recommended alternatives).

<details>
    <summary><b>Click here to expand script</b></summary>

```python
import sys
from datetime import datetime

import requests
import uvicorn
from requests import Session
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import Response
from starlette.testclient import TestClient

app = Starlette()


@app.route("/", methods=["POST"])
async def endpoint(request: Request):
    payload = await request.json()
    assert isinstance(payload, dict)
    return Response("success")


def _speed_test(session: Session, url: str):
    payload = {"payload": "a" * 100_000_000}
    start = datetime.utcnow()
    response = session.post(url=url, json=payload)
    elapsed = datetime.utcnow() - start
    assert response.status_code == 200
    assert response.content == b"success"
    print(elapsed)


def asgi_test():
    client = TestClient(app)
    _speed_test(client, "/")


def uvicorn_test():
    session = requests.Session()
    _speed_test(session, f"http://127.0.0.1:8000/")


def main():
    if "--asgi-test" in sys.argv:
        asgi_test()
        # 0:00:00.650825
    elif "--uvicorn-test" in sys.argv:
        uvicorn_test()
        # 0:00:17.502396
        # cProfile:
        # Name   Call Count   Time (ms)  Own Time (ms)
        # body      391         16670        16649
    else:
        uvicorn.run(app)


if __name__ == "__main__":
    main()
```
</details>

------

This script can perform three actions:
* Start the uvicorn server if run without arguments
[...]

PR #653, opened 2019-10-02
Performantly build large request bodies

dmontagu:
This PR replaces the quadratically-scaling `body += chunk` with the linearly-scaling `body = b"".join(chunks)`.

This leads to a speed up of `20x` in the script from #651 (which involves a 100MB payload). (The time currently scales quadratically with the payload size, so this large factor makes sense.)

Even for payloads as small as 5MB, in my testing, this change causes the server to handle the request ~15-20% faster.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -161,7 +161,7 @@ class URLPath(str):
     Used by the routing to return `url_path_for` matches.
     """
 
-    def __new__(cls, path: str, protocol: str = "", host: str = "") -> str:
+    def __new__(cls, path: str, protocol: str = "", host: str = "") -> "URLPath":
         assert protocol in ("http", "websocket", "")
         return str.__new__(cls, path)  # type: ignore
 
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -382,7 +382,7 @@ class TestClient(requests.Session):
         self.app = asgi_app
         self.base_url = base_url
 
-    def request(
+    def request(  # type: ignore
         self,
         method: str,
         url: str,
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -178,10 +178,10 @@ class Request(HTTPConnection):
 
     async def body(self) -> bytes:
         if not hasattr(self, "_body"):
-            body = b""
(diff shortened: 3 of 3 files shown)


starlette/datastructures.py
```diff
@@ -161,7 +161,7 @@ class URLPath(str):
     Used by the routing to return `url_path_for` matches.
     """
 
-    def __new__(cls, path: str, protocol: str = "", host: str = "") -> str:
+    def __new__(cls, path: str, protocol: str = "", host: str = "") -> "URLPath":
```
dmontagu, line 164: Mypy 0.730 wants the return type of `__new__` to be the type of the class.

starlette/testclient.py
```diff
@@ -382,7 +382,7 @@ def __init__(
         self.app = asgi_app
         self.base_url = base_url
 
-    def request(
+    def request(  # type: ignore
```
dmontagu, line 385: Mypy sees the `url` parameter of the superclass method `Session.request` as a `Union[str, bytes]`, so it is complaining about the signature here. But the `urljoin` call below will fail if it `self.base_url` is a `str` and `url` is a `bytes`, so we probably shouldn't annotate it as `Union[str, bytes]`.

For what it's worth, mypy does let you use error codes now, so we could now make the `# type: ignore` more specific if desired (it would become `# type: ignore[override]`). I have no preference here.

dmontagu:
@tomchristie I can split this into a separate pull request for the two minor mypy annotation updates if desired.

lovelydinosaur:
I'm okay with this yeah, looks good!

At another point we might want to consider hard-pinning mypy - it's not ideal to have unrelated changes in the tooling impact on our PRs.

rpkilby:
For anyone looking for an explanation as to what's happening: https://stackoverflow.com/q/44487537/1103124

gvbgduh:
@rpkilby the main reason for that is that string is the immutable type in python. That means it can't be changed, so for the simple concatenation, it has to allocate new memory, copy the original string and the additional one than create a new reference to the object. It's quite a bit of overhead.

PR #656, opened 2019-10-03
Allow multiple headers of the same key

### LINKED ISSUE #650: Multiple headers of the same name
HTTP spec allows multiple headers of the same key to be present. This is why libraries such as h11 accept a list of headers instead of dictionaries. 

[HTTP 1.1, section 4.2](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

Currently starlette dosen't support this because it is just running `value.encode()` on the value in the header. 

I propose that the header dictionary be able to contain lists as values as well, in which case starlette will add a header for each value. 
This is most common when using the `set-cookie` header, when a backend wants to add multiple cookies, but not have the one header exceed too big of an individual header length. 

for example:
```
headers = {'mykey': ['value1', 'value2']}
```

would produce the following headers:

```
mykey: value1
mykey: value2
```

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/650">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/650/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

nhumrich:
Resolves #650

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -53,16 +53,24 @@ class Response:
             return content
         return content.encode(self.charset)
 
-    def init_headers(self, headers: typing.Mapping[str, str] = None) -> None:
+    def init_headers(self,
+                     headers: typing.Mapping[
+                         str,
+                         typing.Union[str, typing.List[str]]] = None) -> None:
         if headers is None:
             raw_headers = []  # type: typing.List[typing.Tuple[bytes, bytes]]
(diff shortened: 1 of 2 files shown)


nhumrich:
Closing as issue was resolved in discussion.

PR #660, opened 2019-10-08
Add 425 Too Early status code

HarrySky:
I was researching TLS1.3 and found out that client can send `Early-Data` header if request is made before TLS handshake finishes (https://tools.ietf.org/html/rfc8470). Early data request is subject to replay attack.

Starlette is not aware if request is early data (it can be passed from nginx).

I thought of making a first small step - adding `425 Too Early` status code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/425) so this early data can be rejected.

I would like to also add `is_early_data` to request (or something like that) so that request can be more easier denied if it is early data.


HarrySky:
**Idea**: maybe Starlette should deny `Early-Data` requests by default (with `425 Too Early`) and make user explicitly set some flag if they are handled in code.

HarrySky:
It should not matter behind what Starlette is running, early data responses, if they are not rejected outright (in nginx this is default - `ssl_early_data off;`), should be processed by application layer and rejected with `425 Too Early` if replay of this request can pose a security risk.

Case: I want to use speed improvement provided by 0-RTT Data (early data requests), but I have some endpoint that, if replayed, may do some bad things to user's account and I want to reject it if it has `Early-Data="1"` header and get this request after handshake is finished.

HarrySky:
Thanks, @tomchristie :)

I have written middleware for this - [starlette-early-data](https://github.com/HarrySky/starlette-early-data). Can I add it to Third party middleware section in documentation?

Issue #663, opened 2019-10-10
Wrong custom middleware documentation


HarrySky:
Writing middleware for #660 (Too Early) and stumbled upon an error when was following Middleware documentation:

`AttributeError: 'EarlyDataMiddleware' object has no attribute 'dispatch_func'`

I believe that `__init__` method in documentation:

```
class CustomHeaderMiddleware(BaseHTTPMiddleware):
    def __init__(self, app, header_value='Example'):
        self.app = app
        self.header_value = header_value
...
```

Should be used with super:

```
class CustomHeaderMiddleware(BaseHTTPMiddleware):
    def __init__(self, app, header_value='Example'):
        super().__init__(app)
        self.header_value = header_value
...
```

I can open a Pull Request that fixes documentation.

lovelydinosaur:
Yup, your assessment looks correct to me, too.

HarrySky:
Closed via #665

PR #665, opened 2019-10-11
Fix __init__ code in custom HTTP middleware example

HarrySky:
Use `super` in `__init__` to prevent #663 issue when following custom HTTP middleware documentation.


lovelydinosaur:
Thanks!

PR #669, opened 2019-10-11
add support for query parameters in `Route.url_path_for`

### LINKED ISSUE #560: add support to request.url_for for query parameters
I totally misunderstood `url_for`'s API, I was expecting `url_for('some_route', foo='bar')` to return `'http://testserver/my_route/?foo=bar'`.

I understand why that is now, but what is the best way to generate urls like this? 
(for pagination in a rest list api in this case).

Could `url_for` grow support for query parameters in addition to path parameters?

ricardogsilva:
This PR add support for query parameters when resolving a URL with `request.url_for` and friends

From the added tests:

```
assert (
    app.url_path_for("homepage", some_query_param="this") == "/?some_query_param=this")

assert (
    app.url_path_for("homepage", **{"funky[query_param]": "this"}) =="/?funky%5Bquery_param%5D=this")

```

I find this useful in the context of getting a reproducible URL when generating pagination links for my API.

fixes #560

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -126,8 +126,8 @@ class Starlette:
 
         return decorator
 
-    def url_path_for(self, name: str, **path_params: str) -> URLPath:
-        return self.router.url_path_for(name, **path_params)
+    def url_path_for(self, name: str, **path_and_query_params: str) -> URLPath:
+        return self.router.url_path_for(name, **path_and_query_params)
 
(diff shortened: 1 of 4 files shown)


ricardogsilva:
I'm retreating this PR since there is further activity on the issue page (#560) that suggests the starlette maintainers plan to implement this functionality as part of a deeper refactor.

PR #672, opened 2019-10-13
Enable flag parameters

### LINKED ISSUE #670: Feature Request - Empty query string flags
HTTP URIs are allowed to have query strings without any values in them, like `/some-url?flag`. I'll call these 'flags'.

Starlette doesn't appear to provide a way to distinguish between these and not passing the flag at all, so `/some-url?flag` looks identical to `/some-url`.

I was redirected here after submitting this issue to FastAPI: https://github.com/tiangolo/fastapi/issues/610

I can work around it by making my clients pass `/some-url?flag=true` instead, so if it's a deliberate design decision not to support this then I understand.

jodusan:
Closes https://github.com/encode/starlette/issues/670

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -399,7 +399,9 @@ class QueryParams(ImmutableMultiDict):
         if isinstance(value, str):
             super().__init__(parse_qsl(value, keep_blank_values=True), **kwargs)
         elif isinstance(value, bytes):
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
This looks good to me, yup.

Could you:

* Update the test cases, so that we're validating this functionality.
* Run `scripts/lint` to fix the broken code formatting thats preventing the test suite from passing.

jodusan:
@tomchristie  updated!

lovelydinosaur:
Great, thanks!

Issue #284, opened 2018-12-18
Session backends


lovelydinosaur:
Currently sessions always use signed cookies.

Really we'd also like to have cache backend and database backend support, so that we're providing a a common interface to a range of available backends.

lovelydinosaur:
Closing as stale.

PR #676, opened 2019-10-14
Session Backend

akhilputhiry:
All work is done by @alex-oleshkevich in https://github.com/encode/starlette/pull/499
I am just trying to get this merged, since I need this feature https://github.com/encode/starlette/issues/284

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -73,7 +73,7 @@ request through as normal, but will include appropriate CORS headers on the resp
 
 ## SessionMiddleware
 
-Adds signed cookie-based HTTP sessions. Session information is readable but not modifiable.
+Adds sessions supports using various session backends. Session information is readable but not modifiable.
 
 Access or modify the session data using the `request.session` dictionary interface.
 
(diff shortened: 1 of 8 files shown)


blueyed:
@akhilputhiry 
Not much I can do - I have several PRs idling myself.. :/

blueyed:
@akhilputhiry 
JFI: I am using https://github.com/blueyed/starlette/tree/next myself with those being merged there.

theruziev:
@akhilputhiry What about expire session. If I authorise an user and it go away without logout, how remove session in redis or in other backends? May be we need add expiring method in abstract class and realize in specific backends

akhilputhiry:
@theruziev There is already a `remove` method available. Those who want to auto expire can make use of this method. Probably they can run a background job which compares the current time vs time in session (timestamp can be stored as part of session)

theruziev:
@akhilputhiry good point about background job. Or we can set TTL if backend supports it. Redis and Memcached support TTL

gnat:
Just wanted to say very nice PR. Useful for clusters of machines sharing user sessions. I feel like many of us have implemented this independently for our projects.

Kludex:
As you said, duplicated of #499 , so closing this. I'll see if we can take a decision on that PR soon.

Sorry for taking so long.

EDIT: The duplication is on #284.

Issue #604, opened 2019-08-18
Starlette behind reverse proxy, url_for does not return externally valid urls


narzeja:
Suppose I have my application reverse proxied, such that http://my.domain/foo (e.g. Traefik or Nginx) has the upstream http://localhost:8000 (the Starlette app). `url_for` will generate urls relative to the upstream url. How do I make it produce urls relative to the external url (with prefix /foo)?

For other projects, e.g. Flask/Werkzeug, this can be handled by setting `SCRIPT_NAME` in the request environment (scope) if the 'X-Forwarded-Prefix' header is set. It does not look like there is an equivalent option in Starlette.

Am I missing something? Otherwise, consider this a feature request.

narzeja:
Forgive my haste, this is a possible duplicate of or at least related to https://github.com/encode/starlette/issues/592

lovelydinosaur:
See uvicorn's `--root-path` and `--proxy-headers` in order to configure mounted path, and host/scheme configuration.

narzeja:
I am aware of those configuration settings, the scheme and host works fine, but the resulting URL is still not correct.

Example:

```python
from starlette.applications import Starlette
from starlette.responses import PlainTextResponse

async def index(request):
    return PlainTextResponse(request.url_for('index'))

app = Starlette()
app.add_route('/', index, methods=['GET'])
```

I use Traefik to load-balance and use the following rules to direct the incoming connections to the Starlette application: `traefik.frontend.rule=Host:localhost; PathPrefixStrip:/foo/;`.
I use `--root-path /foo --proxy-headers` in uvicorn.

Accessing `http://localhost/foo/` gives the following response: `http://localhost/`, while the expected response should be `http://localhost/foo/`.

sm-Fifteen:
The Uvicorn documentation says that `--proxy-headers` only adds the `X-Forwarded-Proto` (the protocol) and `X-Forwarded-For` (the client IP) headers to populate remote scheme/address info.

Looking at the [proxy-headers middleware, it indeed only checks those two headers](https://github.com/encode/uvicorn/blob/b8a3ab98f7f4d01265f2f91d32e5f8386a8d1003/uvicorn/middleware/proxy_headers.py). `X-Forwarded-Host` (the original hostname, possibly the port as well?) isn't checked, so `scope['server']` does not get updated to match the address the client is expected to try and access on subsequent requests (though `url_for` seems to use something else; changing it does not affect the resulting URL as it probably should).

[...]

PR #681, opened 2019-10-20
[WIP] Account for root_path and forwarded host in make_url_absolute

sm-Fifteen:
See #604. I'm currently trying to write test cases to target the specific issue related to this bug, but I have never contributed to Starlette before and am not entirely sure how to go about solving this in a way that works with the intended design.

The issue is twofold:
* Starlette does not account for its own `root_path` (which is part of the ASGI spec) when making paths into absolute URLs
* Starlette does not account for proxied hostname when making paths into absolute URLs

The former is clear-cut and easily fixed, but I'm not sure how the latter should be handled. The [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and [`Forwarded`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) headers are meant to tell the application that the hostname given in `Host` is not what the client used to access the application, and I don't know if those should be decoded by the URL constructor as is currently the case for `Host` or decoded in advance and written into the scope instead, such as by overwriting `scope['server']`.

### DIFF
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -280,3 +280,45 @@ def test_app_add_event_handler():
         assert not cleanup_complete
     assert startup_complete
     assert cleanup_complete
+
+def test_url_for_forwarded_host():
+    app = Starlette()
+
+    @app.route("/echo")
+    async def echo(request):
+        return PlainTextResponse(request.url_for('echo'))
+    
+    client = TestClient(app)
+    response = client.get("/echo")
+    assert response.text == 'http://testserver/echo'
+
+    response = client.get("/echo", headers={'x-forwarded-host':'starlette.example:8080'})
+    assert response.text == 'http://starlette.example:8080/echo'
+
+class SetRootPathMiddleware:
+    '''
+    A middleware that mimics the effect of `uvicorn --root-path="/prefix"`
+    '''
+    def __init__(self, app, root_path: str) -> None:
+        self.app = app
+        self.root_path = root_path
+
+    async def __call__(self, scope, receive, send) -> None:
+        scope['root_path'] = self.root_path
+        await self.app(scope, receive, send)
+
+def test_url_for_root_path():
+    app = Starlette()
+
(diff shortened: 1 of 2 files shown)


sm-Fifteen:
The ASGI specification says that `scope['server']` is supposed to be the listening address and port for the server, so it probably wouldn't be appropriate to change it. [RFC 7230 §5.4](https://tools.ietf.org/html/rfc7230#section-5.4) also says that `Host` is supposed to change with each proxy that retransmits the information, so that header cannot be relied upon as the source of the hostname for URL generation when operating behind a reverse proxy. [RFC 7239](https://tools.ietf.org/html/rfc7239), which defines the `Forwarded` header mentions the `Via` header as an alternative to `Forwarded` and `X-Forwarded-Host` when the server needs to be aware of proxies between it and the client, so that's 3 potential headers from which that information would have to be extracted if they happen to be present.

The client-facing hostname part could probably be omitted if `url_for` and `make_url_absolute` could be made to not include the hostname, port and protocol information in their result (assuming the root path is correct) since the client is rarely in a position where those fields aren't known and can't be used in subsequent requests (the primary use-case for this is HTML templates).

**EDIT:** The solution most people seem to use with other frameworks and libraries (be it Flask, Rails, h11, PHP, etc.: None of them support any of those headers) seems to be to either require that any upstream proxy servers perserve the value of `Host` as sent by the client ([including uvicorn](https://www.uvicorn.org/deployment/#running-behind-nginx), which I failed to notice until now); to have some kind of middleware override that header with the aforementionned proxy headers; or simply to use static environment variables to keep track of the external hostname.

sm-Fifteen:
Closing, since I now realize that there is no standard way of dealing with figuring out what netloc to use for generated URLs from request data alone, and dealing with `root_path` has its own share of complications given how `URL` currently works (see #690).

PR #683, opened 2019-10-22
Dropping double square brackets in install script

espositofulvio:
The install script works with `sh` but contains double square brackets tests which aren't implemented in `sh` and result in an error. This pull request changes the bracket tests to single bracket.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,12 +1,12 @@
 #!/bin/sh -e
 
 # Use the Python executable provided from the `-p` option, or a default.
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks for this!
We're going to have to hold for a moment - looks like we ought to be pinning our linting tools, since mypy's decided that something that wasn't perviously an error is now an error or something along those lines.

espositofulvio:
Closing this one. I had issues with my local git repo, in the end it was easier delete it and doing it again so I made a second pull request.

PR #704, opened 2019-11-05
Version 0.13

lovelydinosaur:
Documentation still in progress.

This PR does not significantly remove any public API, but it does start to move us to a new prefered style of routing, moving away from the decorator style.

I think using plain route lists is more clearly decoupled, and gives a clearer indication of what's going on under the hood.

For example...

```python
routes = [
    Route('/', homepage),
    Mount('/users', routes=[
        Route('/{username}', user),
    ])
]

```

Similarly, middleware now has a prefered configure-on-init style...

```python
routes = ...

middleware = [
    Middleware(SentryAsgiMiddleware) if settings.SENTRY_DSN else None,
    Middleware(HTTPSRedirectMiddleware) if settings.HTTPS_ONLY else None,
    Middleware(SessionMiddleware, secret_key=settings.SECRET, https_only=settings.HTTPS_ONLY)
]

```

Still to do here...

* [ ] ~Review `exception_handlers` / `server_error`~. - We should defer this to a future release.
* [ ] Update `starlette-example`.
* [x] Update Routing docs.
* [x] Update Middleware docs.
* [x] Update routing examples in docs throughout.

We *may* also want to reivew our `url_for` and `url_path_for` documentation and API as part of this release version. I'm somewhat keen for us to switch things around there, so that path-only URLs are the default style, and there's less surface area for folks to need to care about configuring Uvicorn's `--allow-forward-ips`. For example if we were using this:

* `app.url_for()`, `request.url_for()`, `url_for()` in templates. Return a relative path-only URL.
* `request.absolute_url_for()`, `absolute_url_for()` in templates. Return a fully qualified URL.

I'm also pretty keen to switch things up here so that when `debug=True` we...

* Display helpful routing table info if no route matches during routing.
* Display helpful routing table info when `url_for(...)` fails.

Something else that I think is interesting here is that we're more clearly exposing that alternative subclasses of `BaseRoute` can also be added to the routing table. For example, it'd be interesting to consider if there's any worthwhile documentation work to do in the ecosystem around eg. noting how you can use FastAPI's `APIRoute` inside a regular Starlette application, to point at a handler which gets their dependency injection treatments, alongside other regular `Route` cases, which just get the standard `request` argument.

That line of thought is as much about demystifying how the different bits fit together, about making sure that we can display useful routing debug info for higher level frameworks as well as the plain starlette case, and also about enabling other framework authors to *take the happy path* when building on top of Starlette. (We kinda missed a trick with `Responder` here, which functions differently enough that it's working against the grain.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,20 +61,28 @@ $ pip3 install uvicorn
 
 ## Example
 
+**example.py**:
+
 ```python
 from starlette.applications import Starlette
 from starlette.responses import JSONResponse
-import uvicorn
-
-app = Starlette(debug=True)
+from starlette.routing import Route
 
 
-@app.route('/')
 async def homepage(request):
     return JSONResponse({'hello': 'world'})
 
-if __name__ == '__main__':
-    uvicorn.run(app, host='0.0.0.0', port=8000)
+routes = [
+    Route("/", endpoint=homepage)
+]
+
+app = Starlette(debug=True, routes=route)
+```
+
+Then run the application using Uvicorn:
+
+```shell
+$ uvicorn example:app
 ```
 
 For a more complete example, see [encode/starlette-example](https://github.com/encode/starlette-example).
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -5,77 +5,45 @@ its other functionality.
 ```python
 from starlette.applications import Starlette
 from starlette.responses import PlainTextResponse
+from starlette.routing import Route, Mount, WebSocketRoute
 from starlette.staticfiles import StaticFiles
 
 
-app = Starlette()
-app.debug = True
-app.mount('/static', StaticFiles(directory="static"))
-
-
-@app.route('/')
 def homepage(request):
     return PlainTextResponse('Hello, world!')
 
-@app.route('/user/me')
 def user_me(request):
     username = "John Doe"
     return PlainTextResponse('Hello, %s!' % username)
 
-@app.route('/user/{username}')
 def user(request):
     username = request.path_params['username']
     return PlainTextResponse('Hello, %s!' % username)
 
-
-@app.websocket_route('/ws')
 async def websocket_endpoint(websocket):
     await websocket.accept()
     await websocket.send_text('Hello, websocket!')
     await websocket.close()
 
-
-@app.on_event('startup')
 def startup():
     print('Ready to go')
+
+
+routes = [
+    Route('/', homepage),
+    Route('/user/me', user_me),
+    Route('/user/{username}', user),
+    WebSocketRoute('/ws', websocket_endpoint),
+    Mount('/static', StaticFiles(directory="static")),
+]
+
+app = Starlette(debug=True, routes=routes, on_startup=[startup])
 ```
 
 ### Instantiating the application
 
-* `Starlette(debug=False)` - Create a new Starlette application.
-
-### Adding routes to the application
-
-You can use any of the following to add handled routes to the application:
-
-* `app.add_route(path, func, methods=["GET"])` - Add an HTTP route. The function may be either a coroutine or a regular function, with a signature like `func(request, **kwargs) -> response`.
-* `app.add_websocket_route(path, func)` - Add a websocket session route. The function must be a coroutine, with a signature like `func(session, **kwargs)`.
-* `@app.route(path)` - Add an HTTP route, decorator style.
-* `@app.websocket_route(path)` - Add a WebSocket route, decorator style.
-
-### Adding event handlers to the application
-
-There are two ways to add event handlers:
-
-* `@app.on_event(event_type)` - Add an event, decorator style
-* `app.add_event_handler(event_type, func)` - Add an event through a function call.
-
-`event_type` must be specified as either `'startup'` or `'shutdown'`.
-
-### Submounting other applications
-
-Submounting applications is a powerful way to include reusable ASGI applications.
-
-* `app.mount(prefix, app)` - Include an ASGI app, mounted under the given path prefix
-
-### Customizing exception handling
-
-You can use either of the following to catch and handle particular types of
-exceptions that occur within the application:
-
-* `app.add_exception_handler(exc_class_or_status_code, handler)` - Add an error handler. The handler function may be either a coroutine or a regular function, with a signature like `func(request, exc) -> response`.
-* `@app.exception_handler(exc_class_or_status_code)` - Add an error handler, decorator style.
-* `app.debug` - Enable or disable error tracebacks in the browser.
+::: starlette.applications.Starlette
+    :docstring:
 
 ### Storing state on the app instance
 
@@ -88,6 +56,6 @@ For example:
 app.state.ADMIN_EMAIL = 'admin@example.org'
 ```
 
-### Acessing the app instance
+### Accessing the app instance
 
-Where a `request` is available (i.e. endpoints and middleware), the app is available on `request.app`.  For other situations it can be imported from wherever it's instantiated.
+Where a `request` is available (i.e. endpoints and middleware), the app is available on `request.app`.
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -9,8 +9,10 @@ from starlette.authentication import (
     AuthenticationBackend, AuthenticationError, SimpleUser, UnauthenticatedUser,
     AuthCredentials
 )
+from starlette.middleware import Middleware
 from starlette.middleware.authentication import AuthenticationMiddleware
 from starlette.responses import PlainTextResponse
+from starlette.routing import Route
 import base64
 import binascii
 
@@ -30,21 +32,24 @@ class BasicAuthBackend(AuthenticationBackend):
             raise AuthenticationError('Invalid basic auth credentials')
 
         username, _, password = decoded.partition(":")
-        # TODO: You'd want to verify the username and password here,
-        #       possibly by installing `DatabaseMiddleware`
-        #       and retrieving user information from `request.database`.
+        # TODO: You'd want to verify the username and password here.
         return AuthCredentials(["authenticated"]), SimpleUser(username)
 
 
-app = Starlette()
-app.add_middleware(AuthenticationMiddleware, backend=BasicAuthBackend())
-
-
-@app.route('/')
 async def homepage(request):
     if request.user.is_authenticated:
-        return PlainTextResponse('hello, ' + request.user.display_name)
-    return PlainTextResponse('hello, you')
+        return PlainTextResponse('Hello, ' + request.user.display_name)
+    return PlainTextResponse('Hello, you')
+
+routes = [
+    Route("/", endpoint=homepage)
+]
+
+middleware = [
+    Middleware(AuthenticationMiddleware, backend=BasicAuthBackend())
+]
+
+app = Starlette(routes=routes, middleware=middleware)
 ```
 
 ## Users
@@ -81,7 +86,6 @@ incoming request includes the required authentication scopes.
 from starlette.authentication import requires
 
 
-@app.route('/dashboard')
 @requires('authenticated')
 async def dashboard(request):
     ...
@@ -93,7 +97,6 @@ You can include either one or multiple required scopes:
 from starlette.authentication import requires
 
 
-@app.route('/dashboard')
 @requires(['authenticated', 'admin'])
 async def dashboard(request):
     ...
@@ -107,7 +110,6 @@ about the URL layout from unauthenticated users.
 from starlette.authentication import requires
 
 
-@app.route('/dashboard')
 @requires(['authenticated', 'admin'], status_code=404)
 async def dashboard(request):
     ...
@@ -120,12 +122,10 @@ page.
 from starlette.authentication import requires
 
 
-@app.route('/homepage')
 async def homepage(request):
     ...
 
 
-@app.route('/dashboard')
 @requires('authenticated', redirect='homepage')
 async def dashboard(request):
     ...
@@ -135,7 +135,6 @@ For class-based endpoints, you should wrap the decorator
 around a method on the class.
 
 ```python
-@app.route("/dashboard")
 class Dashboard(HTTPEndpoint):
     @requires("authenticated")
     async def get(self, request):
--- a/docs/background.md
+++ b/docs/background.md
@@ -13,11 +13,12 @@ Signature: `BackgroundTask(func, *args, **kwargs)`
 ```python
 from starlette.applications import Starlette
 from starlette.responses import JSONResponse
+from starlette.routing import Route
 from starlette.background import BackgroundTask
 
-app = Starlette()
 
-@app.route('/user/signup', methods=['POST'])
+...
+
 async def signup(request):
     data = await request.json()
     username = data['username']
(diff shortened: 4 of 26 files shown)


florimondmanca:
Interesting take! I think it gets Starlette closer to a declarative style, which for end users is almost always preferable.

So as I understand, the current imperative API (`.add_route()` etc) will stay in place? I think that staying public and a first-class citizen (though not recommended for "Starlette-native" apps) is important for integrators and people building on top of Starlette.

Just a small note — in most cases Black is probably going to reformat these big lists a bit differently, e.g.:

```python
routes = [
    Route("/", homepage),
    Mount(
        "/users",
        routes=[
            Route("/", users, methods=["GET", "POST"]),
            Route("/{username}", user),
        ],
    ),
]
```

I think it looks OK, but just wanted to get this out there. When I first saw the Channels documentation with those huge lists made out of strangely-named objects, it didn't look very welcoming to be as a beginner. Luckily, names of Starlette's components are fairly short and explicit, so that shouldn't be a problem here. People can still break things down, e.g.

```python
users_routes = [
    Route("/", users, methods=["GET", "POST"]),
    Route("/{username}", user),
]

routes = [
    Route("/", homepage),
    Mount("/users", users_routes),
]
```

In larger projects, they'll most likely import these routes from a different module anyway.

lovelydinosaur:
> Can you confirm the current imperative API (.add_route() etc) will stay in place?

Yes, that's not going away here. However I *do think* that I'm going to stop documenting it everywhere.

Would I like to *eventually* phase it out? *Possibly at some point in the future, maybe*. One of the key things I'm trying to emphasize in Starlette is a low-complexity stack, and components such as the `Starlette` and `Router` classes get *really* simple if we move away from all the `.add_route(...)` and `@route()` futzing.

> Just a small note — in most cases Black is probably going to reformat these big lists a bit differently

Indeed, yes. I'm actually not super wild about that.

The `hostedapi` project routing looks like this at the moment...

```python

routes = [
    Route("/", endpoints.dashboard, name="dashboard", methods=["GET", "POST"]),
    Route("/tables/{table_id}", endpoints.table, methods=["GET", "POST"], name="table"),
    Route(
        "/tables/{table_id}/columns",
        endpoints.columns,
        methods=["GET", "POST"],
        name="columns",
    ),
    Route(
        "/tables/{table_id}/delete",
        endpoints.delete_table,
        methods=["POST"],
        name="delete-table",
    ),
    Route(
        "/tables/{table_id}/upload", endpoints.upload, methods=["POST"], name="upload"
    ),
    Route(
        "/tables/{table_id}/columns/{column_id}/delete",
        endpoints.delete_column,
        methods=["POST"],
        name="delete-column",
    ),
    Route(
        "/tables/{table_id}/{row_uuid}",
        endpoints.detail,
        methods=["GET", "POST"],
        name="detail",
    ),
    Route(
        "/tables/{table_id}/{row_uuid}/delete",
        endpoints.delete_row,
        methods=["POST"],
        name="delete-row",
    ),
    Route("/500", endpoints.error),
    Mount("/static", statics, name="static"),
]
```

I *guess* we could suggest that folks use `# fmt: off`, for a single route-per-line style...

```python
# fmt: off
routes = [
    Route("/", endpoints.dashboard, name="dashboard", methods=["GET", "POST"]),
    Route("/tables/{table_id}", endpoints.table, name="table", methods=["GET", "POST"]),
    Route("/tables/{table_id}/columns", endpoints.columns, name="columns", methods=["GET", "POST"]),
    Route("/tables/{table_id}/delete", endpoints.delete_table, name="delete-table", methods=["POST"]),
    Route("/tables/{table_id}/upload", endpoints.upload, name="upload", methods=["POST"]),
    Route("/tables/{table_id}/columns/{column_id}/delete", endpoints.delete_column, name="delete-column", methods=["POST"]),
    Route("/tables/{table_id}/{row_uuid}", endpoints.detail, name="detail", methods=["GET", "POST"]),
    Route("/tables/{table_id}/{row_uuid}/delete", endpoints.delete_row, name="delete-row", methods=["POST"]),
    Route("/500", endpoints.error),
    Mount("/static", statics, name="static"),
]
# fmt: on
```

That's still a bit heavy, so I'm open to conversations there.

I do think this is a *bit* of an awkward point because I think the style expresses the internals as clearly as possible, but it's not neccessarily how I'd want it to *display*.

Also worth noting that it's much nicer for allowing us to do expressive, complex stuff such as...

```python
routes = [
    Host("api.example.com", name="api", routes=[
        Route("/users/", user_list, name="user-list"),
        Route("/users/{username}", user_detail, name="user-detail"),
    ]),
    Host("www.example.com", routes=[
        Route("/", homepage, name="homepage"),
        Mount("/auth", routes=auth.routes, name="auth"),
        Mount("/statics", app=statics, name="static"),
    ])
]
```

florimondmanca:
Yes, the big list style has value for sure. When I think about it, in the world of frontend frameworks most routing is defined this way, *declaratively*, using either an array or an object (at least, Angular and Vue Router work like this). We haven’t been particularly used to it in the Python backend side, apart from a lighter form of it in Django, but I really like the simplicity behind it.

Also, I didn’t know Starlette supported subdomain routing via Host! That’s a really handy feature.

lovelydinosaur:
> Also, I didn’t know Starlette supported subdomain routing via Host! That’s a really handy feature.

Indeed - it's not actually documented yet (and I need to add support for `routes=`).

In practice it's actually a bit awkward to use in local development, so may need to put up a bit of documentation on how to use tools such as ngrok, or how to edit `/etc/hosts` or something.

sandys:
the routing change makes me ask - do you have any plans for a starlette project generator ? like a "rails new" command. the routing config could be put in a separate file (pretty much like routes.rb) and turn off formatting for black as well. 

As more and more declarative features get included - a project generator would be very very useful.

lovelydinosaur:
@sandys I'm not exactly sure what we'll want to do there.

It's possibly that we want to leave that kind of thing for higher level frameworks like Fast API to explore.

Or, it might be that we just recommend something along these lines...

```shell
$ git clone https://github.com/encode/starlette-example myproject
$ cd myproject
$ rm -r .git
```

(Note that the starlette-example repo needs updating in line with the 0.13 changes)

Or it could be that, yes, it makes sense for use to have a lightweight `starlette init` command.

sandys:
ideally a `starlette init` would be very useful (i daresay far more useful than the documentation in guiding what is best practice and what is not!)

the alternative is the example project. Actually we are treating hostedapi as pretty much that !

florimondmanca:
From past experience, project code generation is not necessarily the best approach when it comes to hinting at best practices.

Sure, it might be *convenient*, but it can turn out to be hard to maintain, and IMO it prevents users from actually *thinking* about how they'd like to set things up. It might make sense for super-opinionated frameworks like Django, but I don't think it'd bring much value to a super lightweight one like Starlette.

I think ideally the documentation should serve as an example, and point to example projects (`starlette-example`/HostedAPI) if users want to see code of real-world-like apps.

Edit: perhaps let's open an issue for this particular topic if there's more discussion needed?

sandys:
i dont agree with the forced usage of "_users should think_" and "_lightweight frameworks should not have generators_"

I think a framework can be lightweight and yet handholds users. A great example of this is alembic. 
Alembic provides `init` and `revision` which helps setup the environment and then create a migration each time. It takes care of checks and checksums, etc

Obviously its @tomchristie 's prerogative here ...but there are sufficient examples of fantastic frameworks that do this.

gvbgduh:
I think it might have sense what @sandys proposes.

The beauty of starlette that it fits many purposes as it's also a perfect level for big or custom projects, it's not that high level to customize existing functionality, so you can have full control of what you do and it's not too low level to have unreasonable boilerplate.

In case the goal is simple and a lightweight app it might end up as everything in one file and in a few, but you generally wouldn't be thinking about the project structure and its template.

Where in case of a bigger project it might be handy to have some guide around (like Django) does. It's also the case where the route decorators become a mess as spread through the whole project. `flask-RESTful` also sort or guides towards this direction.

But be that as it may, It's tricky indeed to come up with such a template that would fit custom purpose projects and higher-level frameworks as well.

euri10:
> From past experience, project code generation is not necessarily the best approach when it comes to hinting at best practices.
> 
> Sure, it might be _convenient_, but it can turn out to be hard to maintain, and IMO it prevents users from actually _thinking_ about how they'd like to set things up. It might make sense for super-opinionated frameworks like Django, but I don't think it'd bring much value to a super lightweight one like Starlette.


+1 on this, I not only fully agree but this is proved on a daily basis, just take a look at issues on https://github.com/tiangolo/full-stack-fastapi-postgresql/issues 

Most questions have nothing to do with the framework: how I add a celery tasks, my traefik container doesn't work, how do I connect to pgadmin, how can I add a user to the db, how can I debug in docker ?

The result is simple, issues are not addressed.
It effectively makes people lazy and fail to search and think.

sandys:
And i genuinely dont see how one can solve it by asking users to RTFM or
think.

This is a ux (user experience) part of a project - there is value in
reducing barrier to adoption.

React does it, rails does it, alembic does it, django does it...as that
tool becomes better, the number of repetitive questions go down.

lovelydinosaur:
Let's move any convo on `init` and layouts over here... https://github.com/encode/starlette/issues/705

tiangolo:
About the `list` of routes vs decorator-style:

I see how this helps to structure parts by their type, all routing is done in a single place, in a list, and the functions that handle each route are separated in another place.

---

On the other side, my argument in pro of decorator style is that it ties things together by the *feature* or *subject* they are related to.

So, the code that handles a username is near the declaration of the route `/user/{username}`.

I find that especially useful with routes with path parameters like:

```Python
@app.route('/user/{username}')
def user(request):
    return PlainTextResponse('Hello, %s!' % username)
```

In this case, the `username` in `'/user/{username}'` is *close* to the one in `request.path_params['username']`, just a few lines above.

If someone was refactoring that to make it `user_name` instead of `username` (just a silly example), after changing the path to `'/user/{user_name}'` it would be easier to remember to also change the parameter extraction to `request.path_params['user_name']`, just because they are visually closer. And if that person forgot to refactor it inside, finding the bug later would probably be easier still.

Having those two related texts naming something in different files could, in some cases, make that refactor a bit more difficult.

This is similar to the philosophy in React, of having code of different types (HTML-ish JSX and JavaScript) close together, in the same file/component, but related in their *subject*, not the type of code. And React-router works in a similar way.

I actually didn't like React until recently, that I started using it and it made sense to me (I used only Angular and Vue before that). But it also seems that that idea (React itself) ended up being quite popular.

---

The preference between these two can quickly become quite subjective and dependent on personal/team taste, so I understand if this is the preferred way for Starlette. I still wanted to share that additional aspect of it, that might be important.

For FastAPI it's quite more relevant, as there's even more "shared information" between the part that declares a route path (the decorator) and the code in the function that handles it, so I'll keep that as the default there.

As long as I can keep using decorators in FastAPI, I'm OK with the decision. It currently uses `.add_route()`, so I would be inclined towards keeping it.

But even if not, if there's a way to update the routes incrementally (I guess some `self.router.routes`), that would be fine for me.

dmontagu:
@tiangolo Yeah, in addition to the obvious awkwardness posed by path parameters, I think the various `response_X` arguments to the FastAPI decorators are useful, and are best included near the actual endpoint function definition (maybe that's what you meant by "shared information").

If those arguments had to be placed in a routes list rather than next to the endpoint definition, I think it would get hard to maintain.

I have some ideas about how FastAPI might be changed to make it easier to follow this approach, but I'll discuss those in https://github.com/tiangolo/fastapi/issues/687.

tiangolo:
Yep, that's what I meant @dmontagu. And, yeah, my plan in FastAPI is to keep decorators.

---

Still, it shouldn't be a problem, FastAPI doesn't have to use the same style as Starlette, it can use it internally as an ASGI toolkit (one of Starlette's its objectives). As long as Starlette supports it.

Also, Starlette's goals are probably more inclined to be a very lightweight, simple, ASGI toolkit, with a small surface and a clean and simple codebase (it's beautiful inside) that can be easily extended and used by many other tools.

While FastAPI's goals don't include being a toolkit for other frameworks, nor having an as-simple-as-possible codebase. The goals are heavily inclined towards the final API developer's "user/development experience". As an example, all the app HTTP methods like `app.get()`, `app.post()` are just syntax sugar, and increase the internal codebase surface/complexity, required tests, etc. Just to improve final developer's experience a bit, reducing their app's codebase surface and complexity. That's also why there are lots and lots of docs, even for some things that are not strictly related to FastAPI itself.

So, it's OK if both projects adopt/suggest different styles in some places, their objectives are probably a bit different.

And also, although decorator vs list styles look quite different, supporting decorator style in FastAPI wouldn't be that difficult while still using Starlette's awesomeness.

Anyway, we can continue the FastAPI specific discussion in that issue as @dmontagu suggests, but I still wanted to bring some attention to that point of the closeness in code of related *subjects*, in case that's something important for Starlette itself.

lovelydinosaur:
@tiangolo - Very much agree with all the above, yup! 😃

And yeah, totally dig the "closeness of related concerns". I think the overall trade-off is just a bit different for Starlette, than it is for FastAPI.

luckydonald:
Coming from Flask instead of Django I really appreciate how you can just open up a `view` file, and instantly read what's served here.

I was just about to switch to FastAPI, and with it this project.
And even with it being not encouraged to do so, I highly prefer the `@app.decorator` approach.

It just suits some projects much better, where getting something up fast (but without being sloppy).


That's something I experienced with flask very well:

First you have an idea. A simple web app or API which should deliver things.
You add a `main.py` add `app = Flask(__name__)` and are good to go to get a first working prototype set up very quickly.
You don't have to set up routes and edit 20 files <sup><sup>*exaggeration</sup></sup> just to get a simple hello world running.

And that's a very pythonic way of doing things.
In for example java you have to first build a 
```java
class Main {
  public static void main(String args[]) {
    // your code here
  }
}
```
just to get anything running.
Python is not like that.

So back to our example. The app we're building is now getting more complex and so does our code.
No problem, we split it up in database modules, templates etc.
And our routes get split up in views, by using blueprints.

So that's growing very good with our app, and if it is getting big there is no problem there.

Some apps however never need to grow, they stay simple.
Imagine a simple API, a personal blog, or a one-shot landing page for a product with just a few email newsletter functions.

That's when setting up a dedicated routing file is more of a burden than a feature.
And that's how all of my projects start out, commercially or private: Small.

That's why I love the decorator way, it's simple and I have to edit only one file.

luckydonald:
I see why forcing the `list` style makes sense for more _let me call it "prepared"_ projects, where you have a clear picture from the beginning how your routing will look like, e.g. when you have to conform to a functional specifications document or similar.

However I'd love to be able to use the old style as I'm now starting with a new project and looking for a good async alternative of `flask` and have also interest in using the promising `FastAPI`.

I can understand the idea of dropping it from the framework for speed gain and less code, which is a valid goal. Maybe that future plan could instead of dropping it completely just refactor them in a separate project build on Starlette (maybe `Starlette-decorators`, I'm bad with names, lol), similar of how FastAPI is a good addition?

luckydonald:
Also could you maybe serve a pre-0.13 version of the docs, to make it easier for those who still like to use (or have to use) the decorators?

PR #707, opened 2019-11-06
Handle trailing slash redirects properly.

### LINKED ISSUE #633: Strip trailing slash ( / ) from URL route strings.
Hey all, an optional trailing slash seems to be pretty standard on most websites, but in Starlette, you must add separate routes to handle them. Example: 

```
routes = [
    Route('/', PlainTextResponse("home")),
    Route('/test', PlainTextResponse("test")),
    Route('/test/', PlainTextResponse("I need to be added separately?!"))
]

```

Could a `route = route.rstrip("/")` please be added to https://github.com/encode/starlette/blob/master/starlette/routing.py in `class Router` before running the match so these routes don't have to be manually added? Thanks!

lovelydinosaur:
Currently we have behavior so that if:

* A `/some-path/`route exists.
* A `/some-path` route does not exist.

Then if a request to `/some-path` is made, we'll redirect to the trailing slash case.

We ought to also handle the converse case, so that if...

* A `/some-path`route exists.
* A `/some-path/` route does not exist.

Then if a request to `/some-path/` is made, we'll redirect to the no trailing slash case.

Closes #633

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -614,22 +614,23 @@ class Router:
             await partial(scope, receive, send)
             return
 
-        if scope["type"] == "http" and self.redirect_slashes and scope["path"] != "/":
+        if scope["type"] == "http" and self.redirect_slashes:
             if scope["path"].endswith("/"):
                 redirect_path = scope["path"].rstrip("/")
             else:
                 redirect_path = scope["path"] + "/"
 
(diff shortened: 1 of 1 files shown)


gnat:
Thank you @tomchristie !!

lovelydinosaur:
Welcome - thanks for raising the issue!

PR #708, opened 2019-11-07
Fix a small typo in CORS middleware documentation

gdetrez:
The HTTP method for preflight requests is [OPTIONS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS)


lovelydinosaur:
Thanks!

PR #709, opened 2019-11-07
Fix RedirectResponse params (to accept same optional params as Response)

plizonczyk:
Also updating docs to properly reflect base Response params.


HarrySky:
@tomchristie any reason this is not being merged? Without this param we have to use a hacky workarounds (like using `Response` directly and providing all the same params that `RedirectResponse` use)

plizonczyk:
Closing, seems it's fixed in the master branch for a while now.

PR #711, opened 2019-11-09
Add masked property to URL and Secret

remarkov:
This can be useful for cases when you would like to put a masked value in the log etc.

Example usage:
`logger.info(f"Database initialized with {settings.DATABASE_URL.masked}")`


lovelydinosaur:
I'd suggest using `logger.info(f"Database initialized with {repr(settings.DATABASE_URL)}")`, right?

PR #714, opened 2019-11-13
Sharing state with submounted Starlette applications

nkoshell:
### Sharing state with submounted Starlette applications

Include an Starlette app:
* `app.mount(prefix, subapp)` - via mounted under the given path prefix
* `app.host(host, subapp)` - via mounted under the given host

```pydocstring
# root state
app.state.foo = 'bar'
app.state.bar = 'baz'
# subapp scoped state
subapp.state.foo = 'baz'
subapp.state.baz = 'foo'
```

```pydocstring
>>> app.state.foo
'bar'
>>> subapp.state.foo
'baz'
>>> subapp.state.bar
'baz'
>>> app.state.baz
...
AttributeError: 'State' object has no attribute 'baz'
>>> subapp.state.baz
'foo'
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+root = true
+
+[*.py]
+max_line_length = 88
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -84,10 +84,42 @@ generic `app.state` attribute.
 
 For example:
 
-```python
+```pydocstring
 app.state.ADMIN_EMAIL = 'admin@example.org'
+# or
+app.state['ADMIN_EMAIL'] = 'admin@example.org'
 ```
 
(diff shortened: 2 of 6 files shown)


lovelydinosaur:
Thanks for the pull request, tho this isn't a pattern I want us to promote.

Generally you should only use `Starlette` as the very top-level of your application. You can submount ASGI applications inside that, but you shouldn't really need those to be complete standalone `Starlette` instances themselves.

(Use `Router` instances if you really need to, but even then you should probably just be mounting a set of routes. I think https://github.com/encode/starlette/pull/704 will help clear this up, since I think our existing routing/mounting docs are confusing, and help lead users into unintended patterns such as this one.)

PR #718, opened 2019-11-14
Drop double square brackets tests in install script

espositofulvio:
sh doesn't support double square brackets tests which are a bash feature.
This commit replace them with single square brackets tests


lovelydinosaur:
Sure thing, thanks.

PR #719, opened 2019-11-15
Added support for publishing wheel files.

### LINKED ISSUE #668: no pypi wheel
would it be possible to add a wheel on pypi ? like starlette-0.12.9-py3-none-any.whl ?

apollo13:
Fixes #668


lovelydinosaur:
Cheers! 👏

PR #727, opened 2019-11-22
Use async version of jinja

RealOrangeOne:
https://jinja.palletsprojects.com/en/2.10.x/api/#async-support

This allows rendering to run on the event loop, which increases performance. It also enables access to async methods when rendering templates.

I'm not _entirely_ happy with how `_TemplateResponse` uses a fully custom constructor, but I think it's probably fine. I've also defaulted the media type to HTML at the argument level rather than providing a default, so the constructor is slightly simpler.

I'm not sure if there are any issues with calling `init_headers` later in the lifecycle (namely inside `__call__`, but again all the existing tests pass, so it's probably fine!

### DIFF
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -1,7 +1,7 @@
 import typing
 
 from starlette.background import BackgroundTask
-from starlette.responses import Response
+from starlette.responses import HTMLResponse
 from starlette.types import Receive, Scope, Send
 
 try:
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
I'm a nope on this.
Templates shouldn't need to make I/O, and shouldn't need to be async - supporting this will just push our users in the wrong direction.
I'm happy to make it *easier* to switch out the templating behavior, and eg. allow uses to customize the env however they wish, but I'm going to take a look at that seperately to this.

RealOrangeOne:
@tomchristie interested to know what direction you're worried about?

I don't know how much of the internals of jinja are made async by this change (template rendering is generally IO bound), but the ability to call async things from template rendering is quite nice, whether it be calling out to upstreams. Obviously that could be passed in via context, but still.

I was toying with making this functionality opt-in, but personally concluded there's very few reasons not to use it, as it's merely an extension rather than enforcing everything be async.

lovelydinosaur:
> interested to know what direction you're worried about?

Specifically - I/O in templates is an anti-pattern.

alex-oleshkevich:
@tomchristie, templates are read from disk which is IO operation:
https://github.com/encode/starlette/blob/master/starlette/templating.py#L27

don't you aware that under a high load it could be a bottleneck?

https://github.com/pallets/jinja/blob/master/jinja2/loaders.py#L180
https://github.com/pallets/jinja/blob/master/jinja2/utils.py#L154

lovelydinosaur:
That's not the same issue as async template rendering.

And *yes* we could be using threadpool'ed loading, which *might* well make sense, although it's actually not super clear in practice.

RealOrangeOne:
It's also unfortunately not obvious whether running Jinja with an async environment causes it to load the templates async too. 

Looking at those code snippets, it looks like it's done sync, but it's possible jinja does some fancy things to switch it out in async context. As I understood it, all async context did was let you use coroutines in context.

alex-oleshkevich:
There is async rendering support in jinja:
https://jinja.palletsprojects.com/en/2.10.x/api/#jinja2.Template.render_async

PR #728, opened 2019-11-22
Show nicer warning when `ujson` isn't installed

RealOrangeOne:
Copying the pattern used in other places, we should show a warning when `ujson` isn't found when trying to use `UJSONResponse`

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -156,6 +156,7 @@ class UJSONResponse(JSONResponse):
     media_type = "application/json"
 
     def render(self, content: typing.Any) -> bytes:
+        assert ujson is not None, "ujson must be installed to use UJSONResponse"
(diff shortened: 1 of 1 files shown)


gvbgduh:
@RealOrangeOne, thanks for the contibution, but i'm afraid it's not the best place to warn the user. 
Optional libs are attempted to be imported at some point and if not successfully, `None` is assigned. That would probably a better place for that.
Or probably try to fallback for the std lib if that's available.
But it's worth unifying it across all optional libs and other projects as well. So, worth raising with @tomchristie.

lovelydinosaur:
This is reasonable enough.

Although worth pointing out that I'm strongly in favor of us moving towards deprecating `UJSONResponse`, and just *documenting* how to use alternate JSONResponse subclasses.

PR #729, opened 2019-11-22
Add compression level in GZipMiddleware

bronekot:
Add the ability to change the compression level. By default, python gzip.GzipFile compresslevel is 9. This commit saves the behavior of the standard library, but allows you to override compresslevel. High values are not recommended for streaming compression.


lovelydinosaur:
Thanks. I’d probably rather we just choose suitable defaults for our users rather than adding additional configuration switches. If you’ve strong thoughts on a better default value, we could certainly consider that.

PR #730, opened 2019-11-22
Changing the compression level in GZipMiddleware

bronekot:
Being one of your users I find the speed of stream compression too slow at times. I strongly believe that adding this feature will benefit a lot of users. It will not interfere with the functionality if not used but will give the users control over the compression level. Sorry for sending this again. This time I made sure it is formatted properly.

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -6,29 +6,31 @@ from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 
 class GZipMiddleware:
-    def __init__(self, app: ASGIApp, minimum_size: int = 500) -> None:
+    def __init__(self, app: ASGIApp, minimum_size: int = 500, compresslevel: int = 9) -> None:
         self.app = app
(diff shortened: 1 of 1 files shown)


lovelydinosaur (commented):
Thanks! I’m totally in favour of us switching the compression level in GZipFile, yup!

But... let’s not bother making it configurable. Having fewer things  for the end user to think about is a good thing! 😃

gnat:
I currently don't include this middleware because of performance. Making this an optional parameter to allow tuning would be amazing! (But at least bringing it down from 9 to a more reasonable level such as 2 would be good.)

scriptator:
What's the status on this pull request? Sounds great to me as I am also experiencing performance issues

Kludex:
I just saw this PR, the same idea was merged on #1128, and it's available on [Starlette 0.15.0](https://www.starlette.io/release-notes/#0150).

@BroneKot Thanks for the PR! :smile: And... Sorry for the long wait. :cold_sweat:

PR #732, opened 2019-11-22
SameSite support for cookies

danielhedren:
With the new secure-by-default approach chrome has taken, we need to be able to set the samesite attribute of cookies in some applications. The default behaviour is left as it was previously, i.e. samesite unset.


starlette/responses.py
```diff
@@ -14,6 +14,11 @@
 from starlette.datastructures import URL, MutableHeaders
 from starlette.types import Receive, Scope, Send
 
+# Workaround for adding samesite support to pre 3.8 python
+from http.cookies import Morsel
+
+Morsel._reserved[str("samesite")] = str("SameSite")  # type: ignore
```
lovelydinosaur: We shouldn't need the `str()` wrappers here.

(Eg. the comments against https://stackoverflow.com/questions/50813091/how-do-i-set-the-samesite-attribute-of-http-cookies-in-python?noredirect=1#comment93090548_50813092 indicate that they were only there for Python 2 support)

danielhedren: Yeah, you're right.

molten-firescar96:
Glad to see someone already made a PR. What else is needed before this is merged in?

lovelydinosaur:
Okay, this is nice, aplogies for the slow uptake on it.

The `None` vs `"none"` case is a bit confusing. Given that browsers are [starting to default to "lax"](https://www.chromestatus.com/feature/5088147346030592) for the unset case ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies)), I wonder if we should:

* Switch to `"lax"` by default.
* Drop the `None` case - enforce that we always set something, that may be either `"strict"` `"lax"`, or `"none"`.

danielhedren:
> 
> 
> Okay, this is nice, aplogies for the slow uptake on it.
> 
> The `None` vs `"none"` case is a bit confusing. Given that browsers are [starting to default to "lax"](https://www.chromestatus.com/feature/5088147346030592) for the unset case ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies)), I wonder if we should:
> 
>     * Switch to `"lax"` by default.
> 
>     * Drop the `None` case - enforce that we always set something, that may be either `"strict"` `"lax"`, or `"none"`.

I think defaulting to "lax" and dropping None is a good idea for the reasons you point out.

lovelydinosaur:
Fantastic stuff, thanks!

docs/responses.md
```diff
@@ -31,7 +31,7 @@ async def app(scope, receive, send):
 
 Starlette provides a `set_cookie` method to allow you to set cookies on the response object.
 
-Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/", domain=None, secure=False, httponly=False)`
+Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/", domain=None, secure=False, httponly=False, samesite=None)`
```
lovelydinosaur: Needs `samesite="lax"` here, too.

lovelydinosaur:
Ace, good work!

zero-shubham:
I don't see SameSite is being added when using response.set_cookie(). tried and tested with -> samesite=None samesite="none" samesite="lax" it should show up in headers when checking inside network tab right?

lovelydinosaur:
You'd need to be testing against master - it's not in the latest release yet.

zero-shubham:
> You'd need to be testing against master - it's not in the latest release yet.

Got it, for now this is my solution -> https://github.com/tiangolo/fastapi/issues/1099

PR #737, opened 2019-11-26
Allow custom filters for Jinja2

eillarra:
We could change `templates.env` directly, but I am not sure what the preferred way is.
If you prefer that we add filters directly changing the `env` property, then at least we should update the documentation.

Option 1 (included in this pull request):

```python
from commonmark import commonmark

def marked_filter(text):
    return commonmark(text)

templates = Jinja2Templates(directory='templates', custom_filters={
    'marked': marked_filter
})
```

Option 2 (only needs updating the documentation):

```python
from commonmark import commonmark

def marked_filter(text):
    return commonmark(text)

templates = Jinja2Templates(directory='templates')
templates.env.filters['marked'] = marked_filter
```

Let me know what you think.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -36,6 +36,20 @@ For example, we can link to static files from within our HTML templates:
 <link href="{{ url_for('static', path='/css/bootstrap.min.css') }}" rel="stylesheet">
 ```
 
+If you want to use [custom filters][jinja2], you will need to update the `env`
+property of `Jinja2Templates`:
+
+```python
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Option 2 sounds good here.

eillarra:
Ok, I have squashed the commits: it only includes a note in the docs now.

alex-oleshkevich:
`Jinja2Templates` also can pass `**kwargs` to the `Environment`.

PR #742, opened 2019-12-04
fix: Incorrect parameter to WebSocketRoute

gurditsbedi:
Possible Typo in docs. Git diff will be able to explain things.


gurditsbedi:
This is interesting. I changed the docs. But the build failed.

Kludex:
This was fixed on #827. @gurditsbedi Thanks for the PR! :tada:

PR #745, opened 2019-12-06
Add missing import in config documentation

graphman65:


JayH5:
Thanks, this seems to be obsolete now.

PR #751, opened 2019-12-13
TestClient timeout simulates http.disconnect

fantix:
So that on the server side, `http.disconnect` can be `receive`d when timed out.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -162,16 +162,24 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
         request_complete = False
         response_started = False
         response_complete = False
+        timeout_called = False
         raw_kwargs = {"body": io.BytesIO()}  # type: typing.Dict[str, typing.Any]
         template = None
         context = None
+
+        def do_timeout() -> None:
+            nonlocal timeout_called, response_started, response_complete
+            timeout_called = True
+            if request_complete:
+                response_started = True
+                response_complete = True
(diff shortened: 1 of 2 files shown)


blueyed:
Looks interesting.  I've quickly tried if it would abort an `await asyncio.sleep(5)`, but that does not appear to be the case.  Would it possible to handle this?  (i.e. that the timeout for `client.get` would ensure that it takes no longer than that)

fantix:
Oh, this patch only added timeout feature to the `TestClient` and the `requests`-powered ASGI "server". That means, on timeout, the `TestClient` will "disconnect" the connection, and an `http.disconnect` event is raised in the ASGI "server". However, it is totally up to the ASGI **application** to handle the `http.disconnect` event - whether abort the current handling task or whatever else. Please see [an example here](https://github.com/uc-cdis/metadata-service/blob/7c7e50ba49cc80cda829dae2a6046a7d1248e7fa/src/mds/app.py#L18-L51) where we abort any pending request handlers when `http.disconnect` is detected.

Kludex:
Thanks for the PR @fantix ! Sorry for taking so long to act here :sweat_smile: 

I think this feature is extremely useful. It's going to be useful to test `StreamingResponse`, a bug we have on `BaseHTTPMiddleware` and also another PR that is coming around that needs a test when the client disconnects.

I'll be closing this issue, and proceed with it on #1446. I've co-authored yourself, as I've just adapted this code.

PR #755, opened 2019-12-17
support Path in Config

trim21:
`open` or `os.path.stat` support `pathlib.Path`, So we coule use `Path` as `env_file`


lovelydinosaur:
Good call, yup.
The test failures are unrelated, and I'm looking into that, so let's pull this in.
Thanks!

Issue #739, opened 2019-12-01
Background tasks exception handler


aviramha:
Currently, if a background task faces an exception, it isn't handled and there's no way to handle it (rather than wrapping the task)
My suggestion is to be able to add exception handler for background tasks, as there is for requests.

lovelydinosaur:
Yup, this is interesting.

So, the exception handler currently only runs for HTTP requests (not WebSocket) and only runs if a response hasn't already been returned. This is because any 500 page returned by the exception handler can't be displayed if we've already returned a response.

However, I think we've got the wrong policy there. Instead we ought to *always* run the exception handler, but only render its response if we haven't already started doing so.

The more tricky bit of that is how websockets should fit in there. I guess we probably want to change the signature from `handler(request, exc)` to `handler (conn, exc)`, and document that `conn` is an `HTTPConnection`, and might be *either one of* of the `Request` or `WebSocket` subclasses.

aviramha:
Sounds good, should I submit a pr?

lovelydinosaur:
Welcome to take a look at it, yup!

aviramha:
I didn't quite understand the implications of the Websocket changes, as there's no defined signature (right now, besides http_exception func).

aviramha:
Please check my PR, if it's OK I'll write tests and documentation.

lovelydinosaur:
Also, just to get a slightly better idea of the context here - what kind of handled exception are you catching here, and what are you doing with it in the exception handler?

The default exception handling is setup so that you can decouple eg. "A page not found exception occured" from "here's how we want to render the 404 page".

When you're using background tasks you don't have anything to render, so it's a bit less clear what sort of use case *handled exceptions* have.

aviramha:
The use case I'm after is Asgi Exception Middleware similar to Sentry.
I want to report all exceptions that happen in the service - including background tasks.
In my current project I'm using a decorator of try catch, instead of my default exception handler that works for request exceptions.

lovelydinosaur:
Right, yeah. So, that's different, although we need some careful documentation around this.

Starlette has *exception handlers* for handling particular exception classes in a specfic way (eg. to raise a 404 Not Found, which isn't an error, but where you do want to bypass the standard flow), and an *error handler* for the last-ditch handling of all unexpected errors.

What you *actually* want here is for the *error handler* to always run...

https://github.com/encode/starlette/blob/master/starlette/middleware/errors.py#L158

[...]

PR #758, opened 2019-12-17
Call exception handlers even if response started

aviramha:
This change allows background tasks exceptions to be caught by exception handlers.
The implication is that handlers are always called, but response is rendered by the handlers only if not rendered yet.
** THIS IS A DRAFT ** I will add tests and document changes after initial approval.

### DIFF
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -81,16 +81,15 @@ class ExceptionMiddleware:
             if handler is None:
                 raise exc from None
 
-            if response_started:
-                msg = "Caught handled exception, but response already started."
-                raise RuntimeError(msg) from exc
-
             request = Request(scope, receive=receive)
             if asyncio.iscoroutinefunction(handler):
                 response = await handler(request, exc)
             else:
                 response = await run_in_threadpool(handler, request, exc)
-            await response(scope, receive, sender)
+
+            if not response_started:
+                # Send the response only if not response started already.
+                await response(scope, receive, sender)
 
     def http_exception(self, request: Request, exc: HTTPException) -> Response:
         if exc.status_code in {204, 304}:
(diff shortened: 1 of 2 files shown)


aviramha:
#739

starlette/exceptions.py
```diff
@@ -81,15 +81,16 @@ def _lookup_exception_handler(
...
-                msg = "Caught handled exception, but response already started."
-                raise RuntimeError(msg) from exc
-
             request = Request(scope, receive=receive)
             if asyncio.iscoroutinefunction(handler):
                 response = await handler(request, exc)
             else:
                 response = await run_in_threadpool(handler, request, exc)
+
+            if response_started:
```
lovelydinosaur: Not clear to me that we should actually even raise this error here, anymore?

lovelydinosaur: I'm going to suggest that we drop it. We're calling the error handler - that's sufficient.

lovelydinosaur: So instead, just...

```python
if not response_started:
    await response(scope, receive, sender)
```

aviramha:
Added tests, 100% coverage :)

tests/test_exceptions.py
```diff
@@ -58,11 +58,26 @@ def test_websockets_should_raise():
 
 
 def test_handled_exc_after_response():
-    # A 406 HttpException is raised *after* the response has already been sent.
-    # The exception middleware should raise a RuntimeError.
-    with pytest.raises(RuntimeError):
```
lovelydinosaur, line 63: I would change the test much more simply here. A handled exception after the response is rendered isn't a problem now, we should just expect a 200 status code. We don't particularly need to check for side-effects.

aviramha: Wouldn't we want to make sure that the exception handler runs? In case the response is already rendered, we get the same result even if it's not called

Kludex:
As stated on #739, this is not something we're looking for. Let's go for #761. 

Thanks for the PR @aviramha ! Also, sorry for the waiting time.

PR #765, opened 2019-12-21
Fix endpoints doc page example

cauethenorio:
The example isn't working because the wrong class was being used as the WS route.


Kludex:
This was already fixed.

Thanks for the PR @cauethenorio , and sorry for the delay! :smile:

PR #766, opened 2019-12-21
Add space so example is copy-pastable

EmilStenstrom:
This change adds a space character to the routing docs so the example is copy-pastable.


lovelydinosaur (commented):
Thanks!

PR #767, opened 2019-12-21
Allow specifying `pip install starlette[aiofiles,itsdangerous]`

EmilStenstrom:
Adding all dependencies as extras_require allows requirements.txt file to use `starlette[aiofiles,itsdangerous]` which groups the dependencies with the reason they are included (as part of starlette). This makes it easier to understand why a certain package was installed.


EmilStenstrom:
This has had no comments in over a year. Is there something I can do to improve the patch? Is it not wanted?

lovelydinosaur:
Personally I'd *probably* prefer we don't do this, no.
It's not a particularly standard style in other python projects.

EmilStenstrom:
@tomchristie Thanks!

PR #777, opened 2020-01-06
Typing fixes

blueyed:
`mypy starlette tests --ignore-missing-imports --check-untyped-defs` passed (except for #780).

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -10,6 +10,14 @@ from starlette.responses import PlainTextResponse, Response
 from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocket
 
+if False:  # TYPE_CHECKING
+    import sys
+
+    if sys.version_info >= (3, 8):
+        from typing import Literal
+    else:
+        from typing_extensions import Literal
+
 
 class HTTPEndpoint:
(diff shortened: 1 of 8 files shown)


starlette/endpoints.py
```diff
@@ -10,6 +10,14 @@
 from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocket
 
+if False:  # TYPE_CHECKING
```
florimondmanca, line 13: There's a built-in flag in `typing` for this :-)
```suggestion
if typing.TYPE_CHECKING:
```

blueyed: Yes, makes sense (given that this is py36+), thanks!
Not going to update it for now though, given its staleness already.

and-semakin:
What's the status of this PR? It contains the fix of one little typing annoyance I was going to suggest.

starlette/testclient.py
```diff
@@ -449,7 +449,7 @@ def websocket_connect(
 
         return session
 
-    def __enter__(self) -> requests.Session:
+    def __enter__(self) -> "TestClient":
```
and-semakin, line 452: Would be nice to have this fix merged.

and-semakin: Here we go #1064

PR #779, opened 2020-01-06
typing: remove unused "type: ignore" comments

blueyed:
Found via:

> mypy --warn-unused-ignores --ignore-missing-imports starlette

Using:

> mypy 0.770+dev.cdd91ba819646b0138b05a8e9180faba4ef0ff92.dirty


lovelydinosaur:
Thanks for this. Wonder if some of these are no longer required because `mypy` has gotten smarter since they were introduced.

Either way it's a nice improvement, ta!

PR #780, opened 2020-01-06
Update/fix type for (Immutable)MultiDict's `*args`

blueyed:
Tested/used in https://github.com/encode/starlette/blob/c80558e0/tests/test_datastructures.py#L325

Not included in https://github.com/encode/starlette/pull/777 since this is not clear to me.

Fixes:

> tests/test_datastructures.py:326: error: Argument 1 to "update" of "MultiDict" has incompatible type "None"; expected "Union[MultiDict, Mapping[Any, Any], List[Tuple[Any, Any]]]"


Kludex:
This is being handled by #1449.

PR #782, opened 2020-01-06
ServerErrorMiddleware: generate_html: do not capture locals

blueyed:
This might causes errors via `repr`, and is not necessary/used in the
first place.

Ref: https://bugs.python.org/issue39228


Kludex:
The reference was closed as "not a bug" on the Python issue tracker.

For the reason above, I'm closing this PR.

PR #785, opened 2020-01-06
Add support for contextlib-like lifecycle handlers.

iksteen:
These changes allow the startup event handlers to be an (async) generator to which control is returned when the shutdown handlers are called. This can be useful when trying to control another object's lifecycle if that object provides a(n) (asynchronous) context manager.

The control will be returned to perform shutdown operations in reverse order of how the lifecycle generators were added.

f.e.

```
async def prepare_database():
    async with db.ConnectionPool() as pool:
        app.state.pool = pool
        try:
            yield  # Control will be returned during shutdown event.
        finally:
            del app.state.pool

app = Starlette(on_startup=[prepare_database])
````

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -5,8 +5,9 @@ is shutting down.
 
 ## Registering events
 
-These event handlers can either be `async` coroutines, or regular syncronous
-functions.
+These event handlers can either be `async` coroutines, or regular synchronous
+functions. In case of the startup event, it can also be an `async` generator
+or a regular synchronous generator.
 
 The event handlers should be included on the application like so:
 
@@ -20,13 +21,17 @@ async def some_startup_task():
 async def some_shutdown_task():
     pass
 
+async def some_startup_generator():
+    async with SomeContextManager():
+        yield  # Control is returned when shutdown is initiated.
+
 routes = [
     ...
 ]
 
 app = Starlette(
     routes=routes,
-    on_startup=[some_startup_task],
+    on_startup=[some_startup_task, some_startup_generator],
     on_shutdown=[some_shutdown_task]
 )
 ```
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -497,13 +497,19 @@ class Router:
         for handler in self.on_startup:
             if inspect.isasyncgenfunction(handler):
                 gen = handler()
-                await gen.__anext__()
+                try:
+                    await gen.__anext__()
+                except StopAsyncIteration:
+                    raise RuntimeError("async generator did not yield")
                 self.lifecycle_generators.append(gen)
             elif asyncio.iscoroutinefunction(handler):
                 await handler()
             elif inspect.isgeneratorfunction(handler):
                 gen = handler()
-                next(gen)
+                try:
+                    next(gen)
+                except StopIteration:
+                    raise RuntimeError("generator did not yield")
                 self.lifecycle_generators.append(gen)
             else:
                 handler()
@@ -527,11 +533,15 @@ class Router:
                     ).__anext__()
                 except StopAsyncIteration:
                     pass
+                else:
+                    raise RuntimeError("async generator did not exit")
             else:
                 try:
                     next(typing.cast(typing.Generator[None, None, None], gen))
                 except StopIteration:
                     pass
+                else:
+                    raise RuntimeError("generator did not exit")
 
     async def lifespan(self, scope: Scope, receive: Receive, send: Send) -> None:
         """
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -535,6 +535,8 @@ class Router:
                     pass
                 else:
                     raise RuntimeError("async generator did not exit")
+                finally:
+                    await typing.cast(typing.AsyncGenerator[None, None], gen).aclose()
             else:
                 try:
                     next(typing.cast(typing.Generator[None, None, None], gen))
@@ -542,6 +544,8 @@ class Router:
                     pass
                 else:
                     raise RuntimeError("generator did not exit")
+                finally:
+                    typing.cast(typing.Generator[None, None, None], gen).close()
 
     async def lifespan(self, scope: Scope, receive: Receive, send: Send) -> None:
         """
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
This is a great idea, yup. I'd come to the same kind of conclusion, too.

I'm not sure what the API here should look like, as it might be that we want to move towards a single generator function as the prefered startup/shutdown style *instead* of `on_startup`, `on_shutdown`.

dmontagu:
+1 for single generator for on startup and on shutdown. There was a lengthy issue on this topic somewhat recently in fastapi where that approach was argued for.

sm-Fifteen:
Careful, [async generator finalization is not guaranteed](https://www.python.org/dev/peps/pep-0533).

If your generator can still yield after the two calls to `__anext__`, the generator will stay pending and the context managers inside it will not be closed. [Async generators don't get cleaned up at all on GC](https://www.python.org/dev/peps/pep-0525/#finalization) unless they're registered to be finalized before the event loop closes with `sys.set_asyncgen_hooks()`. You'll want to hook your generators that way when setting them up in your startup handler so they at least get a chance to run if the interpreter shuts down before the shutdown event is fired (like uvicorn does if Ctrl+C is pressed twice).

And regardless of the above, you'll probably want to call `close()` or `aclose()` yourself on those generators when handling the shutdown event, it's always good practice to do so.

iksteen:
> I'm not sure what the API here should look like, as it might be that we want to move towards a single generator function as the prefered startup/shutdown style _instead_ of `on_startup`, `on_shutdown`.

@tomchristie Let me know what you want the api to look like and I'll adapt it (and try to maintain backwards compatibility with on_startup/shutdown).

> If your generator can still yield after the two calls to `__anext__`, the generator will stay pending and the context managers inside it will not be closed.

@sm-Fifteen That is a matter of protocol. But, it's easily detectable and I'll follow `contextlib`'s example of raising a runtime error when the initial `__anext__` doesn't yield or if the second `__anext__` doesn't result in a `StopAsyncIteration`.

> [Async generators don't get cleaned up at all on GC](https://www.python.org/dev/peps/pep-0525/#finalization) unless they're registered to be finalized before the event loop closes with `sys.set_asyncgen_hooks()`. You'll want to hook your generators that way when setting them up in your startup handler so they at least get a chance to run if the interpreter shuts down before the shutdown event is fired (like uvicorn does if Ctrl+C is pressed twice).

I'll add closing the generators using a `sys.set_asyncgen_hooks()`-hook.

> And regardless of the above, you'll probably want to call `close()` or `aclose()` yourself on those generators when handling the shutdown event, it's always good practice to do so.

Raising a `RuntimeError` if protocol is broken should be enough. Then again, calling `close()`/`aclose()` doesn't hurt.

sm-Fifteen:
> That is a matter of protocol. But, it's easily detectable and I'll follow `contextlib`'s example of raising a runtime error when the initial `__anext__` doesn't yield or if the second `__anext__` doesn't result in a `StopAsyncIteration`.

True, but it's also going to be a problem if you run `__anext__` once on startup and don't get to run it a second time on shutdown, if the event handler gets skipped for whatever reason. It's just an event handler, after all, there are no guarantees on whether or not it will actually be fired, and an ASGI server is supposed to terminate [regardless of whether application shutdown turned out to be successful or not](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#shutdown-failed).

iksteen:
> > That is a matter of protocol. But, it's easily detectable and I'll follow `contextlib`'s example of raising a runtime error when the initial `__anext__` doesn't yield or if the second `__anext__` doesn't result in a `StopAsyncIteration`.
> 
> True, but it's also going to be a problem if you run `__anext__` once on startup and don't get to run it a second time on shutdown, if the event handler gets skipped for whatever reason. It's just an event handler, after all, there are no guarantees on whether or not it will actually be fired, and an ASGI server is supposed to terminate [regardless of whether application shutdown turned out to be successful or not](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#shutdown-failed).

Shouldn't the `set_asyncgen_hooks`-hook catch that scenario? There is of course no guarantee that the event loop will actually terminate immediately after a shutdown failure occurs, but it will eventually. If the shutdown event handler isn't called or doesn't complete, the hook won't be removed either.

sm-Fifteen:
> Shouldn't the `set_asyncgen_hooks`-hook catch that scenario? There is of course no guarantee that the event loop will actually terminate immediately after a shutdown failure occurs, but it will eventually. If the shutdown event handler isn't called or doesn't complete, the hook won't be removed either.

The hook should catch that case, I just wanted to point this out as a correction of my initial statement where I made it sound like it would only be a problem "if your generator can still yield after the two calls to `__anext__`".

iksteen:
I've implemented the protocol checking (missing yield on startup, generator not exiting after shutdown) and closing the generators.

@sm-Fifteen It seems both asyncio's and uvloop's event loops already implement tracking async generators by setting up `sys.set_asyncgen_hooks()`. Adding this to starlette seems redundant and error-prone. IMO, this is something the ASGI server should manage.

lovelydinosaur:
So, I've implemented an alternate approach, that *doesn't* call directly into `__anext__`, but instead runs the function as a regular async generator.

Interested to know if #799 meets our requirements here?

sm-Fifteen:
@tomchristie: I definitely like this iterator-like way of doing things compared to manually calling `__anext__`.

https://github.com/encode/starlette/blob/a742ecda9dd3de0e9f18b7b557d3cd09856692c6/starlette/routing.py#L525-L529

The exception handling around it was a bit confusing to read (it initially looked like `lifespan.startup.failed` would also get returned if an exception got thrown on shutdown), but I think I prefer that method

I also like the way you handle `lifespan` as the overarching mechanism that's responsible for firing `on_startup` and `on_shutdown`, as those two halves of an imaginary context manager being kept apart from each other was really the part that made lifetime-scoped dependencies hard to reason about. Making this hookable is a big plus.

iksteen:
Feel free to close this and continue work on https://github.com/encode/starlette/pull/799.

PR #789, opened 2020-01-09
Typo in configuration docs

2tunnels:
Seems like there is a typo in configuration docs. I assume you want to check length of entire list, not the first element.


lovelydinosaur:
Thanks

PR #792, opened 2020-01-13
Allow non latin-1 filename in FileResponse

### LINKED ISSUE #790: response headers support custom encoding
Hi~ When i try to return a `FileResponse` with Chinese filename, i got such error:

```
UnicodeEncodeError: 'latin-1' codec can't encode character '\u6708' in position 32: ordinal not in range(256)
```

Then i found all header items will be encode with `latin-1`, after i change it manually to `utf-8`, my problem was solved.

e.g.

- responses.Response.init_headers, line 63
```python
raw_headers = [
    (k.lower().encode("latin-1"), v.encode("latin-1"))
    for k, v in headers.items()
]
```

So may be it should encode with `self.charset` or give a custom parameter so that i can choose encoding to `utf-8`?

Thx for the great project~

em92:
Fixes https://github.com/encode/starlette/issues/790

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -7,7 +7,7 @@ import stat
 import typing
 from email.utils import formatdate
 from mimetypes import guess_type
-from urllib.parse import quote_plus
+from urllib.parse import quote, quote_plus
 
 from starlette.background import BackgroundTask
 from starlette.concurrency import iterate_in_threadpool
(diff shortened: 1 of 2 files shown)


starlette/responses.py
```diff
@@ -229,7 +229,13 @@ def __init__(
         self.background = background
         self.init_headers(headers)
         if self.filename is not None:
-            content_disposition = 'attachment; filename="{}"'.format(self.filename)
+            content_disposition_filename = quote(self.filename)
```
sarathsomana: I'm afraid this expression `quote("ä") == "ä"` will return false and interpret the encoding as a `latin-1`.

lovelydinosaur:
This looks pretty reasonable yup.

Any links to other implementations (eg. Flask) that handle this case, so we can double check what they do? Any particular RFC information that's worth us following here?

em92:
```python
from flask import Flask, send_file
app = Flask(__name__)

@app.route('/')
def hello_world():
    return send_file("你好.txt", as_attachment=True)
```

https://github.com/pallets/flask/blob/64ba43411fbfb2e435455b0df2453d7809b7c5a8/src/flask/helpers.py#L614-L626

Issue #649, opened 2019-09-30
lifespan events in sub-applications


octopoulpe:
Hi.
I was wondering if sub-applications (mounted on the main application) should receive startup/shutdown events and call their own handlers.
I saw it is not the case at the moment. Should we document it or implement the dispatching of lifespan events ?

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/649">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/649/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/649/pledge.svg">
</picture>
</a>

gbozee:
Assuming the sub applications are connected using the Router, startup/shutdown events can be implemented as follows

```python
app = Router([Mount("/api", app=service_app), Mount("", app=root_app)])


async def open_database_connection_pool():
    await service_app.service.connect()


async def close_database_connection_pool():
    await service_app.service.disconnect()


app.lifespan.add_event_handler("startup", open_database_connection_pool)
app.lifespan.add_event_handler("shutdown", close_database_connection_pool)

```

octopoulpe:
Hi.
I chose to manage that through the main application handlers.
```
app = Starlette()

sub_applications = {"/cart": cart_app, "/account": account_app, etc...}
for route, sub_app in sub_applications.items():
    app.mount(route, sub_app)
    
@app.on_event("startup")
async def startup():
    for sub_app in sub_applications.values():
        await sub_app.router.lifespan.startup()
```

The same for shutdown. As long as you do not go all 'subsubapplication', I think it is quiet expressive as, in my case, sub application don't know they are "mounted" somewhere.

Shouldn't we, at least, document it ?

lovelydinosaur:
> I was wondering if sub-applications (mounted on the main application) should receive startup/shutdown events and call their own handlers.

Ideally, yes.

lovelydinosaur:
Having taken a look at this it's really quite awkward.

You can't run each submount off the same lifespan task - instead each would need to have their own sub-task. Then you need to make sure you're handling the case where lifespan events are not supported and deal with exception cases from each submount.

[...]

PR #799, opened 2020-01-14
Support generator function to manage startup/shutdown.

lovelydinosaur:
An alternate implementation of #785

- [x] Refactor lifespan to run as an async generator
- [x] Allow `lifespan_context=...` argument to `Starlette` and `Router`.
- [x] Add `app` argument to the lifespan context function.

End result API would be allowing the application to take a single `async` generator function, that can encapsulate any required startup/shutdown logic.

Taking the example in #785, you'd end up with this kind of pattern...

```python
async def lifespan(app):
    async with db.ConnectionPool() as app.state.pool:
        yield

app = Starlette(routes=..., middleware=..., lifespan=lifespan)
```


sm-Fifteen:
Using this approach, I have two questions:
1. Couldn't an application use a sync generator instead of an async one as their lifespan context? After all, [servers are forbidden from accepting connections until the a startup complete event is recieved](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#startup-complete), so there isn't a whole lot of things that could preempt the startup coroutines regardless of whether the functions being called are blocknig or not. There might not be any async code to run on startup or shutdown.
2. Could this be a step in fixing #649?

lovelydinosaur:
1. The async function is sufficient, since it can run either sync or async code. It *might* be graceful to be able to have it accept either kind of function, or it might be unnecessary extra noise?
2. Potentially, yeah.

sm-Fifteen:
>     1. The async function is sufficient, since it can run either sync or async code. It _might_ be graceful to be able to have it accept either kind of function, or it might be unnecessary extra noise?

Having looked into it, you can't loop though a sync and an async iterator with the same kind of loop, so it would mean duplicating the iteration part, which would be unneccessarily noisy. It probably wouldn't be worth it anyway, so nervermind that.

>     2. Potentially, yeah.

That's good if it turns out that way, lifespan has a number of caveats that make it unreliable (#649) or impractical (what this PR attempts to fix) to use right now. That's probably out of scope for this PR, though.

iksteen:
Works for me, just be sure to properly document that lifespan and startup/shutdown are mutually exclusive (at least, in this implementation).

lovelydinosaur:
> Couldn't an application use a sync generator instead of an async one as their lifespan context?

Alrighty, for completeness I've added support for the function to be *either* a generator or an async generator.

rugleb:
When can we expect a release with this feature?

PR #801, opened 2020-01-14
Security patch: Updating CORS origin regex to use a full regex match

### LINKED ISSUE #800: Dangerous example regex for CORS Middleware?
Looking at the docs for CORS Middlware here: https://www.starlette.io/middleware/#corsmiddleware , under the `allow_origin_regex` attribute, the example value is `https://.*\.example\.org`.

However, based on the handler code for this at https://github.com/encode/starlette/blob/ab86530eddfcf56e0f7e5ca56f6ab69c15594a7d/starlette/middleware/cors.py#L90 , it appears `https://www.example.org.hacker.com` would pass as a valid origin, right?

It seems like the example should be `https://.*\.example\.org$`, yes?

josephbuckley:
This patches a security vulnerability with CORS Middleware. By using a regular expression `match`, it leaves it to the implementer to ensure that the passed-in regex properly handles start/end of the string. The example documentation shows `https://.*\.example\.org`, which, when used with `match`, would return True for `https://.*\.example\.org.evil.com`.

Closes #800

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -87,7 +87,7 @@ class CORSMiddleware:
         if self.allow_all_origins:
             return True
 
-        if self.allow_origin_regex is not None and self.allow_origin_regex.match(
+        if self.allow_origin_regex is not None and self.allow_origin_regex.fullmatch(
             origin
         ):
             return True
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Absolutely, thanks!

PR #804, opened 2020-01-16
Added route value to scope

victoraugustolls:
The purpose of this pr is to add the RAW value of a path (route) to the scope.

Recently I tried to get this value and it didn't seem possible. Don't know if I did it the best way possible, but it seems so.

The main reason (at least for me) for having this value available is to aggregate values based on each endpoint in metric systems (Microsoft Application Insights, New Relic, Datadog...).

Please let me know if I need to change something or if this is not desirable!
Thanks!


victoraugustolls:
Hi @tomchristie, would you mind helping me here, if possible?

starlette/routing.py
```diff
@@ -196,7 +196,12 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
                     matched_params[key] = self.param_convertors[key].convert(value)
                 path_params = dict(scope.get("path_params", {}))
                 path_params.update(matched_params)
-                child_scope = {"endpoint": self.endpoint, "path_params": path_params}
+                child_scope = {
+                    "endpoint": self.endpoint,
+                    "path_params": path_params,
+                    "route": self.path,
```
lovelydinosaur: I'd probably rather we were including the `Route` itself, since there's other information on there too that the developer might potentially want access to.

lovelydinosaur:
Needs some thinking about:

Do we need to track any Mount that might have been matched too? Probably, otherwise eg. you can't actually build up the matched path. Perhaps we need eg. `scope["routes"] = <list of Mounts and Route instances>`

victoraugustolls:
Hmm, and inside each of the matching routes can have a filed saying if it was a full match or partial match?

victoraugustolls:
@tomchristie I added the routes to the scope as a draft to what you purposed. If you like this solution, I would gladly add tests. I maintained the `child_scope` so that anyone using it won't have any problems, and this wouldn't become a breaking change.

starlette/routing.py
```diff
@@ -541,17 +544,20 @@ def url_path_for(self, name: str, **path_params: str) -> URLPath:
 
         partial = None
 
+        scope["routes"] = {"full": None, "partial": []}
```
inikolaev: I wonder what would be the use case to have `full` or `partial` match in here? I have a use case similar to yours and what I'm after is the path of the route which was matched and used by router - I'm not really interested whether that was a full or partial match.

victoraugustolls: I’m in the same place as you. I only need the full match, but the partial ones might be need by someone for debug purposes!

victoraugustolls:
@tomchristie what are your thoughts here? Can I proceed writing tests with this solution? Or should I change something?

starlette/routing.py
inikolaev: What is you simply add `route_path` which would only contain the path of the selected router regardless whether it was partial or full match?

victoraugustolls: In the case of the partial path, there can be multiple matches. By doing this way I think it gives more versatility towards multiple use cases in the future!

inikolaev: I think that for both partial and full match the resource name (route path) in the same, according to how it's defined - same `path`, but different `method`, so there should not be multiple route paths.

Having matching routes might be useful though, but it makes it a bit harder to retrieve route path - you need to have enough knowledge about the underlying ASGI framework to do that.

I'm just trying to think what would be a more standard way of doing this across different frameworks. We are using a library which implements tracing middleware for ASGI applications and it's framework agnostic. Right now it has to have knowledge about Starlette in order to retrieve route path, but it would rather not. 

Would be nice if it was part of the ASGI spec :)

victoraugustolls: AFAIK ASGI has no common way of setting this value, is up to the framework to do it. But I might be wrong, is so, please post a reference here so I can read about it! But if that’s the case, we have no better way to do it. And I still think that this way gives more flexibility. Although I would be ok to change it if requested by the owners.

FYI my first solution only added the matched route, this was a requested change so we agree here!

victoraugustolls:
@tomchristie any updates here?

lovelydinosaur:
I'd rather we don't include `{"full": None, "partial": []}` here.

The `routes` entry ought to just be a list of `BaseRoute` instances, which will generally be zero or more `Mount` instances, and then a final `Route` instance.

victoraugustolls:
@tomchristie updated to what I understood about you comment!

victoraugustolls:
Hi @tomchristie and @florimondmanca! I see this pr was mentioned in many places!

Can we get this moving? I will update the PR when I wake up (it's pretty late here) but would like to know if you guys are ok with this approach!

As I see, with this list of routes, the user would need to iterate over the list and join the path values for the Mounts and finally the Route, is that right?

Also, if that's the case, should we save the final full route value too?

victoraugustolls:
PR updated @tomchristie @florimondmanca !

victoraugustolls:
Tagging @euri10 and @JayH5 also as I saw them doing the reviews in other PR's!

Sorry if this is too much.

starlette/routing.py
```diff
@@ -557,17 +560,20 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         partial = None
 
+        scope["routes"] = []
```
florimondmanca, line 563: Hmm, perhaps I'm misunderstanding the aim of this — assuming we want `routes` to contain all `BaseRoute` items (mounts, routes) that have matched while going through the routing system, it seems that currently it will only ever contain a single item, i.e. the first one that matched? In case of a route under a `Mount`-ed app, it will be the `Mount` (even if the mounted app is actually a Starlette app perhaps with its own router), and in case of a regular route then it will be that `Route`. Correct?

florimondmanca: Basing my judgement off https://github.com/encode/starlette/pull/804#issuecomment-590261001 here - Not entirely sure that's something that's actually possible at all, given the ASGI-interoperable design aspects of Starlette.

For example, if you have…

```python
async def other_non_starlette_app(scope, receive, send):
    ...

app = Starlette(routes=[Mount("/subapp", other_non_starlette_app)])
```

Then in this example context using `scope["routes"][-1]` as a way to get the "route that actually responds to the request" might be a bit confusing — it's only guaranteed to be the _last route within this app_ that handles the request _at least partly_.

Don't know if we want to play on words here, or if for the use case you're describing (manipulating route information for telemetry purposes) having partial route info in some cases and marking it as a limitation would be enough?

JayH5:
I think I understand the need to get the route used for a request, but I'm a bit unclear on exactly how the routes would be used, especially when there are multiple routes. Maybe some test cases would make that clearer?

starlette/routing.py
victoraugustolls: > it seems that currently it will only ever contain a single item, i.e. the first one that matched?

Yes, you are right! I didn't proceed with the PR because wanted to discuss the best strategy with the maintainers before and wasn't sure this was it.

victoraugustolls:
> I think I understand the need to get the route used for a request, but I'm a bit unclear on exactly how the routes would be used, especially when there are multiple routes. Maybe some test cases would make that clearer?

What do you mean by multiple routes? Are you referring to having multiple Mounts and then finally a Route?

I didn't write test cases because I wanted to discuss the best strategy to implement this before doing so.

In my mind the ideal is getting a single value from scope that represents the full route. This means of the user have sub apps, sub routers and etc, the value would be there with the full route.

But at the same type, there is a problem here. As routes are only matched after middlewares,and I think this would be the most common place to use the route value, thinking of a way to add middlewares after routes were matched, would be a nice addition.

victoraugustolls:
Just and example of what I do in a middleware today to get this value (it could be improved):

```python
route = None
for r in request.app.router.routes:
    if r.matches(request.scope)[0] == Match.FULL:
        route = r
        break

complete_route = ""
while isinstance(route, Mount):
    complete_route += route.path
    scope = dict(request.scope)
    scope["path"] = scope.get("path").replace(route.path, "")
    matched = False
    for r in route.routes:
        if r.matches(scope)[0] == Match.FULL:
            route = r
            matched = True
            break

    if matched is False:
        route = None

complete_route += (
    route.path
    if route is not None
    else request.url.path
)
```

starlette/routing.py
victoraugustolls: > Not entirely sure that's something that's actually possible at all, given the ASGI-interoperable design aspects of Starlette.

An example would be subapps on a FastAPI app? If so, I posted some code on a comment here that works by iterating the routes and searching for a match, but I don't think the problem you described would happen with FastAPI :/

cdvv7788:
Hi. Currently running into this issue. Any progress on this? Something that needs help?
About the discussion, instead of returning a list of routes, a single string that contains the whole original route would suffice most of the use cases here.

nhtgl:
Hi @victoraugustolls 👋 Can you share the details behind closing this PR? Could be useful in https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1759/. Currently [OpenTelemetry handle this manually](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-fastapi/src/opentelemetry/instrumentation/fastapi/__init__.py#L318-L337).

leameow:
This feature would be useful for any observability tools. Right now it seems that every library have their own hack.

jahanvi-kathiriya:
why was this pull closed?

PR #805, opened 2020-01-16
Add "raw_path" to TestClient request scope

trimailov:
`raw_path` was missing in `TestClient` request's scope, whereas `uvicorn` has `raw_path` in request's scope. This problem makes it impossible to write tests, when parsing raw path is necessary.

I've created this app to prove this:
```
> cat example.py


async def value(request):
    value = request.path_params['value']
    path = request.scope['path']
    raw_path = request.scope['raw_path']
    return PlainTextResponse(
        f'value: {value}\npath: {path}\nraw_path: {raw_path}\n'
    )


app = Starlette(debug=True, routes=[
    Route('/{value:path}', value),
])


def test_app():
    client = TestClient(app)
    response = client.get('/hel%2Flo')  # /hel/lo
    result = "value: hel/lo\npath: /hel/lo\nraw_path: /hel%2Flo\n"
    assert response.text == result
```

Launch the app with `uvicorn`:

```
> env/bin/uvicorn example:app
```

Test it with `curl`:

```
> curl localhost:8000/hel%2flo
value: hel/lo
path: /hel/lo
raw_path: b'/hel%2flo'
```

Test app with `TestClient`:

```
> env/bin/python -c "import example; example.test_app()"
  File "<string>", line 1, in <module>
  File "/Users/justas/work/tilaajavastuu/starlette_test/example.py", line 28, in test_app
    response = client.get('/hel%2Flo')  # /hel/lo
<...>
  File "/Users/justas/work/tilaajavastuu/starlette_test/example.py", line 14, in value
    raw_path = request.scope['raw_path']
KeyError: 'raw_path'
```

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -134,7 +134,7 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
             scope = {
                 "type": "websocket",
                 "path": unquote(path),
-                "raw_path": path,
+                "raw_path": path.encode(),
                 "root_path": self.root_path,
                 "scheme": scheme,
                 "query_string": query.encode(),
@@ -151,7 +151,7 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
             "http_version": "1.1",
             "method": request.method,
             "path": unquote(path),
-            "raw_path": path,
+            "raw_path": path.encode(),
             "root_path": self.root_path,
             "scheme": scheme,
             "query_string": query.encode(),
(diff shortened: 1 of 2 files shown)


trimailov:
Apparently similar PR was made in #635.

starlette/testclient.py
```diff
@@ -134,6 +134,7 @@ def send(
             scope = {
                 "type": "websocket",
                 "path": unquote(path),
+                "raw_path": path,
```
lovelydinosaur: The ASGI spec mandates this as bytes, so `path.encode()` I guess?

trimailov: You are correct. Fixed it.

Kludex:
@trimailov Thanks for the PR! :tada: 

Sorry for taking so long! Would you mind rebasing it?

Kludex:
Closed in favor of #1445.

Author's commit are preserved.

PR #820, opened 2020-02-07
Add Starlette.create_router method

TiemoKieft:
This makes it easier to change the `Router` class used by `Starlette`. The specific use case is that I want to be able to use a dependency injection framework by overriding `Router.handle` in a custom router.

The current solution is to create a custom `Starlette` and copy-paste the constructor, which is not the best solution. This would make it much robuster.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -48,12 +48,20 @@ class Starlette:
     ) -> None:
         self._debug = debug
         self.state = State()
-        self.router = Router(routes, on_startup=on_startup, on_shutdown=on_shutdown)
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Thanks, but not a piece of public API that I want us to introduce.

alex-oleshkevich:
@tomchristie maybe it makes sense to pass the `router_class` constructor argument?

lovelydinosaur:
I probably don't want us adding *any* API to Starlette at the moment. I'm going to eventually be able to get around to coming back to directing things here, but haven't got the time just yet.

PR #821, opened 2020-02-07
Fix naive debugger line numbers in ServerErrorMiddleware

retnikt:
Previously the error middleware in debug mode would naively calculate the line numbers without knowledge of the position in the file. This meant that if the exception occurred near the top or bottom of the file, the line numbers and the highlighted line would be incorrect. The named tuple returned by `inspect.getinnerframes` contains an `index` field which is not naive and will result in correct line numbers. This change also simplifies the code somewhat.

Below is a short demonstration of the problem:

```python
def error(): return 4/0                                 # 1
import uvicorn                                          # 2
from starlette.applications import Starlette            # 3
                                                        # 4
app = Starlette(debug=True)                             # 5
                                                        # 6
@app.route("/")                                         # 7
async def route(_):                                     # 8
        error()                                         # 9
                                                        # 10
uvicorn.run(app, port=2351, debug=True)                 # 11
```
screenshot

I didn't open an issue for this since I thought it's a pretty simple bug-fix. I also haven't added any tests - I'm not sure they're really necessary, but correct me if I'm wrong.

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -178,23 +178,21 @@ class ServerErrorMiddleware:
             raise exc from None
 
     def format_line(
-        self, position: int, line: str, frame_lineno: int, center_lineno: int
+        self, index: int, line: str, frame_lineno: int, frame_index: int
     ) -> str:
         values = {
             "line": line.replace(" ", "&nbsp"),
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Great stuff, thanks!

PR #825, opened 2020-02-10
More debugger improvements

retnikt:
While working on #821 I noticed several things that could be improved in the debugger:
- more HTML escaping (see also #599, #601)
- code is monospaced


lovelydinosaur:
Thanks!

This PR is making me consider if we should switch the implementation over to just using Jinja, and having a plain text variant if that's not available.

retnikt:
I did wonder... but having practically no concrete requirements is a pretty good selling point

PR #832, opened 2020-02-19
Link to black GitHub repository

Colin-b:
Instead of using "don't be evil" search engine


lovelydinosaur:
Ha, looks like that kinda ended up in there by accident.
Thanks!

PR #833, opened 2020-02-19
Ensure pyyaml is installed before actually using it

Colin-b:
To avoid non user-friendly error message (NoneType does not have attribute whatever)

### DIFF
--- a/starlette/schemas.py
+++ b/starlette/schemas.py
@@ -89,6 +89,8 @@ class BaseSchemaGenerator:
         if not docstring:
             return {}
 
+        assert yaml is not None, "`pyyaml` must be installed to use parse_docstring."
+
         # We support having regular docstrings before the schema
         # definition. Here we return just the schema part from
         # the docstring.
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Would it be reasonable to move the assertion check into `SchemaGenerator.__init__` instead?

I personally don't think that we need unit tests for "does this assertion fire" - the cost of adding that extra code in the codebase doesn't seem to justify being overly protective.

Colin-b:
I don't use SchemaGenerator but if you are ok with the creation of the dunder init method within BaseSchemaGenerator I am willing to do it.

Do you want me to remove the test case as well then ?

lovelydinosaur:
> but if you are ok with the creation of the dunder init method within BaseSchemaGenerator I am willing to do it.

Okay probably best not then. Might be some folks using `BaseSchemaGenerator`, but using a different (overridden) behavior for the docstrings.

In that case I'd suggest leaving the check where it currently is, but dropping the test case.

Colin-b:
> In that case I'd suggest leaving the check where it currently is, but dropping the test case.

Done

lovelydinosaur:
Great, thanks!

PR #835, opened 2020-02-19
Add check on content being falsy before dumping the content into JSON / UJSON dumps

NoutieH:
json.dumps(None) results to "null" and encodes to b"null" which is not the correct return value.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -143,6 +143,8 @@ class JSONResponse(Response):
     media_type = "application/json"
 
     def render(self, content: typing.Any) -> bytes:
+        if not content:
+            return b""
         return json.dumps(
             content,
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
> json.dumps(None) results to "null" and encodes to b"null" which is not the correct return value.

That's open to intpretation. I'd probably suggest *not* return `JSONResponse` if you want to return an empty response body. (Eg. maybe use `return Response()` in that case.)

NoutieH:
I considered your option as well, but finally thought: What should an empty JSONResponse be responding into? and proposed the above because I thought it would improve the package. Nevertheless I am not too happy with the tiny bit of performance loss for each valid json response in the proposed solution

Issue #20, opened 2018-07-12
Server Sent Events


lovelydinosaur:
Helpers for sending SSE event streams over HTTP connections.

Related resources:

* https://github.com/encode/uvicorn/commit/07c6c18945ab4f4167118e717e197adce3aee3da#r29630623
* https://github.com/aio-libs/aiohttp-sse

kesavkolla:
Any update on this?  It would be really helpful for building realtime applications

lovelydinosaur:
Not yet - tho I'd be happy to help guide anyone who's interested in taking on a pull request for it.

Kamforka:
I'm trying to do a simple working example for server side events in `starlette` without luck, maybe if  you could check what's wrong with my logic I could refine it, and then raise a PR for a new `EventResponse` class.

This is my code:
```python
from asyncio.queues import Queue

import uvicorn
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import JSONResponse, StreamingResponse


class SSE:
    def __init__(self, data, event=None, event_id=None, retry=None):
        self.data = data
        self.event = event
        self.id = event_id
        self.retry = retry

    def encode(self):
        message = f"data: {self.data}"
        if self.event is not None:
            message += f"\nevent: {self.event}"
        if self.id is not None:
            message += f"\nid: {self.id}"
        if self.retry is not None:
            message += f"\nretry: {self.retry}"
        message += "\r\n\r\n"
        return message.encode("utf-8")


app = Starlette(debug=True)
app.queues = []


@app.route("/subscribe", methods=["GET"])
async def subscribe(request: Request):
    async def event_publisher():
        while True:
            event = await queue.get()
            yield event.encode()

    queue = Queue()
    app.queues.append(queue)

    headers = {
        "Content-Type": "text/event-stream",
        "Cache-Control": "no-cache",
        "Connection": "keep-alive",
        "X-Accel-Buffering": "no",
    }

    return StreamingResponse(content=event_publisher(), headers=headers)


@app.route("/publish", methods=["POST"])
async def publish(request: Request):
    payload = await request.json()
    data = payload["data"]
    for queue in app.queues:
        event = SSE(data)
        await queue.put(event)

    return JSONResponse({"message": "ok"})


if __name__ == "__main__":
    uvicorn.run("__main__:app", host="0.0.0.0", port=4321, reload=True)
```

[...]

PR #839, opened 2020-02-20
Add listener task to cancel streaming response in case of an early disconnect

Kamforka:
Related to #20

This one is to fix an ongoing stream to an already disconnected client.

All the existing tests are passed, however I'd like to add one to cover a scenario when a `"http.disconnect"` message is sent via the `receive` channel, @tomchristie maybe you could suggest some sort of setup for such a test?

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -1,4 +1,3 @@
-import asyncio
 import hashlib
 import http.cookies
 import inspect
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -11,7 +11,7 @@ from mimetypes import guess_type
 from urllib.parse import quote, quote_plus
 
 from starlette.background import BackgroundTask
-from starlette.concurrency import iterate_in_threadpool
+from starlette.concurrency import iterate_in_threadpool, run_until_first_complete
 from starlette.datastructures import URL, MutableHeaders
 from starlette.types import Receive, Scope, Send
 
@@ -209,13 +209,10 @@ class StreamingResponse(Response):
         await send({"type": "http.response.body", "body": b"", "more_body": False})
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        done, pending = await asyncio.wait(
-            [self.stream_response(send), self.listen_for_disconnect(receive)],
-            return_when=asyncio.FIRST_COMPLETED,
+        await run_until_first_complete(
+            (self.stream_response, {"send": send}),
+            (self.listen_for_disconnect, {"receive": receive}),
         )
-
-        for task in pending:
-            task.cancel()
 
         if self.background is not None:
             await self.background()
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -1,3 +1,4 @@
+import asyncio
 import hashlib
 import http.cookies
 import inspect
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Okay, this is looking decent yup!

I happened to inroduce a [`run_until_first_complete`](https://github.com/encode/starlette/blob/dd764eced233df9356fa3a0a04a2754263fc54c8/starlette/concurrency.py#L14) since I was looking at [a similar pattern for working with websockets](https://github.com/encode/broadcaster/blob/3e8a8bc1dca9c31c33562e8a8802aa089e7688f7/example.py#L20).

I reckon we ought to use that here, and have `.listen_for_disconnect` and `.stream_response_body` methods. That'd cover us in any wierd broken cases where disconnect is never sent for some reason.

Kamforka:
@tomchristie Sure, I'll adjust the code this evening.

Just to be clear, you meant something like this:

```python3
await run_until_first_complete(self.stream_response(send), self.listen_for_disconnect(receive))
```

Right?

lovelydinosaur:
@Kamforka Yup, I'll be along these lines...

```python
await run_until_first_complete(
    (self.stream_response, {"send": send}),
    (self.listen_for_disconnect, {"receive": receive})
)

**Comment by Kamforka** [comment]
@tomchristie added the changes.

However I have one observation. Is there any particular reason for the signature of `run_until_first_complete`?

```python3
async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None
```

I think the following signature would be more intuitive:
```python3
async def run_until_first_complete(tasks: typing.Sequence[typing.Awaitable]) -> None:
```
> Note: `typing.Awaitable` might not be the best type here but you get my point.

So in our particular case one could call it like:
```python3
await run_until_first_complete(
    [self.stream_response(send), self.listen_for_disconnect(receive)]
)
```

As far as I can tell the functionality is the same, however we get less boilerplate with better typing and auto-completion experience. What do you think?

starlette/responses.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
```
lovelydinosaur: Looks like we can drop this now, right?

Kamforka: Oh yeah

lovelydinosaur:
Good suggestion yup, although I’m also keeping half an eye here on trio compatibility, which it wouldn’t play so well with.

Kamforka:
What do you mean by trio compatibility?

lovelydinosaur:
> What do you mean by trio compatibility?

We might want a future version of Starlette to support being run on Trio https://trio.readthedocs.io/en/stable/ as an alternative to Asyncio. The API that Trio uses for starting tasks doesn't allow passing instantiated coroutines (for some good reasons) so keeping this constraint in our own codebase would make it easier for us support Trio at some point.

lovelydinosaur:
And, great stuff, thank you!

Kamforka:
> 
> 
> > What do you mean by trio compatibility?
> 
> We might want a future version of Starlette to support being run on Trio https://trio.readthedocs.io/en/stable/ as an alternative to Asyncio. The API that Trio uses for starting tasks doesn't allow passing instantiated coroutines (for some good reasons) so keeping this constraint in our own codebase would make it easier for us support Trio at some point.

Got it, thanks for the clarification!

PR #841, opened 2020-02-20
adding authlib middleware

aogier:
Hello! I'm writing a Starlette-based SSO application that will protect a series of applications written in many languages.  
As such, I've found using authlib's jwt more viable than itsdangerous mainly because developers can consume my claims/data using JWT libraries that already exists for many languages. I'd be glad if you could consider to include my third party middleware to your manual, this PR add a link to the project. What do you think?

Thank you for starlette, and have a nice day! :)


lovelydinosaur (approved):
Fantastic! 👍

Issue #847, opened 2020-02-27
Middleware Request parse Hangs forever


nikordaris:
I'm currently trying to integrate some metrics tooling in a middleware where I need to inspect the request body. The problem I am seeing is that this triggers the request to hang forever. The following is how you can reproduce it.

```py
app = Starlette()
class ParseRequestMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint):
        # payload = await request.json()  # This will cause test() to hang
        response = await call_next(request)
        payload = await request.json()  # Hangs forever
        return response


async def test(scope: Scope, receive: Receive, send: Send):
    request = Request(scope=scope, receive=receive)
    data = await request.json()
    response = Response(status_code=405)
    await response(scope, receive, send)


app.mount("/test", test)
app.add_middleware(ParseRequestMiddleware)
```

I'm still pretty new to starlette but it seems to be an issue with the fact that the middleware and ASGI app use different Request objects and therefore can't share the Request._body cache. Which means they exhaust the stream on the first read and the second read isn't correctly detecting that the stream is exhausted and so gets stuck in that [`while True`](https://github.com/encode/starlette/blob/master/starlette/requests.py#L178) loop. I'm not sure I haven't dove deeper than the Request object. I'm not sure where the code is for the `receive` function that gets passed into the ASGI App is.

nikordaris:
So i confirmed that if I use the same Request object it works. I added a middleware to store the request object into a ContextVar and if I use that instead of the `receive` and `send` passed into the ASGI app then the request doesn't hang anymore. This might be enough of a workaround for me right now but the more apps I add (especially third party ones) the more of a problem this will become.

Update:
Running more tests today I had hoped to reduce the number of places I have to ignore the send/receive args but it looks like [BaseHTTPMiddleware](https://github.com/encode/starlette/blob/master/starlette/middleware/base.py#L24) also creates a new Request object. So basically anyone who needs to share the request body data should get the request object from ContextVars to ensure the stream is only read once and doesn't get exhausted.

I'm still unclear whether this is all by design and the developer is expected to manage a cache of the data streams. It seems like at minimum it shouldn't hang and it should be raising an error instead of getting stuck in an infinite loop. I would greatly appreciate some insights here whenever you have a moment @tomchristie

nikordaris:
Ok I think I found a solution here. See PR #848

nikordaris:
[...]

PR #848, opened 2020-02-27
Added request body cache to Request.receive

nikordaris:
Adds a request body cache to the receive function that is passed to ASGIApp children. This addresses the root cause of an infinite loop issue described in #847 but does not address the infinite loop itself. There is still a potential need to set a timeout around calling the receive() inside of stream()


nikordaris:
This is no longer a viable solution with the changes to StreamResponse in v0.13.3. I have an idea of how to cache the body and form in the scope instead of the instance of Request and keep the receive function pure. I'll open a different PR for this solution.

PR #862, opened 2020-03-15
More detail on how to use convertors

simonw:
It wasn't immediately clear to me how to use them.


lovelydinosaur:
Great, thanks Simon!

PR #867, opened 2020-03-18
Ignore malformed cookie

### LINKED ISSUE #866: [bug] Invalid cookie name leads to exception
When handling a request with an invalid cookie name (does not conform to RFC2109) starlette raises an exception. i.e iam/cookiename
This is because Starlette uses Python's stdlib cookie library, which is very strict.
I do understand the strictness, but in real life scenarios you receive such malformed cookies and I want to handle those requests.
My suggestion for a solution would be to catch those exceptions and ignore the invalid cookie. 
** EDIT **
I just realized stdlib is used for the whole cookie header, hence can't ignore only one cookie.
I'll create a PR for ignoring the whole cookie on such case, but maybe we should create our own Cookie/Morsel class and override the methods to ignore such error in the inbound case?

aviramha:
Fixes #866

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -285,6 +285,23 @@ def test_request_cookies():
     assert response.text == "Hello, cookies!"
 
 
+def test_invalid_cookie():
+    async def app(scope, receive, send):
+        request = Request(scope, receive)
+        if not request.cookies:
+            response = Response("ok", media_type="text/plain")
+        else:
+            response = Response("not", media_type="text/plain")
+        await response(scope, receive, send)
+
+    client = TestClient(app)
+    response = client.get("/", cookies={"invalid/cookie": "test", "valid": "test2"})
+    assert response.text == "ok"
+
+    response = client.get("/", cookies={"valid": "test2"})
+    assert response.text == "not"
+
+
 def test_chunked_encoding():
     async def app(scope, receive, send):
         request = Request(scope, receive)
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -288,18 +288,12 @@ def test_request_cookies():
 def test_invalid_cookie():
     async def app(scope, receive, send):
         request = Request(scope, receive)
-        if not request.cookies:
-            response = Response("ok", media_type="text/plain")
-        else:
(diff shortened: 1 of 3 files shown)


tests/test_requests.py
```diff
@@ -285,6 +285,23 @@ def test_request_cookies():
...
+        else:
+            response = Response("not", media_type="text/plain")
+        await response(scope, receive, send)
+
+    client = TestClient(app)
+    response = client.get("/", cookies={"invalid/cookie": "test", "valid": "test2"})
+    assert response.text == "ok"
+
+    response = client.get("/", cookies={"valid": "test2"})
+    assert response.text == "not"
```
florimondmanca: Nits:

- The fact that "ok" corresponds to "cookies are invalid" is a bit counter-intuitive to me.
- We already test the valid case, so we can probably focus this test on "what happens if we pass invalid cookies?".

For example…

```suggestion
def test_invalid_cookie():
    async def app(scope, receive, send):
        request = Request(scope, receive)
        response = JSONResponse({"cookies": request.cookies})
        await response(scope, receive, send)

    client = TestClient(app)
    response = client.get("/", cookies={"invalid/cookie": "test", "valid": "test2"})
    assert response.json() == {"cookies": {}}
```

florimondmanca (commented):
Nice little improvement!

Left a small suggestion.

I'm wondering if a next step might be to only drop cookies that are invalid — would require switching manual parsing of cookies, similar to [what Django does](https://github.com/django/django/blob/master/django/http/cookie.py) (it used to use `SimpleCookie` too). But it can be an incremental improvement — the nice thing here is that we won't be failing with an exception anymore.

aviramha:
Thanks for the comments @florimondmanca ! 
Initially I wanted to change the usage of SimpleCookie but thought it's a big change.
If we're up for it I'll be more than happy to add it.

florimondmanca (approved):
Thanks!

florimondmanca:
@aviramha Can you maybe draft a separate PR with those changes on top of the ones here, so we can see how big of a chunk that would be?

aviramha:
@tomchristie What do you think?

gbgduh (approved):

PR #868, opened 2020-03-18
[DRAFT] Implement RobustCookie ignoring invalid cookies.

aviramha:
@florimondmanca  #867 #866

### DIFF
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -73,6 +73,8 @@ Cookies are exposed as a regular dictionary interface.
 
 For example: `request.cookies.get('mycookie')`
 
+Cookies are ignored in case of an invalid cookie. (RFC2109)
+
 #### Body
 
 There are a few different interfaces for returning the body of the request:
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -1,10 +1,10 @@
 import tempfile
 import typing
+import http.cookies
 from collections import namedtuple
 from collections.abc import Sequence
 from shlex import shlex
 from urllib.parse import SplitResult, parse_qsl, urlencode, urlsplit
-
 from starlette.concurrency import run_in_threadpool
 from starlette.types import Scope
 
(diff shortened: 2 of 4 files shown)


aviramha:
I'm trying to push this to Python's upstream, if it gets accepted it'll probably reach us very late, but anyway:
https://github.com/python/cpython/pull/19058

florimondmanca:
Thanks @aviramha! I'm not sure about dragging an entire (almost-)re-implementation of `SimpleCookie` in here though…

Could we get away with a simplified `parse_header(str) -> dict` function, dropping any intermediate `Cookie`-like interface, similar to what Django does?

https://github.com/django/django/blob/fc0fa72ff4cdbf5861a366e31cb8bbacd44da22d/django/http/cookie.py#L10-L26

aviramha:
> Thanks @aviramha! I'm not sure about dragging an entire (almost-)re-implementation of `SimpleCookie` in here though…
> 
> Could we get away with a simplified `parse_header(str) -> dict` function, dropping any intermediate `Cookie`-like interface, similar to what Django does?
> 
> https://github.com/django/django/blob/fc0fa72ff4cdbf5861a366e31cb8bbacd44da22d/django/http/cookie.py#L10-L26

I disliked my implementation also, but I wanted to keep the usage of SimpleCookie.
I'll push an update soon.

aviramha:
@florimondmanca I've checked the Django implementation and it's pretty flawful - for example if you try parsing this using SimpleCookie vs Django you get very different results:
`asd=";more_cookie=1;;;lol";poop=1`
I think sticking to stdlib's implementation is best, as it's battle driven and parsing issues -> security issues.

florimondmanca:
Okay, thanks.

Still not convinced this is worth it; being able to read _some_ cookies (ie all except invalid ones) seems like a pretty specific use case, at least not as critical as preventing server crashes on invalid cookies. (Edit: reworded.) WDYT?

aviramha:
> Okay, thanks.
> 
> Still not convinced this is worth being able to read _some_ cookies (ie all except invalid ones). Not as enticing of a use case than avoiding server crashes due to failing on parsing. WDYT?

I tend to agree.

PR #870, opened 2020-03-20
Add ORJSONResponse

hidaris:
Add ORJSONResponse, including docs and tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ Starlette does not have any hard dependencies, but the following are optional:
 * [`pyyaml`][pyyaml] - Required for `SchemaGenerator` support.
 * [`graphene`][graphene] - Required for `GraphQLApp` support.
 * [`ujson`][ujson] - Required if you want to use `UJSONResponse`.
+* [`orjson`][orjson] - Required if you want to use `ORJSONResponse`.
 
 You can install all of these with `pip3 install starlette[full]`.
 
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -112,6 +112,25 @@ async def app(scope, receive, send):
     await response(scope, receive, send)
 ```
 
+### ORJSONResponse
+
+Another JSON response class that uses the optimised `orjson` library for serialisation.
+
+`orjson` is a fast, correct JSON library for Python. It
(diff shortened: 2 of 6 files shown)


lithammer:
Sorry for the drive-by comment. Maybe it's just me, but I don't think flavour of the month JSON libraries belong in Starlette core, this applies to `UJSONResponse` as well. Next up someone wants a `SIMDJSONResponse` class[1]. Where should the line be drawn?

Another aspect to consider is the maintenance burden of implementing custom response classes like this in core, because that means a contract has been signed. So if orjson becomes unmaintained tomorrow, Starlette would have to go through a deprecation cycle to remove it. Or keep it forever... Bugs in the renderer(s) would likely end up as a Starlette issue as well, instead of in the upstream JSON library.

But enough complaining and on to more constructive feedback. It's evident by this pull request that it's pretty trivial to implement your own custom `JSONResponse` class, so maybe just a section in the documentation describing how to do it is enough? Then someone can (and probably will) publish a package to PyPI that Starlette later can link to, e.g. starlette-orjson. Maybe even a starlette-contrib repo under the encode org? 🤷‍♂️ 

Another option is to make the default `JSONResponse` more pluggable/configurable, similar to how you can configure renderers in Django REST Framework[2]. Personally, I like this approach.

Obviously this is just my random opinion in the matter, so feel free to ignore it 🙂 

[1] https://github.com/TkTech/pysimdjson
[2] https://www.django-rest-framework.org/api-guide/settings/#default_parser_classes

erewok:
I agree with @lithammer on this completely. We use RapidJSON on our projects, but I have no desire to see it directly incorporated into Starlette.

lithammer:
FYI I created an issue related to this after posting my comment: https://github.com/encode/starlette/issues/901

abersheeran:
I agree with @lithammer @erewok on this completely.

By the way, maybe `UJSONResponse` should also be removed.

It is up to the users to decide whether to use a high-performance JSON parser. All we need to do is give them a good interface for easy customization.

lithammer:
> By the way, maybe `UJSONResponse` should also be removed.

See https://github.com/encode/starlette/issues/901.

Issue #715, opened 2019-11-13
Using Custom JSON Encoders and Decoders?


erewok:
Hello,

I have recently discovered Starlette and am about to deploy a couple of small services using it and I have to say that I am pretty excited about this framework. I've been looking for the simple functionality provided by `BackgroundTask` for awhile now. Thanks so much for all the work that has gone into this project!

I am writing this issue because one of the things I typically do with HTTP APIs returning JSON is swap out the JSON libraries used for my own selection. I don't typically do this for performance (although it's a nice side benefit) so much as to control the ways in which JSON encoding and decoding happens. For example, I find that I'm dealing with UUIDs and `datetime`s a lot and it's nice to have them automatically encoded and decoded. I also tend to use [rapidjson](https://github.com/python-rapidjson/python-rapidjson) because it's quick and configurable (and it seems to see more development activity than [ujson](https://github.com/esnme/ultrajson), which hasn't been updated in 3 years...)

Maybe an example would be helpful. In Starlette, I can construct a `Response` class like this:

```python
import json
import typing

import rapidjson
from starlette.responses import JSONResponse


class CustomJSONEncoder(json.JSONEncoder):
    encode = rapidjson.Encoder(
        skip_invalid_keys=False,
        ensure_ascii=False,
        indent=None,
        sort_keys=False,
        number_mode=rapidjson.NM_NATIVE,
        datetime_mode=rapidjson.DM_ISO8601,
        uuid_mode=rapidjson.UM_CANONICAL)


class RapidJSONResponse(JSONResponse):
    media_type = "application/json"

    def render(self, content: typing.Any) -> bytes:
        return CustomJSONEncoder().encode(content).encode("utf-8")
```

And then, in a handler, I can use it like this:

```python
async def index(request):
    conf = request.app.settings
    return RapidJSONResponse(
        {
            "status": http_helpers.Success.status,
            "version": conf.version,
            "test_uuid": uuid.uuid4(),
            "test_datetime": datetime.datetime.utcnow(),
            "message": f"Welcome to {conf.name}",
        }
    )
```

And everything just works:

```sh
In [1]: import requests

In [2]: requests.get("http://localhost:8000/").json()
Out[2]:
{'status': 'success',
 'version': 'local-dev',
 'test_uuid': 'c774a5d5-26c8-4292-9d42-a4e77dcd8a50',
 'test_datetime': '2019-11-13T05:10:21.610705',
 'message': 'Welcome to Starlette!'}
```

[...]

PR #875, opened 2020-03-24
Customizable request class

alex-oleshkevich:
This allows users to change the default `Request` class.

There are following customization options:
1. Set `request_class` on Route constructor:
```python
Route('/', endpoint, request_class=MyRequest)
```
This has higher precedence than `request_class` defined in HTTPEndpoint class.

2. Set `request_class` in HTTPEndpoint:
```python
class MyEndpoint(HTTPEndpoint):
    request_class = MyRequest

    def get(self, request: MyRequest): ...
```

### DIFF
--- a/docs/endpoints.md
+++ b/docs/endpoints.md
@@ -48,6 +48,20 @@ app = Starlette(routes=routes)
 HTTP endpoint classes will respond with "405 Method not allowed" responses for any
 request methods which do not map to a corresponding handler.
 
+
+#### Custom request class
+
+You can set `request_class` attribute to change default request class:
+
+```python
+class Homepage(HTTPEndpoint):
+    request_class = MyRequest
+
+    async def get(self, request):
+        return PlainTextResponse(f"Hello, world!")
+```
+
+
 ### WebSocketEndpoint
 
 The `WebSocketEndpoint` class is an ASGI application that presents a wrapper around
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -216,3 +216,15 @@ app = Router(routes=[
     ])
 ])
 ```
+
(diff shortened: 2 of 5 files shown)


florimondmanca:
Dropping by out of curiosity — could you explicit some practical use cases that would motivate the ability to swap the request class? Maybe this is linked to an existing issue?

em92:
> could you explicit some practical use cases that would motivate the ability to swap the request class?

I would store this to `MyRequest` constructor. https://github.com/em92/quakelive-local-ratings/blob/879c354c4f8d25965c93f6a1c84e68c48f1aabb4/qllr/endpoints.py#L53-L68

alex-oleshkevich:
I am building an application that has a custom Request class. This class has extra methods related to it (input validation, helpers to get values from GET/POST,  user IP, and many other helpers). 

In order to make use of it, I have to completely extend Route/HTTPEndpoint/request_response names. It is a huge overhead here.

Also, it is useful for frameworks built on top of Starlette. They often have own request classes (look at HttpRequest in Django and Request in DRF).

erewok:
I like this idea. I am in favor of figuring out how to add it. It would also close this issue I opened awhile ago related to having custom JSON decoders: https://github.com/encode/starlette/issues/715.

Kludex:
@alex-oleshkevich After this whole time, do you still want this feature? Is it something you'd want to discuss?

alex-oleshkevich:
Yes, i use custom request time each time. My class has some extra properties and utilities.

alex-oleshkevich:
There can be a workaround with a simple decorator that inspects view signature and injects a needed request instance.

Issue #619, opened 2019-08-27
GraphQL support should be a third party application.


lovelydinosaur:
I'd like our graphql support to move out into being a third party package.

We should then document:

* The third party package it's been moved to.
* Links to ASGI implementations with ariadne, tartiflette, strawberry.

Questions:

* Which of Ariadne, Tartiflette & Strawberry currently have provide an ASGI application, that Starlette can mount?
* Are there any other GraphQL implementations I've missed here?
* Anyone keen to take on the GraphQL implementation as a third party package?

lovelydinosaur:
* Ariadne ASGI app: https://ariadnegraphql.org/docs/asgi
* Tatiflette: https://github.com/tartiflette/tartiflette-starlette

dhirschfeld:
`strawberry` asgi app:
* https://github.com/strawberry-graphql/strawberry/blob/master/docs/05_integrations/asgi.md

euri10:
> `strawberry` asgi app:
> * https://github.com/strawberry-graphql/strawberry/blob/master/docs/05_integrations/asgi.md

`​@​strawberry​.​field` forever : 💗 !

JayH5:
Just commenting to add some updated links to libraries that can provide GraphQL support to Starlette applications. These will be in the Starlette documentation as well:
* [Ariadne](https://ariadnegraphql.org/docs/asgi)
* [`tartiflette-asgi`](https://tartiflette.github.io/tartiflette-asgi/)
* [Strawberry](https://strawberry.rocks/docs/integrations/asgi)
* [`starlette-graphene3`](https://github.com/ciscorn/starlette-graphene3)

RoyTrudell:
> > `strawberry` asgi app:
> > 
> > * https://github.com/strawberry-graphql/strawberry/blob/master/docs/05_integrations/asgi.md
> 
> `​@​strawberry​.​field` forever : 💗 !

Dead Link

patrick91:
> > > `strawberry` asgi app:
> > > 
> > > * https://github.com/strawberry-graphql/strawberry/blob/master/docs/05_integrations/asgi.md
> > 
> > 
> > `​@​strawberry​.​field` forever : 💗 !
> 
> Dead Link

Updated link is here: https://strawberry.rocks/docs/integrations/asgi ☺️

cglacet:
[...]

PR #876, opened 2020-03-26
Fix graphql error import bug

yukukky:
There is no graphql.execution.executors.asyncio in GraphQL-core3.

If you're using GraphQL-core3, the import of the AsyncioExecutor fails.
Then, format_error and GraphQLError, which are also present in GraphQL-core3, are handled without being imported

As an amendment, I propose to attach a try to each import.

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -10,13 +10,22 @@ from starlette.types import Receive, Scope, Send
 
 try:
     import graphene
-    from graphql.execution.executors.asyncio import AsyncioExecutor
-    from graphql.error import format_error as format_graphql_error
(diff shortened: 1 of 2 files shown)


JayH5:

RoyTrudell:
@yukukky  I've fixed the issue by mokey patching starlette.graphql
Ugly but since this is being deprecated I think its a necessary evil. 

`import starlette.graphql as gql
from graphql.error import GraphQLError, format_error
gql.GraphQLError = GraphQLError
gql.format_graphql_error = format_error
`

RoyTrudell (commented):
This would be a good addition to the code base.

PR #880, opened 2020-03-27
Add safelisted request headers to CORSMiddleware

vgerak:
As the [spec](https://fetch.spec.whatwg.org/#cors-safelisted-request-header) suggests and the [documentation](https://www.starlette.io/middleware/#corsmiddleware) mentions:

> The Accept, Accept-Language, Content-Language and Content-Type headers are always allowed for CORS requests.

This PR aims to always allow the CORS-safelisted request headers.

PR #881, opened 2020-03-27
Fix safelisted CORS headers implementation

florimondmanca:
Cleanup for #880 (seems like the build had not run there).


florimondmanca:
Hmm… Looks like Travis CI builds are not triggering… 🤔 Status is green though: https://www.traviscistatus.com/

Edit: actually, the build *did* trigger: https://travis-ci.org/github/encode/starlette/builds/667806087

But GitHub is not showing it here… This has been the case for the past 3 days (eg #875 has no build either), and [GitHub status is green](https://www.githubstatus.com/)… Has anything changed in the Travis/GH integration in this repo?

lovelydinosaur:
Merged since this looks nice it it's own right.

Nope nothing has changed our side.

lovelydinosaur:
Badge is updated now, at least.

PR #889, opened 2020-04-03
:memo: make sure all task has cancel in run_until_first_complete

Dustyposa:
When I use `run_until_first_complete`, I meet some error( tasks hadn't be canceled).
Have three  reason for  why I do this PR.
1. Improve compatibility. According to the [document](https://docs.python.org/3/library/asyncio-task.html?highlight=asyncio%20wait#waiting-primitives), 

> Deprecated since version 3.8: If any awaitable in aws is a coroutine, it is automatically scheduled as a Task. Passing coroutines objects to wait() directly is deprecated as it leads to confusing behavior.

So I explicitly declared an task by `asyncio.create_task`

2. According to [document](https://docs.python.org/3/library/asyncio-task.html?highlight=asyncio%20wait#asyncio.Task.cancel). 
> This arranges for a CancelledError exception to be thrown into the wrapped coroutine on the next cycle of the event loop.

`task.cancel` **can't make sure all task has be canceled.**
For some situation, that's will raise some error.
So, I add `asyncio.wait` to **make sure** all tasks will be canceled.

This is my first PR for `starlette`, has any advice please tell me.
Thanks in advance. :smile:

### DIFF
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -12,10 +12,11 @@ T = typing.TypeVar("T")
 
 
 async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    tasks = [handler(**kwargs) for handler, kwargs in args]
+    tasks = [asyncio.create_task(handler(**kwargs)) for handler, kwargs in args]
     (done, pending) = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
     [task.cancel() for task in pending]
     [task.result() for task in done]
+    await asyncio.wait(waiting)
 
 
 async def run_in_threadpool(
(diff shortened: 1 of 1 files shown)


erewok:
Hello, and thanks for submitting this Pull Request. 

Wrapping the `coroutines` as `Task`s seems fine and does seem in line with what the docs state, so this seems like a useful thing to do:

```python
tasks = [create_task(handler(**kwargs)) for handler, kwargs in args]  # type: ignore
```

The latter part, I'm unsure about:

```python
    if pending:
        await asyncio.wait(pending)
```

I haven't seen this myself: can you explain why it would be necessary or recommended to `await` cancelled Tasks?

JayH5:
Going to close this out as we've fixed one of the two issues this PR fixes and w.r.t. the question:
> I haven't seen this myself: can you explain why it would be necessary or recommended to await cancelled Tasks?

...we haven't heard a response.

> For some situation, that's will raise some error.

Consider raising an issue that shows the case where this can happen and we can reopen this if necessary.

Issue #898, opened 2020-04-15
Consider something more lenient for cookie parsing


erewok:
Hello,

I have used Starlette in a handful of small services in the past few months and have been really happy with it. It's a great project. 

Last week I ran into something that I wanted to raise as a potential issue here, because it may be worth considering.

We have an application that uses Okta's Sign-In widget for authentication (which is an OpenID Connect thing). Okta does its Oauth2 callback-redirect and they set some cookies, at least one of which is well outside the spec of HTTP cookies. Anyway, in Starlette the `request.cookies` attribute ends up pretty mangled as a result.

Here's an example of what I'm talking about using the same [cookie-parsing technique in Starlette](https://github.com/encode/starlette/blob/6cb333a978dbb9cd4a06773ead2e9934517fd40c/starlette/requests.py#L88), which relies on Python's standard library module `http.cookies`:

```python
In [1]: cookieval = """okta-oauth-nonce=validAsciiblabla; okta-oauth-state=validAsciiBlabla; okta-oauth-redirect-params={"responseType":"code","state":"somestate",
   ...: "nonce":"somenonce","scopes":["openid","profile","email","phone"],"urls":{"issuer":"https://subdomain.okta.com/oauth2/authServer","authorizeUrl":"https://s
   ...: ubdomain.okta.com/oauth2/authServer/v1/authorize","userinfoUrl":"https://subdomain.okta.com/oauth2/authServer/v1/userinfo"}}; importantCookie=importantValu
   ...: e; sessionCookie=importantSessionValue"""

In [2]: import http.cookies

In [3]: cookie = http.cookies.SimpleCookie()

In [4]: cookie.load(cookieval)

In [5]: list(cookie.items())
Out[5]:
[('okta-oauth-nonce', <Morsel: okta-oauth-nonce=validAsciiblabla>),
 ('okta-oauth-state', <Morsel: okta-oauth-state=validAsciiBlabla>)]
```

Notice in the example above, `cookieval` contains 5 important cookies (including the session cookie set by Starlette's `SessionMiddleware`), but only two are successfully parsed by this method.

I have seen a lot of wacky stuff thrown into cookies, so it may be a good idea to provide some method to parse them more leniently in the Starlette project.

[...]

PR #900, opened 2020-04-15
Switch to a more lenient cookie parsing method

erewok:
This PR is for issue #898, requesting a more lenient cookie-parsing technique for Starlette. @tomchristie requested a failing test to start, so that's what has been included here. I reused the example from my issue. Happy to modify or update as desired by the maintainers.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -1,5 +1,5 @@
 import asyncio
-import http.cookies
+from http import cookies as http_cookies
 import json
 import typing
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


erewok:
Also, for reference, here's the latest spec I have found, [RFC 6265](https://tools.ietf.org/html/rfc6265#page-13), which obsoletes the prior spec. This document offers the following parsing syntax:

```
   cookie-header = "Cookie:" OWS cookie-string OWS
   cookie-string = cookie-pair *( ";" SP cookie-pair )
   cookie-pair       = cookie-name "=" cookie-value
   cookie-name       = token
   cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
   cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
 token             = <token, defined in [RFC2616], Section 2.2>
```



According to this spec, the Starlette cookie parser (and Python's `http.cookies.SimpleCookie.load`) are not properly handling cookies.

erewok:
I also found a related discussion in [Tornado](https://github.com/tornadoweb/tornado/issues/633), where they [also copied what Django is doing](https://github.com/bwangelme/tornado/commit/f5ea83121cfdf0b93e78872bb465488cc23494ad) after finding that `SimpleCookie` would not work for the project.

erewok:
I can add all of the tests they used in Tornado as well, which should give pretty broad coverage of various scenarios.

erewok:
I have a working version of cookie parsing which will pass all of these tests. I am stuck on mypy telling me the following:

```
starlette/requests.py:50: error: Module has no attribute "_unquote"
Found 1 error in 1 file (checked 32 source files)
```

We have two options here, I believe: 

- Ignore this mypy error, or
- Pull the entirety of this `_unquote` function from the Python standard library into this codebase (this is what Tornado did, but not for mypy: they write that they "did not want to have to depend on non-public interfaces", which seems like a reasonable opinion to me, honestly).

I will wait to commit and push what I have until feedback from the maintainers on proceeding with this PR. I believe we have a pretty comprehensive perspective on what other major Python projects are doing in this area and what the typical path is and so I think this is the right approach for Starlette. 

Feedback welcome.

erewok:
I decided to push my changes so they can be compared with other solutions. I would appreciate resolving this in Starlette, so we can pull our own custom cookie-parsing out of our projects. Let me know what you think.

erewok:
I can't reproduce this failing test locally and it's unrelated to any modules that I modified. Any suggestions?

lovelydinosaur:
Seems great yup!

I'll have to look into the failing database test separately.
Ended up not being noticed previously as travis wasn't reporting correctly - will look into switching to GitHub actions for CI too.

lovelydinosaur:
> Let me know what you think.

Seems good to merge, right? (Or did I misunderstand something?)

lovelydinosaur (approved):

erewok:
Yes, I thought it was ready but that failing test really threw me off! If you're sure it's unrelated, then I think it's ready to merge. (This PR closes #898)

PR #902, opened 2020-04-18
Minor improvements to convertors

JayH5:
* typing: Make convertors generic over their return types.
* PathConvertor can inherit from StringConvertor
* Use r-strings for regex pattern literals.


starlette/convertors.py
```diff
@@ -1,67 +1,66 @@
...
 
-    def convert(self, value: str) -> typing.Any:
+class Convertor(Generic[T]):
+    regex = r""
+
+    def convert(self, value: str) -> T:
         raise NotImplementedError()  # pragma: no cover
 
-    def to_string(self, value: typing.Any) -> str:
+    def to_string(self, value: Any) -> str:
```
JayH5, line 14: It's not really clear to me what the type of `value` is. Possibly it can also be `T`, but the semantics of `to_string` are not documented so I'm not sure.

starlette/convertors.py
```diff
@@ -1,67 +1,66 @@
...
         assert value, "Must not be empty"
         return value
 
 
-class PathConvertor(Convertor):
-    regex = ".*"
-
-    def convert(self, value: str) -> typing.Any:
-        return str(value)
+class PathConvertor(StringConvertor):
```
lovelydinosaur: I'd probably prefer that we *don't* subclass `StringConvertor` here.

Subclassing `Convertor` instead is a simpler, flatter, style that simply declares "we implement the (stubbed-out) Convertor interface".

Subclassing `StringConvertor` instead adds an indirection, when we could just implement a trivial `convert` method explicitly.

lovelydinosaur:
Neat bit of work - thanks.

It's a bit of a mixed bag if the `Generic` typing is an improvement or not, since the implementation style of the way that convertors are used (create them, given the path pattern, use them to populate `request.path_params`) means that the typing information provided in (say) `class FloatConvertor(Convertor[float]):` *doesn't actually* ever become exposed anyplace.

Net result... we *might* just want to stick with simplicity and continue to *under-specify* the return type, as we currently do.

Not sure. 🤷‍♂️

JayH5:
Thanks for the thoughtful review. I think I agree that this doesn't add terribly much except checking that the convertors return the expected type. Seems like we're all still figuring out when typing is most useful in Python.

I've reverted the change that made PathConvertor inherit from StringConvertor. I'll leave this for now but feel free to merge or close. Otherwise in a few days I'll close this. ~~(The tests are failing now for some seemingly unrelated reason, though 😞)~~

PR #903, opened 2020-04-18
Add a UUID convertor

JayH5:
This is a very common type of path parameter. Django implements this and I've copied their regex for it: https://github.com/django/django/blob/3.0.5/django/urls/converters.py#L25-L32

This should maybe be merged after #902.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -43,6 +43,7 @@ You can use convertors to modify what is captured. Four convertors are available
 * `str` returns a string, and is the default.
 * `int` returns a Python integer.
 * `float` returns a Python float.
+* `uuid` return a Python `uuid.UUID` instance.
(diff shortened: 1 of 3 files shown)


erewok:
I support this idea. We have one of these in our projects: it's a pretty common requirement for us.

lovelydinosaur (approved):
All very neatly done, thanks so much!

Needs a docs update too, at https://github.com/encode/starlette/blob/master/docs/routing.md#path-parameters but otherwise good to go. 😃

JayH5:
Thanks @tomchristie. I believe I did update the docs where you are pointing to?

Also, I have now checked, and Werkzeug does implement a UUID converter since version 0.10.

lovelydinosaur:
Oh, of course you did. That’s so  much, lovely little PR!

PR #904, opened 2020-04-19
Include exception type and value in plain-text error traceback

calvinwyoung:
This tiny PR adds the exception type and value to the plain-text error traceback when `debug=True`.

Currently, the `ServerErrorMiddleware` plain-text response returns the error traceback, but does *not* return the error type or message, which can make debugging more challenging. This PR simply adds in those components.


JayH5:
Thanks for the PR. I see #1031 was a duplicate of this. There wasn't much between the two PRs, but #1031 updated the test case as well so I went for that.

PR #909, opened 2020-04-22
Docs: Update middleware.md - CORS max_age 600

derekbekoe:
Update max_age documentation from 60 to the actual default value of 600.
https://github.com/encode/starlette/blob/master/starlette/middleware/cors.py#L23


lovelydinosaur:
Fantastic, thanks!

PR #912, opened 2020-04-23
Fix infinite loop when using StreamingResponse with ASGI client

### LINKED ISSUE #908: Starlette 0.13.3 hangs with StreamingResponse & httpx client
When using a `StreamingResponse` and, by extension, any custom middleware (any subclass of `BaseHTTPMiddleware`) with Starlette 0.13.3, with the async httpx client directly connected to the app (i.e. `httpx.AsyncClient(app=...)`), Starlette responses never complete. This did not happen with Starlette 0.13.2.

**Minimal-ish test case:**
With `starlette==0.13.3`, `httpx==0.12.1`, and pytest + pytest-asyncio:
```python
import httpx
import pytest
from starlette.responses import StreamingResponse


@pytest.fixture
def app():
    def data():
        yield b"data"

    async def stream(request):
        return StreamingResponse(data())

    return Starlette(routes=[Route("/stream", stream)])


@pytest.fixture
async def client(app):
        yield client


@pytest.mark.asyncio
async def test_middleware(client):
    response = await client.get("http://example.com/stream")
```

**Expected behaviour:**
The test passes

**Actual behaviour:**
The `await client.get()` call never returns/completes.

**Stacktrace:**
If you hit ctrl-c while the test is stuck you get a long stacktrace. But the important bits I believe are:
```
File "/Users/jamie/.virtualenvs/tempenv-51153646622a/lib/python3.7/site-packages/starlette/responses.py", line 202, in listen_for_disconnect
    message = await receive()
  File "/Users/jamie/.virtualenvs/tempenv-51153646622a/lib/python3.7/site-packages/httpx/_dispatch/asgi.py", line 78, in receive
    body = await request_body_chunks.__anext__()
```

**Suspected cause:**
Starlette calls both the `receive()` (to check for disconnects) and `send()` (to send the response) concurrently in its `StreamingResponse` type (since #839). It tries to wait for either of those two options to complete (using `asyncio.wait({tasks}, return_when=asyncio.FIRST_COMPLETED)`). Because the ASGI dispatch in httpx is generally effectively synchronous, both `send()` and `receive()` may never return control to the event loop when called which means that if _either_ task that is waited on does not complete, then `asyncio.wait()` will not return and the program will remain stuck in an infinite loop inside the `StreamingResponse.listen_for_disconnect` method.

JayH5:
Fixes #908

Without this, the `run_until_first_complete` function never returns when using a StreamingResponse with a client like httpx directly connected to the ASGI app. This is because the `listen_for_disconnect` method will loop forever receiving the same message like:
```
{'type': 'http.request', 'body': b'', 'more_body': False}
```
...without control ever going back to the event loop.

I'm not sure of a better solution to this as it seems like this is necessary when working with a combination of a `while True` loop and `asyncio.wait()`.


starlette/concurrency.py
```diff
@@ -11,8 +12,15 @@
 T = typing.TypeVar("T")
 
 
+def _create_task(coro: Coroutine) -> asyncio.Task:
```
JayH5: ~~This is not going to pass the coverage requirements 😞 not sure what to do about that?~~

Added 2 `no cover` comments 😬

JayH5: Technically, this `create_task` change is not necessary to fix the bug and I can revert this commit if wanted.

See this for why the change: https://docs.python.org/3/library/asyncio-task.html#asyncio-example-wait-coroutine

JayH5: Oh I see #889 attempted to tackle this too, with a slightly different approach.

JayH5: I've removed this change from this PR as it's technically not required.

florimondmanca (approved):
Just got hit by this bug myself as soon as I plug a middleware built with `BaseHTTPMiddleware` into my middleware chain, as those middleware classes return `StreamingResponse` instances.

Tried this fix locally and it works like a charm. Arguably not the most elegant but it makes sense that we should give the event loop some air in this infinite loop.

lovelydinosaur:
Doesn't make any sense to me - we *are* already yielding control to the event loop `message = await receive()`.

JayH5:
@tomchristie as far as I understand, the `await` in that `await receive()` call only _allows_ control to be yielded to the event loop, it doesn't actually _do_ that yielding.

If the `receive()` call is able to do its work without any blocking operations--as is the case when using the httpx client that always has ASGI messages readily available--then control is never yielded.

---
@florimondmanca I'm curious about what client you were using? Also httpx?

---
I'm not super-confident in this PR as it's not clear to me whether it should be the job of the client (httpx) to be doing this yielding in its ASGI backend (hard to do because of async abstractions there). But this is the best I have so far.

florimondmanca:
@JayH5 Yes, HTTPX with the AsyncClient.

@tomchristie As Jay said - the await doesn’t actually guarantee that the event loop does task switching somewhere.

Trio has an explicit notion of “checkpoints” for this and I assume it wouldn’t be a problem there, but asyncio is more tricky to get right sometimes because of this.

FWIW I remember we already had to resort to sleep(0) for some asyncio fixes in the past. We have one in HTTPX for a Python 3.6 compatibility issue somewhere that was exactly caused by this problem (asyncio not having the chance to do a round of its event loop).

florimondmanca:
@JayH5  I agree that this might be more of a client / HTTPX concern.

In normal conditions (eg when uvicorn is calling the app) the receive() call waits on a socket, which triggers event loop checkpoints.

I assume HTTPX should emulate that and make sure that it doesn’t provide a secretly synchronous receive() function - eg because the body iterator actually draws its chunks from a list or similar.

JayH5:
Fixed on the client side in encode/httpx#919

PR #913, opened 2020-04-23
Check isort in CI

JayH5:


lovelydinosaur (approved):
Thanks!

Cross checked this against `scripts/lint`, and it's the same behaviour, soon first glance it makes sense to have this in.

However I'd previous dropped it in https://github.com/encode/starlette/commit/1f7d597c160663c5d63c7b6a9b81e5096dc46b76#diff-bd0b996d66f03a10364809b532be8f0f - it would have been helpful if my commit was a proper PR with a description attached, but I think it's *sometimes* the case that being strict on the check there was sometimes introducing issues, so I'd left it in `scripts/lint`, but removed it as a strict check.

Having said that I'm happy for it to go back in, and we can remove it if there's issues again.

Ideally we'd probably want to reference `scripts/lint` from `scripts/test` so that we're not duplicating the behaviour, but that's an issue for a separate PR to consider.

PR #924, opened 2020-04-30

peso:
As mentioned on Gitter: This patch fixes paths that contain regex special chars, like parenthesis. It is very simple. When function compile_path copies stuff that is not part of a parameter, all special regex chars are prefixed with a backslash.


peso:
Does not pass automated tests.

euri10:
it's black that fails. just run scripts/lint after you installed your venv with scripts/install and you should be gtg

PR #925, opened 2020-04-30

peso:
As mentioned on Gitter: This patch fixes paths that contain regex special chars, like parenthesis. It is very simple. When function compile_path copies stuff that is not part of a parameter, all special regex chars are prefixed with a backslash.


lovelydinosaur:
Thank for this! I was wondering if there's a more standard way of doing this, and found `re.escape` in the stdlib. https://docs.python.org/3/library/re.html#re.escape

Would we be able to use that instead or does it not do quite the same thing?

peso:
re.escape is a better solution. I will make a new patch.

lovelydinosaur:
Sure thing, thanks. It'd be a good idea to add a test case too.

One nice approach can be to add the test case in a commit first, wait for the test suite to run and fail, then add the escaping fix and push that too, and wait for the test suite to run and pass.

That'd then give us a really high confidence in the fix.

Issue #388, opened 2019-02-13
MultipartParser: option to write files to temporary directory on file upload


vchendrix:
Would you consider enhancing `formparsers.MultipartParser` to allow the option to write to a temporary file  instead using `SpooledTemporaryFile`?  Maybe allow the user to specify the temporary file type?

lovelydinosaur:
Perhaps yeah. What’s the motivation? It’s possible that we’ll look at having a generic interface for parsing the request body with `request.parse()`, rather than individual .json and .form methods.

vchendrix:
Thanks for the response!  We are a data archive for earth and environmental science data.   We allow our users to upload up to 1GB of data via HTTP.  The large data transfers ( upto TB or more of data) will be handled in other non-http methods.  We want the files to be written to a temporary directory so that we may perform a number of securiry and quality checks.  Also under load we will not be able to handle all that data in memory but we do have lots of storage space to work with.  

A `request.parse()` actually sounds great because one of the things we need to do is a `checksum` on each file.  It would be really great to do this during parse as the data is coming in. 

Thanks for your consideration.

lovelydinosaur:
So, I guess this won't be resolved until we have a generic answer to "configurable parsers".
In the meantime I guess you may want to deal with this explicitly - subclass `MultipartParser`, and use `files = MyParser().parse(request.stream)` or something.

simonw:
Subclassing `MultipartParser` is a bit messy right now if you want to do this, because you end up needing to copy-and-paste the entire `parse()` method just to change the one line that references `UploadFile`:
https://github.com/encode/starlette/blob/eeb6e549b6bed325e69dc09d88040aed0050f2ea/starlette/formparsers.py#L160-L242

simonw:
Here's a suggestion for a way to make it much easier to subclass this behaviour: https://github.com/simonw/starlette/commit/d53f89d39b4d6451d82e679fffbe0bea342296fc

Does that feel like a good solution? I'm not sure if that's a clean pattern or not.

lovelydinosaur:
Probably sensible, yup.

lovelydinosaur:
Going to close this off for now with the same justification as https://github.com/encode/starlette/issues/849

Out of scope. If ya need more complex parsing than we currently provide, you need to do so within your own project.

CarliJoy:
I think the only background of the question was to be able to change "SpooledTemporaryFile" to "NamedTemporaryFile" while Uploading files. 
[...]

PR #930, opened 2020-05-04
Proposal: Custom Request Classes using type annotations

erewok:
This PR is a proof-of-concept exploring support for custom request classes in Starlette (related issues: #715, #388). I wanted to open this PR as a discussion point: there are certainly other ways to implement custom request classes in Starlette. 

To use a custom request class under the changes here, you can simply provide a custom request class type annotation for your request handler.

Here's an example:

```python
class CustomRequest(Request):
    @property
    def custom(self):
        return "Something Custom"


app = Starlette()


@app.route("/regular")
async def reqular_req(request):
    return PlainTextResponse(request.__class__.__name__)


@app.route("/custom")
async def custom(request: CustomRequest):
    return PlainTextResponse(f"{request.__class__.__name__}: {request.custom}")
```

Results can be seen here:

```
In [1]: import requests                                                                        

In [2]: requests.get("http://localhost:8000/regular").content                                  
Out[2]: b'Request'

In [3]: requests.get("http://localhost:8000/custom").content                                   
Out[3]: b'CustomRequest: Something Custom'
```

## Changes

- Use `request` arg type annotation to instantiate custom request classes.
- Add tests for the above.

## Notes

- The module `starlette.middleware.errors` has not been changed.
- No documentation changes have been made here; this PR is for gathering feedback.
- This code will throw a `TypeError` if there _is_ a type annotation for the `request` arg and that annotation _is not_ a subclass of `starlette.requests.Request`.

## Discussion

I am unsure how people will respond to this proposal. I had a couple of other ideas, but this one seemed ergonomic for end users because it's a small code change, and allows adding any custom request class _for any particular endpoint_ where one is desired. I have tried to break down some of the pros and cons below.

### Pros

- Small code change for users: add a type annotation where a custom request class is desired.
- Allows custom-request classes _per handler_ instead of globally.
- Makes explicit the request class being used for a particular handler.
- Does not affect existing users or anyone who doesn't care about having custom request classes (this seems like a requirement for any solution to this problem).

### Cons

- Potentially surprising and unpythonic: runtime behavior changes based on a type annotation (although, to be fair, other libraries are doing things like this now: [Pydantic](https://pydantic-docs.helpmanual.io/) and [ecological](https://pypi.org/project/ecological/) come to mind).
- Looks for annotation _specifically_ for the arg named `"request"`, which means this wouldn't work: `def a_handler(req: CustomRequest):` because arg-name is not "request". (We could alternately look at first arg and pull annotation for it)
- Doesn't handle the bug where someone types `def a_handler(req: List[str])` or something that's not callable; another way to say this is that it takes advantage of the fact that the type annotation _is_ a class definition.
- Does not globally change requests: some users may wish to have a single place to swap the Request class for _all_ handlers (this can be solved in other ways, I think).

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -22,9 +22,17 @@ class HTTPEndpoint:
         return self.dispatch().__await__()
 
     async def dispatch(self) -> None:
-        request = Request(self.scope, receive=self.receive)
-        handler_name = "get" if request.method == "HEAD" else request.method.lower()
+        method = self.scope["method"]
+        handler_name = "get" if method == "HEAD" else method.lower()
         handler = getattr(self, handler_name, self.method_not_allowed)
+
+        request_class = handler.__annotations__.get("request", Request)
+        if not issubclass(request_class, Request):
+            raise TypeError(
+                "Custom Request classes must subclass `starlette.requests.Request`"
+            )
+
+        request = request_class(self.scope, receive=self.receive)
         is_async = asyncio.iscoroutinefunction(handler)
         if is_async:
             response = await handler(request)
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -85,7 +85,13 @@ class ExceptionMiddleware:
                 msg = "Caught handled exception, but response already started."
                 raise RuntimeError(msg) from exc
 
-            request = Request(scope, receive=receive)
+            request_class = handler.__annotations__.get("request", Request)
+            if not issubclass(request_class, Request):
+                raise TypeError(
+                    "Custom Request classes must subclass `starlette.requests.Request`"
+                )
+
+            request = request_class(scope, receive=receive)
             if asyncio.iscoroutinefunction(handler):
                 response = await handler(request, exc)
             else:
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -21,7 +21,13 @@ class BaseHTTPMiddleware:
             await self.app(scope, receive, send)
             return
 
-        request = Request(scope, receive=receive)
+        request_class = self.dispatch_func.__annotations__.get("request", Request)
+        if not issubclass(request_class, Request):
+            raise TypeError(
+                "Custom Request classes must subclass `starlette.requests.Request`"
+            )
+
+        request = request_class(scope, receive=receive)
         response = await self.dispatch_func(request, self.call_next)
         await response(scope, receive, send)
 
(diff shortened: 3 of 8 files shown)


aviramha:
I like this approach but I am a FastAPI user, and I wonder how this could be interfaced there. I assume I would want specific mounts to be handled with my custom request, and not sure if typing is a convinent way.

lovelydinosaur:
Thanks. I don't think this fits with the general approach in Starlette, and introspecting the annotated types is a bit brittle. (Eg. it can break if using generators.)

It's *feasible* that we could support customising the request class, but I think it'd make sense to talk over the design first if so.

erewok:
Sounds good to me. It was easier to show the idea than describe it in this case, and I intended this for discussion, so I don't mind closing.

erewok:
After experimenting with this, I started to think that offering custom request classes for any arbitrary handler is a powerful and uncommon feature that users are likely to take advantage of. Further, I suspect that changing the request object would be the first obvious thing for users to do, especially when they are unfamiliar with the inner workings of Starlette and they want something other than default behavior. In the discussion about modifying the `FormParser` class, for instance, someone mentions that customizing there feels too low-level, and I tend to sympathize with that.

As the framework is already using dependency injection for the request object, it seems like a natural place for users to specify what to inject. 

But I was also thinking about making an argument that the `Route` or `Mount` constructors could take to achieve the same thing. That answer would likely be more explicit, less magical, more pythonic than this example and it would be easy to apply globally or for particular routes.

aviramha:
I agree with you @erewok but I think the comment belongs to the issue https://github.com/encode/starlette/issues/715 so it wont be missed out!

lovelydinosaur:
> As the framework is already using dependency injection for the request object, it seems like a natural place for users to specify what to inject.

We're not using dependency injection for the request instance. I think the only place we're introspecting function signatures is the `@requires` decorator, and that's something that I'd like to eventually see deprecated in favour of specifying required scopes in the Routes themselves.

erewok:
I see. If that's the case, then customizing the request object, if it's something that is desirable to do, most likely makes sense in the Routes as well, I think. Alternately, you could also make it so users could specify tooling there instead of swapping the whole request class. For instance, your request-parsing customizability suggestion could go there? It's also true that allowing people to customize the request may offer so much freedom and extensibility that people would use it for things that are better solved elsewhere, such as authentication or data validation and then their implementations could end up too strictly wedded to a particular implementation of the `Request` class in Starlette.

PR #931, opened 2020-05-05
Support OData entity key routes

peso:
Adds a test that demonstrates the issue, then a fix that escapes regex meta chars in route paths.


peso:
The lint script is unhappy with a blank line that was deleted by the test script.
I'll try again, this time running lint after test

PR #932, opened 2020-05-05

peso:
Adds a test that demonstrates the issue, then a fix that escapes regex meta chars in route paths.

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -83,16 +83,9 @@ def uuid_converter(request):
     return JSONResponse({"uuid": str(uuid_param)})
 
 
-# OData string entity key
-@app.route("/strentity('{param:str}')", name="strentity")
-def strentity(request):
-    text = request.path_params["param"]
-    return JSONResponse({"str": text})
-
-
-# OData integer entity key
-@app.route("/intentity({param:int})", name="intentity")
-def intentity(request):
+# Route with chars that conflict with regex meta chars
+@app.route("/path-with-parentheses({param:int})", name="path-with-parentheses")
+def path_with_parentheses(request):
     number = request.path_params["param"]
     return JSONResponse({"int": number})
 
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Great, thanks! I've retitled the issue to make the intent more clear.

tests/test_routing.py
```diff
@@ -83,6 +83,20 @@ def uuid_converter(request):
...
+def strentity(request):
+    text = request.path_params["param"]
+    return JSONResponse({"str": text})
+
+
+# OData integer entity key
+@app.route("/intentity({param:int})", name="intentity")
+def intentity(request):
+    number = request.path_params["param"]
+    return JSONResponse({"int": number})
```
lovelydinosaur, line 90: Let's just have a single test case, and be more generic in how we phrase it. (I'd never heard of OData URLs)
That'll be enough to test the escaping functionality.

For example, maybe this?...

```python
@app.route("/escaped-brackets({param:int})", name="escaped-brackets")
def escaped_brackets(request):
    """
    A URL route that has some characters that need to be regex-escaped.
    We want this route to match URL strings such as " /escaped-brackets(42)"
    """
    number = request.path_params["param"]
    return JSONResponse({"int": number})
```

lovelydinosaur (changes requested):
Great stuff.

I've one suggestion on slimming the test down slightly, and phrasing it as less specific to "OData".
You can just include extra commits on this pull request, rather than having to issue an entirely new pull request. (We always automatically squash pull requests down to a single commit in GitHub anyways.)

tests/test_routing.py
peso: I think that brackets look like this [ ], and function docstring is not used elsewhere in this module, but comments are. Also, from the user point of view there should be no escape characters in the specified path. That starlette use regex internally is an implementation detail. 
How about this?
```python
# Route with chars that conflict with regex meta chars
@app.route("/path-with-parentheses({param:int})", name="path-with-parentheses")
def path_with_parentheses(request):
    number = request.path_params["param"]
    return JSONResponse({"int": number})
```

peso: I'm fine with just one test case and to remove reference to OData.
And, I was not aware that parentheses were also known as round brackets. I'm not a native English speaker.

peso:
How about the commit message?
It still contains a reference to OData - should it be changed?

lovelydinosaur (approved):
Looks great now, thanks!

florimondmanca:
Hi there,

Coming here from the chat where we're discussing #1010.

I'm curious about what the rationale was to accept this as a bug fix, ie what was the bug initially?

I understand the motivation was to allow regex special characters to be included in URL patterns.

But couldn't users already achieve this by escaping those characters manually?

For example: `/path-with\(parens\)`

Now as of 0.13.5 users cannot use pure regex matching anymore. I'm not sure this was an intended feature before, but if so, should we consider reverting this PR and instead documenting this "my route pattern contains regex characters" use case?

lovelydinosaur:
@florimondmanca So any confusion here is because, Starlette was never intended to accept regexs in it's `Route` format - we only support the parameter style... https://www.starlette.io/routing/

PR #933, opened 2020-05-06
Thread pool is no longer used for files in memory

### LINKED ISSUE #819: Maybe the thread pool should no longer be used for reading and writing files?
I read performance check in https://github.com/mosquito/aiofile/issues/18. Compared to synchronization overhead, it seems that the overhead of the thread pool is greater, maybe we need some performance tests.

And I noticed that aiofiles has not responded to any questions or requests from the community for ten months, which is dangerous. If you insist on reading and writing files asynchronously, maybe starlette should be switched to aiofile?

If there is any need to modify this, I will be happy to do it.

abersheeran:
Closes #819 

Here I made a judgment, when the file is in memory, no longer use the thread pool to operate. This can save a lot of thread switching time.


lovelydinosaur (commented):
This is a pretty smart approach yup.

I'm not sure if we want `in_memory` to be a public API.
Also I guess we *could* consider changing its check to this?...

```python
rolled_to_disk = getattr(self.file, '_rolled', True)
return not rolled_to_disk
```

tests/test_datastructures.py
```diff
@@ -210,6 +213,17 @@ def test_queryparams():
     assert QueryParams(q) == q
 
 
+@pytest.mark.asyncio
+async def test_upload_file():
+    UploadFile.spool_max_size = 1024
```
lovelydinosaur: We're changing this class attribute across all the test cases. Probably doesn't really matter, but we could alternately do something like...

```python
TestUploadFile(UploadFile):
    spool_max_size = 1024
```

And use that?
Also perhaps we should assert that the `.read` is returning the expected data?

abersheeran: I agree. I will modify the test code here.

lovelydinosaur (approved):
Looks good - thanks!

PR #935, opened 2020-05-07
Middleware should treat `body` in response ASGI messages as an optional key.

euri10:
Fix for an issue sent on gitter by @tomasbedrich, the test added shows it:

if you try a HEAD on a static file and have a middleware that does nothing you get a KeyError 'body' in https://github.com/encode/starlette/blob/97257515f8806b8cc519d9850ecadd783b3008f9/starlette/middleware/base.py#L55

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -294,7 +294,7 @@ class FileResponse(Response):
             }
         )
         if self.send_header_only:
-            await send({"type": "http.response.body"})
+            await send({"type": "http.response.body", "body": b"", "more_body": False})
         else:
             async with aiofiles.open(self.path, mode="rb") as file:
                 more_body = True
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -294,7 +294,7 @@ class FileResponse(Response):
             }
         )
         if self.send_header_only:
-            await send({"type": "http.response.body"})
+            await send({"type": "http.response.body", "body": b'', "more_body": False})
         else:
             async with aiofiles.open(self.path, mode="rb") as file:
                 more_body = True
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -294,7 +294,7 @@ class FileResponse(Response):
             }
         )
         if self.send_header_only:
-            await send({"type": "http.response.body", "body": b"", "more_body": False})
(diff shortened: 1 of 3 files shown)


lovelydinosaur:
The ASGI spec defines "body" as *Optional, defaults to `""`*, so *really* we ought to change this line...

https://github.com/encode/starlette/blob/97257515f8806b8cc519d9850ecadd783b3008f9/starlette/middleware/base.py#L55 to `yield message.get("body", b"")`

tests/test_staticfiles.py
```diff
@@ -20,6 +23,27 @@ def test_staticfiles(tmpdir):
     assert response.text == "<file content>"
 
 
+def test_staticfiles_head_with_middleware(tmpdir):
```
lovelydinosaur, line 26: I guess it'd be worth linking to this issue in the docstring, since it's not obvious what the point of this test is otherwise.

euri10: done

euri10:
aplied your comments @tomchristie 
much better that way, I didnt know that asgi spec subtelty !

starlette/responses.py
```diff
@@ -294,7 +294,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             }
         )
         if self.send_header_only:
-            await send({"type": "http.response.body", "body": b"", "more_body": False})
+            await send({"type": "http.response.body"})
```
lovelydinosaur: Actually I *do* still think it's probably worth this being `send({"type": "http.response.body", "body": b"", "more_body": False})` since that's more explicit. (Even though it's not required now.)

lovelydinosaur:
Thanks for your time on this!

lovelydinosaur (approved):
I'm okay to approve this now. I think it's *probably* worth updating the `send` to the more explicit "no body" version, but not going to block the PR on it, and would be happy enough either way.

starlette/responses.py
euri10: ok I put it back :)

euri10:
yw, thanks to @tomasbedrich for the report :)

PR #938, opened 2020-05-09
Allow render to access status code

walt-w:
Minor change the sequence for init status and body, so render can access status code

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -42,8 +42,8 @@ class Response:
         media_type: str = None,
         background: BackgroundTask = None,
     ) -> None:
-        self.body = self.render(content)
         self.status_code = status_code
+        self.body = self.render(content)
         if media_type is not None:
             self.media_type = media_type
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Could you include an example of how you're using this?

I'm not convinced sure if moving the `self.body = ...` would be an improvement. It feels a bit like relying too much on internal implementation details. Also if we did move it, then we might instead want to move it further down, to just before the `.init_headers` call.

walt-w:
Sorry for slow response. For example, I want the render to response like `{"status": 404, "message": "not found"} `automatically for kinds of status code, for now I can only pass the whole response dict manually. 
For move it before the `.init_headers`, it's a good idea. 
So all the question is, should render be able to access `Response` properties ?

walt-w:
I've updated to move the render just before `.init_headers`

lovelydinosaur (approved):
Sure, I'm okay with this.

Issue #794, opened 2020-01-13
starlette/authentication.py  bug rereport


xidianwlc:
```
request = kwargs.get("request", args[idx])
```

should change to 

```
# first find request in kwargs
request = kwargs.get("request", None)
# second find request in args
if request is None and len(args) > 0:
     request = args[idx]
```

exception when use decorator requires

Traceback (most recent call last):
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/fastapi/applications.py", line 140, in __call__
    await super().__call__(scope, receive, send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/applications.py", line 134, in __call__
    await self.error_middleware(scope, receive, send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/errors.py", line 178, in __call__
    raise exc from None
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/errors.py", line 156, in __call__
    await self.app(scope, receive, _send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/authentication.py", line 48, in __call__
    await self.app(scope, receive, send)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/base.py", line 25, in __call__
    response = await self.dispatch_func(request, self.call_next)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette_exporter/middleware.py", line 40, in dispatch
    raise e
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette_exporter/middleware.py", line 37, in dispatch
    response = await call_next(request)
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/base.py", line 45, in call_next
    task.result()
  File "/data/server/pyenv/py3.7.4/lib/python3.7/site-packages/starlette/middleware/base.py", line 38, in coro
[...]

PR #942, opened 2020-05-12
Be more lenient with route arguments in AuthencationMiddleware 'requires' decorator

hanneskuettner:
As mentioned in https://github.com/encode/starlette/issues/794#issuecomment-627365118 the `@requires` decorator is too strict as to where it expects arguments to be. It always expects there to be at least one non-keyword argument that is the request. 

With a custom decorator (that is similar to the dependency injection in FastAPI) this assumption breaks. Looking at the code it looks like the `@requires` decorator already supports getting the request from the keyword arguments but still expects there to be a non-empty `args` list.

So I added back the default value of `None` in case there are no `args` passed to the route call. It will still fail at runtime if no request parameter is given.

I also added tests to the test suite that cover all three cases.


v0y (approved):
very nice MR with very nice tests, pls merge asap

kurzacz (approved):

erewok (approved):

hanneskuettner:
@erewok Any chance there will be a release in the near future that incorporates this fix?

Issue #771, opened 2019-12-29
Cannot access POST data from AuthenticationBackend


gnat:
Common use case: Retrieve POST data sent by an HTML login `<form>` to validate within `AuthenticationBackend` middleware.

- Passing usernames / passwords over GET with passwords visible in the URL is not acceptable.
- Passing Authorization headers along with a form requires Javascript, a poor solution.

Any common techniques or help appreciated.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/771">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/771/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/771/pledge.svg">
</picture>
</a>

gnat:
One way I've worked around this temporarily is retrieving the form data upfront via middleware.

I feel this is not well documented (or intuitive?) for such a common use case and AuthenticationBackend should get the ability to access the `form()` method on its own. Or some other solution perhaps, since multiple calls to `form()` will hang. See https://github.com/encode/starlette/issues/495 

```
from starlette.middleware.base import BaseHTTPMiddleware

class PostToState(BaseHTTPMiddleware):
    """
    Pulls post / form data, loads into current request state.
    """
    async def dispatch(self, request, call_next):
        post = await request.form()
        request.state.post = post
        response = await call_next(request)
        return response
```

gnat:
Note to anyone using the middleware workaround: There's a sizeable performance hit for using `await request.form()` on pages that do not need it. You can restrict this call to only pages which need to check POST data within the `AuthenticationBackend` by including a check like this in your `dispatch()` method of your PostToState middleware:

`if request['path'] in ('/login', '/signup', '/password_reset'):`

Again, would love a more permanent, performant solution from a core dev, but I'm not sure what the best approach would be.

Would love to hear of anyone else tackling the issue of retrieving POST data in your AuthenticationBackend.

Kludex:
Is this still an issue?

gnat:
Yes, it's entirely related to #944 #1519

An Authentication check requires retrieving POST data.

[...]

PR #944, opened 2020-05-12
Fixes Request.json() Hangs Forever

### LINKED ISSUE #847: Middleware Request parse Hangs forever

nikordaris:
This fixes #847 by using the Request scope to store consumed stream data so that new instances of Request can access the Request stream content. Previously `await Request(...).json()` followed by another `await Request(...).json()` would hang forever because the `receive` stream was exhausted and `await receive()` doesn't timeout after message `{"more_body": False}` is received. This PR is focused only on enabling `Request.body()`, `Request.form()` and `Request.json()` to return the cached data as originally intended and does not address the issue of `await receive()` hanging forever inside of `Request.stream()`


lwinkler-cloudflight:
is there a reason nobody had a look at this yet? seems fairly straight forward and fixes an issue that we keep stumbling over.

akarpodinis:
Any love for this issue?  It's been quiet in here for months.

graingert:
is this fixed by the anyio port?

pandrzejewski37 (approved):

pawelswiecki:
Any updates?

starlette/requests.py
```diff
@@ -199,8 +199,8 @@ def receive(self) -> Receive:
         return self._receive
 
     async def stream(self) -> typing.AsyncGenerator[bytes, None]:
-        if hasattr(self, "_body"):
-            yield self._body
+        if "body" in self.scope:
+            yield self.scope["body"]
```
graingert: HTTPScope doesn't have a `["body"]` key: https://github.com/django/asgiref/blob/main/asgiref/typing.py#L65-L78 so you can't assign to it

gnat:
Would be really nice to see this go in so multiple calls to `await request.form()` etc. are managed automatically--- I'm currently storing the result in `request.state` in the interim so I can make use of it in middlewares, as well as endpoints.

gnat:
Possible to give this a closer look? @tomchristie  This would close the following tickets: 

* https://github.com/encode/starlette/issues/495
* https://github.com/encode/starlette/issues/771
* https://github.com/encode/starlette/issues/847
* https://github.com/encode/starlette/issues/493

Kludex:
This PR is not compliant with the [ASGI HTTP Connection Scope](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope). It doesn't feel right to modify the scope dictionary.

EDIT: My comment is actually invalidated by the one below :)

gnat:
If i'm not mistaken `state` is not compliant either, yet Starlette defines it in scope and it's extremely useful. 

I highly appreciate the team being extra critical of additions to keep Starlette minimalist, but there are going to be more and more issues filed about this as Starlette increases in popularity. It's far too common of a roadbump people are actively running into.

moonso:
Are there any progress or discussions going on around this?

starlette/requests.py
```diff
@@ -222,39 +222,36 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
...
-                chunks.append(chunk)
-            self._body = b"".join(chunks)
-        return self._body
+        if "body" not in self.scope:
+            self.scope["body"] = b"".join([chunk async for chunk in self.stream()])
+        return self.scope["body"]
 
     async def json(self) -> typing.Any:
-        if not hasattr(self, "_json"):
+        if "json" not in self.scope:
```
lovelydinosaur: Can we avoid adding the `"json"` and `"form"` cases here, and *only* persist the body?

(It's not terrible if there are some situations where we're parsing it twice, and we could *consider* if we also want to deal with that as well or not as a separate issue)

Kludex:
@nikordaris Are you still interested in helping with this PR?

starlette/requests.py
nikordaris: Given I'm not longer on this tech stack y'all might want to consider this optimization as a separate PR given this is continuing a pattern that already existed within the class itself. The only difference here is moving it to the `scope`

nikordaris:
I left a comment on Tom's comment. From my perspective I'm not seeing anything else for me to change with this PR. Let me know if there is anything simple I can tweak on this PR to help it get merged. As I'm no longer on this tech stack I won't likely have time to do any major modifications to this PR

starlette/requests.py
nikordaris: i'm guessing the correct thing to do here then is to set the body inside of `scope['extensions']`? i'm not sure what the namespace for it should be. request_cache? `self.scope['extensions']['request_cache']['body']`?

https://asgi.readthedocs.io/en/latest/specs/main.html#extensions

nikordaris:
Ok, i just read through all the comments above and am slowly remembering all the context around this as i've not looked into this in over a year. I can see there is an issue with following the ASGI spec for `scope` but starlette is already breaking that with the use of `state`. I _think_ the correct solution here is to use scope [extensions](https://asgi.readthedocs.io/en/latest/specs/main.html#extensions) to store this stuff (technically i think state should probably also be in there). i can update this PR to move the body data to `scope['extensions']['request_cache']['body']` which i think will resolve the concerns around this. i can go ahead and just store the `body` and `form` and take the json load hit as per @tomchristie's comment. i think the question is what should the namespace be under `extensions`? should it be `starlette`? or more generic `request_cache`? i'm leaning towards the generic `request_cache`, but i also don't really care since i don't really have any skin in this anymore. whenever y'all decide on this `@me` so it'll show up in my github notifications.

moonso:
+1 for request_cache

gnat:
No newcomer is going to want to hunt down `request_cache` just so their request wont hang.

@nikordaris's original PR works how newcomers will expect this to work out of the box- You get the correct behaviour without having to do anything.

nikordaris:
@gnat I'm not sure I'm following your concern. I would expect the fact it is stored in `request_cache` within extensions would be transparent to new comers. They would just use `Request.json` to access the json data. If extensions.request_cache had a body it would use that if not it would pull it from the stream. I wouldn't expect users of Starlette to know anything about the request_cache extension.

gnat:
Thank you for the clarification @nikordaris That sounds totally fine.

starlette/requests.py
nikordaris: moved it to `scope["extensions"]["starlette"]["stream"]`

starlette/requests.py
nikordaris: json, body and form are not consolidated under a single raw bytes stream cache now

nikordaris:
Ok, i got nerd sniped on this. I moved all the caching stuff to the `Request.stream()`. This will now cache the raw bytes stream instead of caching it after parsing it for either body, json, or form. This removes the need to track and raise a stream consumed runtime error. I also moved the cache store to property methods so that it can easily be updated to point to wherever y'all want it to. Right now i'm pointing it to `scope["extensions"]["starlette"]["stream"]`. I went with `starlette` as the namespace here because of other framework conventions where the namespace is the owner of the data. In this case, the owner is `starlette`. Whereas a third-party library that worked on-top of starlette would reserve their own extensions space under their name to signify their ownership and control of the data. With that said, if y'all want to name it to something else or just throw it directly on the scope, i can update this PR to point it to the preferred location. just `@me` in case i miss any responses to this PR. From my perspective, these changes address all the issues raised in the comments above and is ready to merge as long as there are no objections to the location within `scope` for the cached bytes stream. @tomchristie

lovelydinosaur:
Caching the stream is not what we want to do.

In fact we particularly *don't* want to do that, because if we're *streaming* the request we're deliberately ingesting it in a way that doesn't store the entire stream in memory at any one time.

I'd really like to see any changes here string from a POV of narrowing down the exact behaviour that's problematic to a simple-as-possible example case, so that we can properly talk through what's actually needed.

lsmith77:
> I'd really like to see any changes here string from a POV of narrowing down the exact behaviour that's problematic to a simple-as-possible example case, so that we can properly talk through what's actually needed.

My use case was handling of uncaught exceptions before sending the data to Sentry I wanted to get access to certain parameters that were part of a POST request to pass such data in a more structured way to Sentry.

nikordaris:
@tomchristie it is already caching in the stream but only specifically when the `Request.body` caches it. This creates a non-intuitive experience for the user. I'm struggling to know what you want here. Your previous comment wanted to consolidate the json, body and form into a single cache variable. Doing it at the raw stream is the only way without major refactors with the Form parsers. What I can do is move the stream caching out of `Request.stream` and into a private function that wraps the stream and then have `Request.body` and `Request.form` call it instead of the stream. This would leave the public stream function as a pure Generator. I do think the current caching for the body needs to be removed though if we want to make it a pure Generator.

lovelydinosaur:
>  it is already caching in the stream but only specifically when the Request.body caches it

Exactly, yes. If you access `request.body` then you're working with the entire payload in memory. At that point you may as well cache it. You've accessed it, it's in memory, and you might as well keep it there for the duration of the request/response.

> I'm struggling to know what you want here.

What I'd like is for the issue to be well described, with a simple-as-possible "this behaviour isn't what I want, when I do *this*", so we can work through to a better resolution than caching the stream.

Having said that I've taken another look at #847 and that's pretty much the sort of thing I'm asking for.

The resolution to that (it seems to me) would be to cache the body. (Rather than the stream.)

nikordaris:
Ok. Where did you want the cache stored in the scope? In extensions? Directly on the scope? If extensions what namespace? starlette? request_cache? @tomchristie 

I'll do one more pass at this and then I'm done. I'm not on this tech stack anymore and cant invest any more time on this. Someone else will have to fork my PR and resubmit it with any follow-up changes.

My approach will be to cache the raw stream only for body and form. If stream is used directly it will use the cache if it exists. If the stream is consumed and there is no cache then it will raise a runtime error.

nikordaris:
Ok, @tomchristie this is ready for review. The `Request.stream` no longer caches. If `Request.body` or `Request.form` is called it will cache the raw bytes stream into `scope.extensions.starlette.stream`. If `Request.stream` is called followed by `Request.body`, a runtime error will be thrown indicating the stream has been consumed. If `Request.body` or `Request.form` is called followed by `Request.stream`, no error is raised and the cached stream will be returned. I think this lines up with your concerns around a large stream getting stored in memory while still preserving the convenience of leveraging the cache from body or form. Once you decide where in the `scope` you want to store the cached stream, i'll update that location and then this PR should be complete.

LiranRoffman (approved):

lovelydinosaur:
Just bumping by to mention that this *is* on my personal backlog at the moment, but I don't want to deal with it until I can give it some proper attention, since it's a bit of a fiddly area.

Please do feel free to nudge me next week if I don't already get onto looking at it before then.

starlette/requests.py
```diff
@@ -229,34 +252,29 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
...
+            parse_options_header is not None
+        ), "The `python-multipart` library must be installed to use form parsing."
+        content_type_header = self.headers.get("Content-Type")
+        content_type, options = parse_options_header(content_type_header)
+        if content_type == b"multipart/form-data":
+            multipart_parser = MultiPartParser(self.headers, self._cache_stream())
+            self._form = await multipart_parser.parse()
+        elif content_type == b"application/x-www-form-urlencoded":
+            form_parser = FormParser(self.headers, self._cache_stream())
+            self._form = await form_parser.parse()
```
adriangb: If I understand this PR correctly, this would be behavior change: currently `FormParser` consumes the stream and since large files end up as `UploadFile` memory usage is O(1) respect to request body size. Since this now calls `_cache_stream()`, which stores all of the bytes in memory, memory usage would become `O(n)`. Luckily this is not inconsistent with the rest of the PR if you just revert `self._cache_stream()` here to `self.stream()` (I think).

I do agree that it would be nice to have the _option_ to not consume the stream when creating the form, but I think that should be it's own PR.

starlette/requests.py
```diff
@@ -205,10 +205,33 @@ def method(self) -> str:
...
+            self._stream = _stream
+
     async def stream(self) -> typing.AsyncGenerator[bytes, None]:
-        if hasattr(self, "_body"):
-            yield self._body
-            yield b""
+        if hasattr(self, "_stream"):
+            for chunk in self._stream:
+                yield chunk
             return
```
adriangb, line 235: It took me a sec to process what is calling what here. I wonder if we could instead do something like:


```python
class CachedStream:
    def __init__(self, receive: Receive) -> None:
         self.chunks = list()
    
    async def __aiter__(self) -> AsyncIterator[bytes]:
        # read the stream and save the chunks
        ...

    @staticmethod
     def from_asgi(self, scope: Scope, receive: Receive) -> BodyStream:
         # if stream exists in scope, retrieve it from there, if not create an store
         ....
```

And then use that in `Request`?

Not sure if that's better, just an idea 😃

.gitignore
```diff
@@ -2,6 +2,7 @@
 test.db
 .coverage
 .pytest_cache/
+.python-version
```
adriangb: This is now duplicating what is in the default branch:
https://github.com/encode/starlette/blob/38185f39de3f92b2781826b6e0d07ca52b6c9b77/.gitignore#L11

```suggestion
```

starlette/requests.py
nikordaris: this is a good catch. i agree that i think solving the form caching should be a separate PR. i'll revert to the original behavior and scope this PR to only solving the body/json caching issue.

starlette/requests.py
nikordaris: ya, i can't really speak to what is better or not. This PR has attempted to follow the existing coding patterns and make as small a change as possible. i think any kind of clean up like this should be part of a separate PR to make stream caching more general purpose and a first class citizen, which is a larger scope than this PR and my commitment to this project.

lovelydinosaur:
Okay, so I'd been pushing back against this approach because it changes the nature of streaming the request.

If we cache all the data we get when we stream a request, then we're always going to store the entire body in memory. That's find in cases where you're accessing the body (because you needed it all in memory in that cases), or in cases where you're accessing the request as JSON (because, same). Tho it's not ideal in cases where you're accessing `request.form()`, because in that case we're careful that the request parsing deliberately *doesn't* store the entire body in memory.

There's an alternate approach that'd essentially be to raise an error there's an attempt to stream the request more than once but that *does* cache the result of `.body` and `.form`.

However at this point, I'm kinda thinking that we should possibly just suck it up, and accept the trade-off because it results in a much simpler set of "here's what you can and can't do when repeatedly accessing the request content".

Anyone?

adriangb:
> If we cache all the data we get when we stream a request, then we're always going to store the entire body in memory. That's fine in cases where you're accessing the body (because you needed it all in memory in that cases), or in cases where you're accessing the request as JSON (because, same). Tho it's not ideal in cases where you're accessing `request.form()`, because in that case we're careful that the request parsing deliberately _doesn't_ store the entire body in memory.

I think I agree with what you say, but I'm not understanding the relationship to this PR.
Let me echo back to you my interpretation to see if I'm missing anything.

My understanding is that currently in Starlette if you do `Request.body()` (or anything that calls that, like `Request.json()`) the body gets cached. On the other hand, if you do `Request.stream()` (or anything that calls that, like `Request.form()`) then the body is _not_ cached in memory. From my understanding of this PR that is preserved, the only difference is that when the body _is_ cached it is cached into the ASGI scope instead of the Request instance. Right @nikordaris ?

More generally, I think it would make sense to have an API like `Request.form(consume=False)` and `Request.json(consume=True)` (with defaults to the current behavior) that allow users to control this. That would I think aleviate confusion because instead of having to dig into the implementation details of `Request.json` or `Request.form` (or look at some footnote in the docs) you can see just from the function signature what's going on. Not only that, but if you're looking for this info it's probably because you want the opposite behavior, so you also already know how to change it. But again, this would have to be another PR.

lovelydinosaur:
Okay - this looks good - I misunderstood slightly on first pass. (Thanks for clearing that up for me @adriangb)

I think we also need to be careful to preserve the `RuntimeError("Stream consumed")` case.

I've created #1519 as a similar approach that also handles the stream consumed state.

nikordaris:
@tomchristie it looks like this PR should be closed out in favor of #1519 now, correct?

gnat:
Thank you @nikordaris for his valiant effort on this!! Looking forward to seeing the resolution to this in #1519

PR #945, opened 2020-05-13
RedirectResponse now accepts optional background parameter

### LINKED ISSUE #941: RedirectResponse does not accept background argument
The [RedirectResponse](https://github.com/encode/starlette/blob/e57fa2a6b1b19dd118c26a21b2f5bd4dbe46c64a/starlette/responses.py#L174) class does not accept the `background` parameter.

My use case is I am submitting a form and then doing an `HTTP 303 See Other` ([ref](https://httpstatuses.com/303)), redirecting the user to a page specific to the output of the form. That page allows the user to download the processed data, once the background task completes it.

🤔 Perhaps this is intentional?

bartek:
With this change, a `background` task can now be sent during a RedirectResponse.

This can be useful when an application wishes to run a background task after a form submission, which usually includes a redirect (e.g. to prevent resubmission)

Closes #941

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -173,9 +173,15 @@ class UJSONResponse(JSONResponse):
 
 class RedirectResponse(Response):
     def __init__(
-        self, url: typing.Union[str, URL], status_code: int = 307, headers: dict = None
+        self,
+        url: typing.Union[str, URL],
+        status_code: int = 307,
+        headers: dict = None,
+        background: BackgroundTask = None,
     ) -> None:
-        super().__init__(content=b"", status_code=status_code, headers=headers)
+        super().__init__(
+            content=b"", status_code=status_code, headers=headers, background=background
+        )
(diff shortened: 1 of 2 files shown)


bartek:
Likely out of scope for this PR, but I have noticed the repetition of testing background tasks in the other unit tests. Any opinions on how you want these helper functions (E.g. `async def numbers`) organized/consolidated @tomchristie ?

lovelydinosaur:
> Likely out of scope for this PR, but I have noticed the repetition of testing background tasks in the other unit tests. 

Yup, so. I'd change the `__init__.py` here, but I *wouldn't* bother changing the test case. Let's just revert that bit.
It's just adding extra testing code without testing anything valuable that we don't already test for on the superclass.

bartek:
@tomchristie Agreed, makes sense! I've removed the amendments to the existing test.

lovelydinosaur:
Ace, thanks!

lovelydinosaur (approved):

PR #949, opened 2020-05-17
fix: add more type hints

woile:
missing type hints from mypy --strict mode. 
Reduced from `275` -> `157`

Hey people, I've been playing with hints and `--strict` mode in mypy, this is the result, feel free to merge or reject it.

## Some notes

There are many `Callable` missing the parameters, this errors is present a lot:
```
error: Missing type parameters for generic type "typing.Callable"
```

----------

Importing just the `typing` module adds a lot of noise, I was wondering if it would be better to just import what's needed from it. Let's compare:

```python
exception_handlers: typing.Optional[
    typing.Dict[typing.Union[int, typing.Type[Exception]], typing.Callable]
] = None
```

```python
exception_handlers: Optional[Dict[Union[int, Type[Exception]], Callable]] = None
```
I'd be willing to make some PR in the files I consider have long type definitions if wanted.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -38,14 +38,16 @@ class Starlette:
     def __init__(
         self,
         debug: bool = False,
-        routes: typing.Sequence[BaseRoute] = None,
-        middleware: typing.Sequence[Middleware] = None,
-        exception_handlers: typing.Dict[
-            typing.Union[int, typing.Type[Exception]], typing.Callable
+        routes: typing.Optional[typing.Sequence[BaseRoute]] = None,
+        middleware: typing.Optional[typing.Sequence[Middleware]] = None,
+        exception_handlers: typing.Optional[
+            typing.Dict[typing.Union[int, typing.Type[Exception]], typing.Callable]
         ] = None,
-        on_startup: typing.Sequence[typing.Callable] = None,
-        on_shutdown: typing.Sequence[typing.Callable] = None,
-        lifespan: typing.Callable[["Starlette"], typing.AsyncGenerator] = None,
+        on_startup: typing.Optional[typing.Sequence[typing.Callable]] = None,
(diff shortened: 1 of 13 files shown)


tiangolo (approved):
LGTM! :tada: 

This is great!

In fact, `Optional` type checking is now the default: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-strict-optional

So, yeah, I think we should totally have these types well defined.

---

About using:

```Python
from typing import X
```

I'm totally on board with it, that is also the default convention: https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=convention#the-typing-module

> One final note: when adding types, the convention is to import types using the form `from typing import Iterable` (as opposed to doing just `import typing` or `import typing as t` or `from typing import *`).

woile:
Cool cool, if everyone is on board, I will start sending patches refactoring the typing import.

I will also make an investigation about `Callable`, I think they should have a clear signature, right?

tiangolo:
Yeah, let's wait to see if others (I guess mainly @tomchristie ) agree with importing `from typing`.

About the types for `Callable`, it's here: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas

lovelydinosaur:
So, the footprint on this PR is pretty large, and I'm not sure I'm convinced by the verbose `Optional` cases.

A sensible thing to do in cases like this is almost always to try to separate out differing aspects and treat them each as individual PRS. That way we could have a look at the other bits of this in isolation.

Kludex:
@tomchristie Do you still have the same opinion as your last message?

In case you do, I'm closing this.
In case you don't, I'll need to ask for rebase. :kissing:

aminalaee:
I think it should be closed now.

PR #954, opened 2020-05-26
closes #953 add possibility to ovewrite graphiql html

### LINKED ISSUE #953: [Feature] Allow overriding of qraphiql html
`graphiql` uses sources, that are embedded via jsdekivr.net. In some networks hosts like these are prohibited, or there is no internet access at all. On the other side there are several other, enhanced graphql frontends as proposed in #753 or #698 

A simple approach is to add possibility to overwrite the default graphiql html code.

lioman:
If `graphiql` parameter of app is set to a str, we can use it to overwrite the built in template

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ test.db
 .mypy_cache/
 starlette.egg-info/
 venv/
+.vscode
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -100,3 +100,23 @@ routes = [
 
 app = Starlette(routes=routes)
 ```
+
+## Disable or overwrite Graphiql
+
+Disable graphiql py setting graphiql to False
+
+```python
+GraphQLApp(schema=graphene.Schema(query=Query), graphiql=False)
+```
+
+
+graphiql is rendered via a default template.
+If you want to overwrite it set the you template on app creation. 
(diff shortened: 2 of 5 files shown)


lioman:
closes #953

lioman:
Not sure how to push this, or help with review. Would be nice to have this in starlette.

erewok:
It seems like a potentially useful configurable parameter to add, but the type `[bool, str]` is one that I find confusing, especially in combination with the `isinstance` call later on: that variable is being coerced into a wide range of responsibilities. Consider how you would support, for instance, passing a `str` that represents possibly a template itself (such as the default in the code) or a filepath which could be retrieved and used as a template?

It may be helpful to look at other libraries as well.  For instance, [graphql-server](https://github.com/graphql-python/graphql-server) has a config which [for Flask](https://github.com/graphql-python/graphql-server/blob/master/docs/flask.md#supported-options-for-graphqlview) offers the following options:

```
- graphiql: If True, may present GraphiQL when loaded directly from a browser (a useful tool for debugging and exploration).
- graphiql_version: The graphiql version to load. Defaults to "1.0.3".
- graphiql_template: Inject a Jinja template string to customize GraphiQL.
- graphiql_html_title: The graphiql title to display. Defaults to "GraphiQL".
```

JayH5:

PR #966, opened 2020-06-10
Added check for missing static files with ".html" when in HTML mode

sevberg:
Similar to how the StaticFiles object treats unspecified "index.html" and "404.html" requests when in HTML mode, this change would additionally account for the scenario when a request is made for a static file without specifying a ".html" extension. It would only be active when the developer chooses to be in HTML mode, and the '.html' extension would only checked for in the event that the initially-provided path ended with a "FileNotFoundError".

Without this update (or something similar to it), it would not be practical to use Starlette as a backend to serve static files developed with an SSG such as NextJS.


Kludex:
I think it's how @tomchristie's said on https://github.com/encode/starlette/pull/963#issuecomment-640518376.

You can inherit `StaticFiles` and modify the `lookup_path()` in a way that suits you!

In case you still think this should be on Starlette, please open a GitHub issue, so we can discuss, or let's talk on our Gitter server.

Issue #975, opened 2020-06-15
TestClient as async context manager


retnikt:
I use pytest in combination with [pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio), so all of my unit tests are coroutines. Currently I have a fixture that replaces the `TestClient`'s existing `__enter__` and `__exit__` methods to be usable asynchronously directly:
```python
@pytest.fixture
async def client():
    tc = TestClient(app)
    tc.send_queue = asyncio.Queue()
    tc.receive_queue = asyncio.Queue()
    tc.task = asyncio.create_task(tc.lifespan())
    await tc.wait_startup()
    try:
        yield tc
    finally:
        await tc.wait_shutdown()
```
It would be nice if the TestClient provided `__aenter__` and `__aexit__` methods to do this automatically as an [async context manager](https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with). I can provide a pull request if you want.

retnikt:
The issue with the TestClient currently, is that it expects the only async part of the whole system to be the Starlette app. It is based on `requests`, which expects everything to be fully synchronous, so there has to be a lot of complex translation between the paradigms.

I have decided it would be easier to use [`httpx`](https://github.com/encode/httpx) for my purposes, which supports using an ASGI app like the TestClient, and is fully async.

If anyone else wants to take this up, they're welcome to.

Kilo59:
I agree IMO. `TestClient` should be able use `httpx` in lieu of `requests`.
In several of my projects I keep having to include `requests` in my development dependencies purely for the `TestClient`.
In some cases I've done monkey-patching to make `TestClient` actually use `httpx`, so I wouldn't think it would be that difficult to implement.

But having `requests` and `httpx` both part of project tends to create dependency conflicts among the transitive dependencies that they both share.

I would use the built-in httpx feature of talking to ASGI apps directly but `httpx.Client` only works with `WSGI` apps and I like the convenience of making non-async calls from `TestClient`.

# Edit:
Looks like there's already a long-running discussion about this.
https://github.com/encode/starlette/issues/652

lovelydinosaur:
> Looks like there's already a long-running discussion about this

Yes. Closing this off in favour of https://github.com/encode/starlette/issues/652 and https://github.com/encode/starlette/pull/1376

PR #976, opened 2020-06-16
Allow TestClient to be used as an async context manager (#975)

retnikt:


retnikt:
Dammit! I should have known it wouldn't be that easy...

retnikt:
See [comment](https://github.com/encode/starlette/issues/975#issuecomment-644647580) on original issue

PR #978, opened 2020-06-16
Change BugsnagMiddleware URL

2tunnels:
This version is tested better, adds additional information to every event (like scope details and versions of installed packages) and doesn't use legacy methods.
Also has CI with linting and testing.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -266,7 +266,7 @@ Access can be restricted to specific users or to members of specific GitHub orga
 
 A middleware class for capturing Prometheus metrics related to requests and responses, including in progress requests, timing...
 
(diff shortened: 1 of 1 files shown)


florimondmanca:
Hello! How specific to Starlette is this middleware? Wondering if we could have it in https://github.com/florimondmanca/awesome-asgi.

I understand the 3rd party list in the Starlette docs isn't updated very often and I believe awesome-asgi is a more natural place for projects like these.

I do have an opinion in that projects listed in awesome-asgi should be as framework agnostic as possible — because most often it's relatively easy to achieve.

Looking at the source code, this project looks no different (doesn't seem to manipulate any kind of Starlette specific data). :)

So if you'd like to repackage it for ASGI as a whole and submit there... happy to review PRs!

PR #989, opened 2020-06-28
Fix hanging graphql tests

vladmunteanu:
Use pytest-asyncio event loop when creating AsyncioExecutor in tests, through a fixture.


lovelydinosaur:
Fantastic, thank you!

lovelydinosaur (approved):

PR #998, opened 2020-07-13
fix wsgi PATH_INFO encoding

### LINKED ISSUE #997: WSGI Middleware Error in non-ascii path
image
Get error: 
image

image

I glanced at the code and the problem is here:

https://github.com/encode/starlette/blob/ced8c350c459f22381ae00064e204b4620aa3d58/starlette/middleware/wsgi.py#L17

Maybe need to use the following code instead

```python
"PATH_INFO": scope["path"].encode("utf8").decode("latin1"), 
```

abersheeran:
fix https://github.com/encode/starlette/issues/997

### DIFF
--- a/starlette/middleware/wsgi.py
+++ b/starlette/middleware/wsgi.py
@@ -13,7 +13,7 @@ def build_environ(scope: Scope, body: bytes) -> dict:
     """
     environ = {
         "REQUEST_METHOD": scope["method"],
-        "SCRIPT_NAME": scope.get("root_path", ""),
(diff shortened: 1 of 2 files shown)


JayH5:
👋 thanks @abersheeran, if you have time would it be possible to add a test? I see this change has been made in a few other ASGI projects. We could have a test similar to [the one in hypercorn](https://gitlab.com/pgjones/hypercorn/-/commit/62ca3a41cc4d0d1b0e0d42f9d3fdacefc67adb71)

JayH5 (approved):
Didn't know I could add the test myself and now have. Thanks for the pull request.

PR #999, opened 2020-07-14
Use isort v5

JayH5:
There's more that can be done to make the CI scripts/config more similar to httpcore/httpx but this at least should fix CI for now.


florimondmanca (approved):
💜 Thank you!

PR #1003, opened 2020-07-19
Add type annotation to WebSocketEndpoint.encoding

trickeydan:
mypy currently infers this type as `None`, which is obviously incorrect, so we should annotate this explicitly.


florimondmanca (approved):
Yup, thanks!

PR #1007, opened 2020-07-20
Use os.PathLike in StaticFiles for directory

### LINKED ISSUE #1004: StaticFiles directory argument should take a PathLike instead of a str
It would be preferable to allow the `StaticFiles` app to use an `os.PathLike` so you can also provide `pathlib.Path` instances.  I can work on a PR if there's agreement.

kevinastone:
This allows using `pathlib.Path` in addition to `str` for configuring
the base directory of the static files in line with how python3.6+
handles filesystem operations.

Fixes #1004

### DIFF
--- a/docs/staticfiles.md
+++ b/docs/staticfiles.md
@@ -5,7 +5,7 @@ Starlette also includes a `StaticFiles` class for serving files in a given direc
 
 Signature: `StaticFiles(directory=None, packages=None, check_dir=True)`
 
-* `directory` - A string denoting a directory path.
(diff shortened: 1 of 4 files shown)


erewok:
Hi @kevinastone. This looks like a good addition. Thank you for putting it together. 

As part of this PR, would you be up for updating the [documentation on staticfiles](https://github.com/encode/starlette/blob/master/docs/staticfiles.md) to include a note or two on using `Path`s in addition to strings?

In addition, can you pull latest master into your branch? 

Thanks again!

erewok (approved):

erewok:
Thanks again!

PR #1010, opened 2020-07-25
Fix optional route bug

### LINKED ISSUE #1008: router.get("/?") no longer works
I think this is related to https://github.com/encode/starlette/pull/932/files

Formerly you could use router.get("/?") to represent router.get("") & router.get("/"), but now it returns a 404 error as of the latest version

I'm seeing this in FastApi (mentioned here https://github.com/tiangolo/fastapi/issues/1648 )

JulienRobitaille:
Optional routes are not working anymore because the `?` gets escaped by the change introduced in: https://github.com/encode/starlette/commit/f12e237da5ff4e19531d71340b2208f613206882

This code handle that scenario by looking at the last character of the path before doing the escape.
When the last character is a `?` it will do the escape on the path and prevent the escape of the needed `?`.

I've added test to verify that optional routes are now working alongside the regexp meta character in path changes. 

If you feel this change is not welcome, please close the PR but since it was a "very clearly well identified bug" It felt appropriate.

This PR fixes https://github.com/encode/starlette/issues/1008

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -117,7 +117,12 @@ def compile_path(
 
         idx = match.end()
 
-    path_regex += re.escape(path[idx:]) + "$"
+    # Prevent the escape of ? for route ending with ?
+    if path[-1] == "?":
+        path_regex += re.escape(path[idx:-1]) + "?$"
+    else:
+        path_regex += re.escape(path[idx:]) + "$"
+
     path_format += path[idx:]
 
     return re.compile(path_regex), path_format, param_convertors
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -90,6 +90,12 @@ def path_with_parentheses(request):
     return JSONResponse({"int": number})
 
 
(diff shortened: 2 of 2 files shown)


lovelydinosaur:
So, you probably don't want an optional trailing slash. Starlette will handle trailing slash redirects appropriately for you, so you can add either `/optional-path/` or `/optional-path` depending on which you want the canonical form to be, and you'll automatically get redirect behaviour from the other variant.

evanlurvey:
I personally like this slash patcher more. Doesn't create dup docs or dup routes just fixes the regex.
```
def slash_patcher(app: FastAPI):
    app.router.redirect_slashes = False
    for r in app.routes:
        if r.path_regex.pattern.endswith("/$"):
            r.path_regex = re.compile(r.path_regex.pattern.replace("/$", "/?$"))
        else:
            r.path_regex = re.compile(r.path_regex.pattern.replace("$", "/?$"))
```

PR #1011, opened 2020-07-27
DOCUMENTATION -- Correction in websockets.md

Yaakov-Belch:
Correct the exception class from `starlette.websockets.Disconnect` to `starlette.websockets.WebSocketDisconnect`


lovelydinosaur (approved):
Thanks!

PR #1013, opened 2020-07-28
Added Range support to Staticfiles

kevinastone:
Allows staticfiles to serve assets with range header support to serve
partial fragments.

This is necessary when serving media assets used in HLS or Dash
playlists that rely on Range header support to serve fragments from a
larger media container.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -234,6 +234,25 @@ class StreamingResponse(Response):
             await self.background()
 
 
+class InputByteRange(typing.NamedTuple):
+    start: int
+    end: typing.Optional[int] = None
+
+    def clamp(self, start: int, end: int) -> "ByteRange":
+        return ByteRange(
+            max(self.start, start), min(self.end, end) if self.end else end
+        )
+
+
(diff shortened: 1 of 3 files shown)


kevinastone:
Let me know what you think and if this functionality should belong in Starlette.  This PR could use some additional cleanup and improvements to better match the spec for Range requests (including multipart ranges and suffix ranges).

Kludex:
@abersheeran You've created something similar somewhere, right? Do you know where? 🥺

abersheeran:
@Kludex Yes, it's at https://github.com/abersheeran/baize/blob/master/baize/asgi.py#L406

Kludex:
Thank you! 👍

Kludex:
Related to https://github.com/encode/starlette/issues/22

lovelydinosaur:
Looking at this now, I think we should probably just say that the functionality `StaticFiles` currently provides is as much as we'd like it to do. There's still scope for someone to implement a more fully functional third-party package that adds extra options here.

PR #1014, opened 2020-07-28
Use cached property to simplify property logic

uriyyo:
Use `cached_property` in order to simplify the logic of properties which should be cached.

From python 3.8 there is a builtin `cached_property` class at `functools` module.
For lower python versions backport class was written.

### DIFF
--- a/starlette/cached_property.py
+++ b/starlette/cached_property.py
@@ -1,0 +1,43 @@
+try:
+    from functools import cached_property  # type: ignore
+except ImportError:  # pragma: no cover
+    import typing
+
+    class cached_property:  # type: ignore
+        def __init__(
+            self, func: typing.Callable, attrname: typing.Optional[str] = None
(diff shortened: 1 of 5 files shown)


erewok:
I don't know if anyone else will agree, but I'm not seeing an obvious win with this change. There's nothing wrong with the idea, but it swaps equivalent code for equivalent code and vendors this feature from python3.8 in order to do so. I am inclined to pass on this change.

uriyyo:
@erewok I totally agree with you. 

I looked at `starlette` code and found those repeatable patterns with cached properties. So, I thought that `cached_property` decorator can eliminate duplicated logic and make it more clear. I am totally okay with closing this PR, but just want to hear your opinion.

erewok:
Thanks, @uriyyo! Thanks for trying to improve starlette and for your understanding.

Issue #1012, opened 2020-07-27
Memory usage streaming large responses


via:
We've been running into memory issues when providing very large async generators to a streaming response.  We have these generators producing large (larger than memory set) responses in a way that allows us to only keep small chunks in memory at a time.  However, it looks like the BaseHTTPMiddleware implementation uses an asyncio queue to store the individual chunks:

https://github.com/encode/starlette/blob/master/starlette/middleware/base.py#L30

This prevents any network backpressure handling -- if the client that is receiving the streaming response is on a slow connection, the queue will happily grow without bound and consume all memory, triggering kernel out-of-memory, when the ideal handling here would be for send to block (yield) when this happens.  I believe this would naturally happen if there were no queue here at all, so I am wondering why it needs to be here?

Would a PR to remove the queueing be accepted?

If not, what is the appropriate way to override this to not use a queue?  We can write our own, but the use of BaseHTTPMiddleware is hardcoded: https://github.com/encode/starlette/blob/519f5750b5e797bb3d4805fd29657674304ce397/starlette/applications.py#L197, leaving only some fairly hacky approaches to preventing this queueing.

euri10:
I kind of remember it has been discussed already here, not sure it's
relevant to your case though

https://github.com/encode/starlette/issues/802

via:
I did see that ticket, but I do think its important to distinguish between the high level backpressure logic that I think that ticket is mostly about (specifically managing concurrent request limits, etc), and network backpressure.  I do not believe that ticket is correct when it suggests that uvicorn/gnuicorn manages the low-level network socket backpressure, as the use of the asyncio.Queue there is completely in starlette, and will happily absorb the entire response payload until it dies.

florimondmanca:
I think queueing is fine, but the queue is unbounded which results in losing the backpressure properties provided by the server. (As you said, it’s allowed to grow forever even if the other side isn’t consuming chunks fast enough.)

In my mind setting a `maxsize` of eg 10 would resolve the memory issues to you’re seeing, because then if the client isn’t consuming chunks fast enough the app will block on `send()` until there’s room in the queue.

Edit: actually as discussed by @erewok in the chat, a `maxsize` of 1 should be just enough. This way we stick to the backpressure behavior of the server as close as we can, and I don't think there's a risk of deadlock. In any case, any fix would be worth validating on a real-life big upload too. :-)

erewok:
[...]

PR #1017, opened 2020-08-01
Streaming with BaseHTTPMiddleware: force background to run after response completes

### LINKED ISSUE #919: Background tasks don't work with middleware that subclasses `BaseHTTPMiddleware`

1. Use the example below

```python
import asyncio
import uvicorn


        return response


app = Starlette(debug=True)
# uncomment to break


@app.route("/")
async def test(_):


if __name__ == '__main__':
    uvicorn.run(app)
```

erewok:
There is one issue ~~are currently two issues~~ with the `call_next` method inside `BaseHTTPMiddleware`: 

- ~~The queue is unbounded, which means that handlers returning `StreamingResponses` will have their responses loaded fully into the queue.~~ (Fixed in #1018)

- The `StreamingResponse` that `BaseHTTPMiddleware` creates to wrap the handler's response is waiting on the completion from the handler to stop streaming, which requires that any background tasks must be run before the response can be completed.

This PR addresses the latter. [Here's a gist](https://gist.github.com/erewok/14da395beb4fe86bedcdc08cd52a232c) which provides an example of running various tests against this branch.

## Important Notes

- ~~This change limits queue `maxsize` to 1 (thanks for the suggestion, @florimondmanca)~~
- The `body_stream` function now explicitly checks whether we are wrapping a `StreamingResponse` and iterates if so or completes the response if not.
- In the straightforward case, the `None` added to the queue is only needed in case of exception, so we opt to complete the response and leave the `None` on the queue (can be seen in `queue._unfinished_tasks` attribute).
- Check `task.result()` **only** for raised exceptions.

## Other Notes

- ~~Added `task_done` calls so that the queue would increment or decrement `_unfinished_tasks` as a sanity check. It's not required for this to work.~~

- Added a streaming and a broken-streaming test for `tests/middleware/test_base.py`.

Closes #919, ~~#1012~~


starlette/middleware/base.py
```diff
@@ -42,18 +42,41 @@ async def coro() -> None:
...
+                    msg is not None
+                    and msg["type"] == "http.response.body"
+                    and "more_body" in msg
+                    and msg["more_body"] is more_body
+                )
+
+            # In non-streaming responses, there will be one message to emit
+            message = await queue.get()
+            queue.task_done()
+            assert message["type"] == "http.response.body"
```
erewok: There is a possibility that an exception is thrown when the body _starts_ in which case this message will also be `None`. I need to handle that case, I believe. I don't have a test case for it yet, so I haven't written the code.

erewok:
I wrote this for Python3.8. I'll amend for 3.6 and 3.7.

starlette/middleware/base.py
```diff
@@ -5,6 +5,12 @@
 from starlette.responses import Response, StreamingResponse
 from starlette.types import ASGIApp, Receive, Scope, Send
 
+try:
+    from asyncio.exceptions import InvalidStateError  # type: ignore
+except ImportError:  # pragma: nocover
+    from asyncio import InvalidStateError
```
JayH5: From the docs it seems like `asyncio.InvalidStateError` should be available on 3.6-3.8? Which versions isn't it available on?

starlette/middleware/base.py
```diff
@@ -42,18 +48,42 @@ async def coro() -> None:
...
         if message is None:
+            queue.task_done()
             task.result()
             raise RuntimeError("No response returned.")
         assert message["type"] == "http.response.start"
 
         async def body_stream() -> typing.AsyncGenerator[bytes, None]:
-            while True:
+            def streaming_predicate(
+                msg: typing.Optional[dict], more_body: bool = True
```
JayH5: `more_body` seems to always be `True`. Why is it made an argument?

starlette/middleware/base.py
```diff
@@ -42,18 +48,42 @@ async def coro() -> None:
...
+                )
+
+            # In non-streaming responses, there will be one message to emit
+            message = await queue.get()
+            if message is not None:
+                queue.task_done()
+                assert message["type"] == "http.response.body"
+                yield message.get("body", b"")
+
+            while streaming_predicate(message, more_body=True):
```
JayH5: If you indented this while loop by one extra step, could you just do:
```python
while message.get("more_body"):
   ...
```

starlette/middleware/base.py
```diff
@@ -42,18 +48,42 @@ async def coro() -> None:
...
                 message = await queue.get()
+                queue.task_done()
                 if message is None:
                     break
                 assert message["type"] == "http.response.body"
                 yield message.get("body", b"")
-            task.result()
+
+            try:
+                task.result()  # check for exceptions and raise if present
```
JayH5: Alternatively:
```python
if task.done():
    task.result()
```
I guess it's more Pythonic to "ask for forgiveness rather than permission" so I'm ok with this, but IMO `if task.done()` is more readable.

starlette/middleware/base.py
```diff
@@ -42,18 +48,42 @@ async def coro() -> None:
...
+            # In non-streaming responses, there will be one message to emit
+            message = await queue.get()
+            if message is not None:
+                queue.task_done()
+                assert message["type"] == "http.response.body"
+                yield message.get("body", b"")
+
+            while streaming_predicate(message, more_body=True):
                 message = await queue.get()
+                queue.task_done()
```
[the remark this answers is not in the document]
erewok: I think based on this, I should probably pull it out. `task_done` decrements a counter that keeps track of all of the things that have been put into the queue (`put` increments the same counter). It'll throw an exception if you call `task_done` _more_ than you've called `put`, which can be useful when tracking different scenarios built on top of a queue like this where only one thing is allowed inside at a time. 

I was using it as record keeping to make sure that the right amount of things had been processed in various cases. I think it's likely to confuse people in the future, though. I will remove these lines.

starlette/middleware/base.py
erewok: Ah, good point. This was a reactionary change based on the failure in 3.6 and 3.7. I originally had it as `asyncio.exceptions.InvalidStateError`, which isn't valid in the earlier versions. That's a good point, though. I can import it from the top-level for all versions.

starlette/middleware/base.py
erewok: I think that is an improvement and I agree it is more readable. Thanks for the suggestion. I always forget that `done` will be `True` even if an exception is raised (in my mind `done` means "finished without throwing an exception.")

erewok:
Thanks, @JayH5 for the review!

starlette/middleware/base.py
erewok: Ah, a holdover from when I was originally experimenting with separating out some code on it. I'll pull that out.

starlette/middleware/base.py
erewok: Oh, I see: under the first check if the message is `None`. Yes, that should work.

starlette/middleware/base.py
erewok: Actually, with your `if task.done()` the import can be removed entirely.

starlette/middleware/base.py
```diff
@@ -47,13 +47,23 @@ async def coro() -> None:
...
-                if message is None:
-                    break
+            # In non-streaming responses, there should be one message to emit
+            message = await queue.get()
+            if message is not None:
                 assert message["type"] == "http.response.body"
                 yield message.get("body", b"")
-            task.result()
+
+                while message.get("more_body"):
```
florimondmanca: The way I'm seeing it, these changes are related to background task support — i.e. making sure that `body_stream()` can terminate even if `app` is not finished yet _(because it's calling into background tasks, right?)_.

So first question: can we isolate the `maxsize=1` change (+ tests) into its own "fix memory usage" PR? (The title of this PR is already pretty explicit that it's doing "two things" :-))

Second question I'd have, is that for this particular part of the code it seems we could reduce the diff by keeping the previous `while True` loop style - WDYT? Probably something like this…

```diff
        async def body_stream() -> typing.AsyncGenerator[bytes, None]:
            while True:
                message = await queue.get()
                if message is None:
                    break
                assert message["type"] == "http.response.body"
                yield message.get("body", b"")
+               if not message.get("more_body"):
+                   break
+           if task.done():
+               # <Comment>
                task.result()
```

florimondmanca (dismissed):
Great digging, thanks! The solutions you've found make sense to me. Just nagging a bit in favor of clearly separating in two dedicated PRs that we could review, merge and keep in the history independently…

starlette/middleware/base.py
erewok: When I started reading this code, I found it confusing, so I think my overall goal in addition to fixing the issues mentioned at top was to make the intentions plain and to leave it as readable as possible. I was hoping to help anyone in the future who comes to read this section. (If that's not achieved, then I would like to improve it.) 

Anyway, that was the reason for switching out the `while True` for a loop that's more explicit: I was hoping it would become a lot more obvious that, as Tom said in the chat the other day, it's "almost always the case" that the first message ends the body and that looping is only necessary in the case of `more_body`. This is subjective, of course, so I'm happy to do whatever the group thinks is best. On the whole, though, I would argue that this section is still subtle no matter what direction we go in and that we need to provide some strong hints to future readers what's going on here either in the way the code's organized or via comments or both.

I don't think I have an opinion on the first point you raised. I'm fine with a single PR because the changes are to this single function, `call_next`, but it doesn't matter to me. I can collapse the commits into two to make it simpler to read the history, though, if that's preferred: one for `maxsize=1` and one for background (do these PRs get squash merged)?

florimondmanca: @erewok

- Sure, thanks for pushing back on my yakshaving 😅 I understand your point about making it easier for new readers to understand the code. To me both ways are clear enough but obviously I've heads down in this for way too long to tell. 😂 So if this new implementation is clearer to you then let's roll with it!
- Yes, I actually think we have squash merge enforced on all PRs, which is mostly why I was encouraging to open two separate PRs. I'm not putting a hard stop on it nor mean to require anyone's time so okay to leave things as is too.

florimondmanca (commented):
Good for me on the queue fix. Again if submitted separately we could merge it right away, but I'll leave that to your appreciation. :)

As for background tasks, one last item I'm not 100% sure about is that it's not guaranteed by the middleware anymore that the `app` task will be awaited. Previously we always called `task.result()` (which AFAIK essentially equals to doing `await task`). But now we don't _always_ do it.

So for example, I'm wondering if asyncio wouldn't raise a "task was not awaited" warning in case of a streaming response w/ a background task. Do you have thoughts? I don't know if we should then pass an additional background task to the constructed StreamingResponse that calls `await task`, to ensure that the task is properly awaited...

erewok:
> As for background tasks, one last item I'm not 100% sure about is that it's not guaranteed by the middleware anymore that the app task will be awaited. Previously we always called task.result() (which AFAIK essentially equals to doing await task). But now we don't always do it.

> So for example, I'm wondering if asyncio wouldn't raise a "task was not awaited" warning in case of a streaming response w/ a background task. Do you have thoughts? I don't know if we should then pass an additional background task to the constructed StreamingResponse that calls await task, to ensure that the task is properly awaited...

This is the part I spent the most time thinking about in looking at this function. One point is that `task.result()` won't actually `await` the result of the task: if it's called *before* the task has completed, an `InvalidStateError` will be thrown and if it's called after a task has been cancelled, it will raise a `CancelledErrror`. The return value wasn't being used before, so it was really present in the code to reraise any exceptions thrown, but the existing function was relying on the fact that the task must have completed already because the loop inside `body_stream` needed a `None` val to stop, and this would _only_ be present after the app had been fully `await`ed.

In this new version, of course, we're relying on the fact that if there is `background` present, then the `app` must _not_ have completed because we want to get through the `body_stream` first and then hand off control. We're also limiting the queue size, which prevents the `await self.app` from making progress until the body has been dealt with completely.  I guess this is why I tend to connect these two issues, queue size and background; it seems like they complicate each other.

I did test a scenario of streaming with a background task (if you look at the gist, you can see the examples I stole from the docs and the issue reported on this), but I never had it throw "task not awaited." I was also looking for cancelled coroutines. It still seems odd to me, though, to schedule a task and then never `await` it.

As for dealing more explicitly with the background, I think it would be pretty involved. The `Response` classes are calling `await` on the background tasks after they call send, so there would need to be some way to listen for or interrupt those background task `await` calls and send them to the wrapped response instead, but all the solutions I can think of to do that are pretty invasive. There isn't an easy way to capture those background tasks in something like a queue in the same way that the middleware is doing for the response bodies by swapping out the `send` function.

Because the changes here are small and because my own tests were successful, I got over my discomfort with not explicitly awaiting the task, but I sympathize with that perspective. 

I will keep trying to break this code.

erewok:
@florimondmanca I think you're right about separating these: it would be good to continue testing the proposed changes included here around fixing the background but the other would be good to address immediately. I created a new PR with _only_ `maxsize` kwarg added: https://github.com/encode/starlette/pull/1018

erewok:
If I run the app defined in [the gist](https://gist.github.com/erewok/14da395beb4fe86bedcdc08cd52a232c), I can send a request for a large byte stream and then cancel it partway through like so:

```sh
~/open_source/starlette stream_response_bg*
starlette ❯ curl http://localhost:8000/streaming-memory-test/1000000000000000000 --output sample_fl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0^C
```

On the application side, everything looks fine and we can make more requests:

```sh
Yield chunk:  6904
Yield chunk:  6905
Yield chunk:  6906
Yield chunk:  6907
Yield chunk:  6908
Yield chunk:  6909
dispatching a!
returning response streaming
INFO:     127.0.0.1:59226 - "GET /s HTTP/1.1" 200 OK
sleeping!
waking up now!
dispatching a!
returning response json
sleeping!
INFO:     127.0.0.1:59228 - "GET /j HTTP/1.1" 200 OK
waking up now!
```

But then if I can cancel the server, I can sometimes get it to throw pending task:

```sh
^CINFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO:     Finished server process [25869]
Task was destroyed but it is pending!
task: <Task pending name='Task-4' coro=<BaseHTTPMiddleware.call_next.<locals>.coro() done, defined at /Users/erewok/open_source/starlette/starlette/middleware/base.py:36> wait_for=<Future cancelled>>
```

This *same* behavior also appears in the `limit_base_qsize` branch, so it's not due to the background changes in this PR: I think it's because the `disconnect` cancels the async generator that streams the body across, but nothing is cancelling the `coro` function on disconnect. 

On master, I can't reproduce, but I think it's a likely due to a quirk of how `coro` is typically fully evaluated by the `body_stream` async generator. I believe the underlying issue is present on master as well, just less likely to appear.

erewok:
I have tested this code in a variety of scenarios and inspected the tasks remaining after constructing responses using `asyncio.Task.all_tasks()` and I believe this PR is ready.

tests/middleware/test_base.py
```diff
@@ -143,3 +143,49 @@ def homepage(request):
 def test_middleware_repr():
     middleware = Middleware(CustomMiddleware)
     assert repr(middleware) == "Middleware(CustomMiddleware)"
+
+
+async def numbers_stream(minimum, maximum):
```
florimondmanca: Should we remove these utility functions and write the app views inline without any dynamicism, one for each test case...?

- `/stream` - Streams successfully.
- `/stream-error-before-body` - Fail with an exception before reading the first body chunk.
- `/stream-error-in-body` - Fails with an exception while the body is being ready.

Right now the tests need some back and forth to read and understand, I understand the motivation for DRY for in this case I think some code repetition with a clear "here's plainly what we are testing against" might be better.

florimondmanca (commented):
It's a fiddly area which is why I'm personally still a bit wary of merging...

Is there any way we can test that this resolves the "don't wait for background task to send the response" case?

tests/middleware/test_base.py
erewok: That's fair. I pushed all the functionality into each tested handler.

erewok:
@florimondmanca that's a fair request. I had intended to write a test that would demonstrate this behavior but I had overlooked it. 

I have added a test now that will fail in master with the following:

```python
        filepath = tmp_path / "background_test.txt"
        filepath.write_text("Test Start")
        response = client.get("/background-after-streaming?filepath={}".format(filepath))
        assert response.headers["Custom-Header"] == "Example"
        assert response.text == "1\n2\n3\n"
        with filepath.open() as fl:
>           assert fl.read() == "handler first"
E           AssertionError: assert 'background last' == 'handler first'
E             - handler first
E             + background last

tests/middleware/test_base.py:228: AssertionError
```

tests/middleware/test_base.py
```diff
@@ -143,3 +147,83 @@ def homepage(request):
...
+        # this should trigger _first_ message being None
+        response = client.get("/broken-streaming-on-start")
+
+    filepath = tmp_path / "background_test.txt"
+    filepath.write_text("Test Start")
+    response = client.get("/background-after-streaming?filepath={}".format(filepath))
+    assert response.headers["Custom-Header"] == "Example"
+    assert response.text == "1\n2\n3\n"
+    with filepath.open() as fl:
+        assert fl.read() == "handler first"
```
florimondmanca, line 259: Last nit: should we break this test into isolated tests for each case?

- Successful streaming
- Broken streaming start
- Broken streaming midway
- Streaming with background task

erewok: Sure, happy to do it. Thanks for staying on top of this PR!

lovelydinosaur:
I'll also try to take a look at this tomorrow. Sounds like it's been a fiddly one so would be good to get another pair of eyes on it before it goes in.

tests/middleware/test_base.py
erewok: I separated the tests and nested the handlers under the test definitions, which should hopefully make them easier to read. I'll leave this unresolved so you can review.

erewok:
I think this PR is going in the same direction as the fix we had to revert in version 0.13.8 (see [issue #1022](https://github.com/encode/starlette/issues/1022), so I think it's probably a non-starter at this point.

In general, I am beginning to wonder if we shouldn't discourage usage of this particular middleware class while we try to come up with a new means for users to write middleware that works with `request` and `response` objects. Indeed, if users don't need a `request` or `response` in their middleware and if they can write middleware that handles the `scope`, `receive,` and `send` args, then I believe they'll probably have more success writing middleware similar to [other examples](https://github.com/encode/starlette/blob/master/starlette/middleware/authentication.py#L14) in the codebase.

PR #1021, opened 2020-08-08
declare encode when open file

szthanatos:
fix the problem about can't open file in windows when the file has non-English words.

Eg: using `Config(".env",)` to open a .env file with Chinese words in windows will be failed with noticing `UnicodeDecodeError: 'gbk' codec can't decode byte 0xb7 in position xxxx: illegal multibyte sequence`.
That's because there is a problem in windows-python.  It ignore the file's encode setting(so set file's charset to 'utf-8' doesn't work), and will using 'gbk' charset to encode any file with Chinese words.


starlette/config.py
```diff
@@ -78,7 +78,7 @@ def get(
 
     def _read_file(self, file_name: typing.Union[str, Path]) -> typing.Dict[str, str]:
         file_values: typing.Dict[str, str] = {}
-        with open(file_name) as input_file:
+        with open(file_name, encoding="utf8") as input_file:
```
graingert, line 81: The filesystem encoding is supposed to be the same as the os.environ encoding. Should this actually be `errors="surrogateescape"`?

Kludex:
Just to give a reference for what @graingert said:

> In text mode, if encoding is not specified the encoding used is platform-dependent: locale.getpreferredencoding(False) is called to get the current locale encoding. (For reading and writing raw bytes use binary mode and leave encoding unspecified.)

This quote can be found on https://docs.python.org/3/library/functions.html#open .

Based on the above, I don't think this PR make sense. It would be cool to discuss this in an issue and see an alternative solution.

PR #1023, opened 2020-08-09
StaticFiles: Fix cache validation bug for deleted files in html mode

jvstme:
Currently, when StaticFiles(html=True) fails to find the requested file and decides to return the 404 error page, it compares the If-Modified-Since of the requested file with the Last-Modified of 404.html. Therefore, if a requested file was deleted but has had the same Last-Modified as that of 404.html, StaticFiles would still return 304 Not Modified, while the logical behavior is to return the error page.

To reproduce the issue manually, create the app and the html files:
```bash
echo "from starlette.staticfiles import StaticFiles; app = StaticFiles(directory='.', html=True)" > main.py
echo "<p>some file</p>" > some.html && echo "<p>404 file</p>" > 404.html
```
(I will assume some.html and 404.html have the same Last-Modified, since they are created by the same command); then start `main:app` and open some.html in the browser. If you then delete some.html and reload the page, the file should still be there.

This PR fixes the issue by returning the 404 response directly from `StaticFiles.get_response`, rather then calling `StaticFiles.file_response`.


JayH5 (approved):
Thanks, this seems like a potentially quite frustrating bug 😓

PR #1026, opened 2020-08-11
Version 0.13.8

erewok:
- Revert Queue maxsize fix in BaseHTTPMiddleware.
- `Staticfiles` constructor now allows `pathlib.Path` (in addition to `str`) for `directory` keyword arg.


JayH5:
IMO the change to `maxsize` of the queue should be a separate PR or at least have an issue about the problem. I understand this was discussed in gitter, but that doesn't provide a good record of why changes were made, especially without a link to that discussion. I'd rather have the changelog link to that issue than trying to explain the problem fully in the changelog.

erewok:
You're right @JayH5. Further, a separate PR for that change would give others more of a chance to comment on it as well. People are unlikely to see it here. Here's a PR with only that change: https://github.com/encode/starlette/pull/1028

erewok:
This PR has been updated to include only release notes and version bump.

docs/release-notes.md
```diff
@@ -1,3 +1,12 @@
...
+  
+  * Moving forward, using  `BaseHTTPMiddleware` middleware classes with streaming responses is **not recommended** See issues [#1022](https://github.com/encode/starlette/issues/1022), [#1012](https://github.com/encode/starlette/issues/1012), and [#919](https://github.com/encode/starlette/issues/919). Further, this middleware class will likely need to be redesigned and for now it is recommended that instead of subclassing `BaseHTTPMiddleware`, users consider writing middleware classes similar to [other examples](https://github.com/encode/starlette/blob/master/starlette/middleware/authentication.py#L14) in the codebase.
```
florimondmanca: I don't think we've taken a stance on this yet, have we?

In any case I don't think this should be in the changelog of this hot fix release.

docs/release-notes.md
```diff
@@ -1,3 +1,12 @@
+## 0.13.8
+
+* Revert `Queue(maxsize=1)` fix for `BaseHTTPMiddleware` middleware classes and streaming responses. 
+  * Before version `0.13.7`, [some users](https://github.com/encode/starlette/issues/1022) were relying on behavior where the response object would be fully evaluated within their middleware based on `BaseHTTPMiddleware`. (Of course, this was a problem for `StreamingResponse`s with this middleware, because they were also being loaded fully into memory.) The `maxsize=1` fix released in version `0.13.7` explicitly prevented the response object from being fully evaluated until later `await` calls, but any users who were dropping this response for any other created inside their middleware were likely to see pending tasks accumulate.
```
florimondmanca: Thoughts on dropping the mdr verbose details? What's important here seems to be that the fix corrected a regression that appeared in 0.13.7. I don't think the nitty gritty technical details are of importance to users yet - we can keep this in our heads for when we take a more thorough review at the state of concurrency and streaming. In other words: should we drop this bullet point? WDYT?

erewok: Sounds good. All of this discussion should be present in the PR for anyone who wants it. I'll drop it from here.

florimondmanca (approved):

PR #1031, opened 2020-08-16
Use format_exception instead of format_tb

Mause:
This gives much more information about the exception, including causes, and the exception message itself, in addition to the trackback


JayH5 (approved):
Thanks 🎉

PR #1033, opened 2020-08-19
gitignore pyenv local and pycharm settings

azinoviev:


euri10:
Use a global gitignore for those, there's no reason the library should take care of that

PR #1035, opened 2020-08-23
request.body() up to 6 times faster, using content length header

avideci:
I ran into a huge performance overhead when HTTP payloads are big.
The bottleneck is ` request.body()` awaitable, which is used in every POST request.

What did I add?
I used the http header "Content-Length", to allocate a bytearray() with the exact size, and wrote the requests bytes into that bytearray buffer.

It appeared to be much more efficient, because the previous code used bytes.append() for every chunk.
These append() calls caused a duplication of the bytes array by the python interpreter, every time it reached it's half capacity, and it is not the best practice when working with bytes.

Whenever Content-Length is present (it should be present in every valid post request),
by pre-allocating the buffer we significantly decreasing the time it takes to `await request.body()`to join the chunks.
Up to 6 times faster than current request.body().

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -226,7 +226,7 @@ class Request(HTTPConnection):
                         chunk_length = len(chunk)
                         body_bytes[body_index : body_index + chunk_length] = chunk
                         body_index += chunk_length
-                    self._body = bytes(body_bytes)
+                    self._body = bytes(body_bytes[:body_index])
                     return self._body
 
             # If the header is not present, reading the chunks until the stream ends.
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -232,7 +232,6 @@ class Request(HTTPConnection):
             chunks = []
             async for chunk in self.stream():
                 chunks.append(chunk)
-
             self._body = b"".join(chunks)
         return self._body
 
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -222,9 +222,9 @@ class Request(HTTPConnection):
 
     async def body(self) -> bytes:
         if not hasattr(self, "_body"):
-            content_length = self.headers.get('content-length')
+            content_length = self.headers.get('Content-Length')
             if content_length:
-                # Allocating a bytearray with a size that matches "content-length" header.
+                # Allocating a bytearray with a size that matches "Content-Length" header.
                 body_bytes = bytearray(int(content_length))
                 body_index = 0
                 async for chunk in self.stream():
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -13,7 +13,6 @@ try:
 except ImportError:  # pragma: nocover
     parse_options_header = None
 
-
 SERVER_PUSH_HEADERS_TO_COPY = {
     "accept",
     "accept-encoding",
@@ -130,21 +129,21 @@ class HTTPConnection(Mapping):
     @property
     def session(self) -> dict:
         assert (
-            "session" in self.scope
+                "session" in self.scope
         ), "SessionMiddleware must be installed to access request.session"
         return self.scope["session"]
 
     @property
     def auth(self) -> typing.Any:
         assert (
-            "auth" in self.scope
+                "auth" in self.scope
         ), "AuthenticationMiddleware must be installed to access request.auth"
         return self.scope["auth"]
 
     @property
     def user(self) -> typing.Any:
         assert (
-            "user" in self.scope
+                "user" in self.scope
         ), "AuthenticationMiddleware must be installed to access request.user"
         return self.scope["user"]
 
@@ -173,7 +172,7 @@ async def empty_send(message: Message) -> None:
 
 class Request(HTTPConnection):
     def __init__(
-        self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send
+            self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send
     ):
         super().__init__(scope)
         assert scope["type"] == "http"
@@ -221,6 +220,28 @@ class Request(HTTPConnection):
             self._body = b"".join(chunks)
         return self._body
 
+    async def body(self) -> bytes:
+        if not hasattr(self, "_body"):
+            content_length = self.headers.get('content-length')
+            if content_length:
+                # Allocating a bytearray with a size that matches "content-length" header.
+                body_bytes = bytearray(int(content_length))
+                body_index = 0
+                async for chunk in self.stream():
+                    chunk_length = len(chunk)
+                    body_bytes[body_index:body_index + chunk_length] = chunk
+                    body_index += chunk_length
+                self._body = body_bytes
+                return self._body
+            else:
+                # If the header is not present, reading the chunks until the stream ends.
+                chunks = []
+                async for chunk in self.stream():
+                    chunks.append(chunk)
+                self._body = b"".join(chunks)
+
+        return self._body
+
     async def json(self) -> typing.Any:
         if not hasattr(self, "_json"):
             body = await self.body()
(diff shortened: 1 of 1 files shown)


starlette/requests.py
```diff
@@ -221,6 +220,28 @@ async def body(self) -> bytes:
...
+                    chunk_length = len(chunk)
+                    body_bytes[body_index:body_index + chunk_length] = chunk
+                    body_index += chunk_length
+                self._body = body_bytes
+                return self._body
+            else:
+                # If the header is not present, reading the chunks until the stream ends.
+                chunks = []
+                async for chunk in self.stream():
+                    chunks.append(chunk)
```
acutaia:
```suggestion
                chunks = [
                    chunk
                    async for chunk in self.stream()
                ]
```
With a list comprehension the performance would be better

acutaia (changes requested):
Add async list comprehension

acutaia (changes requested):
Change line from 232 to 234 with an async list comprehension as suggested in the first commit

starlette/requests.py
avideci: Great :)
Thanks

avideci:
@acutaia The list comprehension is fixed,  and the tests pass.
Can we merge it?

Thanks.

acutaia (approved):

JayH5:
Thanks for the PR. This looks like a solid performance enhancement overall but I have two concerns:
1. You say _"Up to 6 times faster than current request.body()."_ Under what exact conditions is this? Do you have a graph maybe?
2. Pre-allocating a bytearray results in a bytearray initialised with null bytes. If the body is actually smaller than the specified content-length, this could result in us returning a request body with unexpected null bytes. While it's likely that the web server or some other proxy would reject a request with an invalid content-length header, I think this is a case where I'd rather be safe than sorry. A quick fix for this might be to slice the bytearray at the end.. something like `self._body = bytes(body_bytes[:total_length])` where `total_length` is the index of the end of the last chunk.

avideci:
Hey @JayH5 ,

1. The performance improvement is mainly noticeable when working with very big payloads (40mb request body, For example, When using HTTP for Deep Learning).
We noticed significant improvements on both Mac and Linux, with different hardwares, locally and on the cloud.

It was pretty negligible when the payloads are small JSON requests.
In my company, we managed to reduce the HTTP latency from 54.05ms -> 31.50ms for 40MB HTTP payload.

If you would sent body > 200MB, it would not be linear. It would be efficient more than 2 times, and we noticed up to 6 times better performance to request.body(), specially when the CPU bound is overloaded (May regard to memory allocation by the OS?)
Unfortunately I don't have a graph for that. Can we run a benchmark for that using existing code in this repository?

2. I added a commit that uses the body index and slices the bytearray according to your example. I couldn't agree more about that.

Thanks!

starlette/requests.py
```diff
@@ -216,9 +215,22 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
...
+                    body_bytes = bytearray(int(content_length))
+                    body_index = 0
+                    async for chunk in self.stream():
+                        chunk_length = len(chunk)
+                        body_bytes[body_index : body_index + chunk_length] = chunk
+                        body_index += chunk_length
+                    self._body = bytes(body_bytes[:body_index])
+                    return self._body
+
+            # If the header is not present, reading the chunks until the stream ends.
```
JayH5:
```suggestion
            # If the header is not present, read the chunks until the stream ends.
```

starlette/requests.py
```diff
@@ -216,9 +215,22 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
...
+                content_length = self.headers.get("Content-Length")
+                if content_length:
+                    # Allocating a bytearray with a fixed size,
+                    # that matches "content-length" header.
+                    body_bytes = bytearray(int(content_length))
+                    body_index = 0
+                    async for chunk in self.stream():
+                        chunk_length = len(chunk)
+                        body_bytes[body_index : body_index + chunk_length] = chunk
+                        body_index += chunk_length
```
JayH5: Nit (only change if you'd like to): we can do one less addition here.
```suggestion
                        chunk_end_index = body_index + len(chunk)
                        body_bytes[body_index:chunk_end_index] = chunk
                        body_index = chunk_end_index
```

starlette/requests.py
```diff
@@ -216,9 +215,22 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
...
     async def body(self) -> bytes:
         if not hasattr(self, "_body"):
-            chunks = []
-            async for chunk in self.stream():
-                chunks.append(chunk)
+            if hasattr(self, "_headers"):
+                content_length = self.headers.get("Content-Length")
+                if content_length:
+                    # Allocating a bytearray with a fixed size,
+                    # that matches "content-length" header.
```
JayH5: "fixed size" is not technically correct: the bytearray could expand to a larger size.

I'd suggest a comment like: `# Pre-allocate a bytearray large enough to fit "content-length" bytes.`

JayH5 (commented):
This looks good.

I do want to check with some of the other encode maintainers how they think we should deal with request bodies with a size that doesn't match the content-length header (should we reject them or not).

starlette/requests.py
avideci: Sounds good.
I'll commit it.

avideci:
> This looks good.
> 
> I do want to check with some of the other encode maintainers how they think we should deal with request bodies with a size that doesn't match the content-length header (should we reject them or not).

Legitimate,
Let me know if I can help.

I changed the docstring for the bytearray allocation and accepted you requested changes for the bytearray additions.

euri10:
Am I doing something silly below ? takes 22s without the patch, 37s with it, so I'm getting opposite results so far.

```
TOT = 5000000

def test_chunked_encoding():
        body = await request.body()
        response = JSONResponse({"body": body.decode()})

    client = TestClient(app)

    def post_body():
        for i in range(TOT):
            yield b"foo"
            yield "bar"

    t1 = datetime.datetime.utcnow()
    response = client.post("/", data=post_body(), )
    t2 = datetime.datetime.utcnow()
    print(t2-t1)
    assert response.json() == {"body": "foobar"*TOT}


def test_chunked_encoding_preallocate():
        h = request.headers
        body = await request.body()
        response = JSONResponse({"body": body.decode()})

    client = TestClient(app)

    def post_body():
        for i in range(TOT):
            yield b"foo"
            yield "bar"

    t1 = datetime.datetime.utcnow()
    response = client.post("/", data=post_body(), headers={"Content-Length": str(6*TOT)})
    t2 = datetime.datetime.utcnow()
    print(t2-t1)
    assert response.json() == {"body": "foobar"*TOT}
```

JayH5:
@avideci this _looks_ good on the surface but I think it needs more careful benchmarking. Unfortunately we don't have a standardised benchmark for this.

> These append() calls caused a duplication of the bytes array by the python interpreter, every time it reached it's half capacity, and it is not the best practice when working with bytes.

I don't think the append calls are actually the slow part here. In fact, in my testing the old algorithm is faster in the case of more smaller body chunks as in @euri10's example above. I suspect that in cases with many chunks the `b"".join` becomes the dominant part of the algorithm and because that method is so optimised in the standard library that it's actually faster than we can do with a for-loop like you've done here.

I do think tiny chunk sizes like in @euri10's test above are unrealistic in practice, but with the same test and using 3000 chunks with size 87380 bytes ([the default Linux TCP receive buffer size](https://man7.org/linux/man-pages/man7/tcp.7.html)) (total body 250MiB) this new solution is _maybe_ very slightly faster.

At this point, unless you're able to provide a reproducible test case that shows this being significantly faster, I think we're better off leaving things the way they are as it leaves the code a lot simpler. Thanks for bearing with all the review suggestions, but I think this is a lesson in how difficult optimisations are and how important it is to benchmark (I'm not saying your benchmarks of your service are bad or invalid, just that we need benchmarks we can reproduce & verify ourselves).

PR #1040, opened 2020-08-31
configure pytest to be stricter

graingert:
configure pytest to be stricter and fix all the problems that the increased strictness detects,

eg `tests/test_datastructures::TestUploadFile PytestCollectionWarning`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,4 +16,19 @@ profile = black
 combine_as_imports = True
 
 [tool:pytest]
-addopts = --cov-report= --cov=starlette --cov=tests -rxXs
+addopts =
+  --cov-report=term-missing:skip-covered
+  --cov=starlette
+  --cov=tests
+  -rxXs
+  --strict-config
+  --strict-markers
+  --cov-config tests/.ignore_lifespan
+  --cov-fail-under=100
+xfail_strict=True
+filterwarnings=
+    error
+    ignore: "@coroutine" decorator is deprecated.*:DeprecationWarning
+    ignore: Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated.*:DeprecationWarning
+    ignore: The 'context' alias has been deprecated. Please use 'context_value' instead\.:DeprecationWarning
+    ignore: The 'variables' alias has been deprecated. Please use 'variable_values' instead\.:DeprecationWarning
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,9 @@ addopts =
 xfail_strict=True
 filterwarnings=
     error
+    # https://github.com/Tinche/aiofiles/issues/81
     ignore: "@coroutine" decorator is deprecated.*:DeprecationWarning
+    # https://github.com/graphql-python/graphene/issues/1055
     ignore: Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated.*:DeprecationWarning
     ignore: The 'context' alias has been deprecated. Please use 'context_value' instead\.:DeprecationWarning
     ignore: The 'variables' alias has been deprecated. Please use 'variable_values' instead\.:DeprecationWarning
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,6 @@ addopts =
   -rxXs
   --strict-config
   --strict-markers
-  --cov-config tests/.ignore_lifespan
 xfail_strict=True
 filterwarnings=
     # Turn warnings that aren't filtered into exceptions
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,6 @@ addopts =
   --strict-config
   --strict-markers
   --cov-config tests/.ignore_lifespan
-  --cov-fail-under=100
 xfail_strict=True
 filterwarnings=
     error
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ addopts =
   --cov-config tests/.ignore_lifespan
 xfail_strict=True
 filterwarnings=
+    # Turn warnings that aren't filtered into exceptions
     error
     # https://github.com/Tinche/aiofiles/issues/81
     ignore: "@coroutine" decorator is deprecated.*:DeprecationWarning
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -213,13 +213,13 @@ def test_queryparams():
     assert QueryParams(q) == q
 
 
-class TestUploadFile(UploadFile):
+class _TestUploadFile(UploadFile):
     spool_max_size = 1024
 
 
 @pytest.mark.asyncio
 async def test_upload_file():
-    big_file = TestUploadFile("big-file")
+    big_file = _TestUploadFile("big-file")
     await big_file.write(b"big-data" * 512)
     await big_file.write(b"big-data")
     await big_file.seek(0)
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -213,13 +213,13 @@ def test_queryparams():
     assert QueryParams(q) == q
 
 
-class _TestUploadFile(UploadFile):
+class BigUploadFile(UploadFile):
     spool_max_size = 1024
 
 
 @pytest.mark.asyncio
 async def test_upload_file():
-    big_file = _TestUploadFile("big-file")
+    big_file = BigUploadFile("big-file")
     await big_file.write(b"big-data" * 512)
     await big_file.write(b"big-data")
     await big_file.seek(0)


tests/middleware/test_cors.py
```diff
@@ -181,7 +181,7 @@ def test_cors_allow_origin_regex_fullmatch():
     app.add_middleware(
         CORSMiddleware,
         allow_headers=["X-Example", "Content-Type"],
-        allow_origin_regex="https://.*\.example.org",
+        allow_origin_regex=r"https://.*\.example.org",
```
graingert: invalid escape sequences are deprecated

```
E     File "/home/runner/work/starlette/starlette/tests/middleware/test_cors.py", line 184
E       allow_origin_regex="https://.*\.example.org",
E                          ^
E   SyntaxError: invalid escape sequence \.
```

scripts/test
```diff
@@ -10,7 +10,7 @@ export PYTHON_VERSION=`python -c "$VERSION_SCRIPT"`
 
 set -x
 
-PYTHONPATH=. ${PREFIX}pytest --ignore venv --cov-config tests/.ignore_lifespan -W ignore::DeprecationWarning --cov=starlette --cov=tests --cov-fail-under=100 --cov-report=term-missing ${@}
```
graingert: options have been moved to `.pytest.ini[pytest].addopts`

JayH5 (commented):
Hey 👋 thanks for the PR, but I've gone in a slightly different direction to keep things consistent with other encode projects. I believe most of the things in this PR have now been fixed, but to summarise:
- [ ] PytestCollectionWarning on TestUploadFile: still not fixed and still worth fixing (note that an alternative to renaming the class is [adding a `__test__ = False` attribute](https://stackoverflow.com/questions/42268401/how-can-i-get-pytest-to-ignore-test-classes-that-dont-subclass-unittest) to it)
- [ ] "Stricter" pytest: you'd need to explain why this is desirable/what problem it solves for you.
- [x] Trailing comma fixes for black 20 #1042 
- [x] Regex literal in test #1043 
- [x] Pytest config moved to file #1043 (setup.cfg instead of pytest.ini)
- [ ] Filtering deprecation warnings in pytest: I'm not sure we actually want to do this. We probably should keep tabs on these warnings, even if they're in 3rd party libraries.

Thanks again, and I'd be open to this PR just fixing the first issue in this list, and maybe a separate PR for the second point if that's something you can give a stronger reason for.

graingert:
> "Stricter" pytest: you'd need to explain why this is desirable/what problem it solves for you.

It's very possible to accidentally skip tests by adding an `__init__` method to the suite

secondly it's good to find and explicitly list the warnings that you're ignoring, rather than just having them fruitlessly print out at the end of the suite.

thirdly you had a SyntaxWarning that until run with a stricter pytest was not noticed.

graingert:
>  Filtering deprecation warnings in pytest: I'm not sure we actually want to do this. We probably should keep tabs on these warnings, even if they're in 3rd party libraries.

You were previously filtering all of them, in this PR I'm fixing it so you can keep tabs on them

tests/test_datastructures.py
```diff
@@ -213,13 +213,13 @@ def test_queryparams():
     assert QueryParams(q) == q
 
 
-class TestUploadFile(UploadFile):
+class _TestUploadFile(UploadFile):
```
graingert: I strongly dislike `__test__ = False` as it violates:

https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers

> System-defined names, informally known as “dunder” names. These names are defined by the interpreter and its implementation (including the standard library). Current system names are discussed in the Special method names section and elsewhere. More will likely be defined in future versions of Python. **Any use of `__*__` names, in any context, that does not follow explicitly documented use, is subject to breakage without warning.**

there's other options here, like renaming to `ExampleUploadFile`, or nesting it under `test_upload_file()`

setup.cfg
```diff
@@ -16,4 +16,18 @@ profile = black
 combine_as_imports = True
 
 [tool:pytest]
-addopts = --cov-report= --cov=starlette --cov=tests -rxXs
+addopts =
+  --cov-report=
```
graingert: any reason for not using

```suggestion
  --cov-report=term-missing
```

or
```suggestion
  --cov-report=term-missing:skip-covered
```

graingert: getting the missing line numbers is handy for running locally https://pytest-cov.readthedocs.io/en/latest/reporting.html

setup.cfg
```diff
@@ -16,4 +16,18 @@ profile = black
...
 [tool:pytest]
-addopts = --cov-report= --cov=starlette --cov=tests -rxXs
+addopts =
+  --cov-report=term-missing:skip-covered
+  --cov=starlette
+  --cov=tests
+  -rxXs
+  --strict-config
+  --strict-markers
+  --cov-config tests/.ignore_lifespan
```
graingert:
```suggestion
  --cov-config tests/.ignore_lifespan
  --cov-fail-under=100
```

setup.cfg
```diff
@@ -16,4 +16,19 @@ profile = black
...
+  --cov-fail-under=100
+xfail_strict=True
+filterwarnings=
+    error
```
graingert, line 29: and there's an unclosed eventloop, and unclosed socket:

```
Exception ignored in: <function BaseEventLoop.__del__ at 0x7f6cfc2e6f70>
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/asyncio/base_events.py", line 654, in __del__
    _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
Exception ignored in: <socket.socket fd=21, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/_pytest/logging.py", line 680, in _runtest_for
    with catching_logs(
ResourceWarning: unclosed <socket.socket fd=21, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
Exception ignored in: <socket.socket fd=22, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/_pytest/logging.py", line 680, in _runtest_for
    with catching_logs(
ResourceWarning: unclosed <socket.socket fd=22, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
```

graingert: https://github.com/pytest-dev/pytest/issues/5299

setup.cfg
```diff
@@ -16,4 +16,19 @@ profile = black
...
-addopts = --cov-report= --cov=starlette --cov=tests -rxXs
+addopts =
+  --cov-report=term-missing:skip-covered
+  --cov=starlette
+  --cov=tests
+  -rxXs
+  --strict-config
+  --strict-markers
+  --cov-config tests/.ignore_lifespan
+  --cov-fail-under=100
```
JayH5: Can we keep the coverage requirement out of this PR? Users can check that by running `scripts/coverage`. I can see how it's nice to have everything in one pytest command, but this then does kind of obsolete `scripts/coverage` and more importantly makes this setup different from other encode projects.

setup.cfg
JayH5: Hrm..we should probably fix these 🤔 curious that they all seem to be warnings for `family=AddressFamily.AF_UNIX`

starlette/middleware/sessions.py
```diff
@@ -64,7 +64,7 @@ async def send_wrapper(message: Message) -> None:
                 elif not initial_session_was_empty:
                     # The session has been cleared.
                     headers = MutableHeaders(scope=message)
-                    header_value = "%s=%s; %s" % (
+                    header_value = "{}={}; {}".format(
```
JayH5: Please can we keep small changes like this and the others (starlette/routing.py, tests/test_staticfiles.py) out of this PR if they're unrelated to the purpose of improving pytest stuff? We try to keep PRs small and very focussed.

JayH5 (commented):
> secondly it's good to find and explicitly list the warnings that you're ignoring, rather than just having them fruitlessly print out at the end of the suite.

Fair enough, although "it's good to" is not a tangible benefit. I do see a benefit to having an explicit list as it means we know what warnings we expect and when a new one shows up we're more aware of it. On the other hand, somebody then needs to maintain the list of known warnings and it can get stale. I could go either way 🤷‍♂️ 

> thirdly you had a SyntaxWarning that until run with a stricter pytest was not noticed.

Where was that?

---
This PR looks like a good change, thanks. IMO, parts of Starlette could do with some TLC, and using our test/lint tools to highlight those areas and keep things neat and tidy is a good idea.

Please try to keep it to a minimal set of changes, though. If the PR does exactly what it says in the title then I'm happy.

starlette/middleware/sessions.py
graingert: This was originally to fix the SyntaxWarning with `pyupgrade`

tests/test_datastructures.py
graingert: Sure

graingert:
> Where was that?

The missing `r"`

graingert:
> On the other hand, somebody then needs to maintain the list of known warnings and it can get stale.

Yeah it's not so bad as you get notified when you need to add new ones, I've got a ticket for xwarns decorator in pytest too

JayH5:
> If the PR does exactly what it says in the title then I'm happy.

😰 let me be clearer: that doesn't mean add more stuff to the title and then the PR is ok, it means this PR has too much stuff in it. Please try to do one thing at a time in a PR. If a PR has an "and" in the title then it's a good sign that it is doing more than one thing. pyupgrade has nothing to do with pytest and I don't see why it should be included here.

I understand that this is more work for you, but squeezing extra stuff into PRs makes more work for the reviewer. We also squash commits when we merge, so PRs that change multiple unrelated things make it harder to track down the reasons for changes in git history.

graingert:
@JayH5 originally these are all related to the increase in strictness of pytest

graingert:
@JayH5 I've split out the black and pyupgrade changes: https://github.com/encode/starlette/pull/1049 https://github.com/encode/starlette/pull/1048

setup.cfg
graingert: this is the only place coverage is checked in CI and I think it's important to enforce coverage in CI.

setup.cfg
graingert: @JayH5 I think I've got a fix for the requests session not being closed

starlette/testclient.py
```diff
@@ -455,11 +455,12 @@ def __enter__(self) -> requests.Session:
         self.receive_queue = asyncio.Queue()  # type: asyncio.Queue
         self.task = loop.create_task(self.lifespan())
         loop.run_until_complete(self.wait_startup())
-        return self
+        return super().__enter__()
```
graingert: this should probably use an ExitStack to manage the loop and the session

starlette/testclient.py
```diff
@@ -455,11 +455,12 @@ def __enter__(self) -> requests.Session:
...
 
     def __exit__(self, *args: typing.Any) -> None:
         loop = asyncio.get_event_loop()
```
graingert: I think this is causing the problem with the loop not closing:

I think a loop context, based on asyncio.run, that allows intermediate blocking IO might be best as this:

```python
import asyncio.runners
improt contextlib

def _get_running_loop(*args, **kwargs):
    try:
        return asyncio.get_running_loop()
    except RuntimeError:
        return None

@contextlib.contextmanager
def new_loop():
    if _get_running_loop() is not None:
        raise RuntimeError(
            "new_loop() cannot be called from a running event loop"
        )

    with exitstack.ExitStack() as stack:
        loop = events.new_event_loop()
        stack.callback(loop.close)

        asyncio.set_event_loop(loop)
        stack.callback(asyncio.set_event_loop, None)

        @stack.callback
        def _shutdown_asyncgens():
            loop.run_until_complete(loop.shutdown_asyncgens())

        stack.callback(asyncio.runners._cancel_all_tasks, loop)
        yield loop
```

graingert:
>  to keep things consistent with other encode projects

I'd recommend using a custom `pytest-encode` plugin that configures all your settings and depends on any other pytest plugins you need across your ecosystem

tests/test_datastructures.py
graingert: @JayH5 can you make a ` ```suggestion ` or push the change before merging? I don't want to bikeshed the name - just do the minimal fix

setup.cfg
graingert: @JayH5 nope, it made it worse, it looks like a lot of code isn't expecting to be closed properly - eg the _ASGIAdaptor needs to override `.close(`


https://github.com/encode/starlette/issues/1050

tests/test_datastructures.py
JayH5:
```suggestion
class BigUploadFile(UploadFile):
```

tests/test_datastructures.py
```diff
@@ -213,13 +213,13 @@ def test_queryparams():
...
     spool_max_size = 1024
 
 
 @pytest.mark.asyncio
 async def test_upload_file():
-    big_file = TestUploadFile("big-file")
+    big_file = _TestUploadFile("big-file")
```
JayH5:
```suggestion
    big_file = BigUploadFile("big-file")
```

setup.cfg
JayH5: Hrm now I see we skip the coverage script in CI. I agree that coverage should be checked in the CI but I'll need to go check why it was skipped in the other project I copied the scripts from.

I still think this is outside the scope of making pytest stricter, though.

JayH5 (commented):
> I'd recommend using a custom pytest-encode plugin that configures all your settings and depends on any other pytest plugins you need across your ecosystem

That might be nice to have, but unless someone has interest in building/maintaining such a thing it is unlikely to happen for now.

setup.cfg
graingert:
```suggestion
```

setup.cfg
JayH5: Sorry, would you mind explaining the `error` line? What is that ignoring? Removing it doesn't seem to make a difference for me when I try locally.

I also don't seem to get that ResourceWarning every time I run the tests 😕

graingert: > Sorry, would you mind explaining the error line?

@JayH5 of course, The lines in pytest.ini/filterwarnings are passed directly to warnings.simplefilter, https://docs.python.org/3/library/warnings.html#the-warnings-filter the `error` is the same as running with `pytest -Werror`, 

> What is that ignoring?

it doesn't ignore anything, it un-ignores warnings because the default is to ignore them

> Removing it doesn't seem to make a difference for me when I try locally.

yes, this is because I fixed the problems that raised warnings

> I also don't seem to get that ResourceWarning every time I run the tests confused

the ResourceWarning is intermittent because it's GC related

JayH5: Thanks for the explanation.

Would it be possible to add a code comment explaining this briefly? Otherwise it could easily be read as "ignore errors".
```suggestion
    # Turn warnings that aren't filtered into exceptions
    error
```

setup.cfg
```diff
@@ -16,4 +16,18 @@ profile = black
...
+  --cov=starlette
+  --cov=tests
+  -rxXs
+  --strict-config
+  --strict-markers
+  --cov-config tests/.ignore_lifespan
+xfail_strict=True
+filterwarnings=
+    error
+    ignore: "@coroutine" decorator is deprecated.*:DeprecationWarning
```
JayH5, line 31:
```suggestion
    # https://github.com/Tinche/aiofiles/issues/81
    ignore: "@coroutine" decorator is deprecated.*:DeprecationWarning
```

setup.cfg
```diff
@@ -16,4 +16,18 @@ profile = black
...
+    ignore: Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated.*:DeprecationWarning
```
JayH5, line 33:
```suggestion
    # https://github.com/graphql-python/graphene/issues/1055
    ignore: Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated.*:DeprecationWarning
```

JayH5 (commented):
Looks good. I've suggested a couple of comments so we can hopefully better keep track of the deprecation warnings. It looks like the context/variable ones we can maybe fix in Starlette, but for a different issue/PR.

I'm not super clear on where comments can go in setup.cfg files so relying on my syntax highlighter to say those are ok.

setup.cfg
```diff
@@ -16,4 +16,21 @@ profile = black
...
 [tool:pytest]
-addopts = --cov-report= --cov=starlette --cov=tests -rxXs
+addopts =
+  --cov-report=term-missing:skip-covered
+  --cov=starlette
+  --cov=tests
+  -rxXs
+  --strict-config
+  --strict-markers
+  --cov-config tests/.ignore_lifespan
```
JayH5: ok ok last thing (I hope 🤞🏼)

This `. ignore_lifespan` file seems to be obsolete. I'll try clean it up in a separate PR.
```suggestion
```

JayH5 (approved):
We got there eventually 🎉 thanks for your patience

PR #1043, opened 2020-09-05
Update CI scripts to match httpcore

JayH5:
Apologies in advance: this is quite a lot of stuff in not many commits. It was a kind of "copy as much as possible verbatim from httpcore and then make it work" task 😬 

Here's a rough summary of all that this changes:
* Copy in all new `scripts/<>` files from httpcore, adjust any existing to match
* Make Github Actions workflows the same as httpcore (these really should be identical now)
* Add a setup.cfg file, mostly just copying from httpcore again
* Add flake8 to requirements and make it pass (mostly line length issues, but a few other small things to fix)
* Add packaging tools to requirements file since they aren't installed in `scripts/publish` anymore
* Update gitignore to be like httpcore, was missing build/dist/site directories


requirements.txt
```diff
@@ -12,6 +12,7 @@ ujson
 autoflake
 black==20.8b1
 databases[sqlite]
+flake8
```
JayH5, line 15: Not doing flake8-bugbear or flake8-pie for now. They're not passing and this PR is big enough already.

setup.cfg
```diff
@@ -0,0 +1,18 @@
...
+ignore = W503, E203, B305
+max-line-length = 88
+
+[mypy]
+disallow_untyped_defs = True
+ignore_missing_imports = True
+
+[mypy-tests.*]
+disallow_untyped_defs = False
+# check_untyped_defs = True
```
JayH5, line 12: This doesn't pass yet. Not sure whether to remove or add a TODO comment or something?

starlette/routing.py
```diff
@@ -520,20 +520,20 @@ async def lifespan(self, scope: Scope, receive: Receive, send: Send) -> None:
         """
         first = True
         app = scope.get("app")
-        message = await receive()
```
JayH5, line 523: Unused variable picked up by flake8.

tests/test_authentication.py
```diff
@@ -130,7 +130,7 @@ async def app(request):
 @app.route("/dashboard/decorated")
 @async_inject_decorator(additional="payload")
 @requires("authenticated")
-async def decorated_sync(request, additional):
+async def decorated_async(request, additional):
```
JayH5, line 133: These two functions were redefinitions which flake8 picked up. I tried to give them better names.

tests/test_routing.py
```diff
@@ -443,7 +443,11 @@ def test_url_for_with_root_path():
 
 
 double_mount_routes = [
-    Mount("/mount", name="mount", routes=[Mount("/static", ..., name="static")]),
+    Mount(
+        "/mount",
+        name="mount",
+        routes=[Mount("/static", ..., name="static")],  # type: ignore[arg-type]
```
JayH5: Example error without this ignore:
```
tests/test_routing.py:449: error: Argument 2 to "Mount" has incompatible type "ellipsis"; expected "Optional[Callable[[MutableMapping[str, Any], Callable[[], Awaitable[MutableMapping[str, Any]]], Callable[[MutableMapping[str, Any]], Awaitable[None]]], Awaitable[None]]]"
Found 1 error in 1 file (checked 19 source files)
```

setup.cfg
florimondmanca: Probably remove this `tests.*` section and log a ticket to "Add type checking to test suite"?

tests/test_routing.py
florimondmanca: (Nit) Should we pass a stub `app(scope, receive, send)` instead?

scripts/build
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh -e
+
+PYTHONPATH=.
```
florimondmanca: Do we need this? (Or was it necessary in HTTPCore eg just so that unasync could work, or something?)

florimondmanca (approved):
Great! :) A few suggestions but approving in advance...

scripts/build
JayH5: Good catch 🤔 I don't think this'll actually do anything without being `export`ed too.

setup.cfg
JayH5: I created #1045 (open issue number 100 😰). I think it makes sense to leave `mypy-tests.*` in here and have mypy run against the tests directory like we do so that it can still check (very basic) things and at least we don't regress.

florimondmanca (approved):

PR #1044, opened 2020-09-06
Added support for graphene middleware

sostholm:
It's now possible to add graphene middleware to GraphQLApp.
Passes graphql tests.


miguescri:
Will this be merged anytime soon?

cglacet:
You need to make sure checks are validated. 

For example here the file is not properly formated (according to [black](https://github.com/psf/black#installation-and-usage)):

```diff
-                middleware=self.middleware
+                middleware=self.middleware,
```

You can see quickly test how black would re-format this using [this online tool](https://black.now.sh/?version=stable&state=_Td6WFoAAATm1rRGAgAhARYAAAB0L-Wj4AJEANxdAD2IimZxl1N_Wlws4TihG40eSpK2ZJaMnjXCK_k8eKwZgejBILX4YmOFWFsxEEYW7qAXvuA4aGY6qRsWJve0Zy6VsO2xryZJsoBLKkWcBFpGlBng0UYjbgtP_ILMenUJUvIChd3ujO1AXpU5rXFxTs2JYpoFFrkLY_ePrxn87i4xVbDED8b3BZS6pkA4fL-3an5OLmj5KnhG-uAWZtTJaQwSw-paXhHyxHDvOtvSbakv2Oo9ncfsxAOkwS2t5twwUfdPn5nlDccFXKewmBPcIqIkCTAcmITkDALRy_cAu6EI7DyAPC0AAfgBxQQAAPHQbkGxxGf7AgAAAAAEWVo=).

You should install black on your machine then use it for format your code before adding a new commit to this merge request.

starlette/graphql.py
```diff
@@ -128,6 +130,7 @@ async def execute(  # type: ignore
                 executor=self.executor,
                 return_promise=True,
                 context=context,
+                middleware=self.middleware
```
cglacet: Same here

starlette/graphql.py
```diff
@@ -136,6 +139,7 @@ async def execute(  # type: ignore
                 variables=variables,
                 operation_name=operation_name,
                 context=context,
+                middleware=self.middleware
```
cglacet: Same here

starlette/graphql.py
```diff
@@ -26,9 +26,11 @@ def __init__(
         executor: typing.Any = None,
         executor_class: type = None,
         graphiql: bool = True,
+        middleware: list = None,
     ) -> None:
         self.schema = schema
         self.graphiql = graphiql
+        self.middleware = middleware
```
cglacet, line 33: Missing coma at the end of line. 

Using comas even in the last entry (dict/list/tuples/function args/etc.) is nice to avoid having diffs in the future when we add a line right after.

For example: 

```
d = {
    'x': 1, 
    'y': 2
}
```

Then someone adds a new entry in the next commit diff you would have:

```diff
d = {
    'x': 1, 
+   'y': 2,
+   'z': 3
}
```

I'm pretty sure that's the reason why black enforces such a rule.

cglacet (changes requested):
[black auto-formater](https://github.com/psf/black#installation-and-usage)

starlette/graphql.py
sostholm: Hello, 

This should be fixed now.
I've added the commas using the black formatter. :)

sostholm: I think line 33 should not have a comma since it's not defining an argument but simply set the middleware of the object from the argument.

cglacet: Right, my bad. That's why automatic checks are useful 👍.

mangeld:
Plz merge this!

sostholm:
Hello, is anything else required from my side to merge this?

JayH5:

PR #1047, opened 2020-09-06
Remove UJSONResponse

### LINKED ISSUE #901: Consider removing UJSONResponse from core
Sorry if this isn’t the place to discuss this. I created this issue for visibility of a [comment](https://github.com/encode/starlette/pull/870#issuecomment-614258113) I made in a pull request to add a `ORJSONResponse` class (similar to the `UJSONResponse` class). I'm gonna quote it here:

> [...] I don't think flavour of the month JSON libraries belong in Starlette core, this applies to `UJSONResponse` as well. Next up someone wants a `SIMDJSONResponse` class[1]. Where should the line be drawn?
> 
> Another aspect to consider is the maintenance burden of implementing custom response classes like this in core, because that means a contract has been signed. So if orjson becomes unmaintained tomorrow, Starlette would have to go through a deprecation cycle to remove it. Or keep it forever... Bugs in the renderer(s) would likely end up as a Starlette issue as well, instead of in the upstream JSON library.
> 
> But enough complaining and on to more constructive feedback. It's evident by this pull request that it's pretty trivial to implement your own custom `JSONResponse` class, so maybe just a section in the documentation describing how to do it is enough? Then someone can (and probably will) publish a package to PyPI that Starlette later can link to, e.g. starlette-orjson. Maybe even a starlette-contrib repo under the encode org? 🤷‍♂️ 
> 
> Another option is to make the default `JSONResponse` more pluggable/configurable, similar to how you can configure renderers in Django REST Framework[2]. Personally, I like this approach.
>
> [...]
> 
> [1] https://github.com/TkTech/pysimdjson
> [2] https://www.django-rest-framework.org/api-guide/settings/#default_parser_classes

JayH5:
Closes #901 

It's easy enough to implement your own UJSONResponse response class if you need it:
```python
from typing import Any

import ujson

class UJSONResponse(JSONResponse):
    def render(self, content: Any) -> bytes:
        return ujson.dumps(content, ensure_ascii=False).encode("utf-8")
```


euri10:
I'm all in for that removal !
Not a blocker on my side but could we add the example you provided (or even better with orjson ) in `docs/responses.md` ?

euri10 (approved):

JayH5:
@euri10 I wrote a bit about custom JSON serialization in `responses.md`. lmk what you think!

euri10:
> @euri10 I wrote a bit about custom JSON serialization in `responses.md`. lmk what you think!

looks perfect ! :+1:

lovelydinosaur:
Fab yup!
Will need calling out in release notes and a median version bump.
Great stuff, thank you!

erewok:
@JayH5, this would have been a good one to include in the `0.14.0` release. Can we merge it and get it released after the fact as `0.14.1`? It's pretty close to the "breaking-change" semver bump and with the recent release of `0.14.0` most users have probably not upgraded yet?

JayH5:
@erewok yeah sorry, I didn't want it to get in the way of all the fixes in "0.13.9" and then when we switched to 0.14 I forgot about it.

Happy to squeeze it in if you want to do that? I won't be able to organise a release right now (haven't actually done one before), but if you want to merge this and do one be my guest.

erewok:
@JayH5, no worries!

I do think we could take advantage of the "breaking change" version bump to merge this (and probably any other test or documentation changes that are easily to include) and make another release. 

Would you be up for merging this and issuing a release? (For reference: https://www.python-httpx.org/contributing/#releasing)

JayH5:
> (and probably any other test or documentation changes that are easily to include)

I don't see any other obvious candidate PRs. Will give this a go.

PR #1049, opened 2020-09-06
allow black to pass on v20.8b1 and v19.10b0

graingert:
by running black v19.3b0 (we miss you) and then running black v20.8b1


JayH5 (approved):
We're not going to support more than one black version but this is a decent cleanup anyway.

PR #1052, opened 2020-09-07
UploadFile readline() and async iterator

polamayster:
Adds `readline()` to `UploadFile` (usecase: read .csv file headers)
Allows to read upload file line by line:
```python
async for line in upload_file:  
   ...
```


tests/test_datastructures.py
```diff
@@ -224,9 +224,31 @@ async def test_upload_file():
...
+    await big_file.write(b"data-line\n" * 11)
+    await big_file.seek(len(b"big-data") * 513 + 1)
+    assert await big_file.readline() == b"data-line\n"
+    lines = [line async for line in big_file]
+    assert len(lines) == 10
+    assert lines[-1] == b"data-line\n"
     await big_file.close()
 
 
+@pytest.mark.asyncio
```
Kludex:
```suggestion
@pytest.mark.anyio
```

Kludex:
The idea sounds reasonable. I had this issue at some point in the past.

I'll wait for other members' opinions.

Thanks @polamayster ! :)

lovelydinosaur:
I'd be up for us accepting `readline()`, yup. I'd accept a pull request with *just* that.

I'm not sure I'm so convinced about the iterator-which-defaults-to-readline. I get that's what python's stdlib does for files, but I think it's a bit of a weird behaviour. (Not 100% saying "no" to that, but at a minimum we oughta consider it separately)

PR #1054, opened 2020-09-10
Add issue templates

JayH5:
Copying the setup from httpx & uvicorn. `config.yml` and the feature request YAML should be identical to httpx, but the bug template was adapted slightly for Starlette.


florimondmanca (approved):
:100:

PR #1055, opened 2020-09-11
Move lifespan tests out of obsolete middleware test module

JayH5:
I think this is a pre-0.13 leftover. Previously the lifespan stuff was implemented using middleware, and when that changed it seems that the tests weren't moved to a new home and the config for coverage to ignore testing of the middleware wasn't removed. i.e. there was a `tests/middleware/test_lifespan.py` but no `starlette/middleware/lifespan.py`.

This PR moves the `Router`-specific lifespan tests out of `test_lifespan.py` into `test_routing.py`, renaming `test_routed_lifespan` -> `test_lifespan`. The remaining tests in `test_lifespan.py` were specific to applications, not routers but were already covered in the existing tests for application lifespan.


euri10:
so if I follow correctly (from old filename to new one)

1. `test_routed_lifespan` becomes `test_lifespan`
2. `test_raise_on_startup` and `test_raise_on_shutdown` stay the same
> The remaining tests in test_lifespan.py were specific to applications, not routers but were already covered in the existing tests for application lifespan.
3. we lose `test_app_lifespan` and `test_app_async_lifespan`. Both use the old on_event, but it seems to me that `test_app_async_lifespan` is lost along the way as the only test using on_event in test_applications is `test_app_add_event_handler` which uses sync startup and shutdown methods and not async ones

JayH5:
> test_app_async_lifespan  is lost along the way as the only test using on_event in test_applications is test_app_add_event_handler which uses sync startup and shutdown methods and not async ones

@euri10 Thanks for the thorough review 😄. The application doesn't really do much with lifespan functions, it just passes them on to the Router, so I didn't see much point in adding more tests there. I did add a test for the Router with async event callbacks, though.

euri10 (approved):
lgtm !
this makes me think those on_event should be deprecated, but FastAPI still using them and cant use `lifespan=` with it, maybe one day/..!

PR #1056, opened 2020-09-11
Fix graphql deprecation warnings

graingert:


starlette/graphql.py
```diff
@@ -123,19 +123,19 @@ async def execute(  # type: ignore
         if self.is_async:
             return await self.schema.execute(
                 query,
-                variables=variables,
+                variable_values=variables,
```
graingert, line 126: you might instead choose to propagate the rename all the way up?

JayH5: 🤔 I'm in two minds about this. On the one hand, graphene/graphql-core seems to have settled on these names with `_value`. OTOH, whether we want to properly support GraphQL in Starlette is unclear (#619) so I'm hesitant to make more than the most minimal changes to make the warnings go away.

How about we _do_ go to the effort of renaming the parameters of this function (`execute(..., variable_values=None, ...`) so that it remains a very small wrapper around calling the schema? But not do anything more than that (although that might be everything there is to do anyway 😅).  It doesn't seem to be a documented Starlette API anyway, so it should be safe to change.

JayH5 (commented):
Thanks for following up on this.

JayH5:

PR #1057, opened 2020-09-12
Use flake8-bugbear

JayH5:
* Try fix all the errors

### DIFF
--- a/tests/test_authentication.py
+++ b/tests/test_authentication.py
@@ -28,7 +28,7 @@ class BasicAuth(AuthenticationBackend):
         try:
             scheme, credentials = auth.split()
             decoded = base64.b64decode(credentials).decode("ascii")
-        except (ValueError, UnicodeDecodeError, binascii.Error):
+        except (ValueError, binascii.Error):
             raise AuthenticationError("Invalid basic auth credentials")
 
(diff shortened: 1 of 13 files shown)


tests/test_authentication.py
```diff
@@ -28,7 +28,7 @@ async def authenticate(self, request):
         try:
             scheme, credentials = auth.split()
             decoded = base64.b64decode(credentials).decode("ascii")
-        except (ValueError, UnicodeDecodeError, binascii.Error):
```
JayH5, line 31: `issubclass(UnicodeDecodeError, ValueError)` => `True`

lovelydinosaur:
Just flagging this up to run past me before we think about commiting. 😃
I've been a bit -0 some of the flake8 extension plugins - I don't think it's necessarily a clear win.

  - Thanks!

JayH5:
@tomchristie got it! I'm mostly aiming for consistency with httpx/httpcore in Starlette.

I'm also a bit [sceptical](https://github.com/encode/httpcore/pull/156/files#r469387220) of some of these flake8 extensions. Think I'm _mostly_ alright with bugbear, but flake8-pie less so 😕 

Think I'll remove flake8-pie for now, but not in a rush to merge this at all.

PR #1062, opened 2020-09-29
Allow to configure 'except_path' on TrustedHost Middleware

riseryan89:
Now you can configure 'except_path' on TrustedHost Middleware. 

This now can allow your cloud service check the server's health with accessing different private IPs which are internally used in VPC.

For example, if you are using Elastic Beanstalk in AWS. Your Load Balanacers keep checking the servers' health with different IPs. If you don't have the except path configuration, LB might regard your servers down.

### DIFF
--- a/starlette/middleware/trustedhost.py
+++ b/starlette/middleware/trustedhost.py
@@ -1,5 +1,6 @@
 import typing
 
+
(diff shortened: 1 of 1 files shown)


riseryan89:
Will re-contribute after tests

PR #1063, opened 2020-09-30
Make scripts compatible with windows

serhatsdev:
You can run shell scripts on windows via `Git Bash` or `cygwin`. So for the sake of making development easier on windows, i added windows compatibility to the scripts.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -1,9 +1,13 @@
 #!/bin/sh -e
 
 if [ -d 'venv' ] ; then
-    PREFIX="venv/bin/"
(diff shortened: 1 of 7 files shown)


Kludex:
@ssw99 Thanks for the interest! :medal_military: 

The development scripts on [encode](https://github.com/encode) projects follow the same ideas, so changing here without changing the others is something that we'd like to avoid.

The right flow of action here would be first to discuss this topic on our [Gitter channel](https://gitter.im/encode/community?at=5fb8d301f5849839ade4f620) to see the other members' opinions. :smile: 

Once again, thank you for the interest! For housekeeping, and the reason I stated in the last paragraph, I'm closing this PR.

PR #1066, opened 2020-10-06
Create tasks from coroutines before providing to asyncio.wait

JayH5:
Not doing this was marked as deprecated in the Python 3.8 documentation, but now produces a DeprecationWarning on Python 3.9.

Does 1/2 of #889

### DIFF
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -1,18 +1,24 @@
 import asyncio
 import functools
+import sys
 import typing
 from typing import Any, AsyncGenerator, Iterator
 
 try:
-    import contextvars  # Python 3.7+ only.
+    import contextvars  # Python 3.7+ only or via contextvars backport.
 except ImportError:  # pragma: no cover
     contextvars = None  # type: ignore
+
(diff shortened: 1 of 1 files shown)


starlette/concurrency.py
```diff
@@ -1,18 +1,24 @@
...
 import typing
 from typing import Any, AsyncGenerator, Iterator
 
 try:
-    import contextvars  # Python 3.7+ only.
+    import contextvars  # Python 3.7+ only or via contextvars backport.
 except ImportError:  # pragma: no cover
     contextvars = None  # type: ignore
 
+if sys.version_info >= (3, 7):  # pragma: no cover
```
florimondmanca, line 12: (Nit) Should we instead try/except with a comment that the except path will be taken on 3.6?

florimondmanca (approved):
Sounds good! Just a bit if you'd like to address it, but otherwise LGTM.

starlette/concurrency.py
JayH5: try/except ImportError has the downside of not being mypy-friendly. But happy to switch to that if that's more the standard we're going with.

PR #1069, opened 2020-10-07
Update responses.md

paxcodes:
- Fix typo ("Third party middleware" to "...responses")
- Fix w3 link (current link leads to 404; I followed links from archive.org to get the "latest" spec)
- Format it similarly to "Third party middleware" (use h4, use the class name as header)


deliseev (approved):

Kludex (approved):
Good catch! :tada: 

Thanks @paxcodes !

PR #1070, opened 2020-10-10
Use zero copy to send file if can

abersheeran:
https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send


Kludex:
@abersheeran Would you mind rebasing it? It looks like I can't do it.

Also, have you thought about implementing the extension support on `uvicorn`? :kissing:

abersheeran:
All right.

PR #1074, opened 2020-10-15
Fix FormData.get

tricosmo:
FormData can have multiple value for a single key, e.g. multiple file upload. In this case `FormData.get(key)` should return a list not just one element.
```
ipdb> from starlette.datastructures import FormData
ipdb> imd = FormData([(1,1),(2,2), (1,3)])
ipdb> imd.get(1)
[1, 3]
```

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -482,7 +482,7 @@ class FormData(ImmutableMultiDict):
         for key, value in self.multi_items():
             if isinstance(value, UploadFile):
                 await value.close()
-    
+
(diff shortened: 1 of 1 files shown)


tricosmo:
Actually this is not required, so close it.

invokermain:
Out of interest why did you decide this was not required?

victorliun:
Sorry, this has been a while. I can't recall it now. Probably I went on another approach.

lovelydinosaur:
You can use `.get_list(<key>)` here.

PR #1075, opened 2020-10-21
Multidict handles list

tchalupnik:
Make Multidict handling list splitted by comma where key finishes with brackets


JayH5:
Hi there, thanks for your work. However, could you please follow the contributing guidelines and create a feature request or ask in the Gitter chat before opening a pull request?

You can already pass multiple values for a query param key like `a=1&a=2&a=3`.

This PR adds behaviour which some specific web frameworks may support but is not part of any HTTP standard, and in my experience is not widely used. As such, I am not in favour of this change without a strong argument. I'm going to close this PR. You're welcome to still try and get support for your idea, but please try and do that using the proper channels.

PR #1076, opened 2020-10-22
Add 418 I'm A Teapot status code

manuelzander:
`418 IM_A_TEAPOT` status code added in Python 3.9: https://docs.python.org/3.9/whatsnew/3.9.html#http and https://docs.python.org/3.9/library/http.html#http-status-codes


JayH5 (approved):
Why not? 😄 🍵

euri10:
:tea: :coffee:

PR #1077, opened 2020-10-23
TestClient: save event loop when entering context manager

em92:
While running test, asyncio.get_event_loop() in `__exit__` method
can return loop, that differs from that one, that is used in `__enter__` method.

As result it may lead to RuntimeError: got Future <Future pending> attached to a different loop.

Here is error report from my usecase:

```
___________________________________________________________________________________ ERROR at teardown of test_nulled_ratings[shiningforces] ____________________________________________________________________________________

    @fixture(scope="session")
    def test_cli():
        from qllr import app
    
        class PatchedTestClient(TestClient):
            def __enter__(self):
                r = super(PatchedTestClient, self).__enter__()
                self.loop = asyncio.get_event_loop()
                return r
    
            def __exit__(self, *args):
                self.loop.run_until_complete(self.wait_shutdown())
    
        with TestClient(app, raise_server_exceptions=False) as client:
>           yield client

tests/conftest.py:168: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.6/site-packages/starlette/testclient.py:462: in __exit__
    loop.run_until_complete(self.wait_shutdown())
/usr/local/lib/python3.6/asyncio/base_events.py:488: in run_until_complete
    return future.result()
/usr/local/lib/python3.6/site-packages/starlette/testclient.py:487: in wait_shutdown
    message = await self.send_queue.get()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Queue at 0x7f5d4f7f2208 maxsize=0 tasks=1>

    @coroutine
    def get(self):
        """Remove and return an item from the queue.
    
        If queue is empty, wait until an item is available.
    
        This method is a coroutine.
        """
        while self.empty():
            getter = self._loop.create_future()
            self._getters.append(getter)
            try:
>               yield from getter
E               RuntimeError: Task <Task pending coro=<TestClient.wait_shutdown() running at /usr/local/lib/python3.6/site-packages/starlette/testclient.py:487> cb=[_run_until_complete_cb() at /usr/local/lib/python3.6/asyncio/base_events.py:200]> got Future <Future pending> attached to a different loop

/usr/local/lib/python3.6/asyncio/queues.py:167: RuntimeError
```

As you may pay attention, that there is also PatchedTestClient with fix, that looks like in this PR. If I use it instead of original TestClient, no error happens.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -455,11 +455,11 @@ class TestClient(requests.Session):
         self.receive_queue = asyncio.Queue()  # type: asyncio.Queue
         self.task = loop.create_task(self.lifespan())
         loop.run_until_complete(self.wait_startup())
+        self.loop = loop
         return self
 
     def __exit__(self, *args: typing.Any) -> None:
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.wait_shutdown())
+        self.loop.run_until_complete(self.wait_shutdown())
 
     async def lifespan(self) -> None:
         scope = {"type": "lifespan"}


JayH5:
@em92 I'm not necessarily opposed to this change, but have you considered changing the scope of your TestClient fixture to match the scope of the event loop fixture or vice-versa? With pytest-asyncio & similar tools, the event loop fixture usually has function scope.

While it might be nice to be able to support TestClient fixtures with longer scopes, this change would mean that we're hanging on to the first test's event loop for the duration of the test suite and I'm not sure of the side-effects of that.

JayH5:
Going to close this out. You can adjust the scope of the client fixture by separating it from your app fixture:
```python
@fixture(scope="session")
def test_app():
    from qllr import app

    return app


@fixture(scope="function")
def test_client(test_app):
    with TestClient(test_app, raise_server_exceptions=False) as client:
        yield client
```

em92:
Hi, @JayH5 ! 
Sorry for long no-response. This PR ran away from my head.

In my case TestClient is required to have session scope. There is other dependent fixture that deploys demo data to database for the test session.

I can show the code itself.
test_cli fixture: https://github.com/em92/quakelive-local-ratings/blob/23f8c73e7e48d92b24327f6d29b5535d51a79be5/tests/conftest.py#L154-L170
dependent fixture: https://github.com/em92/quakelive-local-ratings/blob/23f8c73e7e48d92b24327f6d29b5535d51a79be5/tests/conftest.py#L186-L188
Service class constructor deploys data for almost all tests: https://github.com/em92/quakelive-local-ratings/blob/23f8c73e7e48d92b24327f6d29b5535d51a79be5/tests/conftest.py#L57-L105

If I set "service" fixture's score to function, that would require to rewrite tests to deploy demo data on every test which is excess and will take more time to run tests.

In any case, those tests worked fine before, when I was using my own [backport of starlette for python 3.5](https://github.com/em92/starlette/). Not sure, what caused those tests to break, when I migrated to original starlette.

So, I suggest to reopen this PR and merge it.

PR #1079, opened 2020-10-26
Correct type of AuthenticationBackend.authenticate parameter

dwd:
As far as I can tell from the examples, this is called with a Request, and not an HTTPConnection.


hidaris:
It's called `HTTPConnection` because we can support both `Websocket`, which inherits from `HTTPConnection`

dwd:
Ah! Sorry, missed that.

PR #1087, opened 2020-11-08
Version 0.14.0

erewok:
### Added

* Starlette now officially supports Python3.9.
* In `StreamingResponse`, allow custom async iterator such as objects from classes implementing `__aiter__`.
* Allow usage of `functools.partial` async handlers in Python versions 3.6 and 3.7.
* Add 418 I'm A Teapot status code.

### Changed

* Create tasks from handler coroutines before sending them to `asyncio.wait`.
* Use `format_exception` instead of `format_tb` in `ServerErrorMiddleware`'s `debug` responses.
* Be more lenient with handler arguments when using the `requires` decorator.


euri10:
not a semver pro but shouldn't that be a 0.14.0 ?

JayH5:
I _thiiink_ this just scrapes under the requirements for a version 0.14.0. There shouldn't be any breaking changes and the additions are all slight relaxations of existing interfaces.

Maybe worth adding a point that this release "officially supports Python 3.9"?

florimondmanca:
Thanks!

Agreed, given the changelog entries this sounds like an obvious candidate for 0.14.0. :-)

JayH5:
But..version numbers are free. So happy to do v0.14.0

erewok:
Sounds fine to me: switch to `0.14.0.` (but I reused the same git branch...)

florimondmanca (approved):

JayH5 (approved):

PR #1088, opened 2020-11-08
Version 0.14.1

JayH5:
## Removed
* `UJSONResponse` was removed (this change was intended to be included in 0.14.0).


JayH5:
OK, this will actually have to wait till tomorrow unless somebody want to take it over. 😴 🛌

erewok:
Looks fine to me. I'd probably include a link in the changelog to the documentation that teaches people how to build their own?

docs/release-notes.md
```diff
@@ -1,3 +1,11 @@
+## 0.14.1
+
+November 8th, 2020
+
+### Removed
+
+* `UJSONResponse` was removed (this change was intended to be included in 0.14.0).
```
florimondmanca:
```suggestion
* `UJSONResponse` was removed (this change was intended to be included in 0.14.0). (Pull #1047)
```

florimondmanca (dismissed):

docs/release-notes.md
erewok: This suggestion works too. I think it would be nice provide as many hints here as possible because I'm sure there are people who will upgrade and be surprised (but at least that surprise should come early...).

@florimondmanca  and @JayH5, what about something like this:

```markdown
* `UJSONResponse` has been removed and a note added to the responses documentation instructing users how to construct their own custom JSON Response classes (see: https://www.starlette.io/responses/)
```

If I saw a line like this in a git log or CHANGELOG and I had been hit by this and if I were unsure where to go next, I *think* I'd have enough of a trail of bread crumbs to follow?

docs/release-notes.md
```diff
@@ -1,3 +1,13 @@
+## 0.14.1
+
+November 9th, 2020
+
+### Removed
+
+* `UJSONResponse` was removed (this change was intended to be included in 0.14.0). Please see the
+  [documentation](https://www.starlette.io/responses/#custom-json-serialization) for how to
```
JayH5, line 8: This link should point to the right place in the docs once the release is done.

euri10 (approved):

docs/release-notes.md
```diff
@@ -1,3 +1,13 @@
+## 0.14.1
+
+November 9th, 2020
+
+### Removed
+
+* `UJSONResponse` was removed (this change was intended to be included in 0.14.0). Please see the
```
florimondmanca, line 7: The additional hint is nice; could we please still add the PR number, for the sake of traceability? (This is what I meant in my last suggestion adding `Pull #N`. I find they make it easier to find back the set of changes and provide an additional way to evaluate what actually changed.)

JayH5: Agree. Was going for brevity/similarity to previous changelog entries, but will rather go for (succinct) detail in the future 👍.

erewok (approved):
Looks great. Thanks for doing it.

PR #1089, opened 2020-11-09
Added a note on websockets library requirement

adnan-st:
`uvicorn` version 0.12+ doesn't seem to include a hard dependency on the `websockets` library anymore so it doesn't install it by default but it is needed for the websockets integration. Solution is to specify `uvicorn==0.11.8` or include `websockets` in `pip install`.

Running the example code from `applications.md` will give the following error when connecting from a websocket client without the `websockets` library installed:
```
WARNING:  Unsupported upgrade request.
```

My fix suggestion was to add a note to the `websockets.md` page but it may be better if worded more elegantly and included on (or before) the `applications.md` page because that is where the problem makes its first appearance.


docs/websockets.md
```diff
@@ -86,3 +86,7 @@ correctly updated.
 
 * `await websocket.send(message)`
 * `await websocket.receive()`
+
+#### Note
+
+The `websockets` library is required for this capability but may not automatically be installed depending on your version of `uvicorn`. Install with `pip3 install websockets` if needed.
```
Kludex, line 92:
```suggestion
!!! note
    To use `WebSocket`, be sure to have a websocket package implementation installed. 
    
    In case you're using `uvicorn`, install either [`websockets`](https://github.com/aaugustin/websockets) or [`wsproto`](https://github.com/python-hyper/wsproto).
```

Kludex:
Would you mind checking my comment, @adnan-st ?

Kludex:
Author is absent, so I'm closing this as it's stale.

Anyone willing to take this over, is more than welcome. :)

PR #1090, opened 2020-11-13
Simple support for range headers

Kojiro20:
With this change `FileResponse` can be used to serve media that use the range header such as HTML5 `audio` and `video` tags.

For example, using FastAPI app container you can support range like this:
```python
@app.get("/media/{file_path}")
async def media(file_path: str, range_header: Optional[str] = Header('bytes=0-', alias="Range")):
    full_path = os.path.join('media', file_path)
    start, end = range_header.strip('bytes=').split('-')
    size = os.stat(full_path)[6]
    end = min(size-1, int(start)+1024)
    return FileResponse(full_path, status_code=206, offset=int(start), headers={
        'Accept-Ranges': 'bytes',
        'Content-Range': 'bytes %s-%s/%s' % (start, end, size),
        'Content-Length': str(size)
    })
```

with HTML:
```html
<!DOCTYPE html>
<html>
  <body>
    <audio preload="auto" tabindex="0" controls="" type="audio/mp3">
      <source src="media/tunes.mp3">
    </audio>
  </body>
</html>
```

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -300,7 +300,7 @@ class FileResponse(Response):
         if self.send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
         else:
-            async with aiofiles.threadpool.open(self.path, mode='rb') as file:
+            async with aiofiles.threadpool.open(self.path, mode="rb") as file:
                 await file.seek(self.offset)
                 more_body = True
                 while more_body:
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -235,7 +235,7 @@ class FileResponse(Response):
 
     def __init__(
         self,
-        path: typing.Union[str, "os.PathLike[str]"],
+        path: str,
         status_code: int = 200,
         headers: dict = None,
         media_type: str = None,
@@ -243,6 +243,7 @@ class FileResponse(Response):
         filename: str = None,
         stat_result: os.stat_result = None,
         method: str = None,
+        offset: int = 0
     ) -> None:
         assert aiofiles is not None, "'aiofiles' must be installed to use FileResponse"
         self.path = path
@@ -266,6 +267,7 @@ class FileResponse(Response):
         self.stat_result = stat_result
         if stat_result is not None:
             self.set_stat_headers(stat_result)
+        self.offset = offset
 
     def set_stat_headers(self, stat_result: os.stat_result) -> None:
         content_length = str(stat_result.st_size)
@@ -298,10 +300,8 @@ class FileResponse(Response):
         if self.send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
         else:
-            # Tentatively ignoring type checking failure to work around the wrong type
-            # definitions for aiofile that come with typeshed. See
-            # https://github.com/python/typeshed/pull/4650
-            async with aiofiles.open(self.path, mode="rb") as file:  # type: ignore
+            async with aiofiles.threadpool.open(self.path, mode='rb') as file:
+                await file.seek(self.offset)
                 more_body = True
                 while more_body:
                     chunk = await file.read(self.chunk_size)
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -235,7 +235,7 @@ class FileResponse(Response):
 
     def __init__(
         self,
-        path: str,
+        path: typing.Union[str, "os.PathLike[str]"],
         status_code: int = 200,
         headers: dict = None,
         media_type: str = None,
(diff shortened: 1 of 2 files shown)


Kojiro20:
In case maintainers reject this, for anyone else who needs to support range header queries in FastAPI or Starlette, you can still achieve the same functionality by extending `FileResponse` like this:
```python
class MediaResponse(FileResponse):
    def __init__(self, *args, **kwargs):
        self.offset = kwargs.pop('offset')
        super(MediaResponse, self).__init__(*args, **kwargs)

        if self.stat_result is None:
            try:
                stat_result = await aiofiles.os.stat(self.path)
            except FileNotFoundError:
                raise RuntimeError(f"File at path {self.path} does not exist.")
            else:
                mode = stat_result.st_mode
                if not stat.S_ISREG(mode):
                    raise RuntimeError(f"File at path {self.path} is not a file.")
        await send(
            {
                "status": self.status_code,
                "headers": self.raw_headers,
            }
        )
        if self.send_header_only:
        else:
            async with aiofiles.threadpool.open(self.path, mode='rb') as file:
                await file.seek(self.offset)
                more_body = True
                while more_body:
                    chunk = await file.read(self.chunk_size)
                    more_body = len(chunk) == self.chunk_size
                    await send(
                        {
                            "type": "http.response.body",
                            "body": chunk,
                            "more_body": more_body,
                        }
                    )
        if self.background is not None:
            await self.background()
```

and using like this:
```python
@app.get("/media/{path}")
async def media(path: str, range_header: Optional[str] = Header('bytes=0-', alias="Range")):
    if '..' in path:
        raise Exception(path + ' is not allowed')

    full_path = os.path.join('media', path)
    start, end = range_header.strip('bytes=').split('-')
    start = int(start)
    size = os.stat(full_path)[6]
    end = min(size-1, start+10)
    return MediaResponse(path=full_path, status_code=206, offset=start, headers={
        'Accept-Ranges': 'bytes',
        'Content-Range': 'bytes %s-%s/%s' % (start, end, size),
        'Content-Length': str(size-start)
    })
```

kevinastone:
This has been in the PR queue for quite a while: #1013

Kojiro20:
@kevinastone I wasn't able to use #1013 via monkey-patching or extension, but I agree it is a better solution to enable this out-of-box through staticfiles. I almost didn't make a PR after seeing how long yours had been waiting. However, I wanted to share my work-around since it could be a while before range header support makes it to FastAPI.

This solution could also enable other functionality that is currently lacking. Chunked file transfer or anything that uses 206 to resume is currently impossible (or incredibly slow using StreamResponse) without some trickery.

kevinastone:
@Kojiro20 Here's a copy of the monkey-patch I'm currently running to support range requests: https://gist.github.com/kevinastone/a6a62db57577b3f24e8a6865ed311463

Kojiro20:
@kevinastone thank you! Will use that if I do another server project in python.

But, TBH, will probably stick with node or java for a few more years after this experience. It's too hard to get real http2 support out of python and I'm very confused about all the busted/incompatible versions for asgi tooling. Even Conda not solved dependency tree for fastAPI + h2 etc 😖 ... Will never touch wsgi again, but solid asgi support feels nascent at best. Watching for python to natively support promises 🤔 – that was about the time node seemed mature.

Seems like Guido letting go of the helm was a mixed blessing. Now everyone is fearless to break APIs 🥳 but slow to adopt upstream changes 🥉

Kojiro20:
For posterity - starlette has a memory leak with `FileResponse`.

I think the fix would involve adding a watchdog of some sort inside the `async with open (...` block to clean up unused connections which haven't sent new data within some timeout. It's currently possible for a client to request data from two files, then ignore one of them while continuing to read from the other. When this happens the ignored file is held open indefinitely. The connection is still open and valid, but nothing cleans up the handle/memory for files which weren't fully transferred.

Honestly curious if anyone runs a python service that doesn't need to be rebooted periodically to stay afloat. I have added something like this to my Dockerfile to mitigate.
```
RUN echo "0 * * * * /usr/bin/pkill -f server.py && /usr/bin/python3 /server.py &" >> /etc/crontab
```

gimebreak:
> @Kojiro20 Here's a copy of the monkey-patch I'm currently running to support range requests: https://gist.github.com/kevinastone/a6a62db57577b3f24e8a6865ed311463

have you done a PR? I think they should update this block of code

somebuckaroo:
> @Kojiro20 Here's a copy of the monkey-patch I'm currently running to support range requests: https://gist.github.com/kevinastone/a6a62db57577b3f24e8a6865ed311463

Not sure if this is still the preferred solution for ranged support for static files, but I think there's an off-by-one error in OpenRange. This is easily verified with curl.

`curl -v -r 1000- http://127.0.0.1:xxxx/somefile`

PR #1091, opened 2020-11-14
 Graphql subscription support added

ghost:
Hi,

I've upgaded graphene to v3 and added support for GraphQl subscriptions. 

That allows one to use subscribe to graphql via websocket.

AsyncioExecutor gets deprecated as Graphene dropped support for the executor argument. I've still added the library from graphene v2.3.2 for someone interested in modifying the code.


ghost:
I've dropped AsyncioExecutor and added a variable to switch between graphene's v3 Schema.execute_async and Schema.execute.

coryvirok:
@NatachaMeow nice! Looking forward to having this merged!

I've been working around this with a monkey patch. You'll likely need to change the `format_graphql_error` import [here](https://github.com/encode/starlette/blob/master/starlette/graphql.py#L13). graphql-core 3 no longer has that method but I've found that monkey patching in [format_error](https://github.com/graphql-python/graphql-core/blob/master/src/graphql/error/graphql_error.py#L199) works.

JayH5:

PR #1095, opened 2020-11-20
Add 208 Already Reported status code

r-m-n:


euri10 (approved):

euri10:
thanks @r-m-n , TIL 208 !

lovelydinosaur:
If we want an authoritative list then we should probably be reaching for https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml right? (I didn't know that existed until https://tools.ietf.org/html/rfc5842 mentioned "the HTTP Status Code Registry")

PR #1096, opened 2020-11-20
Add missing status codes

r-m-n:


euri10 (approved):
lgtm all that were not in https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml are in

418 is unassigned in that list though, I remember we added it recently

lovelydinosaur:
Yup I'm okay with this. At least it's a single definitive place we can point to now.

euri10:
thanks @r-m-n !

PR #1097, opened 2020-11-21
add typing overloads for Config.__call__

sbdchd:
This allows for more precise return types instead of the current `Any`.

We have 3 overload cases here:

1. handles cases where the user provides an explicit `cast` argument.

   ```python
   reveal_type(config("POOL_SIZE", cast=int))
   # note: Revealed type is 'int'

   reveal_type(config("DEBUG", cast=bool, default=None))
   # note: Revealed type is 'Union[bool, None]'
   ```

2. handles no cast or default being passed

   ```python
   reveal_type(config("FOO"))
   # note: Revealed type is 'str'
   ```

3. handles no cast being provided, and the default not being `str`

   ```python
   reveal_type(config("DB_NAME", default=None))
   # note: Revealed type is 'Union[str, None]'

   reveal_type(config("FOO", default=False))
   # note: Revealed type is 'Union[str, bool]'
   ```

rel: https://gitter.im/encode/community?at=5fb88fdf33d7764bde811c46


Kludex (approved):
@sbdchd  Sorry for taking so long!

Thanks for the PR! :tada: 
According to this [message on Gitter](https://gitter.im/encode/community?at=5fb8d301f5849839ade4f620), @florimondmanca was positive about this, and I'm too.

PR #1109, opened 2020-12-08
add basic timeout support for testclient

### LINKED ISSUE #1108: TestClient.request does not honor timeout
### Checklist

### Describe the bug

`timeout` parameter exposed in `testclient.request` has no effect. Looking through the code, it seems as though we use the same interface as `requests` package but only a subset of those features are supported by `testclient`

### To reproduce

```python
import time

mock_service = Starlette()


@mock_service.route("/slow_response")
def slow_response(request):
    time.sleep(0.01)
    return JSONResponse({"mock": "slow example"})


def test_timeout():
    """ no timeout exception raised -- this test passes when it should fail """
    client = TestClient(mock_service, raise_server_exceptions=True)
    response = client.get("/slow_response", timeout=0.001)
    assert response.json() == {"mock": "slow example"}
```

### Expected behavior

A TimeoutException should be raised if the timeout specified in request is shorter than the time the request takes.

### Actual behavior

No timeout exceptions are raised when setting timeout.

### Additional context

In CI/CD context, testing our own endpoints which, during integration testing, may timeout, is very helpful. If this is unsupported, it might be nice to have a list of differences between the interface of requests/testclient (this might include #1102).

I did look a bit how the requests package implemented their timeout exceptions, and to be completely honest I got a bit lost going down the rabbit hole of `requests` -> `urllib3` -> `http.client`.  Since we're using a different adapter for the testclient, it might be some work to reimplement all the way down the timeout exceptions in the same way.

I did notice that the `async_asgi_testclient` works with a timeout. Of course, it requires an async endpoint. But [their implementation](https://github.com/vinissimus/async-asgi-testclient/blob/f42d1f8ce9fd826d80dbe35e3ba123166c6a2b44/async_asgi_testclient/utils.py#L111-L113) might still be useful? 

Found that `httpx` had a similar issue with timeout having no effect when mounting an app with its client.

falkben:
Closes https://github.com/encode/starlette/issues/1108

Note, this does not implement the separate `connection` and `read` timeouts that `requests` supports (will raise a `ValueError` if user supplies a `tuple`. Besides the added complexity of dealing with that, it also didn't make sense to me that there would be a connection timeout when testing a local mounted app.

Not sure if something is needed in the docs about timeout, but maybe I could if desired? https://github.com/encode/starlette/blob/ben%2Ftestclient_timeout/docs/testclient.md


tests/test_testclient.py
```diff
@@ -132,3 +146,20 @@ async def asgi(receive, send):
     with client.websocket_connect("/") as websocket:
         data = websocket.receive_json()
         assert data == {"message": "test"}
+
+
+@pytest.mark.parametrize("endpoint", ["/slow_response", "/async_slow_response"])
+def test_timeout(endpoint):
+    client = TestClient(mock_service, raise_server_exceptions=True)
```
Kludex, line 153: Should we test with `raise_server_exceptions=False` or is it already covered somewhere else?

falkben: Default is `raise_server_exceptions=True`. 

I should perhaps add a test (check for 500 status code) when set to `False`...

Kludex (dismissed):

falkben:
Wondering if there's anything more to do for this? Thanks.

JayH5:
Hey 👋 this seems like a reasonable change, but I have a couple of questions/points:
1. Can you explain more about why you want this functionality? I see in #1108 you mention timing out in CI/CD. Can you explain what you're trying to solve exactly? Is it to prevent tests from hanging?
2. I think the behaviour should be that a timeout error is *always* raised if one occurs, regardless of the value of the `raise_server_exceptions` flag. The flag is for *server* errors, and this timeout is a *client* error.

falkben:
> 
> 
> Hey 👋 this seems like a reasonable change, but I have a couple of questions/points:
> 
>     1. Can you explain more about why you want this functionality? I see in #1108 you mention timing out in CI/CD. Can you explain what you're trying to solve exactly? Is it to prevent tests from hanging?
> 
>     2. I think the behaviour should be that a timeout error is _always_ raised if one occurs, regardless of the value of the `raise_server_exceptions` flag. The flag is for _server_ errors, and this timeout is a _client_ error.

1. Yes, to prevent tests from hanging. Better to fail them if they take too long.
2. I see your point, but I think the line between client/server here is not so clear cut. The client isn't timing out (since the app is mounted directly into the client, it *cannot* time out).  It's the server that's timing out. But, perhaps it's something that should be cleaned up somehow, since it is a bit confusing.

Kludex:
Hi @falkben! :wave: 

Would you mind rebasing it?

falkben:
I started this process in a diff branch but unclear where to put something like `fail_after` from anyio. I get an error message about not being in an async context where I have the fail_after currently, which I thought I would be inside the portal_factory context. Do you have any suggestions? 

https://github.com/encode/starlette/compare/master...falkben:ben/testclient_timeout_rebase?expand=1

lovelydinosaur:
I'm not convinced we necessarily want timeout behaviour in the test client, but either way around it's probably not great timing given #1376 - might make sense to consider this only after that's in?

Kludex:
I'm more inclined towards #1446, which is exclusive considering this one (as the parameter name is the same).

We can consider this PR with another parameter, after #1376, tho.

lovelydinosaur:
I'm going to put on my maintainer-pushback hat here and say "the TestClient doesn't support timeouts".
At least for now, and possibly for the long term.

Existing WSGI frameworks haven't generally supported this, and needing them almost certainly points to flaky tests.

I'd be up for docs pull requests around this if anyone's motivated to do so.

PR #1110, opened 2020-12-09
Document the lifespan event handler parameter

emilmelnikov:
https://github.com/encode/starlette/pull/799 added support for the `lifespan` parameter, but it was not reflected in the documentation.


docs/events.md
```diff
@@ -37,6 +37,25 @@ registered startup handlers have completed.
 The shutdown handlers will run once all connections have been closed, and
 any in-process background tasks have completed.
 
+A single lifespan asyncronous generator handler can be used instead of
```
graingert:
```suggestion
A single lifespan asynccontextmanager handler can be used instead of
```

graingert (changes requested):
lifespan now takes an async context manager factory, eg an async generator function decorated with `@contextlib(2).asynccontextmanager`

docs/events.md
```diff
@@ -37,6 +37,25 @@ registered startup handlers have completed.
...
+separate startup and shutdown handlers:
+
+```python
+from starlette.applications import Starlette
+
+
+async def lifespan(app):
```
graingert, line 50:
```python
@contextlib.asynccontextmanager
async def lifespan(app):
    # acquire async resources
    async with anyio.create_task_group() as app.tg:
        try:
            yield
        finally:
            with anyio.CancelScope(shield=True):
                # release async resources
```

docs/events.md
```diff
@@ -37,6 +37,33 @@ registered startup handlers have completed.
...
+
+@contextlib.asynccontextmanager
+async def lifespan(app):
+    # acquire async resources
+    async with anyio.create_task_group() as app.tg:
+        try:
+            yield
+        finally:
+            with anyio.CancelScope(shield=True):
+                # release async resources
```
graingert: sorry I lead you wrong here - the TaskGroup context manager will capture any CancelledError for you

```suggestion
    async with anyio.create_task_group() as app.tg:
        yield
    # release async resources
```

graingert: I do want to somehow document that `yield` can now throw CancelledError

graingert: sometimes you want to acquire/release resources inside the TaskGroup and sometimes you want to acquire/release them outside the TaskGroup and there's no specific answer as they both have different behaviours with their own pros/cons

image

uSpike: I'm not sure we should have the main documentation for lifespan be setting `app.tg`.  That could be a separate topic, but I think the initial introduction to "make a lifespan" example should be as simple as possible.

```python
@contextlib.asynccontextmanager
async def lifespan(app):
    async with some_async_resource():
        yield
```

graingert: that's fair

graingert:
Blocks https://github.com/encode/starlette/pull/1233

docs/events.md
emilmelnikov: I like the simple lifespan example; we can also add a link to [anyio documentation](https://anyio.readthedocs.io/) on task groups and cancellation.

graingert (approved):

graingert:
Thanks! Apologies for dragging this one off topic in my original review

PR #1111, opened 2020-12-11
Add Origin to Vary header on credentialed CORS response

jcwilson:
According to the [MDN CORS docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Access-Control-Allow-Origin), the `Origin` item should be added to the `Vary` header list when
the `Access-Control-Allow-Origin` is set to an explicit origin value when it could change due to
something like `allow_origins` being the `*` wildcard, a multi-item whitelist or `allow_origin_regex`
being in use.

>If the server specifies a single origin (that may dynamically change based on the requesting origin
as part of a white-list) rather than the "*" wildcard, then the server should also include Origin in
the Vary response header — to indicate to clients that server responses will differ based on the
value of the Origin request header.

The existing code fails to update the `Vary` list when the server is configured to allow all
origins (`*`) and the request has a `Cookie` header (ie. credentialed). In that situation, the
`Access-Control-Allow-Origin` header will be set to the request's `Origin` value. It should
be noted that the code does currently update the `Vary` list when the allowed origins are
defined by an explicit whitelist (even if it only has one value and doesn't actually vary) or
a regex pattern.

It appears this may have just been a simple oversight in the original implementation. This updates
the code to add `Origin` to the `Vary` header under these circumstances. If it was intentionally
omitted, I'd be delighted to learn why.


tests/middleware/test_cors.py
```diff
@@ -245,12 +245,30 @@ def homepage(request):
     assert response.headers["vary"] == "Origin"
 
 
-def test_cors_vary_header_is_properly_set():
+def test_cors_vary_header_is_properly_set_for_credentialed_request():
```
euri10, line 266: should we add another `test_cors_vary_header_is_properly_set_for_not_credentialed_request` ? and have the same setup as here without the Cookie sent ?

jcwilson:
So i added that test, but for some reason the Python 3.8 tests are failing all over the place. If this code is causing that, I'm at a loss to explain how.

jcwilson:
@euri10 This is the other trivial one dealing with CORS improvements.

I think #1113 could use some attention, too. It'd be nice to see all of these CORS updates ship together at some point.

euri10 (approved):

JayH5:
These PRs all look good, thank you. Let's get them merged. I'll need a bit more time to fully understand #1113, though.

PR #1112, opened 2020-12-11
Add HEAD to CORS ALL_METHODS list

jcwilson:
The HEAD method is conspicuously absent from the allowed methods list when `allow_methods="*"` is
used. This doesn't really affect CORS preflight requests, as HEAD requests aren't preflighted by the
browser, but it does prevent the actual cross-origin HEAD response from being read by the calling
app.
[This can catch people off-guard.](https://discuss.encode.io/t/for-cors-middleware-why-is-head-not-included-in-all-methods/939)

This simply adds HEAD to the `ALL_METHODS` list in the CORS middleware module and includes some
additional tests to validate the new behavior.


falkben:
It does seem to be specified in the spec. https://fetch.spec.whatwg.org/ Do you know of any reason it might have been missing?

jcwilson:
Nope, I just discovered it when I found that the `Access-Control-*` response headers weren't present on HEAD requests. I'm not sure why it was omitted in the first place.

tests/middleware/test_cors.py
```diff
@@ -118,6 +118,56 @@ def homepage(request):
...
+    def homepage(request):
+        return PlainTextResponse("Homepage", status_code=200)
+
+    client = TestClient(app)
+
+    headers = {"Origin": "https://example.org"}
+
+    for method in ("delete", "get", "head", "options", "patch", "post", "put"):
+        response = getattr(client, method)("/", headers=headers, json={})
+        assert response.status_code / 100 == 2
```
euri10: any reason to return a 200 and assert on 200+ ?

euri10:
```suggestion
        assert response.status_code == 200
```

jcwilson: I think in an earlier iteration i had my test method returning more method-appropriate status codes (like 201 for POST). I can switch it to be more explicit.

jcwilson:
Hmm, this one's failing in the same way as #1111 with I think errors in the database code.

jcwilson:
@euri10 I forgot about this one. Looks like tests are passing now.

euri10 (approved):

PR #1113, opened 2020-12-11
Set explicit Origin in CORS preflight response if allow_credentials is True and allow_origins is wildcard

jcwilson:
When a non-simple request will include credentials, the browser will likely send a CORS preflight request.
The browser will fail the request unless the preflight response explicitly allows the request's `Origin` in the
`Access-Control-Allow-Origin` response header rather than `*`.

This can be resolved by updating the CORS middleware to respond to preflight requests with the explicit
origin from the request rather than the literal `*` wildcard value when `allow_credentials=True`.

The current workaround is to provide an equivalently permissive `allow_origin_regex` pattern like `".*"`,
but this would appear to be a [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) violation. If one wishes to allow calls from all origins, regardless of the
`allow_credentials` setting, then `allow_origins=["*"]` should be sufficient to express that intent.

The `simple_response()` code already performs similar logic which currently only applies to non-preflight
requests:

```
# If request includes any cookie headers, then we must respond
# with the specific origin instead of '*'.
if self.allow_all_origins and has_cookie:
    headers["Access-Control-Allow-Origin"] = origin
```

This PR just updates the preflight logic to match the simple request logic. The difference is that preflight
requests won't contain cookies so we can't perform the same `has_cookie` check. That's why we'll rely
on the `allow_credentials` setting.

See the relevant MDN CORS docs [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Access-Control-Allow-Credentials):
>When used as part of a response to a preflight request, this indicates whether or not the actual request can
be made using credentials. Note that simple GET requests are not preflighted, and so if a request is made for
a resource with credentials, if this header is not returned with the resource, the response is ignored by the
browser and not returned to web content.

**Note: This could be considered an interface-breaking change and demand a stronger version bump if starlette
follows a strict semver policy.** Servers that currently specify `allow_origins="*"` instruct the browser to
block credentialed CORS calls that require preflight authorization. Adopting this behavior would allow these
calls through and could result in an unexpected relaxation of their current effective security rules. However, it
looks like starlette is still pre-1.0.0, so maybe this just means a minor version bump?


jcwilson:
This is the error one currently gets when attempting to make a CORS request with credentials and `allow_origins=["*"]`

image

jcwilson:
@JayH5 thanks for taking care of the other CORS PRs. Let me know if there's anything I can do to help out with this one. I know CORS is such a headache to context switch to.

jcwilson:
Actually, I think this one could be updated to use the new `allow_explicit_origin()` method. I'll take a stab at that.

starlette/middleware/cors.py
```diff
@@ -131,9 +132,9 @@ def preflight_response(self, request_headers: Headers) -> Response:
         # if we do.
         if failures:
             failure_text = "Disallowed CORS " + ", ".join(failures)
-            return PlainTextResponse(failure_text, status_code=400, headers=headers)
+            return PlainTextResponse(failure_text, status_code=400, headers=dict(headers))
 
-        return PlainTextResponse("OK", status_code=200, headers=headers)
+        return PlainTextResponse("OK", status_code=200, headers=dict(headers))
```
jcwilson: Ugh, I don't like this. `Response` marks `headers` as type `dict`. I could go and update all of the `Response` classes to take `Mapping[Any, Any]` or `Mapping[str, str]`, or even some `Union` type with `Mapping` and `dict`, but that seems way out of scope for this PR.

Maybe I should just revert this and not use `allow_explicit_origin()` in the preflight case, since that's what requires a `MutableHeaders` object?

starlette/middleware/cors.py
```diff
@@ -30,27 +30,32 @@ def __init__(
...
+            preflight_headers["Access-Control-Allow-Origin"] = "*"
         preflight_headers.update(
             {
                 "Access-Control-Allow-Methods": ", ".join(allow_methods),
                 "Access-Control-Max-Age": str(max_age),
             }
         )
         allow_headers = sorted(SAFELISTED_HEADERS | set(allow_headers))
-        if allow_headers and "*" not in allow_headers:
+        if allow_headers and not allow_all_headers:
```
jcwilson, line 58: isn't `allow_headers` redundant here since it's never an empty list? This could be reduced to this, right?:

```
if not allow_all_headers:
    ...
```

JayH5 (commented):
@jcwilson been meaning to say: this looks good to me, but I don't feel that I have enough experience with CORS to say what the best practices are, especially w.r.t. `allow_origins="*"`. If another member of @encode/maintainers is happy to sign-off on this then so am I. Alternatively, if you can find precedence for this behaviour in any commonly used libraries, especially in the Flask or Django space, that would help.

jcwilson:
> Alternatively, if you can find precedence for this behaviour in any commonly used libraries, especially in the Flask or Django space, that would help.

Sure thing. Let me get back to you on that. I've been working more closely with the envoy-proxy implementation lately, but I'll see what else I can dig up.

jcwilson:
I don't believe [envoy-proxy](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/cors/cors_filter.cc#L92...L93) will ever send back `*` on preflight requests, regardless of the allow-credentials setting.

Django via [django-cors-headers](https://github.com/adamchainz/django-cors-headers/blob/main/src/corsheaders/middleware.py#L131...L134) performs nearly identical logic to what this PR contains.

Flask-CORS seems to over-complicate things a bit, but you'd have to specifically force it to try to send the literal `*` value and even then it [checks for that and prevents that misconfiguration](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L368...L373). The actual behavior when configured with `origins="*"` doesn't even look at whether credentials are allowed or not, [it will just reflect back the request's origin](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L133). (If one were to set `send_wildcard` to attempt to get it to send `*`, it would trigger the above error before getting to [that point](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L125).)

JayH5 (approved):
Cool, I'm happy with this, especially with the `django-cors-headers` precedent.

JayH5:
Thank you!

jcwilson:
Thank you, too. And @euri10

PR #1128, opened 2021-01-19
Add compresslevel to GZipMiddleware

andamound:
Default level = 9

Documentation gzip: https://docs.python.org/3/library/gzip.html
Benchmarks: https://tukaani.org/lzma/benchmarks.html


starlette/middleware/gzip.py
```diff
@@ -6,29 +6,36 @@
 
 
 class GZipMiddleware:
-    def __init__(self, app: ASGIApp, minimum_size: int = 500) -> None:
+    def __init__(
+        self, app: ASGIApp, minimum_size: int = 500, compresslevel: int = 1
```
ycd: You should keep it at 9, which is the default value for [GzipFile](https://docs.python.org/3/library/gzip.html#gzip.GzipFile), or this is going to break a lot of existing code.

starlette/middleware/gzip.py
```diff
@@ -6,29 +6,36 @@
...
+                    self.app, self.minimum_size, compresslevel=self.compresslevel
+                )
                 await responder(scope, receive, send)
                 return
         await self.app(scope, receive, send)
 
 
 class GZipResponder:
-    def __init__(self, app: ASGIApp, minimum_size: int) -> None:
+    def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 1) -> None:
```
ycd: Also this one.

ycd (changes requested):

starlette/middleware/gzip.py
andamound: I agree. Ready

Kludex (approved):

Kludex:
@andamound Thanks for your PR! :tada: 

Just for the record, I think the `GZipResponder` class should **not** have a default value on `compresslevel`, as it's meant to be used by the `GZipMiddleware`. But as we expose the API of that module as is, I think it's better to not add breaking changes unnecessarily.

Issue #1126, opened 2021-01-18
The debug mode does not work well


ghost:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

The application shows plain text instead of html page in debug mode when an error occurs.

### To reproduce

I am trying to use the debug mode. Here is my application:

    from starlette.applications import Starlette
    from starlette.responses import JSONResponse
    from starlette.routing import Route

    async def show_index(request):
        return JSONResponse({'hello': current_user})

    app = Starlette(debug=True, routes=[
        Route('/', show_index)])

I start the server this way:

    import uvicorn

    if __name__ == '__main__':
        uvicorn.run(
            'main:app', host='127.0.0.1',
            reload=True, port=5000, log_level='info')

### Expected behavior

I expect an exception and the html page in my browser with debug information, but it shows plain text:
'Internal Server Error'.

### Actual behavior

Browser shows plain text: 'Internal Server Error'

### Debugging material

This is the traceback I see in terminal:

    Traceback (most recent call last):
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
        await self.app(scope, receive, _send)
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
        raise exc from None
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
        await self.app(scope, receive, sender)
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/routing.py", line 582, in __call__
        await route.handle(scope, receive, send)
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/routing.py", line 243, in handle
        await self.app(scope, receive, send)
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/starlette/routing.py", line 54, in app
        response = await func(request)
      File "/home/avm/workspace/abug/main.py", line 7, in show_index
        return JSONResponse({'hello': current_user})
    NameError: name 'current_user' is not defined

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "/home/avm/workspace/abug/venv/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 394, in run_asgi
[...]

PR #1132, opened 2021-01-30
Fix middleware traceback fetching on Python 3.8+, fix ResourceWarnings in TestClient, fix CI build

### LINKED ISSUE #1050: fix Unraisable ResourceWarnings
Hrm..we should probably fix these 🤔 curious that they all seem to be warnings for `family=AddressFamily.AF_UNIX`

_Originally posted by @JayH5 in https://github.com/encode/starlette/pull/1040#discussion_r484068578_

### LINKED ISSUE #1131: Test failure
### Checklist

### Describe the bug

Test failure on NixOS and Fedora during the package build process.

```
============================= test session starts ==============================
platform linux -- Python 3.8.7, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: /build/source, configfile: setup.cfg
plugins: asyncio-0.14.0, cov-2.10.1
collected 240 items                                                            
[...]
=================================== FAILURES ===================================
_______________________________ test_debug_html ________________________________

    def test_debug_html():
            raise RuntimeError("Something went wrong")

        app = ServerErrorMiddleware(app, debug=True)
        response = client.get("/", headers={"Accept": "text/html, */*"})
>       assert response.headers["content-type"].startswith("text/html")

tests/middleware/test_errors.py:42:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = {}, key = 'content-type'

    def __getitem__(self, key):
>       return self._store[key.lower()][1]
E       KeyError: 'content-type'

/nix/store/9bb3iygm8kkdzfsaj3j1j2q845nzja1p-python3.8-requests-2.25.0/lib/python3.8/site-packages/requests/structures.py:54: KeyError
```
Fedora uses different releases

```bash
============================= test session starts ==============================
platform linux -- Python 3.9.1, pytest-6.0.2, py-1.10.0, pluggy-0.13.1 -- /usr/bin/python3
cachedir: .pytest_cache
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/python-starlette/starlette-0.14.1, configfile: setup.cfg
plugins: cov-2.10.1, asyncio-0.14.0, requests-mock-1.7.0, httpx-0.10.0, mock-3.3.1, benchmark-3.2.3, aresponses-2.0.0, forked-1.3.0, xdist-1.34.0, django-3.8.0, vcr-1.0.2, datafiles-2.0, pyfakefs-3.5.8, httpbin-1.0.0, sugar-0.9.3, aiohttp-0.3.0
collecting ... collected 187 items
[...]
```

### To reproduce

Run with enabled unit tests.

### Expected behavior

All tests passing.

### Environment

- OS: NixOS and Fedora 
- Python version: 3.8.7 
- Starlette version: 0.14.1 but as also present in 0.13.8

JayH5:
This would ideally be 3 separate PRs, but all of these issues have broken the build on master so it'd be difficult to demonstrate them working separately. Still, let me know if you'd like me to split things up.

The issues are:
1. Mypy 0.800 has some changes w.r.t module discovery. The mypy run was failing because mypy wanted the test functions in `tests/middleware/*` to have type annotations, despite this being disabled for `tests` submodules. Mypy was not correctly identifying that these tests were submodules due to the lack of an `__init__.py` file. See somebody having a similar problem in python/mypy#9974. An alternative (or something we could _also_ do) would be to add `namespace_packages = True` to the config. Since we have `__init__.py` files everywhere else, I went with that option for now.
2. Something in a dependency (pytest? Python itself?) changed so that we were having a bunch more `ResourceWarning`s detected due to unclosed sockets. These (Unix) sockets I believe were setup by the asyncio event loop as part of its internal machinery. In the `TestClient`, in the `WebSocketTestSession` we weren't closing the event loop used by the worker thread there. I changed the loop creation to be done inside the thread (since it is only used by the thread, and can never be used again once the thread has completed) and closed the loop at the end of its use (also in the thread). Possibly fixes #1050 
3. In `ServerErrorMiddleware` we were using the undocumented `exc_traceback` attribute of `TracebackError`. This attribute was removed for [bpo42482](https://bugs.python.org/issue42482) in Python 3.9.1 & 3.8.7. I switched to using the `__traceback__` attribute of exceptions which I think should have the same value (it at least produces the same HTML for the test case we have). I also added a check that the exception actually had a traceback, which we weren't doing previously but probably should've been. Fixes #1131, #1126

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -217,15 +217,15 @@ class ServerErrorMiddleware:
         traceback_obj = traceback.TracebackException.from_exception(
             exc, capture_locals=True
         )
-        frames = inspect.getinnerframes(
-            traceback_obj.exc_traceback, limit  # type: ignore
-        )
 
         exc_html = ""
         is_collapsed = False
-        for frame in reversed(frames):
-            exc_html += self.generate_frame_html(frame, is_collapsed)
-            is_collapsed = True
+        exc_traceback = exc.__traceback__
+        if exc_traceback is not None:
+            frames = inspect.getinnerframes(exc_traceback, limit)
+            for frame in reversed(frames):
+                exc_html += self.generate_frame_html(frame, is_collapsed)
+                is_collapsed = True
 
         # escape error class and text
         error = (
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -268,7 +268,6 @@ class WebSocketTestSession:
         self.app = app
         self.scope = scope
         self.accepted_subprotocol = None
-        self._loop = asyncio.new_event_loop()
         self._receive_queue = queue.Queue()  # type: queue.Queue
         self._send_queue = queue.Queue()  # type: queue.Queue
         self._thread = threading.Thread(target=self._run)
@@ -293,13 +292,16 @@ class WebSocketTestSession:
         """
         The sub-thread in which the websocket session runs.
         """
+        loop = asyncio.new_event_loop()
         scope = self.scope
         receive = self._asgi_receive
         send = self._asgi_send
         try:
-            self._loop.run_until_complete(self.app(scope, receive, send))
+            loop.run_until_complete(self.app(scope, receive, send))
         except BaseException as exc:
             self._send_queue.put(exc)
+        finally:
+            loop.close()
 
     async def _asgi_receive(self) -> Message:
         while self._receive_queue.empty():


euri10 (approved):
> Something in a dependency (pytest? Python itself?)

It's pytest itself (since 6.2 https://docs.pytest.org/en/stable/usage.html#unraisable) and if you look it happens only on 3.8 and above as expected.
Since we turn warnings into exceptions then this popped off.
Uvicorn has a pending PR to do the same.

florimondmanca (approved):
Thanks so much ❤️

florimondmanca:
@JayH5 I updated the title bc there are 2 items we'll want this PR to be included in the changelog for (fixing tracebacks, fixing `ResourceWarning` instances).

florimondmanca:
Depending on what's currently in `master` I guess we can follow up with a bug fix release?

JayH5:
Thanks @euri10 @florimondmanca ❤️ 

I'll look at doing a release soon.

PR #1133, opened 2021-02-02
Version 0.14.2

JayH5:
### Fixed

* Fixed `ServerErrorMiddleware` compatibility with Python 3.9.1/3.8.7 when debug mode is enabled - [#1132](https://github.com/encode/starlette/pull/1132).
* Fixed unclosed socket `ResourceWarning`s when using the `TestClient` with WebSocket endpoints - #1132.
* Improved detection of `async` endpoints wrapped in `functools.partial` on Python 3.8+ - [#1106](https://github.com/encode/starlette/pull/1106).


docs/release-notes.md
```diff
@@ -1,3 +1,17 @@
+## 0.14.2
+
+February 2, 2021
+
+### Changed
+
+* Fixed `ServerErrorMiddleware` compatibility with Python 3.9.1/3.8.7 when debug mode is enabled -
```
euri10, line 7: I'm not a changelog pro but I'd put these 2 items under Fixed since the functionality does not change

euri10 (approved):
Approved ! thanks @JayH5 . I have one minor comment, you can ignore if you disagree _)

docs/release-notes.md
JayH5: Yeah I had a similar thought 😅 we've only used these subheadings recently and was trying to be consistent with how it's been done in the past. All of these fall under "Fixed", I think, though.

PR #1136, opened 2021-02-06
Fix docs syntax highlighting

JayH5:
At some point syntax highlighting seems to have stopped working, possibly due to changes in `mkdocs-material` which now recommends using highlighting from Python Markdown Extensions rather than CodeHilite: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/

So switch to that, and also clean up the indentation in the config file a bit.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,35 +2,35 @@ site_name: Starlette
 site_description: The little ASGI library that shines.
 
 theme:
-    name: 'material'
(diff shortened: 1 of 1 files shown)


jpic (approved):
Works here!
2021-02-07-114448_1180x668_scrot
2021-02-07-114443_1062x696_scrot

florimondmanca (approved):

PR #1139, opened 2021-02-08
Accept exception_handlers as Mapping

antonagestam:
This change does two things: the first is that it communicates to the caller that the passed dict isn't going to be mutated. The second is that it allows the caller to pass any Mapping without first casting it to a dict, which is very redundant since that already happens inside `__init__`.


Kludex:
image

EDIT: I don't recall why I printed this. :sweat_smile:

JayH5:
I do think `Mapping` is the right type here, but the fact that the key type is a `Union` of two types means that #1142 is still an issue. Mappings and dicts are not covariant in their key types, so really the best we can do is `Mapping[Any, Callable]`

starlette/applications.py
```diff
@@ -41,7 +40,7 @@ def __init__(
         debug: bool = False,
         routes: typing.Sequence[BaseRoute] = None,
         middleware: typing.Sequence[Middleware] = None,
-        exception_handlers: typing.Dict[
+        exception_handlers: typing.Mapping[
             typing.Union[int, typing.Type[Exception]], typing.Callable
         ] = None,
```
aminalaee:
```suggestion
        exception_handlers: typing.Mapping[typing.Any typing.Callable] = None,
```

aminalaee:
@antonagestam Will you update this?

aminalaee (changes requested):

antonagestam:
I'll try to get to addressing this soon.

antonagestam:
Updated to use a contravariant typevar, which makes this also close #1142.

Kludex:
This was closed by #1360.

Thanks for the PR @antonagestam ! :)

antonagestam:
@Kludex Oh, note that I updated this PR to solve the problem without introducing `Any` for the key type, so I still think there's value in integrating some of that work.

Kludex:
Cool! I've opened again! :)

antonagestam:
@Kludex I opened #1456 to supersede this one.

Issue #947, opened 2020-05-16
Background Thread created by WebSocketTestSession should exit when close() is called


devxpy:
Hello!

I've been trying to write tests for a starlette websocket API. 

Only one problem - the server isn't written in a way to exit nicely when a websocket is closed. 
As a result, the tests just sit there, because `websocket_connect()` won't exit until the server does.

By looking at the source, I figured that `WebSocketTestSession` is waiting for the session thread (`_run()`) to complete.

Can we please make it so that it stops the server when `close()` is called?

---

I've cobbled together a small [commit](https://github.com/devxpy/starlette/commit/41907e863220b8b34f6bcc0d8f3a8d1241ded8eb) to make this work. If you're on board with this change, I'd love to upstream this.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/947">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/947/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/947/pledge.svg">
</picture>
</a>

grutts:
I _think_ this is what is causing our issues as well. Every test case that runs adds another thread which sticks around, and the test suite hangs at the end. 

@devxpy I tried to use the TestClient class from your branch but I was still getting the hanging issue. Does it still work for you locally with the latest Starlette?

Anyway, it would be great if #948 could go in!

seanwu1105:
Here is a minimal example for the issue with latest starlette: https://github.com/seanwu1105/starlette-ws-test-issue

There is no way to test a long-running WebSocket with `TestClient` now. Have not checked if https://github.com/encode/starlette/pull/948 solves the issue as it depends on an outdated starlette.

PR #1140, opened 2021-02-09
Added timeout to WebSocketTestSession.receive() methods

ohiliazov:
The purpose of this PR is to make testing websockets easier.

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -80,9 +80,9 @@ May raise `starlette.websockets.WebSocketDisconnect` if the application does not
 
 #### Receiving data
 
-* `.receive_text()` - Wait for incoming text sent by the application and return it.
-* `.receive_bytes()` - Wait for incoming bytestring sent by the application and return it.
-* `.receive_json(mode="text")` - Wait for incoming json data sent by the application and return it. Use `mode="binary"` to send JSON over binary data frames.
+* `.receive_text(timeout=None)` - Wait for incoming text sent by the application and return it.
+* `.receive_bytes(timeout=None)` - Wait for incoming bytestring sent by the application and return it.
+* `.receive_json(mode="text", timeout=None)` - Wait for incoming json data sent by the application and return it. Use `mode="binary"` to send JSON over binary data frames.
 
 May raise `starlette.websockets.WebSocketDisconnect`.
 
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -335,25 +335,25 @@ class WebSocketTestSession:
     def close(self, code: int = 1000) -> None:
         self.send({"type": "websocket.disconnect", "code": code})
 
-    def receive(self) -> Message:
-        message = self._send_queue.get()
+    def receive(self, timeout: float = None) -> Message:
+        message = self._send_queue.get(timeout=timeout)
         if isinstance(message, BaseException):
             raise message
         return message
 
-    def receive_text(self) -> str:
-        message = self.receive()
+    def receive_text(self, timeout: float = None) -> str:
+        message = self.receive(timeout)
         self._raise_on_close(message)
         return message["text"]
 
-    def receive_bytes(self) -> bytes:
-        message = self.receive()
+    def receive_bytes(self, timeout: float = None) -> bytes:
+        message = self.receive(timeout)
         self._raise_on_close(message)
         return message["bytes"]
 
-    def receive_json(self, mode: str = "text") -> typing.Any:
+    def receive_json(self, mode: str = "text", timeout: float = None) -> typing.Any:
         assert mode in ["text", "binary"]
-        message = self.receive()
+        message = self.receive(timeout)
         self._raise_on_close(message)
         if mode == "text":
             text = message["text"]


Kludex:
Why this makes testing websockets easier?

ddemidov:
This could be useful in case the first message received from the client should contain an authorization token.
Timeouts would allow closing non-authorized clients.

MatthewScholefield:
This would be a great addition -- I was just about to make a patch for this until I saw this.

In my case my websocket handler forwards status updates from another service and often that service fails to complete (or error) which causes no websocket messages to be generated. To prevent my GitHub Actions job from timing out when a random test case hangs on a websocket receive call, I need to add a timeout.

lovelydinosaur:
> to make testing websockets easier

I'd like us to take a pass on this.

Flaky tests that need timeouts point towards an issue that needs resolving anyways.

MatthewScholefield:
> Flaky tests that need timeouts point towards an issue that needs resolving anyways.

Actually that's why I think it'd be useful. Right now I just see that the tests all timed out with minimal indication of what line/test-case caused it to hang. If we could add a timeout then on timeouts it could raise an exception and properly fail the test.

Are there any other reasons you're hesitant to include this?

irgolic:
+1 for either non-blocking or timeout option

I'm running a websocket in a separate thread to simulate another user. Currently, there's no way for me to gracefully exit the thread without "smoking it out"; making the server send another message after the `kill_thread` flag has been set.

Edit: I think the way I'm trying to write this also depends on https://github.com/encode/starlette/issues/947
Edit 2: For anyone else running into a similar issue, `ws.portal.call(ws.portal.stop, True)` killed the thread for me

PawelRoman:
> > to make testing websockets easier
> 
> I'd like us to take a pass on this.
> 
> Flaky tests that need timeouts point towards an issue that needs resolving anyways.

@tomchristie 

Are you sure?

I'm coding a turn based strategic wargame using a websocket server. I have 1300+ tests (using unittest). The game server has a single websocket endpoint to which the client connects. Then the client talks to the server over the websocket realizing all in-game actions in form of JSON messages. 

The websocket endpoint is plugged into game rules engine which is extremely complex as it involves handling all game rules and all validations. Thus, the single websocket endpoint can receive and send back DOZENS of all sorts of different JSON messages to connected players. Any modification to any of the game rules may change the structure, or number of the messages received by the client.

Of course I would have to be insane if I tried to test all those rules and validations via a websocket. 95% of my tests are pure unit tests, where I test only the game logic layer. Those tests are pure 'business logic' tests and no I/O involved whatsoever. I have an excellent test coverage of the game rules with unit tests, reaching 98%.

But the remaining 5% tests are kind of integration tests where I simulate an end-to-end scenarios between the client and the server. Those tests are using starlette TestClient, and methods such as `websocket_connect`, `send_json` and `receive_json`.

You would not believe how painful it is to NOT have this timeout parameter on `receive_json` with this kind of environment where the back-end server emits dozens of highly specialized, UI-related messages over the course of each test. 

With a proper architecture, 95% of your tests should be pure unit tests testing business logic in isolation from the network layer. And that's precisely what I'm doing with the game rules engine. But for those tests in the logic layer **you don't need the websocket client**, and that's the point you seem to be missing. 

The tests for which I do need the websocket client are **network layer tests** and those are high-level, system tests where the test function sees the server as a big black box. Here I simulate many different things but some of them include a more complicated flow where players simply play the game. The JSON messages flow coming out from the websocket server over the course of such tests is crazy. And I need to somehow receive them all in the test and sometimes I even assert them. 

My biggest system test is where 2 concurrent players (fired on 2 different asyncio tasks) play full game turn, each making about 10 different actions. Server outputs about 100 different messages over the websocket to each of the connected clients. This tests literally goes through tens of thousands of lines of python code in the game rules engine and requires a synchronization to avoid validation errors (e.g. player makes a move when it's not their turn). 

You would not believe how many problems there are in writing such tests and how often the test would simply freeze on websocket.receive() because of some "innocent" change in the game logic layer which alters the message flow.

One could argue that my logic layer tests are not good enough because they miss many message flow assertions, and those would allow detecting the root cause of each of the the network layer tests freezes. That's true in the theoretical world where programmers have infinite time and no deadlines. From practical standpoint, the primary purpose for the logic layer tests is to assert the changes in the game STATE, not to assert the correctness of the message flow outputted by the game rules engine. Some of those messages may be irrelevant (such as log messages), sometimes the sequence of the ouptut messages is completely irrelevant, sometimes it makes no sense to even assert an actual data returned (or most of it). Message-building and message-sending validation is not so important. The game state evolution is a top priority. So we're back to the network layer tests where we run a complicated scenario and freeze the whole test suite, need to re-create the database (yes, a lot of those system tests involve DB lookups) and so on and so on. All this could be avoided with a simple websocket.receive_json(timeout=1.0). Pretty pretty please, add the timeout parameter.

zmn-hamid:
I gotta +1 this. my situation is similiar to @PawelRoman. A really frustrating problem with no timeout is also that if you use VSCode and run the test via the editor itself, then even if you stop the test mid process (because it froze on receive), it wouldn't cut the connection to database, in my case at least. I have to run the test via terminal and kill the whole process which doesn't look nice at all

PR #1143, opened 2021-02-22
Change value from `file_values` prior to `environ`

elisong:


Kludex:
Hi @elisong, thanks for the PR! :1st_place_medal: 

The current behavior is actually well documented on https://www.starlette.io/config/#configuration-precedence, and I don't see a reason to change it. 

In case you feel like this should be discussed, feel free to raise an issue or provide more information about why we should aim this change. Again, thanks for the interest! :smile:

PR #1145, opened 2021-03-03
Fix code highlight

euri10:
oups, just pushed this PR because httpx/uvicorn are doing the same and see it exists already in https://github.com/encode/starlette/pull/1136

so not sure it's relevant, but in current state docs are not highlighted, locally with this PR it is, so maybe it's because docs from master aren't "live", not sure

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -35,8 +35,8 @@ nav:
 markdown_extensions:
   - mkautodoc
   - admonition
-  - pymdownx.highlight
-  - pymdownx.superfences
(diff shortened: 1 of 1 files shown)


JayH5:
@euri10 do we need this still?

euri10:
I'd say yes, just checked the docs and there's no highlight

JayH5:
#1136 did fix this in an alternative way. We just haven't done a release since then so the live docs aren't showing highlights yet.

PR #1147, opened 2021-03-11
Make session cookie use ASGI root path

### LINKED ISSUE #233: Session cookie should use root path
The session cookie currently uses '/'.
It should really use the ASGI root path instead, in case the application is submounted.

mahmoudhossam:
Fixes #233

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -49,14 +49,17 @@ class SessionMiddleware:
 
         async def send_wrapper(message: Message) -> None:
             if message["type"] == "http.response.start":
+                # Detect ASGI root path
+                path = scope["root_path"] or "/"
                 if scope["session"]:
                     # We have session data to persist.
                     data = b64encode(json.dumps(scope["session"]).encode("utf-8"))
                     data = self.signer.sign(data)
                     headers = MutableHeaders(scope=message)
-                    header_value = "%s=%s; path=/; Max-Age=%d; %s" % (
+                    header_value = "%s=%s; path=%s; Max-Age=%d; %s" % (
                         self.session_cookie,
                         data.decode("utf-8"),
+                        path,
                         self.max_age,
                         self.security_flags,
                     )
@@ -66,7 +69,7 @@ class SessionMiddleware:
                     headers = MutableHeaders(scope=message)
                     header_value = "{}={}; {}".format(
                         self.session_cookie,
-                        "null; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT;",
+                        f"null; path={path}; expires=Thu, 01 Jan 1970 00:00:00 GMT;",
                         self.security_flags,
                     )
                     headers.append("Set-Cookie", header_value)
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -49,7 +49,6 @@ class SessionMiddleware:
 
         async def send_wrapper(message: Message) -> None:
             if message["type"] == "http.response.start":
-                # Detect ASGI root path
                 path = scope.get("root_path", "") or "/"
                 if scope["session"]:
                     # We have session data to persist.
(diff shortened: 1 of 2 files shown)


JayH5:
Thanks, this seems like a reasonable improvement. It does need a test, I think, though.

starlette/middleware/sessions.py
```diff
@@ -49,14 +49,17 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         async def send_wrapper(message: Message) -> None:
             if message["type"] == "http.response.start":
+                # Detect ASGI root path
+                path = scope["root_path"] or "/"
```
lovelydinosaur: Note that ]`root_path` is Optional](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) so this could fail with a KeyError.

Perhaps we want this?...

```suggestion
                path = scope.get("root_path", "") or "/"
```

starlette/middleware/sessions.py
```diff
@@ -49,14 +49,17 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         async def send_wrapper(message: Message) -> None:
             if message["type"] == "http.response.start":
+                # Detect ASGI root path
```
lovelydinosaur: I don't think we need this comment - it's not really adding any value here.

```suggestion
```

lovelydinosaur (commented):
Once the final remaining suggestion is addressed I'm probably okay with pulling this in.

It *could* potentially use a test case, but it's probably also okay enough without.

starlette/middleware/sessions.py
mahmoudhossam: Done

mahmoudhossam:
I'm going to add a test and then we can merge this, thanks for the quick response!

lovelydinosaur (approved):

mahmoudhossam:
I couldn't find a way to set the root path directly on the app so I had to mount another app on the main one and tested that, seems to work as expected.

lovelydinosaur:
Ah fantastic, yup.

devsetgo:
@mahmoudhossam or @tomchristie  This update broke my session checking function to ensure users are logged in. I mount all the sub routes (users, etc..). The request.session has data in it when set, but when I redirect back to the start page then session is now null. I see no documentation on how to have the session across Mounts?

```
# Routes
routes = [
    Mount("/user", routes=
    [
        Route(
        ),
        Route(
            "/password-change",
            methods=["GET", "POST"],
        ),
    ],
    name="user",
    ),

# Session Checking
            logger.error(
            )

        else:
            )
            )

            if current == False:
                logger.error(
                )

            # update datetime of last use
            logger.info(
            )

    return check_login
```

mahmoudhossam:
I think there should be a difference between mounting routes and mounting whole applications.

I'd expect routes to share everything with the parent application, but submounted applications should have their own session.

@tomchristie thoughts?

devsetgo:
> I think there should be a difference between mounting routes and mounting whole applications.
> 
> I'd expect routes to share everything with the parent application, but submounted applications should have their own session.
> 
> @tomchristie thoughts?

I agree, the [documentation ](https://www.starlette.io/routing/) shows using Mount with a sub app (Mount("/static", app=StaticFiles(directory="static"), name="static"), but without defining "app=x" it is just Submounting routes. For the session, it should check if submounted as an app or route.

devsetgo:
@tomchristie ?

mahmoudhossam:
@devsetgo I'd say open a new issue describing your use case and make sure to mention that this PR broke your flow so it'd be easy to track for the maintainers.

This PR is already closed and I don't think anyone will be checking it anymore.

lovelydinosaur:
Okay folks - got a resolution to this here... https://github.com/encode/starlette/pull/1512

PR #1155, opened 2021-03-24
reverse url lookup for custom middleware in subapp fixed

PaxPrz:
Applying custom middleware into sub application will cause no routes to be found when `request.url_for("subapp:route_remain")` is called. It was because the `self.routes` for Mount will be `None` as the Middleware layer wraps it.


Kludex:
Is there an example of a middleware like that?

Kludex:
I'm closing this as author didn't reply, and there's no issue/discussion associated. Feel free to create a discussion or talk on Gitter in case you still want this, I'd be happy to discuss it. :)

PR #1157, opened 2021-03-26
anyio integration

### LINKED ISSUE #811: Trio support.
Via https://gitter.im/python-trio/general?at=5e2fe9b940da694c5edc8d70

It'd be nice to have Starlette support both Trio and Asyncio.

Currently it's an asyncio-only framework, but there's actually not many places where we're tied to asyncio, with the exception of a bunch of "if this callable is a plain function then call into with with `run_in_threadpool`, otherwise await it.".

I think the other places are:

* The WSGI middleware (optional). Currently has an asyncio-based implementation.
* The BaseHTTPMiddleware class (optional), that provides a request/response middleware interface to develop against, rather than directly writing against ASGI. Again, currently has an asyncio-based implemention.
* One weird part of the request model, `.is_disconnected` - that I think is structured as an anti-pattern and probably needs rethinking.
* The graphene-based GraphQL async executor, which probably ought to come out of the core package anyways.
* The existing TestClient, which we oughta replace with httpx now anyways, which does have trio+asyncio.

Short: Tweaking our run_in_threadpool function to use `sniffio` and then either call trio's threadpooling or asyncio's threadpooling would be sufficient as a starting point to start using Starlette with Trio.

uSpike:
Fixes: #811

This adds support for `asyncio` and `trio` by leveraging [anyio](https://anyio.readthedocs.io/) and a little more structured concurrency in spots.

Things to do yet:
- [x] 100% tests passing (there's one failing test at the moment)
- [x] `TestClient` rewritten for anyio, probably with portals.
- [x] `anyio` 3.0 release

This adds a hard dependency for starlette, which didn't exist before, and was a selling point.  In my opinion, the benefit is worth it.  It would be great to use starlette with either `uvicorn` or `hypercorn` using Trio.

This also removes the need for `aiofiles` which is gained for free with `anyio`.

Some drawbacks are that the other integrations such as graphql are only currently compatible with `asyncio`.  But, this does not make those integrations incompatible.  It does give users the ability to use trio without them, and maybe they can be ported in the future as well.

I created a branch for framework benchmarks here: https://github.com/uSpike/FrameworkBenchmarks/tree/starlette-anyio

See results: https://www.techempower.com/benchmarks/#section=test&shareid=c92a1338-0058-486c-9e47-c92ec4710b6a&hw=ph&test=query&a=2

There is a performance penalty to using anyio. The weighted composite score is 173 vs 165, so a 4.5% drop in performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 # Starlette
 
 Starlette is a lightweight [ASGI](https://asgi.readthedocs.io/en/latest/) framework/toolkit,
-which is ideal for building high performance asyncio services.
+which is ideal for building high performance async services.
 
 It is production-ready, and gives you the following:
 
@@ -38,7 +38,8 @@ It is production-ready, and gives you the following:
 * Session and Cookie support.
 * 100% test coverage.
 * 100% type annotated codebase.
-* Zero hard dependencies.
+* Few hard dependencies.
+* Compatible with `asyncio and `trio` backends.
 
 ## Requirements
 
@@ -86,7 +87,7 @@ For a more complete example, see [encode/starlette-example](https://github.com/e
 
 ## Dependencies
 
-Starlette does not have any hard dependencies, but the following are optional:
+Starlette only requires `anyio`, and the following are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
 * [`aiofiles`][aiofiles] - Required if you want to use `FileResponse` or `StaticFiles`.
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ It is production-ready, and gives you the following:
 * 100% test coverage.
 * 100% type annotated codebase.
 * Few hard dependencies.
-* Compatible with `asyncio and `trio` backends.
+* Compatible with `asyncio` and `trio` backends.
 
 ## Requirements
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ For a more complete example, see [encode/starlette-example](https://github.com/e
 Starlette only requires `anyio`, and the following are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
-* [`aiofiles`][aiofiles] - Required if you want to use `FileResponse` or `StaticFiles`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
 * [`python-multipart`][python-multipart] - Required if you want to support form parsing, with `request.form()`.
 * [`itsdangerous`][itsdangerous] - Required for `SessionMiddleware` support.
@@ -170,7 +169,6 @@ gunicorn -k uvicorn.workers.UvicornH11Worker ...
 <p align="center"><i>Starlette is <a href="https://github.com/encode/starlette/blob/master/LICENSE.md">BSD licensed</a> code. Designed & built in Brighton, England.</i></p>
 
 [requests]: http://docs.python-requests.org/en/master/
-[aiofiles]: https://github.com/Tinche/aiofiles
 [jinja2]: http://jinja.pocoo.org/
 [python-multipart]: https://andrew-d.github.io/python-multipart/
 [graphene]: https://graphene-python.org/
--- a/docs/index.md
+++ b/docs/index.md
@@ -82,7 +82,6 @@ For a more complete example, [see here](https://github.com/encode/starlette-exam
 Starlette does not have any hard dependencies, but the following are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
-* [`aiofiles`][aiofiles] - Required if you want to use `FileResponse` or `StaticFiles`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
 * [`python-multipart`][python-multipart] - Required if you want to support form parsing, with `request.form()`.
 * [`itsdangerous`][itsdangerous] - Required for `SessionMiddleware` support.
@@ -161,7 +160,6 @@ gunicorn -k uvicorn.workers.UvicornH11Worker ...
 <p align="center"><i>Starlette is <a href="https://github.com/encode/starlette/blob/master/LICENSE.md">BSD licensed</a> code. Designed & built in Brighton, England.</i></p>
 
 [requests]: http://docs.python-requests.org/en/master/
-[aiofiles]: https://github.com/Tinche/aiofiles
 [jinja2]: http://jinja.pocoo.org/
 [python-multipart]: https://andrew-d.github.io/python-multipart/
 [graphene]: https://graphene-python.org/
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,7 +32,7 @@ It is production-ready, and gives you the following:
 * Session and Cookie support.
 * 100% test coverage.
 * 100% type annotated codebase.
-* Zero hard dependencies.
+* Few hard dependencies.
 
 ## Requirements
 
@@ -79,7 +79,7 @@ For a more complete example, [see here](https://github.com/encode/starlette-exam
 
 ## Dependencies
 
-Starlette does not have any hard dependencies, but the following are optional:
+Starlette only requires `anyio`, and the following dependencies are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -35,7 +35,8 @@ case you should use `client = TestClient(app, raise_server_exceptions=False)`.
 
 `TestClient.async_backend` is a dictionary which allows you to set the options
 for the backend used to run tests.  These options are passed to
-`anyio.start_blocking_portal()`.  By default, `asyncio` is used.
+`anyio.start_blocking_portal()`. See the [anyio documentation](https://anyio.readthedocs.io/en/stable/basics.html#backend-options)
+for more information about backend options.  By default, `asyncio` is used.
 
 To run `Trio`, set `async_backend["backend"] = "trio"`, for example:
 
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -31,6 +31,21 @@ application. Occasionally you might want to test the content of 500 error
 responses, rather than allowing client to raise the server exception. In this
 case you should use `client = TestClient(app, raise_server_exceptions=False)`.
 
+### Selecting the Async backend
+
+`TestClient.async_backend` is a dictionary which allows you to set the options
+for the backend used to run tests.  These options are passed to
+`anyio.start_blocking_portal()`.  By default, `asyncio` is used.
+
+To run `Trio`, set `async_backend["backend"] = "trio"`, for example:
+
+```python
+def test_app()
+    client = TestClient(app)
+    client.async_backend["backend"] = "trio"
+    ...
+```
+
 ### Testing WebSocket sessions
 
 You can also test websocket sessions with the test client.
@@ -71,6 +86,8 @@ always raised by the test client.
 * `.websocket_connect(url, subprotocols=None, **options)` - Takes the same set of arguments as `requests.get()`.
 
 May raise `starlette.websockets.WebSocketDisconnect` if the application does not accept the websocket connection.
+
+`websocket_connect()` must be used as a context manager (in a `with` block).
 
 #### Sending data
 
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio<3,>=2"],
+    install_requires=["anyio>=3.0.0rc1"],
     extras_require={
         "full": [
             "aiofiles",
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0rc1"],
+    install_requires=["anyio>=3.0.0rc2"],
     extras_require={
         "full": [
             "aiofiles",
--- a/setup.py
+++ b/setup.py
@@ -48,10 +48,9 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0rc2"],
+    install_requires=["anyio>=3.0.0rc3"],
     extras_require={
         "full": [
-            "aiofiles",
             "graphene",
             "itsdangerous",
             "jinja2",
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0rc3"],
+    install_requires=["anyio>=3.0.0rc4"],
     extras_require={
         "full": [
             "graphene",
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
+    instal_requires=["anyio<3,>=2"],
     extras_require={
         "full": [
             "aiofiles",
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    instal_requires=["anyio<3,>=2"],
+    install_requires=["anyio<3,>=2"],
     extras_require={
         "full": [
             "aiofiles",
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     packages=find_packages(exclude=["tests*"]),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0rc4"],
+    install_requires=["anyio>=3.0.0,<4"],
     extras_require={
         "full": [
             "graphene",
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,5 +1,7 @@
 import typing
 
+import anyio
+
 from starlette.datastructures import State, URLPath
 from starlette.exceptions import ExceptionMiddleware
 from starlette.middleware import Middleware
@@ -109,7 +111,9 @@ class Starlette:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
-        await self.middleware_stack(scope, receive, send)
+        task_group = scope["task_group"] = anyio.create_task_group()
+        async with task_group:
+            await self.middleware_stack(scope, receive, send)
 
     # The following usages are now discouraged in favour of configuration
     #  during Starlette.__init__(...)
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,6 +1,7 @@
 import typing
 
 import anyio
+from anyio.abc import TaskGroup
 
 from starlette.datastructures import State, URLPath
 from starlette.exceptions import ExceptionMiddleware
@@ -37,6 +38,8 @@ class Starlette:
     Shutdown handler callables do not take any arguments, and may be be either
     standard functions, or async functions.
     """
+
+    task_group: TaskGroup
 
     def __init__(
         self,
@@ -111,8 +114,8 @@ class Starlette:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
-        task_group = scope["task_group"] = anyio.create_task_group()
-        async with task_group:
+        self.task_group = anyio.create_task_group()
+        async with self.task_group:
             await self.middleware_stack(scope, receive, send)
 
     # The following usages are now discouraged in favour of configuration
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,8 +1,5 @@
 import typing
 
-import anyio
-from anyio.abc import TaskGroup
-
 from starlette.datastructures import State, URLPath
 from starlette.exceptions import ExceptionMiddleware
 from starlette.middleware import Middleware
@@ -38,8 +35,6 @@ class Starlette:
     Shutdown handler callables do not take any arguments, and may be be either
     standard functions, or async functions.
     """
-
-    task_group: TaskGroup
 
     def __init__(
         self,
@@ -114,9 +109,7 @@ class Starlette:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
-        self.task_group = anyio.create_task_group()
-        async with self.task_group:
-            await self.middleware_stack(scope, receive, send)
+        await self.middleware_stack(scope, receive, send)
 
     # The following usages are now discouraged in favour of configuration
     #  during Starlette.__init__(...)
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -20,10 +20,10 @@ async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -
         async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
             nonlocal result
             result = await _handler(**_kwargs)
-            await task_group.cancel_scope.cancel()
+            task_group.cancel_scope.cancel()
 
         for handler, kwargs in args:
-            await task_group.spawn(task, handler, kwargs)
+            task_group.spawn(task, handler, kwargs)
 
     return result
 
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -14,7 +14,7 @@ T = typing.TypeVar("T")
 
 
 async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    async def run(handler: typing.Callable[..., typing.Coroutine]) -> None:
+    async def run(handler: typing.Callable[[], typing.Coroutine]) -> None:
         await handler()
         tg.cancel_scope.cancel()
 
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -28,6 +28,7 @@ async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -
 
     return result
 
+
 async def run_in_threadpool(
     func: typing.Callable[..., T], *args: typing.Any, **kwargs: typing.Any
 ) -> T:
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -1,5 +1,4 @@
 import functools
-import sys
 import typing
 from typing import Any, AsyncGenerator, Iterator
 
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -1,33 +1,36 @@
-import asyncio
 import functools
 import sys
 import typing
 from typing import Any, AsyncGenerator, Iterator
+
+import anyio
 
 try:
     import contextvars  # Python 3.7+ only or via contextvars backport.
 except ImportError:  # pragma: no cover
     contextvars = None  # type: ignore
 
-if sys.version_info >= (3, 7):  # pragma: no cover
-    from asyncio import create_task
-else:  # pragma: no cover
-    from asyncio import ensure_future as create_task
 
 T = typing.TypeVar("T")
 
 
 async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    tasks = [create_task(handler(**kwargs)) for handler, kwargs in args]
-    (done, pending) = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
-    [task.cancel() for task in pending]
-    [task.result() for task in done]
+    result: Any = None
+    async with anyio.create_task_group() as task_group:
 
+        async def task(_handler, _kwargs) -> Any:
+            nonlocal result
+            result = await _handler(**_kwargs)
+            await task_group.cancel_scope.cancel()
+
+        for handler, kwargs in args:
+            await task_group.spawn(task, handler, kwargs)
+
+    return result
 
 async def run_in_threadpool(
     func: typing.Callable[..., T], *args: typing.Any, **kwargs: typing.Any
 ) -> T:
-    loop = asyncio.get_event_loop()
     if contextvars is not None:  # pragma: no cover
         # Ensure we run in the same context
         child = functools.partial(func, *args, **kwargs)
@@ -35,9 +38,9 @@ async def run_in_threadpool(
         func = context.run
         args = (child,)
     elif kwargs:  # pragma: no cover
-        # loop.run_in_executor doesn't accept 'kwargs', so bind them in here
+        # run_sync doesn't accept 'kwargs', so bind them in here
         func = functools.partial(func, **kwargs)
-    return await loop.run_in_executor(None, func, *args)
+    return await anyio.run_sync_in_worker_thread(func, *args)
 
 
 class _StopIteration(Exception):
@@ -57,6 +60,6 @@ def _next(iterator: Iterator) -> Any:
 async def iterate_in_threadpool(iterator: Iterator) -> AsyncGenerator:
     while True:
         try:
-            yield await run_in_threadpool(_next, iterator)
+            yield await anyio.run_sync_in_worker_thread(_next, iterator)
         except _StopIteration:
             break
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -13,6 +13,16 @@ except ImportError:  # pragma: no cover
 T = typing.TypeVar("T")
 
 
+async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
+    async def run(handler: typing.Callable[..., typing.Coroutine]) -> None:
+        await handler()
+        tg.cancel_scope.cancel()
+
+    async with anyio.create_task_group() as tg:
+        for handler, kwargs in args:
+            tg.start_soon(run, functools.partial(handler, **kwargs))
+
+
 async def run_in_threadpool(
     func: typing.Callable[..., T], *args: typing.Any, **kwargs: typing.Any
 ) -> T:
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -14,13 +14,14 @@ T = typing.TypeVar("T")
 
 
 async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    async def run(handler: typing.Callable[[], typing.Coroutine]) -> None:
-        await handler()
-        tg.cancel_scope.cancel()
+    async with anyio.create_task_group() as task_group:
 
-    async with anyio.create_task_group() as tg:
-        for handler, kwargs in args:
-            tg.start_soon(run, functools.partial(handler, **kwargs))
+        async def run(func: typing.Callable[[], typing.Coroutine]) -> None:
+            await func()
+            task_group.cancel_scope.cancel()
+
+        for func, kwargs in args:
+            task_group.start_soon(run, functools.partial(func, **kwargs))
 
 
 async def run_in_threadpool(
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -21,7 +21,7 @@ async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -
             task_group.cancel_scope.cancel()
 
         for handler, kwargs in args:
-            task_group.spawn(task, handler, kwargs)
+            task_group.start_soon(task, handler, kwargs)
 
 
 async def run_in_threadpool(
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -13,17 +13,6 @@ except ImportError:  # pragma: no cover
 T = typing.TypeVar("T")
 
 
-async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    async with anyio.create_task_group() as task_group:
-
-        async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
-            await _handler(**_kwargs)
-            task_group.cancel_scope.cancel()
-
-        for handler, kwargs in args:
-            task_group.start_soon(task, handler, kwargs)
-
-
 async def run_in_threadpool(
     func: typing.Callable[..., T], *args: typing.Any, **kwargs: typing.Any
 ) -> T:
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -14,18 +14,14 @@ T = typing.TypeVar("T")
 
 
 async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
-    result: Any = None
     async with anyio.create_task_group() as task_group:
 
         async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
-            nonlocal result
-            result = await _handler(**_kwargs)
+            await _handler(**_kwargs)
             task_group.cancel_scope.cancel()
 
         for handler, kwargs in args:
             task_group.spawn(task, handler, kwargs)
-
-    return result
 
 
 async def run_in_threadpool(
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -18,7 +18,7 @@ async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -
     result: Any = None
     async with anyio.create_task_group() as task_group:
 
-        async def task(_handler, _kwargs) -> Any:
+        async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
             nonlocal result
             result = await _handler(**_kwargs)
             await task_group.cancel_scope.cancel()
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -36,7 +36,7 @@ async def run_in_threadpool(
     elif kwargs:  # pragma: no cover
         # run_sync doesn't accept 'kwargs', so bind them in here
         func = functools.partial(func, **kwargs)
-    return await anyio.run_sync_in_worker_thread(func, *args)
+    return await anyio.to_thread.run_sync(func, *args)
 
 
 class _StopIteration(Exception):
@@ -56,6 +56,6 @@ def _next(iterator: Iterator) -> Any:
 async def iterate_in_threadpool(iterator: Iterator) -> AsyncGenerator:
     while True:
         try:
-            yield await anyio.run_sync_in_worker_thread(_next, iterator)
+            yield await anyio.to_thread.run_sync(_next, iterator)
         except _StopIteration:
             break
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -29,13 +29,12 @@ class BaseHTTPMiddleware:
     async def call_next(self, request: Request) -> Response:
         send_stream, recv_stream = anyio.create_memory_object_stream()
         scope = request.scope
-        task_group = scope["task_group"]
 
         async def coro() -> None:
             async with send_stream:
                 await self.app(scope, request.receive, send_stream.send)
 
-        task_group.start_soon(coro)
+        scope["app"].task_group.start_soon(coro)
 
         try:
             message = await recv_stream.receive()
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -36,7 +36,7 @@ class BaseHTTPMiddleware:
         async def coro() -> None:
             async with send_stream:
                 try:
-                    await self.app(scope, recv_stream.receive, send_stream.send)
+                    await self.app(scope, request.receive, send_stream.send)
                 except BaseException as exc:
                     nonlocal coro_exc
                     coro_exc = exc
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -31,15 +31,9 @@ class BaseHTTPMiddleware:
         scope = request.scope
         task_group = scope["task_group"]
 
-        coro_exc = None
-
         async def coro() -> None:
             async with send_stream:
-                try:
-                    await self.app(scope, request.receive, send_stream.send)
-                except BaseException as exc:
-                    nonlocal coro_exc
-                    coro_exc = exc
+                await self.app(scope, request.receive, send_stream.send)
 
         task_group.spawn(coro)
 
@@ -55,8 +49,6 @@ class BaseHTTPMiddleware:
                 async for message in recv_stream:
                     assert message["type"] == "http.response.body"
                     yield message.get("body", b"")
-                if coro_exc is not None:
-                    raise coro_exc
 
         response = StreamingResponse(
             status_code=message["status"], content=body_stream()
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -27,7 +27,9 @@ class BaseHTTPMiddleware:
         await response(scope, receive, send)
 
     async def call_next(self, request: Request) -> Response:
-        send_stream, recv_stream = anyio.create_memory_object_stream(0, item_type=Message)  # XXX size
+        send_stream, recv_stream = anyio.create_memory_object_stream(
+            0, item_type=Message
+        )  # XXX size
 
         scope = request.scope
         task_group = scope["task_group"]
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -31,11 +31,17 @@ class BaseHTTPMiddleware:
         scope = request.scope
         task_group = scope["task_group"]
 
+        coro_exc = None
+
         async def coro() -> None:
             async with send_stream:
-                await self.app(scope, recv_stream.receive, send_stream.send)
+                try:
+                    await self.app(scope, recv_stream.receive, send_stream.send)
+                except BaseException as exc:
+                    nonlocal coro_exc
+                    coro_exc = exc
 
-        await task_group.spawn(coro)
+        task_group.spawn(coro)
 
         try:
             message = await recv_stream.receive()
@@ -49,6 +55,8 @@ class BaseHTTPMiddleware:
                 async for message in recv_stream:
                     assert message["type"] == "http.response.body"
                     yield message.get("body", b"")
+                if coro_exc is not None:
+                    raise coro_exc
 
         response = StreamingResponse(
             status_code=message["status"], content=body_stream()
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -29,7 +29,7 @@ class BaseHTTPMiddleware:
     async def call_next(self, request: Request) -> Response:
         send_stream, recv_stream = anyio.create_memory_object_stream(
             0, item_type=Message
-        )  # XXX size
+        )
 
         scope = request.scope
         task_group = scope["task_group"]
@@ -43,7 +43,6 @@ class BaseHTTPMiddleware:
         try:
             message = await recv_stream.receive()
         except anyio.EndOfStream:
-            print("WHAT " * 10)
             raise RuntimeError("No response returned.")
 
         assert message["type"] == "http.response.start"
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -4,7 +4,7 @@ import anyio
 
 from starlette.requests import Request
 from starlette.responses import Response, StreamingResponse
-from starlette.types import ASGIApp, Message, Receive, Scope, Send
+from starlette.types import ASGIApp, Receive, Scope, Send
 
 RequestResponseEndpoint = typing.Callable[[Request], typing.Awaitable[Response]]
 DispatchFunction = typing.Callable[
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -22,38 +22,38 @@ class BaseHTTPMiddleware:
             await self.app(scope, receive, send)
             return
 
-        request = Request(scope, receive=receive)
-        response = await self.dispatch_func(request, self.call_next)
-        await response(scope, receive, send)
+        async def call_next(request: Request) -> Response:
+            send_stream, recv_stream = anyio.create_memory_object_stream()
 
-    async def call_next(self, request: Request) -> Response:
-        send_stream, recv_stream = anyio.create_memory_object_stream()
-        scope = request.scope
+            async def coro() -> None:
+                async with send_stream:
+                    await self.app(scope, request.receive, send_stream.send)
 
-        async def coro() -> None:
-            async with send_stream:
-                await self.app(scope, request.receive, send_stream.send)
+            task_group.start_soon(coro)
 
-        scope["app"].task_group.start_soon(coro)
+            try:
+                message = await recv_stream.receive()
+            except anyio.EndOfStream:
+                raise RuntimeError("No response returned.")
 
-        try:
-            message = await recv_stream.receive()
-        except anyio.EndOfStream:
-            raise RuntimeError("No response returned.")
+            assert message["type"] == "http.response.start"
 
-        assert message["type"] == "http.response.start"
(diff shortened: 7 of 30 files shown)


README.md
```diff
@@ -38,7 +38,8 @@ It is production-ready, and gives you the following:
 * Session and Cookie support.
 * 100% test coverage.
 * 100% type annotated codebase.
-* Zero hard dependencies.
+* Few hard dependencies.
+* Compatible with `asyncio and `trio` backends.
```
Kludex:
```suggestion
* Compatible with `asyncio` and `trio` backends.
```

uSpike:
There's a weird bit of code in `starlette/requests.py`:
```python
            async with anyio.move_on_after(
                0.001
            ):  # XXX: too small of a deadline and this blocks
```
The existing code used `asyncio.wait_for(..., 0.00000001)` or some really small number.  I found that with `anyio.move_on_after` I needed a larger number for the deadline or else it would block.  I'm not sure of a better way to handle that.

EDIT: refactored to an event in 4dd8c5de79df2ee56959dba7bc20bbcdbb4d833d

agronholm:
> Is there/does there need to be an anyio equivalent to asyncio.iscoroutinefunction()?

Not really, no. Trio does not do any tricks in that regard. As I understand, starting with Python 3.10 even asyncio doesn't need that.

uSpike:
https://github.com/encode/starlette/pull/1157/commits/03e312ede7858c4031a5e01728e7513855d54763 added a chunk of work on `testclient`.  On my machine, the tests finish but hang with a number of threads blocking.  Not sure what's up there...

EDIT: not a problem anymore

agronholm:
My recommendation is to base this work on AnyIO 3.0, of which the first RC release is imminent. Since this is a new feature, I don't think backwards compatibility is necessary.

agronholm:
On that note, the 3.0.0rc1 pre-release is out now.

starlette/testclient.py
```diff
@@ -281,31 +276,30 @@ def __enter__(self) -> "WebSocketTestSession":
         return self
 
     def __exit__(self, *args: typing.Any) -> None:
-        self.close(1000)
-        self._thread.join()
+        try:
+            self.close(1000)
+        finally:
+            self.portal.stop_from_external_thread()
```
agronholm:
```suggestion
            self.portal.call(portal.stop)
```

starlette/testclient.py
```diff
@@ -452,42 +446,56 @@ def websocket_connect(
...
+        self.portal.call(self.wait_startup)
         return self
 
     def __exit__(self, *args: typing.Any) -> None:
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self.wait_shutdown())
+        try:
+            self.portal.call(self.wait_shutdown)
+        finally:
+            self.portal.stop_from_external_thread()
```
agronholm:
```suggestion
            self.portal.call(portal.stop)
```

starlette/testclient.py
```diff
@@ -452,42 +446,56 @@ def websocket_connect(
...
+            self.task.result()
+            return
+        try:
+            message = await self.stream_receive.receive()
+        except anyio.EndOfStream:
             self.task.result()
+            return
         assert message["type"] == "lifespan.shutdown.complete"
-        await self.task
+        self.task.result()
```
agronholm: What is `self.task` here?

agronholm (changes requested):
These should get you going on AnyIO 3. I had at least one question among the suggestions too.

starlette/testclient.py
```diff
@@ -231,13 +231,8 @@ async def send(message: Message) -> None:
...
 
         try:
-            loop = asyncio.get_event_loop()
-        except RuntimeError:
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
-
-        try:
-            loop.run_until_complete(self.app(scope, receive, send))
+            with anyio.start_blocking_portal(backend_options={"debug": True}) as portal:
```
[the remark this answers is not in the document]
uSpike: This is in here because I was having issues with some tasks not finishing.  I definitely don't plan to leave it :)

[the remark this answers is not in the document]
uSpike: Regarding configuring the portal backend, that's another conversation to have.

starlette/testclient.py
uSpike: I've changed a bit more, but `self.task` should be `self.portal.spawn_task(self.lifespan)`

uSpike:
I've pushed changes to use anyio 3.0.0rc1, thanks @agronholm !

There's some failing tests left:
- test_state_data_across_multiple_middlewares
- test_wsgi_get
- test_wsgi_post
- test_wsgi_exc_info

Those tests seem to not be getting data in `response.text` from `Testclient.get()`

There's another test I've had to skip because it blocks forever: test_graphql_async_old_style_executor, need to investigate that one a bit.

agronholm:
Those test failures don't look too bad. Let me know if you need more help. You can catch me on Gitter.

uSpike:
`starlette/graphql.py:GraphQLApp.__init__` states:
```python
        else:
            # Old style. Use 'executor'.
            # We should remove this in the next median/major version bump.
            self.executor = executor
            self.executor_class = None
            self.is_async = isinstance(executor, AsyncioExecutor)
```
The test for this is blocking forever.  I'd rather not spend too much effort debugging that, since it's slated to be removed.  I will remove it assuming that this PR would bring about a median/major version bump.

uSpike:
I'm down to one test failing.  I've also enabled testing for `asyncio`, `asyncio+uvloop`, and `trio`.

uSpike:
Alright!  Tests are passing and the core library is 100% anyio.

I added a task for `anyio` version 3 to be released, but in the mean time reviews would be much appreciated.

A future PR could convert the fileresponse API to use anyio's file functionality.

uSpike:
I realized that `anyio.start_blocking_portal()` defaults to `asyncio`, so my tests were all running in asyncio.  Whoops!  I added the ability to pass `async_backend` to `TestClient`.  This can also be set via the environment variable `STARLETTE_TESTCLIENT_ASYNC_BACKEND`, and defaults to "asyncio".

Now the tests are actually running `trio`, and some more are failing again.

agronholm:
Looks like the tests are hanging again. Did they work for you locally?

uSpike:
No, they randomly hang locally as well.

uSpike:
Seems like trio doesn't like the way I'm using portals so much:
```
ests/middleware/test_wsgi.py::test_build_environ_encoding[trio] PASSEDException ignored in: <function Nursery.__del__ at 0x7f3488616200>
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1055, in __del__
    assert not self._children
AssertionError: 
Exception ignored in: <coroutine object WSGIResponder.sender at 0x7f3488dd6320>
  File "/home/jordan/src/starlette/starlette/middleware/wsgi.py", line 105, in sender
    async for message in self.stream_receive:
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/abc/_streams.py", line 29, in __anext__
    return await self.receive()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/streams/memory.py", line 86, in receive
    await receive_event.wait()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_core/_eventloop.py", line 88, in get_cancelled_exc_class
    return get_asynclib().CancelledError
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_core/_eventloop.py", line 109, in get_asynclib
    asynclib_name = sniffio.current_async_library()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/sniffio/_impl.py", line 82, in current_async_library
    "unknown async library, or not in async context"
sniffio._impl.AsyncLibraryNotFoundError: unknown async library, or not in async context
Exception ignored in: <function Nursery.__del__ at 0x7f3488616200>
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/trio/_core/_run.py", line 1055, in __del__
    assert not self._children
AssertionError: 
Exception ignored in: <coroutine object WSGIResponder.sender at 0x7f34890567a0>
  File "/home/jordan/src/starlette/starlette/middleware/wsgi.py", line 105, in sender
    async for message in self.stream_receive:
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/abc/_streams.py", line 29, in __anext__
    return await self.receive()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/streams/memory.py", line 86, in receive
    await receive_event.wait()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_core/_eventloop.py", line 88, in get_cancelled_exc_class
    return get_asynclib().CancelledError
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_core/_eventloop.py", line 109, in get_asynclib
    asynclib_name = sniffio.current_async_library()
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/sniffio/_impl.py", line 82, in current_async_library
    "unknown async library, or not in async context"
sniffio._impl.AsyncLibraryNotFoundError: unknown async library, or not in async context
```

```
tests/middleware/test_wsgi.py:52: in test_wsgi_get
    response = client.get("/")
venv/lib/python3.7/site-packages/requests/sessions.py:555: in get
starlette/testclient.py:440: in request
    json=json,
venv/lib/python3.7/site-packages/requests/sessions.py:542: in request
venv/lib/python3.7/site-packages/requests/sessions.py:655: in send
starlette/testclient.py:246: in send
    raise exc from None
starlette/testclient.py:243: in send
    portal.call(self.app, scope, receive, send)
venv/lib/python3.7/site-packages/anyio/abc/_threads.py:177: in call
    return self.spawn_task(func, *args).result()
../../.pyenv/versions/3.7.5/lib/python3.7/concurrent/futures/_base.py:435: in result
    return self.__get_result()
../../.pyenv/versions/3.7.5/lib/python3.7/concurrent/futures/_base.py:384: in __get_result
    raise self._exception
venv/lib/python3.7/site-packages/anyio/abc/_threads.py:126: in _call_func
    retval = await retval
venv/lib/python3.7/site-packages/anyio/_backends/_trio.py:74: in __exit__
    return self.__original.__exit__(exc_type, exc_val, exc_tb)
venv/lib/python3.7/site-packages/trio/_core/_ki.py:159: in wrapper
    return fn(*args, **kwargs)
venv/lib/python3.7/site-packages/trio/_core/_run.py:549: in __exit__
    raise remaining_error_after_cancel_scope
E   RuntimeError: Cancel scope stack corrupted: attempted to exit <trio.CancelScope at 0x7f3488224fa0, active> in <Task 'anyio.abc._threads.BlockingPortal._call_func' at 0x7f34892f1190> that's still within its child <trio.CancelScope at 0x7f3488224590, active, cancelled>
E   
E   This is probably a bug in your code, that has caused Trio's internal state to
E   become corrupted. We'll do our best to recover, but from now on there are
E   no guarantees.
E   
E   Typically this is caused by one of the following:
E     - yielding within a generator or async generator that's opened a cancel
E       scope or nursery (unless the generator is a @contextmanager or
E       @asynccontextmanager); see https://github.com/python-trio/trio/issues/638
E     - manually calling __enter__ or __exit__ on a trio.CancelScope, or
E       __aenter__ or __aexit__ on the object returned by trio.open_nursery();
E       doing so correctly is difficult and you should use @[async]contextmanager
E       instead, or maybe [Async]ExitStack
E     - using [Async]ExitStack to interleave the entries/exits of cancel scopes
E       and/or nurseries in a way that couldn't be achieved by some nesting of
E       'with' and 'async with' blocks
E     - using the low-level coroutine object protocol to execute some parts of
E       an async function in a different cancel scope/nursery context than
E       other parts
E   If you don't believe you're doing any of these things, please file a bug:
E   https://github.com/python-trio/trio/issues/new
```

uSpike:
I think the key part is
```
E     - using [Async]ExitStack to interleave the entries/exits of cancel scopes
E       and/or nurseries in a way that couldn't be achieved by some nesting of
E       'with' and 'async with' blocks
```

Unfortunately the way `starlette.testclient` is written doesn't mandate context scopes, so I think to fix this would require using websocket connect as context managers, among other things.

agronholm:
This reminds me of an important omission in anyio: we need to enforce that the same task is used for both entering and exiting the cancel scope. I need to add this before v3.0 final.

uSpike:
Alright, I realized I was doing dumb stuff in worker threads, so that is all fixed up now.  I've also added some documentation of the changes to `TestClient`.

agronholm:
Is aiofiles actually needed really? It doesn't seem to do much that anyio doesn't offer out of the box.

starlette/concurrency.py
```diff
@@ -1,43 +1,42 @@
...
-    [task.cancel() for task in pending]
-    [task.result() for task in done]
+    async with anyio.create_task_group() as task_group:
+
+        async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
+            await _handler(**_kwargs)
+            task_group.cancel_scope.cancel()
+
+        for handler, kwargs in args:
+            task_group.spawn(task, handler, kwargs)
```
agronholm:
```suggestion
            task_group.start_soon(task, handler, kwargs)
```

setup.py
```diff
@@ -48,9 +48,9 @@ def get_packages(package):
     packages=get_packages("starlette"),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
+    install_requires=["anyio>=3.0.0rc3"],
```
agronholm:
```suggestion
    install_requires=["anyio>=3.0.0rc4"],
```

starlette/middleware/base.py
```diff
@@ -26,34 +27,28 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
-            finally:
-                await queue.put(None)
+            async with send_stream:
+                await self.app(scope, request.receive, send_stream.send)
 
-        task = loop.create_task(coro())
-        message = await queue.get()
-        if message is None:
-            task.result()
+        task_group.spawn(coro)
```
agronholm:
```suggestion
        task_group.start_soon(coro)
```

starlette/middleware/wsgi.py
```diff
@@ -83,31 +84,18 @@ async def __call__(self, receive: Receive, send: Send) -> None:
...
-            if self.exc_info is not None:
-                raise self.exc_info[0].with_traceback(
-                    self.exc_info[1], self.exc_info[2]
-                )
-        finally:
-            if sender and not sender.done():
-                sender.cancel()  # pragma: no cover
+
+        async with anyio.create_task_group() as task_group:
+            task_group.spawn(self.sender, send)
```
agronholm:
```suggestion
            task_group.start_soon(self.sender, send)
```

starlette/testclient.py
```diff
@@ -264,48 +267,59 @@ async def send(message: Message) -> None:
...
-        self.accepted_subprotocol = message.get("subprotocol", None)
 
     def __enter__(self) -> "WebSocketTestSession":
+        self.exit_stack = contextlib.ExitStack()
+        self.portal = self.exit_stack.enter_context(
+            anyio.start_blocking_portal(**self.async_backend)
+        )
+
+        try:
+            self.portal.spawn_task(self._run)
```
agronholm:
```suggestion
            self.portal.start_task_soon(self._run)
```

starlette/testclient.py
```diff
@@ -452,42 +473,55 @@ def websocket_connect(
...
+            anyio.start_blocking_portal(**self.async_backend)
+        )
+        self.stream_send = StapledObjectStream(
+            *anyio.create_memory_object_stream(math.inf)
+        )
+        self.stream_receive = StapledObjectStream(
+            *anyio.create_memory_object_stream(math.inf)
+        )
+        try:
+            self.task = self.portal.spawn_task(self.lifespan)
```
agronholm:
```suggestion
            self.task = self.portal.start_task_soon(self.lifespan)
```

starlette/testclient.py
```diff
@@ -452,42 +473,55 @@ def websocket_connect(
...
+            self.portal.call(self.wait_startup)
```
agronholm, line 492: Just a thought: could `portal.start_task()` be used here?

tests/test_testclient.py
```diff
@@ -118,13 +117,14 @@ async def respond(websocket):
...
             await websocket.accept()
-            asyncio.ensure_future(respond(websocket))
-            try:
-                # this will block as the client does not send us data
-                # it should not prevent `respond` from executing though
-                await websocket.receive_json()
-            except WebSocketDisconnect:
-                pass
+            async with anyio.create_task_group() as task_group:
+                task_group.spawn(respond, websocket)
```
agronholm:
```suggestion
                task_group.start_soon(respond, websocket)
```

agronholm (changes requested):
Added one comment and several changes to match AnyIO 3.0.0rc4.

starlette/testclient.py
uSpike: I think the intention here is that we want to wait for 'wait_startup' to finish before moving on.  And if it fails we want to not continue.

agronholm: Couldn't `self.lifespan()` signal such readiness on its own? Do we need a separate method for that?

uSpike: Well, we want to not leave `TestClient.__enter__` until `wait_startup()` has finished.  I could signal readiness some other way, but I don't understand what we'd gain doing that.

agronholm: Looking at the actual code, it doesn't seem very easy at least.

agronholm: Yeah, never mind.

wagnerflo:
Just adding my +1 for this. The more I read about trio, the more I want to use it instead of asyncio for future projects. Having starlette support it, would be grand!

uSpike:
Anyio 3.0.0 final is released, and this PR is ready to be reviewed.

docs/index.md
```diff
@@ -82,7 +82,6 @@ For a more complete example, [see here](https://github.com/encode/starlette-exam
 Starlette does not have any hard dependencies, but the following are optional:
 
 * [`requests`][requests] - Required if you want to use the `TestClient`.
-* [`aiofiles`][aiofiles] - Required if you want to use `FileResponse` or `StaticFiles`.
```
graingert, line 85: This was removed, but not in the other readme

uSpike: Updated!

uSpike:
@tomchristie @florimondmanca just wondering if there's anything I can do to facilitate a review or discussion?  Workflows are not enabled for this PR, but before they were globally disabled on Github the workflows were passing.

uSpike:
Failures are due to new jinja2 3 release which brings a bunch of deprecation warnings.   I pushed a change to lock `jinja2<3` in requirements.txt so workflows should pass now.

README.md
```diff
@@ -86,10 +87,9 @@ For a more complete example, see [encode/starlette-example](https://github.com/e
 
 ## Dependencies
 
-Starlette does not have any hard dependencies, but the following are optional:
+Starlette only requires `anyio`, and the following are optional:
```
lovelydinosaur, line 88: I'm totally okay with this, yup.

docs/testclient.md
```diff
@@ -31,6 +31,21 @@ application. Occasionally you might want to test the content of 500 error
...
+`TestClient.async_backend` is a dictionary which allows you to set the options
+for the backend used to run tests.  These options are passed to
+`anyio.start_blocking_portal()`.  By default, `asyncio` is used.
+
+To run `Trio`, set `async_backend["backend"] = "trio"`, for example:
+
+```python
+def test_app()
+    client = TestClient(app)
+    client.async_backend["backend"] = "trio"
```
lovelydinosaur, line 46: This is interesting. I like the way you can set the backend, tho the dictionary style isn't immediately obvious to me. Are there *other* keys can be set here?

starlette/applications.py
```diff
@@ -109,7 +111,9 @@ def url_path_for(self, name: str, **path_params: str) -> URLPath:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
-        await self.middleware_stack(scope, receive, send)
+        task_group = scope["task_group"] = anyio.create_task_group()
+        async with task_group:
+            await self.middleware_stack(scope, receive, send)
```
lovelydinosaur: Something we need to be aware of here, that's *kinda* silly, but also is at least worth pointing out...

This is on the critical path, so it'll have some impact on the benchmarking.

Now, I'm acutely aware that the micro-benchmarking is a big ol' pile of nonsense. But also it does, tediously, lead some folks into particular evaluations.

When I started Starlette one of the design arguments that I felt needed to be knocked down was that integrated server+application frameworks would be necessarily "faster" than frameworks with a clean server / application interface split. In this case ASGI.

To that extent it was important that Starlette ought to be able to equal sanic or aiohttp when evaluated with microbenchmarks, however tedious that might be.

Adoptiong this would likely change that.

Which might well be *okay*, but it's something we need to consider.

starlette/concurrency.py
```diff
@@ -1,43 +1,42 @@
...
-    [task.cancel() for task in pending]
-    [task.result() for task in done]
+    async with anyio.create_task_group() as task_group:
+
+        async def task(_handler: typing.Callable, _kwargs: dict) -> Any:
+            await _handler(**_kwargs)
+            task_group.cancel_scope.cancel()
+
+        for handler, kwargs in args:
+            task_group.start_soon(task, handler, kwargs)
```
lovelydinosaur: I think this helps highlight a miserable bit of code-smell.
Might be worth reviewing where we're using `run_until_first_complete`, because it's icky as heck, whichever way we slice it.

starlette/applications.py
```diff
@@ -109,7 +111,9 @@ def url_path_for(self, name: str, **path_params: str) -> URLPath:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
-        await self.middleware_stack(scope, receive, send)
+        task_group = scope["task_group"] = anyio.create_task_group()
```
lovelydinosaur: Would be interesting to know what quart chooses to do wrt. a request/response-wide task group. Eg. if you're running under trio do they provide an anyio TaskGroup or a trio Nursery?

starlette/requests.py
```diff
@@ -251,10 +252,10 @@ async def close(self) -> None:
 
     async def is_disconnected(self) -> bool:
         if not self._is_disconnected:
-            try:
-                message = await asyncio.wait_for(self._receive(), timeout=0.0000001)
-            except asyncio.TimeoutError:
-                message = {}
+            message: Message = {}
+            with anyio.move_on_after(0.001):
+                # XXX: to small of a deadline and this blocks
```
lovelydinosaur: Sounds brittle?

tests/test_websockets.py
```diff
@@ -208,22 +207,24 @@ async def asgi(receive, send):
...
 
         async def asgi(receive, send):
             websocket = WebSocket(scope, receive=receive, send=send)
-            queue = asyncio.Queue()
             await websocket.accept()
             await run_until_first_complete(
-                (reader, {"websocket": websocket, "queue": queue}),
-                (writer, {"websocket": websocket, "queue": queue}),
+                (reader, {"websocket": websocket}),
+                (writer, {"websocket": websocket}),
```
lovelydinosaur: I've not looked into it in detail, but I'd guess we can structure this more along structured concurrency lines, without my silly `run_until_first_complete` nonsense.

Eg. Using the recommended style in anyio... https://anyio.readthedocs.io/en/stable/streams.html?highlight=create_memory_object_stream#memory-object-streams

I'm also curious what the equivalent of `create_memory_object_stream` is in Trio.

lovelydinosaur (commented):
Fantastic work here.

I've taken a bit of a review. It's a tough one for sure. There's a bunch of stuff in Starlette that would be far nicer if it'd been a trio framework originally, and then moved to asyncio later, but that's just the way things have progressed, so.

docs/testclient.md
graingert: trio lets you configure you stuff like `clock=None, instruments=(), restrict_keyboard_interrupt_to_checkpoints=False`

asyncio lets you configure: use_uvloop, debug, policy https://anyio.readthedocs.io/en/stable/basics.html#backend-specific-options

graingert: it could be defined

```python
@dataclass
class TrioOptions:
    backend: Literal["trio"] = "trio"
    clock: trio.abc.Clock | None = None
    instruments: Sequence[trio.abc.Instrument] = ()
    restrict_keyboard_interrupt_to_checkpoints: bool = False

@dataclass
class AsyncioOptions:
   backend: Literal["asyncio"] = "asyncio"
   debug: bool = False
   use_uvloop: bool = True
   policy: asyncio.AbstractEventLoopPolicy | None = None
```

starlette/requests.py
graingert: @uSpike what do you mean by blocks? - are we getting a negative deadline?

agronholm:
> I'm also curious what the equivalent of create_memory_object_stream is in Trio.

https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_memory_channel

tests/test_websockets.py
agronholm: https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_memory_channel

starlette/applications.py
uSpike: I've explored this elsewhere, and for an application lifespan to run background tasks with structured concurrency, it'd be better to have an app-wide task group from which all other tasks are spawned:
```python
diff --git a/starlette/applications.py b/starlette/applications.py
index 10bd075..168b78c 100644
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,6 +1,7 @@
 import typing
 
 import anyio
+from anyio.abc import TaskGroup
 
 from starlette.exceptions import ExceptionMiddleware
@@ -37,6 +38,7 @@ class Starlette:
     Shutdown handler callables do not take any arguments, and may be be either
     standard functions, or async functions.
     """
+    task_group: TaskGroup
 
     def __init__(
         self,
@@ -111,8 +113,8 @@ class Starlette:
 
         scope["app"] = self
-        task_group = scope["task_group"] = anyio.create_task_group()
-        async with task_group:
+        self.task_group = anyio.create_task_group()
+        async with self.task_group:
 
     # The following usages are now discouraged in favour of configuration
diff --git a/starlette/middleware/base.py b/starlette/middleware/base.py
index 16d5064..4a7e58f 100644
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -29,7 +29,7 @@ class BaseHTTPMiddleware:
     async def call_next(self, request: Request) -> Response:
         scope = request.scope
-        task_group = scope["task_group"]
+        task_group = scope["app"].task_group
 
         async def coro() -> None:
             async with send_stream:
```

This then allows you to write a lifespan generator or startup/shutdown events such as:
```python
async def lifespan(app: Starlette):
    app.task_group.start_soon(background_task)
    ...
```

The only issue is, what happens if a task in that task group raises an error?

starlette/concurrency.py
uSpike: It's only used in one other place (`starlette/responses.py`) so I can replace it there.

starlette/applications.py
uSpike: I can provide some benchmarks for this conversation.  Are there any that you recommend as meaningful?

docs/testclient.md
uSpike: Maybe the `backend_options` argument to `anyio.run()` should be typed in this way and we just expose that in Starlette?

starlette/applications.py
graingert: > The only issue is, what happens if a task in that task group raises an error?

The lifespan task is cancelled ~and the server is shutdown. For the apps I write this is the desired behaviour.~

> If an exception is raised when calling the application callable with a lifespan.startup message or a scope with type lifespan, the server must continue but not send any lifespan events.
https://github.com/django/asgiref/blob/main/specs/lifespan.rst#scope

Other users of Starlette might want a task wrapper that catches Exceptions and logs them instead

starlette/requests.py
uSpike: That was discovered early in my development and it may have changed with anyio3 or other changes.  What I found was that  a deadline lower than around 0.001 (such as 0.00001) would cause the tests  to block.  I'll investigate more.

tests/test_websockets.py
uSpike: @tomchristie I was able to simplify this further:
```python
            await websocket.accept()
                task_group.start_soon(reader, websocket)
                task_group.start_soon(writer, websocket)
```

graingert: You can save a task with:

```python
            await websocket.accept()
                task_group.start_soon(reader, websocket)
                await writer(websocket)
```

starlette/requests.py
uSpike: I was able to refactor this into an event with 4dd8c5de79df2ee56959dba7bc20bbcdbb4d833d

tests/test_websockets.py
uSpike: Fixed in dde5079bdf59c1f974dff30d888ba1062faf9145

docs/testclient.md
uSpike: I'm not sure it's best to define specific backend options to starlette.  anyio could add new backends (twisted?) or the existing backends could support new options with new anyio versions.  It seems valuable to me to be as agnostic to the version of anyio as possible.

Seems to me that the best approach would be to use a `TypedDict` in `anyio.run()` for `backend_options`.  Any thoughts @agronholm ?

starlette/responses.py
```diff
@@ -221,10 +217,14 @@ async def stream_response(self, send: Send) -> None:
...
-            (self.listen_for_disconnect, {"receive": receive}),
-        )
+        async with anyio.create_task_group() as task_group:
+
+            async def wrap(coro: typing.Callable[..., typing.Awaitable]) -> None:
+                await coro()
+                task_group.cancel_scope.cancel()
+
+            task_group.start_soon(wrap, partial(self.stream_response, send))
+            task_group.start_soon(wrap, partial(self.listen_for_disconnect, receive))
```
graingert: you can save a task here
```suggestion
            await wrap(partial(self.listen_for_disconnect, receive))
```

starlette/applications.py
graingert: here's a little demo of what quart_trio does:


```python
import httpx
import anyio
import quart
import quart_trio

import logging

logger = logging.getLogger(__name__)

blueprint = quart.Blueprint("blueprint", __name__)


class MyCrashError(Exception):
    pass


@blueprint.route("/")
async def index():
    async def crash():
        raise MyCrashError

    quart.current_app.nursery.start_soon(crash)
    return b""


def create_app():
    app = quart_trio.QuartTrio(__name__)
    app.register_blueprint(blueprint)
    return app


async def amain():
    async with anyio.create_task_group() as tg:
        tg.start_soon(create_app().run_task)
        async with httpx.AsyncClient() as client:
            await client.get("http://localhost:5000/")
            (await client.get("http://localhost:5000/")).raise_for_status()


anyio.run(amain, backend="trio")
```

```
[2021-05-13 19:04:22,398] Running on http://127.0.0.1:5000 (CTRL + C to quit)
[2021-05-13 19:04:22,414] 127.0.0.1:56808 GET / 1.1 200 0 2337
[2021-05-13 19:04:22,416] ERROR in app: Exception on request GET /
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart_trio/app.py", line 131, in handle_request
    return await self.full_dispatch_request(request_context)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart_trio/app.py", line 153, in full_dispatch_request
    result = await self.handle_user_exception(error)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart_trio/app.py", line 168, in handle_user_exception
    return await super().handle_user_exception(error)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart/app.py", line 943, in handle_user_exception
    raise error
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart_trio/app.py", line 151, in full_dispatch_request
    result = await self.dispatch_request(request_context)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart/app.py", line 1511, in dispatch_request
    return await self.ensure_async(handler)(**request_.view_args)
  File "/home/graingert/projects/demo-quart-trio/foo.py", line 22, in index
    quart.current_app.nursery.start_soon(crash)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/trio/_core/_run.py", line 988, in start_soon
    GLOBAL_RUN_CONTEXT.runner.spawn_impl(async_fn, args, self, name)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/trio/_core/_run.py", line 1429, in spawn_impl
    raise RuntimeError("Nursery is closed to new arrivals")
RuntimeError: Nursery is closed to new arrivals
[2021-05-13 19:04:22,414] ASGI Framework Lifespan error, continuing without Lifespan support
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/hypercorn/trio/lifespan.py", line 29, in handle_lifespan
    await invoke_asgi(self.app, scope, self.asgi_receive, self.asgi_send)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/hypercorn/utils.py", line 239, in invoke_asgi
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart/app.py", line 1697, in __call__
    await self.asgi_app(scope, receive, send)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart/app.py", line 1723, in asgi_app
    await asgi_handler(receive, send)
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/quart_trio/asgi.py", line 148, in __call__
    break
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/trio/_core/_run.py", line 815, in __aexit__
    raise combined_error_from_nursery
  File "/home/graingert/projects/demo-quart-trio/foo.py", line 20, in crash
    raise MyCrashError
MyCrashError
[2021-05-13 19:04:22,419] 127.0.0.1:56808 GET / 1.1 500 290 2909
  File "/home/graingert/projects/demo-quart-trio/foo.py", line 40, in <module>
    anyio.run(amain, backend="trio")
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/anyio/_core/_eventloop.py", line 55, in run
    return asynclib.run(func, *args, **backend_options)  # type: ignore
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/trio/_core/_run.py", line 1932, in run
    raise runner.main_task_outcome.error
  File "/home/graingert/projects/demo-quart-trio/foo.py", line 37, in amain
    (await client.get("http://localhost:5000/")).raise_for_status()
  File "/home/graingert/.virtualenvs/testing39/lib/python3.9/site-packages/httpx/_models.py", line 1405, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: 500 Server Error:  for url: http://localhost:5000/
For more information check: https://httpstatuses.com/500
```

starlette/responses.py
```diff
@@ -221,10 +217,14 @@ async def stream_response(self, send: Send) -> None:
...
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        await run_until_first_complete(
-            (self.stream_response, {"send": send}),
-            (self.listen_for_disconnect, {"receive": receive}),
-        )
+        async with anyio.create_task_group() as task_group:
+
+            async def wrap(coro: typing.Callable[..., typing.Awaitable]) -> None:
+                await coro()
+                task_group.cancel_scope.cancel()
```
graingert:
```suggestion
            async def wrap(func: typing.Callable[[], typing.Coroutine]) -> None:
                await func()
                task_group.cancel_scope.cancel()
```

starlette/applications.py
uSpike: @tomchristie I created a branch here: https://github.com/uSpike/FrameworkBenchmarks/tree/starlette-anyio

See results: https://www.techempower.com/benchmarks/#section=test&shareid=c92a1338-0058-486c-9e47-c92ec4710b6a&hw=ph&test=query&a=2

There is a performance penalty to using anyio.  The weighted composite score is 173 vs 165, so a 4.5% drop in performance.

I'll add this result to the top comment.

uSpike: Interestingly, the multiple queries test was ~25% faster with anyio.

starlette/middleware/base.py
```diff
@@ -26,34 +27,27 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
-            finally:
-                await queue.put(None)
+            async with send_stream:
+                await self.app(scope, request.receive, send_stream.send)
 
-        task = loop.create_task(coro())
-        message = await queue.get()
-        if message is None:
-            task.result()
+        scope["app"].task_group.start_soon(coro)
```
graingert: ~Can this use a task_group created in this async function?~
Ah it's for the body_stream async generator

starlette/applications.py
uSpike: @graingert that's interesting.  A similar test app in Starlette would return "/" without error but raise an exception on the server, using uvicorn.

```python

async def crash():
    raise Exception()

async def home(request):
    request.app.task_group.start_soon(crash)
    return JSONResponse({"ok": "True"})

app = Starlette(routes=[Route("/", home)])
```

```
INFO:     127.0.0.1:38388 - "GET / HTTP/1.1" 200 OK
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/uvicorn/protocols/http/h11_impl.py", line 396, in run_asgi
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
  File "./starlette/applications.py", line 120, in __call__
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 528, in __aexit__
    raise exceptions[0]
  File "/home/jordan/src/starlette/venv/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 559, in _run_wrapped_task
    await coro
  File "./testapp.py", line 6, in crash
    raise Exception()
Exception
```

Using hypercorn, it crashes and never responds.

docs/testclient.md
agronholm: AnyIO won't even try to keep up with the options provided by upstream. Is there a problem with that?

uSpike: With that in mind, it seems simplest to me to keep the options as a dictionary with options referenced to https://anyio.readthedocs.io/en/stable/basics.html#backend-specific-options

What I have could probably be refined and documented better, though.

starlette/applications.py
graingert: @uSpike I think this task_group should be scoped to the dispatch, and not attached to the app: https://github.com/uSpike/starlette/pull/1

uSpike: @graingert yep that makes sense to me, I'll get that in today.

starlette/applications.py
uSpike: I ran the benchmark again against sanic and aiohttp, see https://www.techempower.com/benchmarks/#section=test&shareid=a2023abc-ee86-458f-a0e9-849b5c9f282f&hw=ph&test=composite&a=2

graingert (changes requested):

starlette/applications.py
agronholm: I only see starlette and starlette-anyio. How do I run these benchmarks locally?

uSpike: @agronholm if you click the other test tabs they should show up.

You can run locally with
```
./tfb --test starlette starlette-anyio sanic aiohttp
```

uSpike: It's hard to get reproducible results on my local machine.  I ran it again now with @graingert changes to make task groups per-request, and now starlette-anyio is much closer to starlette performance: https://www.techempower.com/benchmarks/#section=test&shareid=d1edb7b5-09f0-43c0-a4cb-cca51f87801c

graingert: task groups were per request before, they were just attached to the app

starlette/testclient.py
```diff
@@ -365,6 +380,14 @@ def receive_json(self, mode: str = "text") -> typing.Any:
 class TestClient(requests.Session):
     __test__ = False  # For pytest to not discover this up.
 
+    #: These options are passed to `anyio.start_blocking_portal()`
+    async_backend = {
+        "backend": "asyncio",
+        "backend_options": {},
+    }  # type: typing.Dict[str, typing.Any]
```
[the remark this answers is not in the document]
uSpike: Nowhere else in this file was variable annotation being used, but I'll change this one since starlette depends on python 3.6+

starlette/requests.py
uSpike: It seems I resolved the wrong conversation, my bad.

uSpike: I'm not quite sure what to do here.  The original code seems brittle to me as well:
```python
try:
    message = await asyncio.wait_for(self._receive(), timeout=0.0000001)
except asyncio.TimeoutError:
    message = {}
```
~What is the original intention here?~

uSpike: The comment `# XXX: to small of a deadline and this blocks` is incorrect.  Too small of a value for `anyio.move_on_after` and it seems that `self._receive()` doesn't have enough time to execute and `self._disconnected` is not set, and the tests fail.

uSpike: @JayH5 @tomchristie  OK, this is (actually) fixed now.  @graingert helped me to understand that the test client `async def receive()` needs to be non-trapping (meaning it doesn't `await` the `Event`) when `response_complete` is already set.  This allows a zero deadline in `is_disconnected` which is much cleaner.

starlette/testclient.py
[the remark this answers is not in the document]
uSpike: @JayH5 this is fixed in 5667a4b

docs/index.md
```diff
@@ -82,7 +82,6 @@ For a more complete example, [see here](https://github.com/encode/starlette-exam
 Starlette does not have any hard dependencies, but the following are optional:
```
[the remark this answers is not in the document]
uSpike: Fixed in 19685db

starlette/requests.py
JayH5: Amazing, thanks

starlette/testclient.py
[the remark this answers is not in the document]
JayH5: Yeah I know it's a bit inconsistent in the code base, just don't want to add more comment-style annotations.

JayH5:
I'll try give this a proper look over the weekend but so far it looks really good, thanks!

starlette/testclient.py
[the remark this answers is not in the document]
uSpike: @JayH5 @tomchristie I opened https://github.com/encode/starlette/pull/1189 to update annotations across the whole project.  If this could be merged soon-ish then the workflows for this PR could be run without approval.

JayH5 (commented):
> I'll try give this a proper look over the weekend but so far it looks really good, thanks!

Sorry, I'm a couple weeks late 😢. Life happens! And the weather's been really good!

I'm really happy with this and think it's a good improvement. Trio support is great, but on top of that I think this cleans up some tricky code in Starlette. I have a couple of minor questions, but think this is pretty much ready to merge. I'd also like to get a final sign-off from @tomchristie if possible.

starlette/middleware/base.py
```diff
@@ -21,45 +22,38 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await self.app(scope, receive, send)
             return
 
-        request = Request(scope, receive=receive)
-        response = await self.dispatch_func(request, self.call_next)
-        await response(scope, receive, send)
+        async def call_next(request: Request) -> Response:
+            send_stream, recv_stream = anyio.create_memory_object_stream()
```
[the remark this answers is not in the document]
graingert: do you have a test case for this?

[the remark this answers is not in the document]
graingert: Ah we can cancel the task group after we return the response, fixing both problems

starlette/middleware/base.py
```diff
@@ -21,45 +22,38 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+            response = StreamingResponse(
+                status_code=message["status"], content=body_stream()
+            )
+            response.raw_headers = message["headers"]
+            return response
+
+        async with anyio.create_task_group() as task_group:
+            request = Request(scope, receive=receive)
+            response = await self.dispatch_func(request, call_next)
+            await response(scope, receive, send)
```
graingert, line 56:
```suggestion
            await response(scope, receive, send)
            task_group.cancel_scope.cancel()
```

starlette/middleware/base.py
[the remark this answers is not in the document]
uSpike: Added in 63cfcb9ef029406e39852a09592199309a139f22 along with a test for https://github.com/encode/starlette/issues/1022

docs/testclient.md
```diff
@@ -31,6 +31,21 @@ application. Occasionally you might want to test the content of 500 error
 responses, rather than allowing client to raise the server exception. In this
 case you should use `client = TestClient(app, raise_server_exceptions=False)`.
 
+### Selecting the Async backend
+
+`TestClient.async_backend` is a dictionary which allows you to set the options
+for the backend used to run tests.  These options are passed to
+`anyio.start_blocking_portal()`.  By default, `asyncio` is used.
```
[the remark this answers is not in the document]
uSpike: Added in 6208ca5e356cb9acc605723e5b05e91bfbe1c391

JayH5 (dismissed):
Thanks for your patience & hard work on this.

@tomchristie are you happy with this? If so, I'd like to merge this and release Starlette 0.15.

JayH5:
🤔 I see we have a mypy failure for `StapledObjectStream`

uSpike:
> I see we have a mypy failure for StapledObjectStream

Yeah, I think it's because anyio uses `dataclasses` on <3.7 but we need to install `types-dataclasses` (added in e0c99676119f97d1e35e995a9b33649a69d5da68)

uSpike:
@graingert suggested on gitter the following changes:

```python
if sys.python_version >= (3, 8):
    from typing import Protocol, runtime_checkable
else:
    from typing_extensions import Protocol, runtime_checkable

@runtime_checkable
class ACloseAble(Protocol):
    async def aclosing(self) -> None: ...
```
```diff
@@ -186,8 +197,10 @@ class StreamingResponse(Response):
         media_type: str = None,
         background: BackgroundTask = None,
     ) -> None:
-        if isinstance(content, typing.AsyncIterable):
+        if isinstance(content, typing.AsyncGenerator):
             self.body_iterator = content
+        elif isinstance(content, ACloseAble):
+            self.body_iterator = WrapAiterable(content)
         else:
             self.body_iterator = iterate_in_threadpool(content)
         self.status_code = status_code
@@ -209,10 +222,11 @@ class StreamingResponse(Response):
                 "headers": self.raw_headers,
             }
         )
-        async for chunk in self.body_iterator:
-            if not isinstance(chunk, bytes):
-                chunk = chunk.encode(self.charset)
-            await send({"type": "http.response.body", "body": chunk, "more_body": True})
+        async with aclosing(self.body_iterator) as aiter:
+            async for chunk in aiter:
+                if not isinstance(chunk, bytes):
+                    chunk = chunk.encode(self.charset)
+                await send({"type": "http.response.body", "body": chunk, "more_body": True})

         await send({"type": "http.response.body", "body": b"", "more_body": False})

**Comment by uSpike** [comment]
Go home github, you're drunk!

**Comment by uSpike** [comment]
0ef6efa84b2907120c1ebda89ebbf025812f95d9 added support (as suggested by @graingert) for response iterators that are aclose-able.  If `aclose` is defined then it will always be called.  I also added a test for it.

**Review by graingert — APPROVED** [review]

**Comment by lovelydinosaur** [comment]
@encode/maintainers - I'm okay with this. I think it'd be beneficial if we can start to make sure we've got a good community flow to managing a bunch of the encode projects, because personally I'm only really able to *fully* dedicate my time to one major project at once, and right now that's `httpx`.

So...

* Have we got any blockers on one of y'all being able to merge this?
* Have we adequately documented our release process, do we have good visibility onto how maintainers can do this?

**Comment by JayH5** [comment]
> 0ef6efa added support (as suggested by @graingert) for response iterators that are aclose-able. If aclose is defined then it will always be called. I also added a test for it.

@uSpike this seems reasonable, but I don't understand how it directly relates to this PR? Please justify its necessity for the anyio integration, or move it to a separate PR.

**Comment by graingert** [comment]
@JayH5 trio warns when a async generator is left open

**Comment by JayH5** [comment]
@tomchristie 
> * Have we got any blockers on one of y'all being able to merge this?

Not afaik. I've added one last question now

> * Have we adequately documented our release process, do we have good visibility onto how maintainers can do this?

I've done a couple of releases. It's not well documented in Starlette itself, but the process [documented for httpx](https://www.python-httpx.org/contributing/#releasing) works just fine.

**Comment by JayH5** [comment]
@graingert 
> @JayH5 trio warns when a async generator is left open

What kind of warning? One we can get pytest to fail on?

File: starlette/responses.py [hunk 30]
```diff
@@ -34,6 +35,29 @@ def guess_type(
...
+    # This is taken from 3.10 contextlib.aclosing and modified to only
+    # call thing.aclose() if it is ACloseAble.
+    def __init__(self, thing: T):
+        self.thing = thing
+
+    async def __aenter__(self) -> T:
+        return self.thing
+
+    async def __aexit__(self, *exc_info: typing.Any) -> None:
+        if isinstance(self.thing, ACloseAble):
```
JayH5: I'm not sure I see the advantage of `ACloseAble` over just `hasattr(self.thing, "aclose")`? We haven't used protocols in Starlette yet and I'm not against it but not sure it's justified here either.

graingert: mypy doesn't know about hasattr

graingert:
I'm happy for the aclosing change to be backed out to not delay this pr

starlette/responses.py
JayH5: To me this is doing a little too much to appease the linter.

uSpike: I've reverted these changes

uSpike:
> > [0ef6efa](https://github.com/encode/starlette/commit/0ef6efa84b2907120c1ebda89ebbf025812f95d9) added support (as suggested by @graingert) for response iterators that are aclose-able. If aclose is defined then it will always be called. I also added a test for it.
> 
> @uSpike this seems reasonable, but I don't understand how it directly relates to this PR? Please justify its necessity for the anyio integration, or move it to a separate PR.

I also do not want to delay this PR so I've rolled back to before those changes (they're still available at https://github.com/uSpike/starlette/tree/anyio_bak).  I'd rather make a PR in the future to cover aclose-able iterators.  @JayH5 this should be back to the state at which you approved.

starlette/concurrency.py
```diff
@@ -1,43 +1,31 @@
...
 
-if sys.version_info >= (3, 7):  # pragma: no cover
-    from asyncio import create_task
-else:  # pragma: no cover
-    from asyncio import ensure_future as create_task
 
 T = typing.TypeVar("T")
 
 
-async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
```
JayH5, line 20: This is technically an API removal so I was wondering if anything used this. It looks like [some things do](https://sourcegraph.com/search?q=context:global+lang:Python+run_until_first_complete&patternType=literal), but not too many. FastAPI re-exports it but nobody seems to use that export.

Is there an easy anyio equivalent we can either implement here or put in the changelog? (I'm not suggesting updating the changelog in this PR, but I'll need to compile a changelog for 0.15 soon.)

Could we maybe keep this function for Starlette 0.15 but deprecate it? I don't think we really want this in our API because we don't use it much ourselves and it seems fairly rare that others do either. OTOH, the more stuff the next Starlette release breaks, the less people likely to use it.

graingert: something like this?

```python
async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
    async def run(handler, kwargs):
        await handler(**kwargs)
        tg.cancel_scope.cancel()

    async with anyio.create_task_group() as tg:
        for handler, kwargs in args:
            tg.start_soon(run, handler, kwargs)
```

uSpike: That's reasonable.  @graingert mentioned this too on gitter as well, so I'll add a compatible API back in.

@JayH5 how would we deprecate the API?  It's not documented, so would we show a `DeprecationWarning`?

graingert: is it so bad to leave it without deprecating it? it's a handy function.

uSpike: IMO a function like this is not documented and is an internal helper for concurrency.  It doesn't seem appropriate to me to expose it and support it as a first-class API of Starlette, long term.  I think users of Starlette should look elsewhere for concurrency helpers.

JayH5: I've thought about it some more, and while I 100% agree with you, @uSpike, I think let's add it back but _not_ deprecate it for now. That should leave this PR completely API-compatible except for the GraphQL change which is fine because that was old deprecated stuff and we're going to remove the whole thing soon anyway.

In some later version of Starlette we should go through the code more systematically and make APIs public or private. I think httpx went through this process at some point.

JayH5: (I guess the other API change is the testclient websocket context manager thing but there's a clear technical reason for that)

uSpike: @JayH5 added in 2b9dd22173e7ddb552e08503433c700601a417d0 with a test case since it is an api function used elsewhere.

starlette/concurrency.py
```diff
@@ -13,6 +13,16 @@
 T = typing.TypeVar("T")
 
 
+async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
+    async def run(handler: typing.Callable[..., typing.Coroutine]) -> None:
```
graingert:
```suggestion
    async def run(handler: typing.Callable[[], typing.Coroutine]) -> None:
```

uSpike: Yeah just added that lol, thanks.

uSpike: 643d107a88f1e26b0c0485b8ca155488f75fabf1

starlette/concurrency.py
```diff
@@ -1,43 +1,41 @@
...
-    (done, pending) = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
-    [task.cancel() for task in pending]
-    [task.result() for task in done]
+    async def run(handler: typing.Callable[[], typing.Coroutine]) -> None:
+        await handler()
+        tg.cancel_scope.cancel()
+
+    async with anyio.create_task_group() as tg:
+        for handler, kwargs in args:
+            tg.start_soon(run, functools.partial(handler, **kwargs))
```
JayH5: The scope of `tg` is a bit confusing here, can we do it like it's done in `StreamingResponse` (and maybe use this function in `StreamingResponse`)?:
```suggestion
    async with anyio.create_task_group() as task_group:
        async def run(func: typing.Callable[[], typing.Coroutine]) -> None:
            await func()
            task_group.cancel_scope.cancel()
        
        for func, kwargs in args:
            task_group.start_soon(run, functools.partial(func, **kwargs))
```

JayH5: This diff is confusing, but I hope you get what I mean 🙈

uSpike: Sure, I can address that.  IMO I'd rather not use `run_until_first_complete` in `StreamingResponse` if the plan is to deprecate/remove it in the future.

uSpike: @JayH5  addressed in d0ca3f2bca14aa58d813e52b691b8ffd462a602b

JayH5: Cool, we'll see if/how we end up deprecating it but I'm ok with also not using it.

JayH5 (approved):
I think it's time 🚀

uSpike:
Cheers @JayH5 and thank you everyone for your help/work on this.  I think this is an important step in bringing structured concurrency to the mainstream and I'm grateful for the opportunity to contribute.  :heart:

agronholm:
Just one question: how soon can we expect to see this in a release?

graingert:
@agronholm I think it's all but released now: https://github.com/encode/starlette/pull/1202

PR #1163, opened 2021-04-04
Stream with utf8 filename

### LINKED ISSUE #790: response headers support custom encoding

Yasu-umi:
Fixes: https://github.com/encode/starlette/issues/790#issuecomment-667881654

This adds support filename for StreamingResponse  same as FileResponse.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -34,6 +34,18 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
+# Build Content Disposition Header from filename
+def build_content_disposition_header(filename: str) -> str:
+    content_disposition_filename = quote(filename)
+    if content_disposition_filename != filename:
+        content_disposition = "attachment; filename*=utf-8''{}".format(
+            content_disposition_filename
+        )
+    else:
+        content_disposition = 'attachment; filename="{}"'.format(filename)
+    return content_disposition
+
+
 class Response:
     media_type = None
     charset = "utf-8"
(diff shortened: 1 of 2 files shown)


addisonElliott:
Running across this bug. I am definitely interested in getting this merged!

starlette/responses.py
```diff
@@ -34,6 +34,18 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
+# Build Content Disposition Header from filename
```
Kludex: No need, the function name is pretty clear. :)

```suggestion
```

Kludex (commented):
I'm looking at the history, and it looks like conceptually `StreamingResponse` shouldn't be accepting `filename`.

@tomchristie Can you confirm?

adriangb:
> I'm looking at the history, and it looks like conceptually `StreamingResponse` shouldn't be accepting `filename`.

Would this work for you @Yasu-umi ?

```python
import os
from tempfile import TemporaryDirectory

from starlette.responses import FileResponse

def test_streaming_response_with_japanese_filename() -> None:
    filename = "こんにちは.txt"  # "Hello.txt" in Japanese
    content = b"1, 2, 3, 4, 5"
    with TemporaryDirectory() as dirname:
        filepath = os.path.join(dirname, filename)
        with open(filepath, mode="wb") as f:
            f.write(content)
        client = TestClient(FileResponse(filepath, filename=filename))
        response = client.get("/")
    expected_disposition = (
        "attachment; filename*=utf-8''%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF.txt"
    )
    assert response.content == content
    assert response.headers["content-disposition"] == expected_disposition
```

Yasu-umi:
@adriangb thanks for example!
I solved this by below codes (just overwrite header).
```python
def create_file_response(*, stream: BytesIO, filename: str, **kwargs: Any) -> StreamingResponse:
  res = StreamingResponse(stream, **kwargs)
  res.raw_headers.append((
    b'content-disposition',
    f"attachment; filename*=utf-8''{quote(filename)}".encode('latin-1'),
  ))
  return res
```
As Kludex taught me, if conceptually `StreamingResponse` shouldn't be accepting `filename`, should I close this pull req & issue ?

adriangb:
Yes, please close it

PR #1164, opened 2021-04-14
use quote instead of quote_plus for RedirectResponse location header

### LINKED ISSUE #1162: RedirectResponse call to `quote_plus` replaces spaces & `%20`'s with `+` character
### Checklist

### Describe the bug

`RedirectResponse` replaces url paths with spaces or `%20` with the `+` character. This happens as the `location` header is written after passing the url to `quote_plus`: https://github.com/encode/starlette/blob/master/starlette/responses.py#L181

If part of the path was a path param (with a space in it), this then modified those path params.

We currently encounter this with trailing slash redirects, which return a RedirectResponse.

### To reproduce

```python


async def homepage(request):
    name = request.path_params["name"]
    return JSONResponse({"hello": name})


app = Starlette(routes=[Route("/{name}/greeting", homepage)])


client = TestClient(app)


def test_redirect_homepage():
    name = "bob%20ross"
    url = f"{name}/greeting"
    for u in [url, url + "/"]:
        resp = client.get(u)
        assert url in str(resp.url)
```

### Expected behavior

Expect that the location header would contain the url to redirect to without modification. If needed, it could replace the actual space character with `%20`, instead, but it's not clear to me that is necessary, after looking through the http specs.

This could be accomplished with swapping `urllib.quote_plus` for `urllib.quote` in `RedirectResponse`.

### Actual behavior

Spaces are being replaced with the `+` character. 

### Environment

- OS:  macOS
- Python version:  3.8.8
- Starlette version:  0.13.6 (also separately tested 0.14.2)

falkben:
Closes #1162

use `quote` instead of `quote_plus` for `RedirectResponse` location header

adjust safe characters: rem. duplicate `&` symbol

add test for redirect quoting

Used django's redirect quote call as a reference: https://github.com/django/django/blob/23fa29f6a6659e0f600d216de6bcb79e7f6818c9/django/utils/encoding.py#L129

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -7,7 +7,7 @@ import sys
 import typing
 from email.utils import formatdate
 from mimetypes import guess_type as mimetypes_guess_type
-from urllib.parse import quote, quote_plus
+from urllib.parse import quote
 
 from starlette.background import BackgroundTask
 from starlette.concurrency import iterate_in_threadpool, run_until_first_complete
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -178,7 +178,7 @@ class RedirectResponse(Response):
         super().__init__(
             content=b"", status_code=status_code, headers=headers, background=background
         )
-        self.headers["location"] = quote_plus(str(url), safe=":/%#?&=@[]!$&'()*+,;")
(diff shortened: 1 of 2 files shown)


JayH5 (approved):
Thank you!

PR #1166, opened 2021-04-21
starlette's setup.py creates a broken top_level.txt record in the wheel file

wimglenn:
The current setup file has a hand-rolled `get_packages` function which returns the following packages:

```
>>> get_packages("starlette")
['starlette', 'starlette/middleware']
```

This results in an [incorrect `top_level.txt` file to be generated in the wheel's dist-info directory](https://setuptools.readthedocs.io/en/latest/deprecated/python_eggs.html#top-level-txt-conflict-management-metadata):

```
$ cat starlette-0.14.2.dist-info/top_level.txt
starlette
starlette/middleware
```

"starlette/middleware" is not a Python identifier, breaking some automated tooling at $EMPLOYER.

After the change in this PR, the correct packages list passed to setup is:

```
['starlette', 'starlette.middleware']
```

Which is resulting in the correct `top_level.txt` file being generated in the wheel:

```
starlette
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@
 import os
 import re
 
-from setuptools import setup
+from setuptools import setup, find_packages
 
 
 def get_version(package):
@@ -23,17 +23,6 @@ def get_long_description():
         return f.read()
 
 
-def get_packages(package):
-    """
-    Return root package and all sub-packages.
-    """
-    return [
-        dirpath
-        for dirpath, dirnames, filenames in os.walk(package)
-        if os.path.exists(os.path.join(dirpath, "__init__.py"))
-    ]
-
-
 setup(
     name="starlette",
     python_requires=">=3.6",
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Yup looks sensible.

Here's the difference:

```pycon
>>> get_packages("starlette")
['starlette', 'starlette/middleware']
>>> find_packages(exclude=["tests*"])
['starlette', 'starlette.middleware']
```

We *could* alternately tweak a line in the `get_packages` implementation, like so...

```python
return [
    dirpath.replace(os.path.sep, ".")
    for dirpath, dirnames, filenames in os.walk(package)
    if os.path.exists(os.path.join(dirpath, "__init__.py"))
]
```

But.... it probably makes sense to just lean on `find_packages` instead.

lovelydinosaur (approved):

wimglenn:
@tomchristie Thank you for the quick turnaround on this. Several other encode projects have the same issue, is it okay if I create similar PR's for those others?

PR #1169, opened 2021-04-27
Allow to silently handle exceptions in ServerErrorMiddleware

arugifa:
I am using Starlette in association with Sentry, and I recently wanted to return the event ID generated by Sentry when an internal server error happens.

For this purpose, I created a small error handler looking like that:

```python
def catch_internal_server_errors(request: Request, exc: Exception) -> JSONResponse:
    event_id = sentry_sdk.capture_exception(exc)
    payload = {"event_id": event_id}

    return JSONResponse(payload, 500)
```

That I connected to my app like this:

```python
app.add_exception_handler(Exception, catch_internal_server_errors)
```

It is working well. However, all exceptions are recorded twice in Sentry:

1. By my error handler,
2. By Sentry itself, which automatically records all unhandled exceptions.

This happens because the `ServerErrorMiddleware` always reraise exceptions, even if they are processed with a dedicated handler.

That's why I though about updating this middleware, so it doesn't reraise caught exceptions by default, and thus avoids errors to be recorded twice in Sentry.

Let me know what you think about. If you like it, then I will add the missing tests :rainbow:


jcwilson:
Not sure if it's viable here, but have you looked into using the sentry "before-send" hook to filter exceptions you do not wish to send? Maybe there's some way to identify and drop the events from the re-raise.
https://docs.sentry.io/platforms/python/configuration/filtering/

arugifa:
> Not sure if it's viable here, but have you looked into using the sentry "before-send" hook to filter exceptions you do not wish to send? Maybe there's some way to identify and drop the events from the re-raise.
> https://docs.sentry.io/platforms/python/configuration/filtering/

Unfortunately, in this case, all possible errors / exceptions should be caught. And the event ID generated by Sentry returned as part of a 500 HTTP response :wink:

lovelydinosaur:
I'm going to close this off, as I don't think it's a pattern that makes much sense for most folks.

PR #1170, opened 2021-04-27
Eliminated code duplication in middleware/gzip.py

ShahriyarR:
Before refactoring, the code portion below was repeated 3 times. Decreased to one.
```
await self.send(message)
```


Kludex (commented):
Thanks for the PR @ShahriyarR ! :1st_place_medal: 

This kind of PRs are difficult to review. :thinking: 

I'm usually in favor of removing duplicated code, but having the `pass` statement makes me think twice. 

For that reason, I'm closing this PR. Don't feel discouraged to open similar PRs, we are always willing to improve our code source! :smile:

PR #1172, opened 2021-04-27
Move the duplicated code in routing.py to their own functions

ShahriyarR:
There are cases where the same code blocks were used twice.
Intuitively I have moved the repeated logic to separate functions.
Ideally, there can be some shared state for those, but for now, it is enough.
It will be great if you can provide the text for docstring for the further explanation :)


fuadaghazada (approved):

lovelydinosaur:
Closing because let's avoid code churn.
There's not a clear enough win here.

PR #1176, opened 2021-05-01
[typing] Allow non-class callables to be used as middleware

decorator-factory:
1. Right now, it's possible to pass a sequence of tuples `(callable, options)` where `callable` is a callable accepting `**options` and returning an ASGI app. But the type annotation on `Starlette` says that one has to pass a sequence of `Middleware` objects.
This PR allows passing in a sequence of tuples as well as a list of `Middleware` objects

2. Type annotations only allow classes to be used as middleware, but it's possible to provide any callable that returns an ASGI app. This PR fixes that as well.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -9,6 +9,12 @@ from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 
+MiddlewareList = typing.Sequence[typing.Union[
+    Middleware,
+    typing.Tuple[typing.Callable[..., ASGIApp], typing.Dict[str, typing.Any]]
+]]
+
+
 class Starlette:
     """
     Creates an application instance.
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
Thanks. Honestly I'd rather prefer we don't add extra options here.
Keeping things constrained to one particular style helps keep everything just a little bit more simple.

decorator-factory:
That's understandable. But why can't I use a function as an ASGI app? It works just fine at runtime.

graingert:
I'm in favour of this being a `Callable[[], asgiref.typing.ASGI3App]`

Kludex:
I'll go with Tom on the `applications.py` - too complex. But I think the `cls` parameter needs a change of type on the `Middleware.__init__`.

> I'm in favour of this being a Callable[[], asgiref.typing.ASGI3App]

We don't have asgiref here yet.

I'll close this based on @tomchristie 's and my own comments, and the fact that the goal of the PR was to attack the `middleware` parameter (`cls` was just a side effect). Thanks for effort @decorator-factory .

PR #1177, opened 2021-05-04
Add better exception msg for duplicated param names

uriyyo:
Raise `ValueError` with information about duplicated names instead of `sre_constants.error`.

For instance:
```py
app = Starlette(
    routes=[
        Route("/api/users/{id}/groups/{id}", route),
    ]
)
```

Will raise:
```py
ValueError: Duplicated param name id at path /api/users/{id}/groups/{id}
```
Instead of:
```
sre_constants.error: redefinition of group name 'id' as group 2; was group 1 at position 42
```

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -107,6 +107,7 @@ def compile_path(
     """
     path_regex = "^"
     path_format = ""
+    duplicated_params = set()
 
     idx = 0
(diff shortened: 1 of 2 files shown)


lovelydinosaur (approved):
Really nicely done. 👍

PR #1179, opened 2021-05-10
RFC: Add StaticResources file

DavidVentura:
This PR only exists to facilitate discussion around the idea of using pkg_resources to serve files directly


Kludex:
Following @tomchristie's comment, I'm closing this.

We will gladly accept a PR with "that we can then link to from the main Starlette docs"! :smile:

PR #1180, opened 2021-05-12
Optional custom GraphiQL welcome message/default query

ChristianMertes:
Adds the `graphiql_welcome_message` parameter to the `GraphQLApp` constructor, which – if set – replaces the standard `# Welcome to GraphiQL` block GraphiQL greets its users with, to allow a custom message as a comment and also an example query for the API you developed to get the user started.

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -34,9 +34,11 @@ class GraphQLApp:
         executor: typing.Any = None,
         executor_class: type = None,
         graphiql: bool = True,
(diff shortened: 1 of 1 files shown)


JayH5:
Thanks for your PR. Unfortunately, we've decided to deprecate GraphQL built in to Starlette. Please see #619

lovelydinosaur:
Thanks for your triage work @JayH5 🙏

PR #1181, opened 2021-05-12
added way to have custom context for graphql

ChristianMertes:
The [Graphene docs recommend](https://docs.graphene-python.org/en/latest/execution/execute/) that a custom context can be passed to a query. Starlette uses this capability but seems to hide it from the user.
This PR adds a context parameter to the `GraphQLApp` class to allow users to inject their own context in addition (but not overwriting) Starlette's context.

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -34,9 +34,11 @@ class GraphQLApp:
         executor: typing.Any = None,
         executor_class: type = None,
         graphiql: bool = True,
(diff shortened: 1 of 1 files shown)


Kludex:
GraphQL is deprecated on Starlette. I guess this will be closed soon :kissing:

JayH5:
Thanks for your PR. Unfortunately, we've decided to deprecate GraphQL built in to Starlette. Please see #619

PR #1183, opened 2021-05-12
Use non-deprecated jinja2.pass_context in Jinja 3.0+

JayH5:
The deprecation warning from Jinja when using `contextfunction` is currently causing CI to fail because our pytest config is pretty strict.

Use the non-deprecated decorator when available. This maintains support for older versions of `jinja2`. Alternative options:
* Drop support for `jinja2<3`
* Emit our own deprecation warning for `jinja2<3`

### DIFF
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -6,6 +6,12 @@ from starlette.types import Receive, Scope, Send
 
 try:
     import jinja2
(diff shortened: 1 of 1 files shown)


lovelydinosaur (approved):
👍

PR #1189, opened 2021-05-27
Updates to python 3.6+ syntax

uSpike:
- Updates type hint comments to use annotations (`com2ann`)
- Updates various syntax to python 3.6+ syntax (`pyupgrade --py36-plus`)

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -231,7 +231,7 @@ class CommaSeparatedStrings(Sequence):
         return f"{class_name}({items!r})"
 
     def __str__(self) -> str:
-        return ", ".join([repr(item) for item in self])
+        return ", ".join(repr(item) for item in self)
 
 
 class ImmutableMultiDict(typing.Mapping):
@@ -649,7 +649,7 @@ class MutableHeaders(Headers):
         self["vary"] = vary
 
 
-class State(object):
+class State:
     """
     An object that can be used to store arbitrary state.
 
@@ -659,7 +659,7 @@ class State(object):
     def __init__(self, state: typing.Dict = None):
         if state is None:
             state = {}
-        super(State, self).__setattr__("_state", state)
+        super().__setattr__("_state", state)
 
     def __setattr__(self, key: typing.Any, value: typing.Any) -> None:
         self._state[key] = value
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -258,7 +258,7 @@ class ImmutableMultiDict(typing.Mapping):
             )
 
         if not value:
-            _items = []  # type: typing.List[typing.Tuple[typing.Any, typing.Any]]
+            _items: typing.List[typing.Tuple[typing.Any, typing.Any]] = []
         elif hasattr(value, "multi_items"):
             value = typing.cast(ImmutableMultiDict, value)
             _items = list(value.multi_items())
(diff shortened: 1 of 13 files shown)


starlette/datastructures.py
```diff
@@ -231,7 +231,7 @@ def __repr__(self) -> str:
         return f"{class_name}({items!r})"
 
     def __str__(self) -> str:
-        return ", ".join([repr(item) for item in self])
+        return ", ".join(repr(item) for item in self)
```
Kludex, line 234: Good catch.

starlette/middleware/trustedhost.py
```diff
@@ -53,7 +53,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             if found_www_redirect and self.www_redirect:
                 url = URL(scope=scope)
                 redirect_url = url.replace(netloc="www." + url.netloc)
-                response = RedirectResponse(url=str(redirect_url))  # type: Response
+                response: Response = RedirectResponse(url=str(redirect_url))
```
Kludex: Is it needed here?

starlette/testclient.py
```diff
@@ -268,8 +268,8 @@ def __init__(self, app: ASGI3App, scope: Scope) -> None:
         self.app = app
         self.scope = scope
         self.accepted_subprotocol = None
-        self._receive_queue = queue.Queue()  # type: queue.Queue
-        self._send_queue = queue.Queue()  # type: queue.Queue
+        self._receive_queue: queue.Queue = queue.Queue()
```
Kludex: Are those needed?

starlette/middleware/trustedhost.py
uSpike: Yes, or alternatively:
```python
            response: Response
            if found_www_redirect and self.www_redirect:
                url = URL(scope=scope)
                redirect_url = url.replace(netloc="www." + url.netloc)
                response = RedirectResponse(url=str(redirect_url))
            else:
                response = PlainTextResponse("Invalid host header", status_code=400)
```

starlette/testclient.py
uSpike: Yes, otherwise:
```
+ mypy starlette tests
starlette/testclient.py:271: error: Need type annotation for '_receive_queue'
starlette/testclient.py:272: error: Need type annotation for '_send_queue'
```

JayH5: I think these (and `asyncio.Queue`) are (or at least were) needed because the elements are typed, but only in the mypy typestubs on Python 3.9+.

So on Python 3.9 you can do:
```python
foo: queue.Queue[str] = queue.Queue()
```

but on earlier versions you had to do
```python
foo: "queue.Queue[str]" = queue.Queue()
```
mypy wants the element type but can't infer it (or that it's `Any`) unless you explicitly set the type.

starlette/middleware/trustedhost.py
uSpike: Without this annotation:
```
+ mypy starlette tests
starlette/middleware/trustedhost.py:58: error: Incompatible types in assignment (expression has type "PlainTextResponse", variable has type "RedirectResponse")
```

starlette/testclient.py
JayH5: We could maybe do this to make it clearer:
```suggestion
        self._receive_queue: "queue.Queue[typing.Any]" = queue.Queue()
```

starlette/middleware/trustedhost.py
Kludex: I usually do what you suggested on the comment (I think it makes more sense). But the current way it's fine.

starlette/testclient.py
Kludex: Not here, but in the future, should we consider using [those](https://github.com/django/asgiref/blob/main/asgiref/typing.py)? 

Like: `self._receive_queue: "queue.Queue[ASGIReceiveEvent]" = queue.Queue()`

starlette/middleware/trustedhost.py
uSpike: I made the change because I agree it makes more sense.

Kludex (approved):

JayH5 (approved):
Great, thanks

PR #1199, opened 2021-06-13
Deduplicate failure text in CORS preflight response

laggardkernel:
Break the loop to t avoid appending string `headers` into `failures` multiple times. Otherwise it causes duplication in CORS response text (not response header).

```diff
        elif requested_headers is not None:
            for header in [h.lower() for h in requested_headers.split(",")]:
                if header.strip() not in self.allow_headers:
                    failures.append("headers")
+                    break
        ...
        if failures:
            failure_text = "Disallowed CORS " + ", ".join(failures)
            return PlainTextResponse(failure_text, status_code=400, headers=headers)
```

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -129,6 +129,7 @@ class CORSMiddleware:
             for header in [h.lower() for h in requested_headers.split(",")]:
                 if header.strip() not in self.allow_headers:
                     failures.append("headers")
+                    break
 
         # We don't strictly need to use 400 responses here, since its up to
         # the browser to enforce the CORS policy, but its more informative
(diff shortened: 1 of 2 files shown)


JayH5:
Thanks. Any way we can add a test for this?

laggardkernel:
It's such a trivial change, just break the loop to avoid appending a value into the same list multiple times. Do we really need a test for it? @JayH5

laggardkernel:
Anyway, I added a test and force pushed the commit.

JayH5 (approved):
> It's such a trivial change, just break the loop to avoid appending a value into the same list multiple times. Do we really need a test for it?

I didn't completely understand this at first so just thought bug == test please 🙂. Probably wasn't super necessary but can't hurt, thanks.

PR #1200, opened 2021-06-15
Fixed error when response headers missing

abersheeran:
https://github.com/abersheeran/asgi-ratelimit/issues/14


euri10 (approved):
easy one, from the [spec](https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event) about the headers key:

> Optional; if missing defaults to an empty list.

PR #1204, opened 2021-06-18
Use coverage directly instead of pytest-cov

JayH5:
Similar to encode/uvicorn#809, encode/httpcore#218, encode/httpx#1353

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,9 +17,6 @@ combine_as_imports = True
 
 [tool:pytest]
 addopts =
-  --cov-report=term-missing:skip-covered
-  --cov=starlette
-  --cov=tests
   -rxXs
   --strict-config
   --strict-markers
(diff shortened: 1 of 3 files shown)


Kludex (approved):
Should we keep the `skip-covered`? :thinking:

JayH5:
> Should we keep the skip-covered? 🤔

This used to apply to a coverage report printed as part of the pytest run. Now no report is printed, but `scripts/coverage` does print the same report and includes `--skip-covered`.

setup.cfg
```diff
@@ -32,3 +29,6 @@ filterwarnings=
     ignore: Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated.*:DeprecationWarning
     ignore: The 'context' alias has been deprecated. Please use 'context_value' instead\.:DeprecationWarning
     ignore: The 'variables' alias has been deprecated. Please use 'variable_values' instead\.:DeprecationWarning
+
+[coverage:run]
+include = starlette/*, tests/*
```
euri10:
```suggestion
include = starlette/*, tests/*
omit = venv/*
```

JayH5: I don't actually understand why we need `omit = venv/*` since that doesn't fall under the `include` paths?

graingert: It would be better to use source_pkgs = starlette

euri10: > I don't actually understand why we need `omit = venv/*` since that doesn't fall under the `include` paths?

I blindlessly looked at what we do in uvicorn and indeed I now dont see why we do that :hamburger:

JayH5:
Have tested `source_pkgs` a bit and seems to do what we want so I'm going ahead with this.

PR #1206, opened 2021-06-20
♻️ Use asgiref typing instead of starlette types

Kludex:
Since `uvicorn` added `asgiref` as a dependency (for typing), we probably want to do the same here.

I didn't discuss this yet, so I'm not sure about the opinion of other members, feel free to comment if you see this. I'm going to write more about this later on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     packages=find_packages(exclude=["tests*"]),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0,<4"],
+    install_requires=["anyio>=3.0.0,<4", "asgiref>=3.3.4,<4"],
     extras_require={
         "full": [
             "graphene; python_version<'3.10'",
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,12 +1,18 @@
 import typing
 
+from asgiref.typing import (
+    ASGI3Application,
+    ASGIReceiveCallable,
+    ASGISendCallable,
+    Scope,
+)
+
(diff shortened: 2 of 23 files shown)


JayH5:
Not against this eventually as it'd be nice to have shared types between encode projects. However, a first step might be to just re-export the `asgiref` types as the ones we use under `starlette.types`. That would greatly reduce the changes and make it a lot easier for our users to upgrade.

Kludex:
> However, a first step might be to just re-export the asgiref types as the ones we use under starlette.types.

I thought about that, but I think the `Message` will be an issue, as there's no compatibility with `asgiref`. `Message` would be something like `Union[ASGIReceiveEvent, ASGISendEvent]`, but this will probably cause a lot of `mypy` conflicts.

That being said, I'm going to try your suggestion in a few days. 👍

Kludex:
Ok. I tried that.

```python
import typing

from asgiref.typing import Scope  # noqa: F401
from asgiref.typing import (
    ASGI3Application,
    ASGIReceiveCallable,
    ASGIReceiveEvent,
    ASGISendCallable,
    ASGISendEvent,
)

Message = typing.Union[ASGIReceiveEvent, ASGISendEvent]

Receive = ASGIReceiveCallable
Send = ASGISendCallable

ASGIApp = ASGI3Application
```
And I got:
```bash
Found 518 errors in 19 files (checked 62 source files)
```
:laughing: :rofl:

Kludex:
I'll close this because I'll not have time to solve the conflicts here. There are ~280. 

If we don't do this, when using `uvicorn.run`, we'll always have to ignore the `mypy` error now that `uvicorn` is almost fully `mypy` compliant.

PR #1202, opened 2021-06-18
Prepare version 0.15.0

JayH5:
- [x] Write the changelog
- [x] #1157 
- [x] #1201 
- [x] Anything else we want to squeeze in?? #1200 ✅ 
- [x] Bump the version

PR #1208, opened 2021-06-22
Update mkdocs.yml to fix syntax highlighting

lewoudar:
Fixed syntax highlighting

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -35,7 +35,8 @@ nav:
 markdown_extensions:
   - mkautodoc
   - admonition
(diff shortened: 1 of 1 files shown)


JayH5:
Hey, I believe this is actually fixed on master, we just haven't done a release in a while so the docs haven't been updated. That should change with #1202.

lewoudar:
Ok, I think it's been a while since there was this problem, so I hope it will be fixed soon
And thanks for anyio support!

JayH5:
Yep this is fixed 👍

PR #1210, opened 2021-06-23
TestClient accepts backend and backend_options as arguments

### LINKED ISSUE #1209: allow passing anyio backend options via TestClient constructor
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [x] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

I want to be able to:

```python
def test_foo(anyio_backend_name, anyio_backend_options):
    with starlette.testing.TestClient(
        create_app(),
        backend=anyio_backend_name,
        backend_options=anyio_backend_options,
    ) as client:
        ...
```

allow passing anyio backend options via TestClient constructor

None

## Additional context

this is needed so I can easily run the fastapi test suite on both trio and asyncio

uSpike:
As implemented in #1157, `TestClient.async_backend` was a class-level attribute.  This made it easy to monkeypatch for use in starlette tests, but makes it difficult/impossible to run different test client backends concurrently (in threads).

This PR changes TestClient to accept a `backend` and `backend_options` arguments which allow a user to specify the backend options for only that TestClient instance.  The class attribute will remain as the default for these arguments.

Fixes #1209


starlette/testclient.py
```diff
@@ -381,13 +381,11 @@ def receive_json(self, mode: str = "text") -> typing.Any:
 
 class TestClient(requests.Session):
     __test__ = False  # For pytest to not discover this up.
-
-    #: These options are passed to `anyio.start_blocking_portal()`
+    #: These are the default options for the constructor arguments
```
graingert, line 384: You should be able to move the type annotation onto the assignment in `def __init__`

graingert: Oh I see this is a ClassVar

graingert (approved):
Thanks for the fix, I think it's worth removing the classvar and using a functools.partial object for the tests

starlette/testclient.py
```diff
@@ -381,13 +381,11 @@ def receive_json(self, mode: str = "text") -> typing.Any:
...
     async_backend: typing.Dict[str, typing.Any] = {
```
JayH5, line 385: Do we need this class-level declaration still? Class vars always confuse me, but this means somebody can do:
```python
TestClient.async_backend["backend"] = "trio"
```
to default all future `TestClient` instances to use the Trio backend, right?

I know removing this would be slightly changing the API, however in the code examples we gave we only modified instance attributes and that usage should stay the same. What's your thinking around this?

graingert: I've removed it in my followup PR: https://github.com/encode/starlette/pull/1211

graingert:
merged in #1211

PR #1211, opened 2021-06-23
TestClient accepts backend and backend_options as arguments to constructor

graingert:


setup.py
```diff
@@ -37,7 +37,10 @@ def get_long_description():
     packages=find_packages(exclude=["tests*"]),
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
-    install_requires=["anyio>=3.0.0,<4"],
+    install_requires=[
+        "anyio>=3.0.0,<4",
+        "typing_extensions; python_version < '3.8'",
```
graingert, line 42: this dep is already in place implicitly via: https://github.com/agronholm/anyio/blob/c90169c0bbf0adc53234bde218e265f1b9157797/setup.cfg#L35-L36

uSpike:
It seems you've removed the parametrization of `anyio_backend` that ran the tests on both trio and asyncio.

graingert:
> It seems you've removed the parametrization of `anyio_backend` that ran the tests on both trio and asyncio.

anyio_backend is set to `["trio", "asyncio"]` by default:https://github.com/agronholm/anyio/blob/c90169c0bbf0adc53234bde218e265f1b9157797/src/anyio/pytest_plugin.py#L140-L142

uSpike:
Hm you're right, the default for anyio_backend is to use all backends: https://github.com/agronholm/anyio/blob/3.2.0/src/anyio/pytest_plugin.py#L140

By deleting it you're effectively testing on all backends supported by anyio (but also not explicitly passing options).

tests/conftest.py
```diff
@@ -7,22 +8,16 @@
...
-    ],
-    autouse=True,
-)
-def anyio_backend(request, monkeypatch):
-    monkeypatch.setattr(TestClient, "async_backend", request.param)
-    return request.param["backend"]
+@pytest.fixture
+def no_trio_support(anyio_backend_name):
+    if anyio_backend_name == "trio":
+        pytest.skip("Trio not supported (yet!)")
```
graingert, line 14:
```suggestion
        pytest.skip("Trio not supported (yet!)")

@pytest.fixture(params=["asyncio", "trio"])
def anyio_backend(request):
    return request.param
```

@uSpike I could add this back in?

starlette/testclient.py
```diff
@@ -91,11 +97,16 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         await instance(receive, send)
 
 
+class _AsyncBackend(TypedDict):
```
JayH5, line 100: I'm not really convinced this `TypedDict` is really worth it over just `Dict[str, Any]` 🤔. Definitely debatable (and I guess a matter of taste), but what class of bug is this really helping with?

tests/conftest.py
```diff
@@ -7,22 +8,16 @@
 collect_ignore = ["test_graphql.py"] if sys.version_info >= (3, 10) else []
 
 
-@pytest.fixture(
-    params=[
```
JayH5, line 11: Question: what happened to these params? Where do the values for `anyio_backend_name`/`anyio_backend_options` come from now? (What am I missing?)

starlette/testclient.py
graingert: mypy processes this for the `anyio.start_blocking_portal(**self.async_backend)`

tests/conftest.py
graingert: anyio provides a default set of parameters "trio" and "asyncio"

starlette/testclient.py
Kludex: Enables autocomplete functionalities (depending on plugins..), and type checking tools are able to detect invalid keys e.g. mypy.

tests/conftest.py
graingert: There's another thread here about it https://github.com/encode/starlette/pull/1211#issuecomment-867152021

JayH5: Thanks. So `anyio_backend_options` is going to be `{}` always it looks like?

starlette/testclient.py
JayH5: > mypy processes this for the anyio.start_blocking_portal(**self.async_backend)

I didn't know it could do that. That is kinda neat.

> Enables autocomplete functionalities (depending on plugins..), and type checking tools are able to detect invalid keys e.g. mypy.

Sure, but..it's an internal interface with 2 keys...

Anyway, this isn't blocking my approval, just my 2c.

tests/conftest.py
JayH5: I do think this is a little non-obvious, though, so maybe just a comment with a link to `https://anyio.readthedocs.io/en/stable/testing.html#specifying-the-backends-to-run-on` would be good?

JayH5 (approved):
Thanks. Only thing is the PR title--does this count as "monkeypatching"?

starlette/testclient.py
```diff
@@ -18,6 +19,11 @@
 from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocketDisconnect
 
+if sys.version_info >= (3, 8):  # pragma: no cover
+    from typing import TypedDict  # pragma: no cover
+else:  # pragma: no cover
+    from typing_extensions import TypedDict  # pragma: no cover
```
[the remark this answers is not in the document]
graingert: let's see what happens

graingert:
> Thanks. Only thing is the PR title--does this count as "monkeypatching"?


isn't that what this is?
```
TestClient.async_backend["backend"] = "trio"
```

JayH5:
> isn't that what this is?

🤔 I guess technically it is monkey patching since it's modifying a class attribute at runtime. Always thought of monkey patching as a "bigger hammer" where you modify a method at runtime and potentially do so in a way that wasn't intended by the original author.

PR #1213, opened 2021-06-23
 ensure TestClient requests run in the same EventLoop as lifespan 

graingert:
Ensure TestClient requests run in the same EventLoop as lifespan. The way a TestClient context interacts with app lifespan and route task is important to support app-wide task groups.

Before, a new thread was started for every interaction with the TestClient, most critically lifespan was run in a different event loop than any subsequent request. Not only does this violate the ASGI spec, it also means that resources (locks, databases and task groups) created in lifespan cannot be used in subsequent requests.

After this change each use of the TestClient synchronous context manager creates and caches a BlockingPortal and any further interactions in this context use that cached blocking portal. Interactions outside of this context continue to acquire a new blocking portal for each interaction.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -12,7 +12,7 @@ import typing
 from concurrent.futures import Future
 from urllib.parse import unquote, urljoin, urlsplit
 
-import anyio
+import anyio.abc
 import requests
 from anyio.streams.stapled import StapledObjectStream
 
@@ -20,9 +20,15 @@ from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocketDisconnect
 
 if sys.version_info >= (3, 8):  # pragma: no cover
-    from typing import TypedDict
+    from typing import Protocol, TypedDict  # pragma: no cover
 else:  # pragma: no cover
-    from typing_extensions import TypedDict
+    from typing_extensions import Protocol, TypedDict  # pragma: no cover
+
+
+class _PortalFactoryType(Protocol):
+    def __call__(self) -> typing.ContextManager[anyio.abc.BlockingPortal]:
+        ...  # pragma: no cover
+
 
 # Annotations for `Session.request()`
 Cookies = typing.Union[
@@ -106,14 +112,14 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
     def __init__(
         self,
         app: ASGI3App,
-        async_backend: _AsyncBackend,
+        portal_factory: _PortalFactoryType,
         raise_server_exceptions: bool = True,
         root_path: str = "",
     ) -> None:
         self.app = app
         self.raise_server_exceptions = raise_server_exceptions
         self.root_path = root_path
-        self.async_backend = async_backend
+        self.portal_factory = portal_factory
 
     def send(
         self, request: requests.PreparedRequest, *args: typing.Any, **kwargs: typing.Any
@@ -162,7 +168,7 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
                 "server": [host, port],
                 "subprotocols": subprotocols,
             }
-            session = WebSocketTestSession(self.app, scope, self.async_backend)
+            session = WebSocketTestSession(self.app, scope, self.portal_factory)
             raise _Upgrade(session)
 
         scope = {
@@ -252,7 +258,7 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
                 context = message["context"]
 
         try:
-            with anyio.start_blocking_portal(**self.async_backend) as portal:
+            with self.portal_factory() as portal:
                 response_complete = portal.call(anyio.Event)
                 portal.call(self.app, scope, receive, send)
         except BaseException as exc:
@@ -285,20 +291,18 @@ class WebSocketTestSession:
         self,
         app: ASGI3App,
         scope: Scope,
-        async_backend: _AsyncBackend,
+        portal_factory: _PortalFactoryType,
     ) -> None:
         self.app = app
         self.scope = scope
         self.accepted_subprotocol = None
-        self.async_backend = async_backend
+        self.portal_factory = portal_factory
         self._receive_queue: "queue.Queue[typing.Any]" = queue.Queue()
         self._send_queue: "queue.Queue[typing.Any]" = queue.Queue()
 
     def __enter__(self) -> "WebSocketTestSession":
         self.exit_stack = contextlib.ExitStack()
-        self.portal = self.exit_stack.enter_context(
-            anyio.start_blocking_portal(**self.async_backend)
-        )
+        self.portal = self.exit_stack.enter_context(self.portal_factory())
 
         try:
             _: "Future[None]" = self.portal.start_task_soon(self._run)
@@ -396,6 +400,7 @@ class WebSocketTestSession:
 class TestClient(requests.Session):
     __test__ = False  # For pytest to not discover this up.
     task: "Future[None]"
+    portal: typing.Optional[anyio.abc.BlockingPortal] = None
 
     def __init__(
         self,
@@ -418,7 +423,7 @@ class TestClient(requests.Session):
             asgi_app = _WrapASGI2(app)  #  type: ignore
         adapter = _ASGIAdapter(
             asgi_app,
-            self.async_backend,
+            portal_factory=self._portal_factory,
             raise_server_exceptions=raise_server_exceptions,
             root_path=root_path,
         )
@@ -429,6 +434,16 @@ class TestClient(requests.Session):
         self.headers.update({"user-agent": "testclient"})
         self.app = asgi_app
         self.base_url = base_url
+
+    @contextlib.contextmanager
+    def _portal_factory(
+        self,
+    ) -> typing.Generator[anyio.abc.BlockingPortal, None, None]:
+        if self.portal is not None:
+            yield self.portal
+        else:
+            with anyio.start_blocking_portal(**self.async_backend) as portal:
+                yield portal
 
     def request(  # type: ignore
         self,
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
+import anyio.abc
 import requests
 from anyio.streams.stapled import StapledObjectStream
 
 from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocketDisconnect
 
 if sys.version_info >= (3, 8):  # pragma: no cover
-    from typing import TypedDict
+    from typing import Protocol, TypedDict  # pragma: no cover
```
```suggestion
    from typing import Protocol, TypedDict
```

starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
 else:  # pragma: no cover
-    from typing_extensions import TypedDict
+    from typing_extensions import Protocol, TypedDict  # pragma: no cover
```
```suggestion
    from typing_extensions import Protocol, TypedDict
```

tests/test_testclient.py
```diff
@@ -14,15 +25,56 @@ def mock_service_endpoint(request):
...
+        app.shutdown_loop = get_identity(counter)
+
     @app.route("/")
     def homepage(request):
         client = test_client_factory(mock_service)
         response = client.get("/")
         return JSONResponse(response.json())
 
+    @app.route("/thread")
+    async def thread(request):
```
graingert:
```suggestion
    @app.route("/loop_id")
    async def loop_id(request):
```

tests/test_testclient.py
```diff
@@ -46,9 +98,48 @@ def test_use_testclient_in_endpoint(test_client_factory):
...
+    assert client.get("/thread").json() == 1
+    assert client.get("/thread").json() == 2
+
+    first_task = app.startup_task
+
+    with client:
+        # the TestClient context can be re-used, starting a new lifespan task
+        # in a new thread
+        assert client.get("/thread").json() == 3
+        assert client.get("/thread").json() == 3
```
graingert:
```suggestion
        assert client.get("/loop_id").json() == 0
        assert client.get("/loop_id").json() == 0

    # that thread is also the same as the lifespan thread
    assert app.startup_loop == 0
    assert app.shutdown_loop == 0

    # lifespan events run in the same task, this is important because a task
    # group must be entered and exited in the same task.
    assert app.startup_task is app.shutdown_task

    # outside the TestClient context, new requests continue to spawn in new
    # eventloops in new threads
    assert client.get("/loop_id").json() == 1
    assert client.get("/loop_id").json() == 2

    first_task = app.startup_task

    with client:
        # the TestClient context can be re-used, starting a new lifespan task
        # in a new thread
        assert client.get("/loop_id").json() == 3
        assert client.get("/loop_id").json() == 3
```

JayH5 (commented):
This gets into some pretty low-level AnyIO stuff that I'm going to have to read up on.. I'll try get back to this tomorrow. 😴

tests/test_testclient.py
```diff
@@ -1,11 +1,22 @@
...
 
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.responses import JSONResponse
 from starlette.websockets import WebSocket, WebSocketDisconnect
 
+if sys.version_info >= (3, 7):
+    from asyncio import current_task as asyncio_current_task  # pragma: no cover
+else:
+    asyncio_current_task = asyncio.Task.current_task  # pragma: no cover
```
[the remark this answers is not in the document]
graingert: that's `# pragma: no branch` right?

starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
-    from typing import TypedDict
+    from typing import Protocol, TypedDict
 else:  # pragma: no cover
-    from typing_extensions import TypedDict
+    from typing_extensions import Protocol, TypedDict
+
+
+class _PortalFactoryType(Protocol):
+    def __call__(self) -> typing.ContextManager[anyio.abc.BlockingPortal]:
+        ...  # pragma: no cover
```
```suggestion
_PortalFactoryType = typing.Callable[
    [], typing.ContextManager[anyio.abc.BlockingPortal]
]
```

starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
+import anyio.abc
 import requests
 from anyio.streams.stapled import StapledObjectStream
 
 from starlette.types import Message, Receive, Scope, Send
 from starlette.websockets import WebSocketDisconnect
 
 if sys.version_info >= (3, 8):  # pragma: no cover
-    from typing import TypedDict
+    from typing import Protocol, TypedDict
```
```suggestion
    from typing import TypedDict
```

starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
 else:  # pragma: no cover
-    from typing_extensions import TypedDict
+    from typing_extensions import Protocol, TypedDict
```
```suggestion
    from typing_extensions import TypedDict
```

starlette/testclient.py
```diff
@@ -12,17 +12,23 @@
...
+
+
+class _PortalFactoryType(Protocol):
+    def __call__(self) -> typing.ContextManager[anyio.abc.BlockingPortal]:
```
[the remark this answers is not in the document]
graingert: yep

starlette/testclient.py
```diff
@@ -430,6 +435,16 @@ def __init__(
         self.app = asgi_app
         self.base_url = base_url
 
+    @contextlib.contextmanager
+    def _portal_factory(
+        self,
+    ) -> typing.Generator[anyio.abc.BlockingPortal, None, None]:
```
uSpike, line 441:
```suggestion
    ) -> typing.Iterator[anyio.abc.BlockingPortal]:
```

starlette/testclient.py
```diff
@@ -490,29 +505,35 @@ def websocket_connect(
...
-        try:
-            self.task = self.portal.start_task_soon(self.lifespan)
-            self.portal.call(self.wait_startup)
-        except Exception:
-            self.exit_stack.close()
-            raise
+        with contextlib.ExitStack() as stack:
+            self.portal = portal = stack.enter_context(
+                anyio.start_blocking_portal(**self.async_backend)
+            )
```
uSpike, line 511:
```suggestion
            self.portal = portal = stack.enter_context(self._portal_factory())
```

uSpike: Why not reuse portal factory here?

tests/test_testclient.py
```diff
@@ -14,15 +25,56 @@ def mock_service_endpoint(request):
...
+            raise RuntimeError("must be called from a running task")  # pragma: no cover
+        return task
+    raise RuntimeError(f"unsupported asynclib={asynclib_name}")  # pragma: no cover
+
+
+def create_app(test_client_factory, counter=itertools.count()):
     app = Starlette()
 
+    @app.on_event("startup")
+    async def get_startup_thread():
```
uSpike: Should this be `get_startup_task`?  I don't think we're working with threads here.

tests/test_testclient.py
```diff
@@ -14,15 +25,56 @@ def mock_service_endpoint(request):
...
+        app.startup_task = current_task()
+        app.startup_loop = get_identity(counter)
+
+    @app.on_event("shutdown")
+    async def get_shutdown_thread():
```
uSpike: Same here, `thread`->`task`?

uSpike (commented):
Looks good to me, I had a few suggestions.  I have some confusion over the mixed use of the terms "thread" and "task".  Could you please clarify?

starlette/testclient.py
graingert: ah I didn't want any chance of recursion

tests/test_testclient.py
graingert: actually this should all be lifespan_context

starlette/testclient.py
uSpike: mypy still passes though?

~Or is https://docs.python.org/3/library/typing.html#typing.ContextManager more appropriate?~

uSpike: Nevermind on typing.ContextManager, derp

uSpike: The typeshed definition for `contextmanager` takes an `Iterable`: https://github.com/python/typeshed/blob/master/stdlib/contextlib.pyi#L37

graingert: it's a bug: https://github.com/python/typeshed/issues/2772

uSpike: Alrighty, never-mind then

uSpike:
~Is the guarantee of having the same task run lifespan startup/shutdown only (currently) valid for the TestClient?  What I'm concerned about is that someone may write tests for their app which use a TaskGroup in the lifespan, and that works fine in the tests, but then would fail if ran in production via `uvicorn` for example.~

starlette/testclient.py
uSpike: That's fair.

graingert:
@uSpike uvicorn has to work this way because it only uses one eventloop per process. TestClient was broken because it uses multiple eventloops per process, which I think isn't technically valid under the lifespan asgi prototocol.

uSpike:
~Right, but you're also checking that the same _task_ is used for lifespan events, and using a task group to check that as well.  I don't think that will currently work outside of these tests, am I wrong?~  Edit: i'm wrong

graingert:
@uSpike lifespan is specified in asgi as always being a single task. In fact it's stricter - `Callable[[Scope, Rx, Tx], Awaitable[None]]`

uSpike:
OK as discussed on gitter I was wrong about lifespan events being triggered in separate tasks.

uSpike (approved):
I'm ok with this but I'd still wait for @JayH5 before merging

tests/test_testclient.py
```diff
@@ -14,15 +25,54 @@ def mock_service_endpoint(request):
...
+        if task is None:
+            raise RuntimeError("must be called from a running task")  # pragma: no cover
+        return task
+    raise RuntimeError(f"unsupported asynclib={asynclib_name}")  # pragma: no cover
+
+
+def create_app(test_client_factory, counter=itertools.count()):
+    async def lifespan_context(app):
+        app.startup_task = current_task()
+        app.startup_loop = get_identity(counter)
```
JayH5: Since these attributes are only used in one test, would it make more sense to not modify this `create_app` function that is used by other tests and instead do this inside the test function? Could then use `nonlocal`s or something instead of assigning attributes to `app`.

JayH5 (commented):
Overall LGTM. I would be grateful for a bit more context in the PR description, though. I understand that you & @uSpike have had conversations about this on gitter, but it'd be good to have a clearer explanation here like: "before things were like _this_ and this was a problem because _reasons_, so after this change things are like _that_"...if you get what I mean.

It's difficult for someone who's not following the various threads to read this PR and understand what the actual improvement is.

starlette/testclient.py
```diff
@@ -490,29 +505,35 @@ def websocket_connect(
...
+
+            self.stream_send = StapledObjectStream(
+                *anyio.create_memory_object_stream(math.inf)
+            )
+            self.stream_receive = StapledObjectStream(
+                *anyio.create_memory_object_stream(math.inf)
+            )
+            self.task = portal.start_task_soon(self.lifespan)
+            portal.call(self.wait_startup)
+            self.portal = portal
```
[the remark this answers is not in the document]
graingert: Oh woops it doesn't need it

```suggestion
```

starlette/testclient.py
graingert, line 511:
```suggestion
            )

            @stack.callback
            def reset_portal() -> None:
                self.portal = None
```

tests/test_testclient.py
```diff
@@ -14,15 +25,54 @@ def mock_service_endpoint(request):
...
+
+    if asynclib_name == "asyncio":
+        task = asyncio_current_task()
+        if task is None:
+            raise RuntimeError("must be called from a running task")  # pragma: no cover
+        return task
+    raise RuntimeError(f"unsupported asynclib={asynclib_name}")  # pragma: no cover
+
+
+def create_app(test_client_factory, counter=itertools.count()):
```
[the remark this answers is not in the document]
graingert: Yes it's intentional so the thread count increases globally unless a counter is passed in

tests/test_testclient.py
graingert: Yeah that makes sense, although I want to assign the taskgroup to app as a demo of how this would be used

JayH5 (approved):
Nice! This is cool. Thanks for the updated PR description.

PR #1216, opened 2021-06-25
👷 Add Python venv cache

Kludex:
Following the same idea as https://github.com/encode/httpx/pull/1680 and https://github.com/encode/uvicorn/pull/1095.

This PR aims to add a cache to the Python environment.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -21,13 +21,25 @@ jobs:
       - uses: "actions/setup-python@v2"
         with:
           python-version: "${{ matrix.python-version }}"
+
+      - uses: actions/cache@v2.1.6
+        id: cache-requirements
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'setup.py') }}
+
       - name: "Install dependencies"
         run: "scripts/install"
+        if: steps.cache-requirements.outputs.cache-hit != 'true'
+
       - name: "Run linting checks"
         run: "scripts/check"
+
       - name: "Build package & docs"
         run: "scripts/build"
+
       - name: "Run tests"
         run: "scripts/test"
+
       - name: "Enforce coverage"
         run: "scripts/coverage"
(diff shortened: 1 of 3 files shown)


Kludex:
I'll put on "hold" until the `httpx` PR is merged, I want to see if @tomchristie has other concerns.

requirements.txt
```diff
@@ -1,31 +1,25 @@
-# Optionals
-graphene; python_version<'3.10'
-itsdangerous
-jinja2
-python-multipart
-pyyaml
-requests
+-e .[full]
```
JayH5, line 1: Do we really want to do `[full]`? These dependencies then won't be pinned and so will only be updated implicitly when the cache is invalidated.

.github/workflows/test-suite.yml
```diff
@@ -21,13 +21,25 @@ jobs:
       - uses: "actions/setup-python@v2"
         with:
           python-version: "${{ matrix.python-version }}"
+
+      - uses: actions/cache@v2.1.6
```
JayH5, line 25: What's the reason for asking for such a specific version?

JayH5 (commented):
1. The implication here is that we now pin our dev dependencies and somebody updates them every once in a while? (Seems we're not so keen on dependabot because it creates too many PRs)
2. Before/after "benchmarks" would be appreciated 🙏  although I understand you're waiting for this to move further along with httpx for now, so no rush.

JayH5:
I think overall I don't really have a problem with the build time as things are (~3min or less) so I'm not sure this is worth the extra complexity?

lovelydinosaur:
> I'll put on "hold" until the `httpx` PR is merged, I want to see if @tomchristie has other concerns.

Echoing https://github.com/encode/uvicorn/pull/1095#issuecomment-869564569 Agreed. Let's take things one step at a time. 😀

> I think overall I don't really have a problem with the build time as things are (~3min or less) so I'm not sure this is worth the extra complexity?

I do also share this outlook somewhat. Not definitively, but tentatively.

Kludex:
I'm closing this, as we still have to discuss on HTTPX.

PR #1124, opened 2021-01-11
Return 404 Not Found for accessing file as directory pattern

xKerman:
fixes #1123

PR #1220, opened 2021-06-27
Handle staticfiles OSError exceptions

### LINKED ISSUE #1123: `StaticFiles` causes Internal Server Error when user accesses existing files as directory
### Checklist

### Describe the bug

`StaticFiles` causes Internal Server Error when user accesses existing files as directory (e.g. `/static/somefile.txt/foobar`)

### To reproduce

1. create virtual env and activate it with `python -m venv venv && source venv/bin/activate`
2. install dependencies with `pip install starlette uvicorn aiofiles`
3. setup application as follows:

directory structure:
```
.
├── poc.py
├── static
│   └── sample.txt
└── venv
```

poc.py
```py
# code from https://www.starlette.io/staticfiles/
from starlette.routing import Mount


routes = [
    Mount("/static", app=StaticFiles(directory="static"), name="static"),
]

```

4. run application with `uvicorn poc:app`
5. access `http://127.0.0.1:8000/static/sample.txt/foo`

### Expected behavior

`StaticFiles` returns "404 Not Found" HTTP response (as Apache HTTP Server and Nginx does).

### Actual behavior

`StaticFiles` returns "500 Internal Server Error" HTTP response.

### Debugging material

console log with tracebacks:
```
INFO:     Started server process [13052]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     127.0.0.1:64288 - "GET /static/sample.txt/foo HTTP/1.1" 500 Internal Server Error
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 394, in run_asgi
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc from None
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/routing.py", line 582, in __call__
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/routing.py", line 392, in handle
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/staticfiles.py", line 97, in __call__
    response = await self.get_response(path, scope)
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/staticfiles.py", line 114, in get_response
    full_path, stat_result = await self.lookup_path(path)
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/starlette/staticfiles.py", line 154, in lookup_path
    stat_result = await aio_stat(full_path)
  File "/Users/xkhorasan/programs/test_starlette/venv/lib/python3.9/site-packages/aiofiles/os.py", line 13, in run
    return await loop.run_in_executor(executor, pfunc)
  File "/usr/local/Cellar/python@3.9/3.9.1_2/Frameworks/Python.framework/Versions/3.9/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
NotADirectoryError: [Errno 20] Not a directory: '/Users/xkhorasan/programs/test_starlette/static/sample.txt/foo'
```

### Environment

- OS: macOS
- Python version: 3.9.1
- Starlette version:  0.14.1

### Additional context

Apache HTTP Server and Nginx treat this case (access existing file as directory) as "404 Not Found".
Samples:
* Apache HTTP Server: https://httpd.apache.org/images/httpd_logo_wide_new.png/foo
* Nginx: https://nginx.org/download/nginx-1.18.0.tar.gz/foo

aminalaee:
Closes #1123 Issue
And can also closes #1124 old PR

This will handle `staticfiles` exceptions:
- If directory not found throws `404` (file not found was already handled)
- if permission denied on file/folder returns `401`
- if any other `OSError` exceptions happen, returns `500`


tests/test_staticfiles.py
```diff
@@ -279,3 +281,59 @@ def test_staticfiles_cache_invalidation_for_deleted_file_html_mode(tmpdir):
...
+    path = os.path.join(tmpdir, "example.txt")
+    with open(path, "w") as file:
+        file.write("<file content>")
+
+    app = StaticFiles(directory=tmpdir)
+    client = TestClient(app)
+
+    with mock.patch(
+        "starlette.staticfiles.StaticFiles.lookup_path"
+    ) as mock_lookup_path:
```
aminalaee: I haven't seen any usage of `unittest.mock` in the project, but I couldn't find any better way to create not-so-simple `OSError` exceptions.

starlette/staticfiles.py
```diff
@@ -156,8 +162,8 @@ async def lookup_path(
             try:
                 stat_result = await anyio.to_thread.run_sync(os.stat, full_path)
                 return full_path, stat_result
-            except FileNotFoundError:
-                pass
+            except OSError:
+                raise
```
JayH5: This try/except seems unnecessary now?

tests/test_staticfiles.py
JayH5: Is there any `OSError` that might cause this? Maybe a broken symlink or would that just 404?

aminalaee: I guess that would be a FileNotFound. I'll give it another try but all other unhandled `OSError` exceptions happened on writing a file or internal sys calls.

starlette/staticfiles.py
```diff
@@ -156,8 +162,8 @@ async def lookup_path(
             try:
                 stat_result = await anyio.to_thread.run_sync(os.stat, full_path)
```
graingert: os.path.realpath, os.path.commonprefix all perform blocking IO so this whole function should be in one run_sync call

sorry this is a review of unchanged code!

tests/test_staticfiles.py
graingert: statting a file you can't access gives you PermissionError:

```
>>> os.stat("/boot/bar")
  File "<stdin>", line 1, in <module>
PermissionError: [Errno 13] Permission denied: '/boot/bar'
```

graingert: ah you already have a test for that

starlette/staticfiles.py
aminalaee: I've changed `lookup_path` to a sync method and changed it's calls to `await anyio.to_thread.run_sync(self.lookup_path, path)`

tests/test_staticfiles.py
aminalaee: Yeah, it looks like all other exceptions are network or sys call exceptions https://docs.python.org/3/library/exceptions.html#os-exceptions

JayH5: Yeah, I don't have any other ideas for testing this without a mock 🤔 

Would it be possible to use pytest's monkeypatch functionality for this? Wouldn't make much difference, but at least we'd be sticking to "one test framework". But if too much effort don't bother.

aminalaee: @JayH5 I think that'd be better. I'll give it a try

aminalaee: @JayH5 Changed that, please check again if that's what you meant.

JayH5: Thanks yes that is what I meant. Sorry for the slow response 😕 

I've been thinking about this again and I'm not sure we actually want to return 500 from within the static files implementation. Should we not just allow/raise the exception since it is so unexpected? IMO, the `ServerErrorMiddleware` (and similar) should really be the only thing in Starlette that returns a 500 response.

aminalaee: @JayH5
That makes sense. What about the other exceptions?
Like 404 and 401, can't `ExceptionMiddleware` handle those?

JayH5: You do make a good point 🤔 I'm not sure why we don't raise an `HTTPException` in those cases. I'd be open to that change (if it's as straightforward as it seems) either in this PR or another.

aminalaee: @JayH5  I played around this a bit and changed that.
I think we can raise `HTTPException` instead of returning `PlaintTextResponse`s but there's a difference. If we raise `HTTPException` then we need to can't use `StaticFiles` alone as we need `Starlette` middlewares to handle the exception which `StaticFiles` doesn't have.

lovelydinosaur (commented):
@aminalaee - Looks neatly done. Happy to leave this up to your judgement.

starlette/staticfiles.py
```diff
@@ -109,9 +105,30 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
+            full_path, stat_result = await anyio.to_thread.run_sync(
+                self.lookup_path, path
+            )
+        except (FileNotFoundError, NotADirectoryError):
+            if self.html:
+                # Check for '404.html' if we're in HTML mode.
+                full_path, stat_result = await anyio.to_thread.run_sync(
+                    self.lookup_path, "404.html"
+                )
+                if stat_result is not None and stat.S_ISREG(stat_result.st_mode):
```
Kludex: I also prefer "is not None" in cases like this, but just below we have the same logic in a different format. :thinking: 
```suggestion
                if stat_result and stat.S_ISREG(stat_result.st_mode):
```
What do you think?

starlette/staticfiles.py
```diff
@@ -109,9 +105,30 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
+                    return FileResponse(
+                        full_path,
+                        stat_result=stat_result,
+                        method=scope["method"],
+                        status_code=404,
+                    )
```
Kludex, line 126: Why are we not using `self.file_response` here?

aminalaee: Makes sense, changed that.

aminalaee:
@Kludex I'm not sure if this is related to this PR. It was working alright, and the tests are from before this PR.

starlette/staticfiles.py
Kludex: This was actually a question, because the `self.file_response` does more than that: 
image

Kludex: check the `self.is_not_modified` logic. It's the reason why ur tests are failing.

aminalaee: Good catch, I didn't remember that change :D
Now that I take a look, looks like the answer to the question is "No". We can't use `self.file_response` as it will cache `404` responses. We have to create `FileResponse` there.

Kludex: People usually assume that I'm suggesting things, when I'm innocently asking questions hahaha

Thanks!

Kludex (approved):

lambdaq:
Heads up for anyone using  `lookup_path()`: it's now changed to a sync method in `starlette>=0.17.1`

I am using `await self.lookup_path` in my code this PR breaks it lol. 

I know it's my bad to use an undocumented internal method, but I just don't want to handle path on my own.

PR #1228, opened 2021-07-02
Response.delete_cookie accepts cookie options

### LINKED ISSUE #1207: request.delete_cookie fails to delete samesite='none' cookies
### Checklist

- [Y] The bug is reproducible against the latest release and/or `master`.
- [Y] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Calling `response.delete_cookie` in a route handler results in a response with a set-cookie header that has `sameSite='lax'`.

This set-cookie header is ignored by browsers if the cookie was being used in a cross-site context (i.e., the cookie was originally set via an xhr response with samesite='none'). The end result is that the cookie is not deleted.

### To reproduce

```python
# api.py

def home(request):
    return PlainTextResponse('Hello world!')

def login(request):
    response = PlainTextResponse('Logged in')
    response.set_cookie('token', 'abc', max_age=18000, httponly=True, secure=True, samesite='none')
    return response

def logout(request):
    response = PlainTextResponse('Logged out')
    response.delete_cookie('token')
    return response

routes = [
    Route('/', home),
    Route('/login', login, methods=['GET', 'POST']),
    Route('/logout', logout, methods=['GET', 'POST']),
]

middleware = [
    Middleware(
        CORSMiddleware,
        allow_origins=['http://localhost:9001'],
        allow_credentials=True,
        allow_methods=['*'],
        allow_headers=['*']
    )
]

app = Starlette(debug=True, routes=routes, middleware=middleware)
```

```html
<!-- index.html -->

<!DOCTYPE html>
<html>
  <head>
    <title>Starlette CORS cookie test</title>
  </head>
  <body>
    <div id="content"><button id="loginBtn">Login</button></div>

    <script>
      // CHANGE AS NECESSARY
      const API_URL = "https://dbdcec9f8fd9.ngrok.io";

      document.querySelector("#loginBtn").addEventListener("click", login);

      async function login() {
        await fetch(`${API_URL}/login`, {
          method: "POST",
          credentials: "include",
        });
        document.querySelector(
          "#content"
        ).innerHTML = `<button id="logoutBtn">Logout</button>`;
        document.querySelector("#logoutBtn").addEventListener("click", logout);
      }

      async function logout() {
        await fetch(`${API_URL}/logout`, {
          method: "POST",
          credentials: "include",
        });
        document.querySelector(
          "#content"
        ).innerHTML = `<button id="loginBtn">Login</button>`;
        document.querySelector("#loginBtn").addEventListener("click", login);
      }
    </script>
  </body>
</html>
```

1. Start the API server locally with `uvicorn api:app --port 9000`
2. Make the API available at a different site. An easy way is to use ngrok with `ngrok http 9000`. Let's say the API is now at https://dbdcec9f8fd9.ngrok.io
3. Change `API_URL` in index.html to point to the API
4. Serve index.html with `python -m http.server 9001`
5. Navigate to http://localhost:9001 and click the login button
6. Open devtools->Application and notice that the cookie is set
7. Click the logout button
8. Open devtools->Application and notice that the cookie is still set
9. (chrome-only) In devtools->Network, the response headers section for the /logout request has a warning indicator whose tooltip says the set-cookie header has been blocked. See screenshot below

Screenshot_20210622_133455

### Expected behavior

Cookie is deleted

### Actual behavior

Cookie is not deleted

### Environment

- OS: Linux
- Python version: 3.8.5
- Starlette version: 0.14.2

BookGin:
Though `functools.partial` could be used, it needs to be instantiated somewhere. Otherwise it will introduce the overhead of function object every time.

Let's just follow [werkzeug's implementation](https://github.com/pallets/werkzeug/blob/af160e0b6b7ddd81c22f1652c728ff5ac72d5c74/src/werkzeug/sansio/response.py#L250-L281).


Kludex:
@BookGin What do you mean by: "Though `functools.partial` could be used, it needs to be instantiated somewhere."?

Thanks for the PR! 🎉

BookGin:
Like this pseudo-code, so we need to instantiate the partial function somewhere.....

```python
from functools import partial
def delete_cookie(
       self,
       key: str,
       path: str = "/",
       domain: str = None,
       secure: bool = False,
       httponly: bool = False,
       samesite: str = "lax",
   ) -> None:
   if not hasattr(delete_cookie, '_func'):
       delete_cookie._func = partial(self.set_cookie, max_age=0, expires=0)
   return delete_cookie._func(path, domain, secure, httponly, samesite)
 ```
 
Never mind. I don't think it's more readable. Let's just keep things simple.

Kludex (approved):

Kludex:
@BookGin Thanks for the PR! :tada:

PR #1232, opened 2021-07-03
TestClient handle exceptions thrown from lifespan before shutdown event

graingert:


starlette/testclient.py
```diff
@@ -514,63 +518,59 @@ def __enter__(self) -> "TestClient":
...
-                *anyio.create_memory_object_stream(math.inf)
-            )
-            self.stream_receive = StapledObjectStream(
-                *anyio.create_memory_object_stream(math.inf)
+            startup_event = portal.call(anyio.Event)
+            fut, task_status = portal.start_task(self._lifespan, startup_event)
+
+            @stack.push
+            def wait_shutdown(*exc_info: object) -> bool:
+                portal.call(startup_event.set)
```
```suggestion
            shutdown_event = portal.call(anyio.Event)
            fut, task_status = portal.start_task(self._lifespan, shutdown_event)

            @stack.push
            def wait_shutdown(*exc_info: object) -> bool:
                portal.call(shutdown_event.set)
```

starlette/testclient.py
```diff
@@ -514,63 +518,59 @@ def __enter__(self) -> "TestClient":
...
                 self.portal = None
 
-            self.stream_send = StapledObjectStream(
-                *anyio.create_memory_object_stream(math.inf)
-            )
-            self.stream_receive = StapledObjectStream(
-                *anyio.create_memory_object_stream(math.inf)
+            shutdown_event = portal.call(anyio.Event)
+            fut, task_status = portal.start_task(self._lifespan, shutdown_event)
+
```
graingert, line 567: I'd like to do

```python
def done(*_):
    if fut.exception()
        os.kill(os.getpid(), signal.SIGUSR1)

fut.add_done_callback(done)

@stack.push
def wait_shutdown(*exc_info: object) -> bool:
    fut.remove_done_callback(done)
```

so lifespan failures interrupt the test suite

graingert: or maybe just closing the portal would do, then at least new requests would fail

starlette/testclient.py
```diff
@@ -514,63 +518,59 @@ def __enter__(self) -> "TestClient":
...
-
-        async def receive() -> typing.Any:
-            message = await self.stream_send.receive()
-            if message is None:
-                self.task.result()
-            return message
+    async def _lifespan(
+        self, shutdown_event: anyio.Event, *, task_status: anyio.abc.TaskStatus
+    ) -> None:
+        failures = []
```
graingert, line 587: I don't like this - I think this should be a MemoryObjectStream and a task consuming errors and raising them

I can't raise the assertion directly in `async def send(` because it might not be run in the current task

starlette/testclient.py
```diff
@@ -514,63 +518,59 @@ def __enter__(self) -> "TestClient":
...
+
+        async def send() -> typing.AsyncGenerator[None, typing.Any]:
+            task_status.started((yield))
+            await shutdown_event.wait()
```
graingert: I think this is the wrong way around. I'm working around a slight abuse of the asgi spec.

when you run:

```python
@contextlib.asynccontextmanager
async def lifespan(app):
    async with anyio.create_task_group() as app.tg:
        yield


app = Starlette(lifespan=lifespan)

async def raise_(e):
    raise e

class MyError(Exception): pass

@app.route("/fail")
def _(*_):
    app.tg.start_soon(raise_, MyError())
```

requesting "http://.../fail" will:
* raise an exception in the taskgroup
* which will cancel the lifespan task cancelling `await receive()`
* which will cause the MyError() to replace the cancelled error from the `async with anyio.create_task_group()` cmgr
* which will be sent as a "lifespan.shutdown.failed" - before the "lifespan.shutdown" event is received.

graingert:
```suggestion
            shutdown_event.set()
```

this way each subsequent request can check `shutdown_event.is_set()` before running

starlette/testclient.py
graingert: sort of like this?

```python
async def run(shutdown_event, app, scope, receive, send):
    return await race(
        functools.partial(app, scope, receive, send),
        shutdown_event.wait(),
    )   

portal.call(run, shutdown_event, self.app, scope, receive, send)
```

starlette/testclient.py
```diff
@@ -514,63 +518,59 @@ def __enter__(self) -> "TestClient":
...
+            while True:
+                failures.append((yield))
+
+        async def receive() -> typing.AsyncGenerator[typing.Any, None]:
+            yield {"type": "lifespan.startup"}
+            await shutdown_event.wait()
```
graingert, line 599:
```suggestion
            await self.portal.sleep_until_stopped()
```

starlette/testclient.py
graingert:
```suggestion
            await self.portal.stop()
```

graingert: https://github.com/agronholm/anyio/pull/334/files

tests/test_testclient.py
```diff
@@ -147,6 +148,72 @@ async def loop_id(request):
...
+            assert client.get("/throw").json() == "good"
+            with pytest.raises(CancelledError):
+                client.get("/sleep_forever")
+            with pytest.raises(ClosedError):
+                client.get("/sleep_forever")
+        except BaseException as e:  # pragma: no cover
+            try:
+                client.__exit__(*sys.exc_info())
+            except BaseException as f:
+                raise ExceptionGroup((e, f)) from None
```
graingert, line 208: I want to move this ExceptionGroup into testclient but constructing it isn't public

tests/test_testclient.py
```diff
@@ -147,6 +148,72 @@ async def loop_id(request):
...
+    for _ in range(2):  # context manager can be re-used
+        try:
+            client.__enter__()
+            assert client.get("/ping").json() == "pong"
+            assert client.get("/throw").json() == "good"
+            with pytest.raises(CancelledError):
+                client.get("/sleep_forever")
+            with pytest.raises(ClosedError):
+                client.get("/sleep_forever")
+        except BaseException as e:  # pragma: no cover
```
graingert, line 204: this happens if there's an AssertionError and a MyError simultaneously. If you use `with client:` then the AssertionError gets swallowed and replaced with MyError

starlette/testclient.py
graingert: but I like having no concurrency in lifespan itself

starlette/testclient.py
```diff
@@ -514,63 +563,58 @@ def __enter__(self) -> "TestClient":
...
-        finally:
-            await self.stream_send.send(None)
-
-    async def wait_startup(self) -> None:
-        await self.stream_receive.send({"type": "lifespan.startup"})
-
-        async def receive() -> typing.Any:
-            message = await self.stream_send.receive()
-            if message is None:
-                self.task.result()
```
graingert, line 550: this is a blocking wait on its own task. it's very likely to hang

starlette/testclient.py
```diff
@@ -514,63 +563,58 @@ def __enter__(self) -> "TestClient":
...
-                *anyio.create_memory_object_stream(math.inf)
-            )
-            self.task = portal.start_task_soon(self.lifespan)
-            portal.call(self.wait_startup)
+            fut, (task_status, shutdown_event) = portal.start_task(self._lifespan)
 
             @stack.callback
-            def wait_shutdown() -> None:
-                portal.call(self.wait_shutdown)
+            def wait_shutdown() -> bool:
```
graingert, line 569: what should we do here if an exception is raised from this block?

I previously had this cancel the `fut.cancel()` but then that means I can't retrieve any failures from the task.

graingert: ah we can avoid the fut interface provided by anyio and cancel the task ourselves
```python
async def task(*, task_status):
    with anyio.CancelScope() as scope:
        task_status.started(scope)
        try:
            await anyio.sleep_forever()
        except anyio.get_cancelled_exc_type():
            raise MyError

fut, scope = portal.start_task(task)
time.sleep(0.1)
portal.call(scope.cancel)
assert isinstance(fut.exception(), MyError)
```

uSpike:
Could you please edit the title/description to specify that this only affects TestClient?  A bit more context in the description as to what you're trying to achieve would help as well.

graingert:
> Could you please edit the title/description to specify that this only affects TestClient? A bit more context in the description as to what you're trying to achieve would help as well.

I'm trying to make it so that if an exception is thrown into a taskgroup spanning the lifespan_context the exception is thrown out of the TestClient context as quickly as possible.

It does this by cancelling any concurrent requests of the TestClient and preventing any new requests until the TestClient is closed

Kludex:
@graingert Are you still interested in this, or can I close it?

We still don't know what we are going to do with the `TestClient` on Starlette (if it's going to be deprecated/removed or switch to `httpx`).

Kludex:
@graingert Closing this as stale.

You're a member, so if you think we should reopen, go ahead. 😗

graingert:
Oh yeah this needs doing lol

graingert:
Wait no. Doing everything via httpx is obviously better

PR #1236, opened 2021-07-04
utf-8 instead of latin-1 encoding

TomaszSzyborski:
Using utf-8 instead of latin-1 encoding as other than latin-based languages e.g. Polish using in various places e.g. cookies - produced 500 Internal Server Error. 
Chaning encoding fixes this issue.

### DIFF
--- a/tests/test_staticfiles.py
+++ b/tests/test_staticfiles.py
@@ -63,7 +63,8 @@ def test_staticfiles_with_package(test_client_factory):
     client = test_client_factory(app)
     response = client.get("/example.txt")
     assert response.status_code == 200
-    assert response.text == "123\n"
+    assert "123" in response.text
+    assert response.text.endswith("\n")
 
 
 def test_staticfiles_post(tmpdir, test_client_factory):
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,8 @@ types-PyYAML
 types-dataclasses
 pytest
 trio
+anyio
+graphene
 
 # Documentation
 mkdocs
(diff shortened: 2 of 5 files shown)


tests/test_staticfiles.py
```diff
@@ -63,7 +63,8 @@ def test_staticfiles_with_package(test_client_factory):
     client = test_client_factory(app)
     response = client.get("/example.txt")
     assert response.status_code == 200
-    assert response.text == "123\n"
+    assert "123" in response.text
+    assert response.text.endswith("\n")
```
graingert, line 67:
```suggestion
    assert response.text == "123\n"
```

JayH5:
I think this change would be better as a discussion topic, as the pull request template encourages you to do.

You also need to be specific about what (cookies/headers/query strings) you want to change the encoding for, and you'll need to back up your claim with some kind of HTTP standard. For example, RFC 7230 says that headers should be ISO-8859-1 (i.e. `latin1`), hence why Starlette works that way.

It is important that Starlette supports languages such as Polish, but please start this out with a discussion topic where you point out where you're facing incompatibility with character sets and we can work from there. I'm going to close this PR because this does not follow our contributing guidelines.

graingert:
Some related URLs

https://dzone.com/articles/utf-8-in-http-headers
https://github.com/encode/starlette/pull/1163
https://bugzilla.mozilla.org/show_bug.cgi?id=601933#c2

graingert:
here's a discussion: https://github.com/encode/starlette/discussions/1238

PR #1241, opened 2021-07-11
Add `routes=routes` to schemas documentation

soasme:
I was following the guide but found it was not working unless I add `routes=routes` to the app instance.


Kludex (approved):
Nice catch! Thanks @soasme ! :tada:

Issue #1116, opened 2020-12-24
Error are not reported using CORS middleware


Faylixe:
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

When using CORS middleware, if calling endpoint lead to any unexpected error, the response sent would not hold proper headers leading to the classic `TypeError: Origin XXX is not allowed by Access-Control-Allow-Origin.` error message from browser. 

### To reproduce

```python
from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware

api = FastAPI()
api.add_middleware(
    CORSMiddleware,
    allow_origins=['*'],
    allow_credentials=True,
    allow_methods=['*'],
    allow_headers=['*'])


@api.get('/')
def index():
    return """
    <html>
        <body>
              <div id="container"></div>
             <script>
                fetch('/error').catch(function(reason) { document.getElementById('container').innerHTML = reason }); 
            </script>
        </body>
    </html>
    """


@api.get('/error')
def errorcase():
    raise ValueError
```

```html

```
### Expected behavior

This is misleading as we expect to received a proper HTTP 500 Internal Server Error response at least.

### Actual behavior

The returned response does not hold proper CORS Header leading to wrong response in browser.

### Environment

- Python version:  3.7 (latest)
- Starlette version: latest

Kludex:
You said that you're in the latest Starlette... But FastAPI is not up-to-date with Starlette.

It works as supposed with `starlette==0.13.6`.

Faylixe:
> You said that you're in the latest Starlette... But FastAPI is not up-to-date with Starlette.
> 
> It works as supposed with `starlette==0.13.6`.

Well I still have the issue with my current stack :

```bash
$ pip freeze
fastapi==0.61.2
h11==0.11.0
h2==4.0.0
hpack==4.0.0
Hypercorn==0.11.1
hyperframe==6.0.0
priority==1.3.0
prometheus-client==0.7.1
pydantic==1.7.2
starlette==0.13.6
starlette-prometheus==0.7.0
toml==0.10.2
typing-extensions==3.7.4.3
uvloop==0.14.0
wsproto==1.0.0
```

Starlette version is `0.13.6` as suggested. If the bug still appear is there any chance it would be related to the process supervisor used (Hypercorn 0.11.1) ?

sjpeterson:
[...]

PR #1243, opened 2021-07-17
Apply CORS response headers to unhandled exception responses

jcwilson:
*This is pretty awful code, but it demonstrates a possible solution for the CORS response header problem (#1116)*

I don't like how we have to make `starlette.errors.ServerErrorMiddleware` aware of `CORSException` but it does provide a convenient way to convey the CORS response headers to the outer layer of the middleware onion.


jcwilson:
Whoops! Didn't mean to make this against the `encode/starlette` repo yet.

PR #1244, opened 2021-07-17
Check if default value is callabe and perform call resault if it is

snosratiershad:
For example:
```python
import secrets

config = Config(".env")
SECRET_KEY_CHARS = "abcdefghijklmnopqrstuvwxyz0123456789!@#$%^&*(-_=+)"


def get_random_secret_key(length=50, allowed_chars=SECRET_KEY_CHARS):
    """
    Return a 50 character random string usable as a SECRET_KEY setting value.
    """
    return ''.join(secrets.choice(allowed_chars) for i in range(length))

SECRET_KEY: str = config("SECRET_KEY", cast=str, default=get_random_secret_key)
# And not SECRET_KEY: str = config("SECRET_KEY", cast=str, default=get_random_secret_key())
```


lovelydinosaur:
Let's take a pass on this and keep it simple.

If you want the value of the callable as a default, then call that *first* and assign it to a constant, that you use as the default.

The issue with this kind of additions is that they *always* end up pushing further and further on. We get folks expecting to be able to use async callables, or using a construct that can be called but doesn't necessarily return `True` when inspected with `callable(value)`. Etc. etc.

Simple and constrained is okay.

PR #1245, opened 2021-07-17
Added server timing

eznix86:
Enables the browser to get server timing in debug mode:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing


starlette/middleware/servertiming.py
```diff
@@ -0,0 +1,106 @@
...
+    @property
+    def duration(self):
+        if self._start:
+            return int(((self._end or time.time()) - self._start) * 1000)
+        else:
+            return 0
+
+
+def get_tickers() -> typing.List[Ticker]:
+    return _thread_local.__dict__.setdefault("tickers", [])
```
graingert: It's idiomatic to subclass threading.local and use a ClassVar to set the default

eznix86: Thanks for the review, corrected.

eznix86:
Also I noticed, that functions are executed in a worker, is it only during tests ? @graingert

starlette/middleware/servertiming.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+
+    def __init__(self, app: ASGIApp, allow_origins: typing.Sequence[str] = ()) -> None:
+        self._allow_origins = allow_origins
+        super().__init__(app, dispatch=self.dispatch)
+
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint):
+
+        defaultTimer = Ticker()
+        defaultTimer.start()
```
graingert, line 86: I'm not sure why you're using a threadlocal here, can't you store the time as a float here?

eznix86: Actually, what I'm trying to achieve is every function which have a ticker, will append the threadlocal, so we have a list of tickers, which I can send via server-timing header.

Ex. I have the endpoint /app which contains a query to database, a serialization.

the total time to complete /app is lets say 3 seconds, can be split into a more detailed list.
One can decorate the query to database and put a ticker around a serialization method, thus we will end up with a detailed view:
app - 3 sec
query - 2 sec
serialization - 1 sec

Where I'm stuck is the functions inside the endpoint are executed asynchronously, is there a way to get share the elapsed time of each tickers ?

graingert:
I think this would be better as a discussion detailing your precise usecase so we can suggest solutions. I think a threadlocal isn't going to work here, you may need to introspect your TaskGroup tree for example

eznix86:
> I think this would be better as a discussion detailing your precise usecase so we can suggest solutions. I think a threadlocal isn't going to work here, you may need to introspect your TaskGroup tree for example

Which platform is available to discuss on this matter and yes, I noticed it (a bit late) that threadlocal isn't the good fit for it

graingert:
There's https://github.com/encode/starlette/discussions
And https://gitter.im/encode/

starlette/middleware/servertiming.py
```diff
@@ -0,0 +1,121 @@
...
+        self.name = name
+        self.description = description
+        self._start = self._end = None
+
+    def start(self):
+        self._start = time.time()
+        _add_ticker(self)
+
+    def end(self):
+        self._end = time.time()
```
polys, line 21: Have you considered using `time.monotonic()` instead of `time.time()`?
It still returns fractional seconds but it's slightly safer.

lovelydinosaur:
Hiya @theArtechnology - thanks for getting stuck in.

We're not adding new middleware to the Starlette core project at the moment. One of the project's strengths is that new middleware and other components can generally be in the form of third party packages, or code examples / gists, rather than having to be part of the core package itself.

Your `ServerTiming` middleware class would be a great candidate [for an independent package](https://www.starlette.io/third-party-packages/).

Things we need to figure out how to do better here:

* Guiding new users towards discussion-first workflows, so that folks aren't working on pull requests that then get closed off.
* Doing a good job of promoting third party packages, and communicating the "if Starlette allows something to be done without it being in core, then it probably shouldn't be in core" message.

Issue #1250, opened 2021-07-22
Could  JSONResponse  provide  JsonEncoder class option on    JSONResponse.__init__ () ?


lishukan:
Hi,dear  commiter:
     When I use the class  -- <JsonResponse>  to serialize(jsonify) a response, I find that I can not assign the special JsonEncoder .  Sometime i have to rewrite the JsonResponse class  ,like  following code
`class JSONResponse(Response):
    media_type = "application/json"

    def __init__(self,*args,**kwargs):
        self.encoder = kwargs.get('encoder',None)
        super().__init__(*args)

    def render(self, content: typing.Any) -> bytes:
        return json.dumps(
            content,
            ensure_ascii=False,
            allow_nan=False,
            indent=None,
            separators=(",", ":"),
            cls=self.encoder,
        ).encode("utf-8")`

 i think it's  necessary that provide custom jsonencoder function.

Kludex:
It feels weird to add the encoder on the `init`, because you usually want to reuse the Response class, and with this you'll have to use the `encoder=Encoder` on every endpoint. :thinking: 

I'm going to take the lead here, and close this issue. Below is the current alternative for the proposed snippet:

```python
class CustomJSONResponse(Response):
    media_type = "application/json"
    encoder = Encoder

    def render(self, content: typing.Any) -> bytes:
        return json.dumps(
            content,
            ensure_ascii=False,
            allow_nan=False,
            indent=None,
            separators=(",", ":"),
            cls=self.encoder,
        ).encode("utf-8")
```

In case someone disagrees with my statement, feel free to start a discussion on our Gitter or reopen the issue.

Kludex:
Related #715

PR #1251, opened 2021-07-26
Add a encoder argument for the function __init__ of class - JsonRespo…

lishukan:
…nse (#1250)

This argument have default value :None. And this argument is  related to argument 'cls' of function - json.dumps(...cls=)


lishukan:
When I use the class -- to serialize(jsonify) a response, I find that I can not assign the special JsonEncoder . Sometime i have to rewrite the JsonResponse class 。

lishukan:
sometime, we  have to rewrite JsonResponse class .for example , when you want to  use json to serialize  some special object ,like   datetime.datetime or Decimal  ....

lishukan:
so I write  provide this pull request   for  serializing those special object .  you can  code your  encoder  class ,and add it to 
 args of   JsonResponse.__init__ .

Kludex:
Explanation on why it's closed on https://github.com/encode/starlette/issues/1250#issuecomment-993882247

@lishukan Thanks for the PR.

Issue #886, opened 2020-03-31
objects defining async __call__ not run/awaited when used in on_startup


cjw296:
Simple reproducer:

```
def test_async_startup_hangs():

    class Foo:
        async def __call__(self):
            raise Exception()

    app = Starlette(on_startup=[Foo()])

    with TestClient(app) as client:
        pass
```

No exception is raised when it should be. This, however, raises an exception as expected:

```
def test_async_startup_hangs():

    async def foo():
        raise Exception()

    app = Starlette(on_startup=[foo])

    with TestClient(app) as client:
        pass
```

euri10:
not sure about that but seems like on_startup expects a list of callables,
`    app = Starlette(on_startup=[Foo().__call__])` will trigger your exception

cjw296:
An instance of `Foo()` is callable:

```
$ python
Python 3.7.2 (default, Dec 29 2018, 00:00:04) 
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> class Foo():
...     async def __call__(self):
...         pass
... 
>>> obj = Foo()
>>> callable(obj)
True
```

Explicitly specifying `__call__` as you have feels like a workaround for this bug rather than a solution.

florimondmanca:
@cjw296 An immediate workaround: wrap your callable object into a function?

```python
class Foo:
    async def __call__(self):
        raise Exception()

foo = Foo()

async def on_startup():
    await foo()

 app = Starlette(on_startup=[on_startup])
with TestClient(app) as client:  # Raises `Exception`.
        pass
```

The long-term solution would be to fix the detection of async functions in the lifespan logic to something like:

```python
# In an utils module
def iscoroutinefunction(obj):
    if inspect.iscoroutinefunction(obj):
        return True
    if hasattr(obj, '__call__') and inspect.iscoroutinefunction(obj.__call__):
        return True
    return False

# routing.py
if iscoroutinefunction(func):
    await func()
else:
    ... # Sync, defer to threadpool.
```
 Not sure where that's in the codebase exactly, but it should do the trick.

cjw296:
Yeah, agreed on both fronts, but this almost feels like a deficiency in python itself; it feels like there should be an `inspect.returnscoroutinewhencalled` to catch these cases, rather than assuming everything is a function or a method.

florimondmanca:
[...]

PR #1256, opened 2021-07-29
Fixed - async __call__ not run/awaited in the BackgroundTask

dpr-0:
Reference to #886


Kludex (commented):
On `starlette/endpoints.py` we also use the same `asyncio.iscoroutinefunction`. We probably want to use the same `iscoroutinefunction` there as well.

Maybe we should move the function to other module?

dpr-0:
Yeah, so should I work on this branch or checkout to a new branch for this `iscoroutinefunction` function?

Kludex:
It can be here.

dpr-0:
> It can be here.

I see. Thanks for reviewing this pull request. Considering current module names might be not completely suitable for this function, it seems like there can be a new module like `utils.py` for this function.

Kludex:
I guess the function was moved, so we could also do the replacement on the `starlette/endpoints.py`?

dpr-0:
> I guess the function was moved, so we could also do the replacement on the `starlette/endpoints.py`?

Hi, I am back for this thread. Should I do the replacement now?

Kludex:
I'm not sure. I think we should look for #1444, but I don't know the opinion of the other members.

dpr-0:
> I'm not sure. I think we should look for #1444, but I don't know the opinion of the other members.

Yeah, I understand that, please let me know if there is any update about this thread. Thanks.
Maybe there is other tickets I can be assigned right now.

Kludex:
@dpr-0 Thanks for this PR! Also, sorry for taking long to take action here.

I'm going to push forward to #1444, so I'll be closing this.

dpr-0:
@Kludex No problem, Thanks for fixing this issue at #1444 !

Kludex:
Thank you for the patience here! 🙇‍♂️

PR #1257, opened 2021-08-02
Try to propagate contexvars from lifespan to endpoints

adriangb:
The goal is that, from a users perspective, it endpoints should be run as if they were a task being executed from within the `yield of a lifespan.

### DIFF
--- a/starlette/context.py
+++ b/starlette/context.py
@@ -1,0 +1,54 @@
+from typing import Any, Optional
+
+try:
+    from contextvars import Context, ContextVar, copy_context
+except ImportError:  # pragma: no cover
+    Context = object  # type: ignore
+    ContextVar = copy_context = None  # type: ignore
+
+
+def _set_cvar(cvar: ContextVar, val: Any) -> None:
+    cvar.set(val)
+
+
+class CaptureContext:
+    def __init__(self) -> None:
+        if Context is not None:
+            self.context = Context()
+        else:
+            self.context = None
+
+    def __enter__(self) -> "CaptureContext":
+        if copy_context is not None:
+            self._outer = copy_context()
+        else:   # pragma: no cover
+            self._outer = None
+        return self
+
+    def sync(self) -> None:
+        if self._outer is None:  # pragma: no cover
+            return
+        final = copy_context()
+        for cvar in final:
+            if cvar not in self._outer:
+                # new contextvar set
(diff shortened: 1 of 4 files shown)


adriangb:
@graingert this is in reference to https://github.com/adriangb/anydep/issues/1#issuecomment-890117334

~Currently this is not working; I feel like I'm missing some fundamental understanding of ASGI servers and/or context var propagation. If someone with more knowledge in this area has any ideas that'd be much appreciated.~

Working now

starlette/routing.py
```diff
@@ -647,6 +651,19 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
 
         partial = None
 
+        if self.user_ctx is not None:
+            current_ctx = contextvars.copy_context()
+            for cvar in self.user_ctx:
+                if cvar in current_ctx:
+                    # not set by the user, skip to avoid modifying
+                    # vars used by event loops, servers, etc.
+                    continue
```
adriangb: @graingert note that without this check, I was running into issues with Trio (not asyncio) here: https://github.com/python-trio/trio/blob/2865222187a6b51f7a1df54a8c4445574c86183d/trio/_core/_run.py#L2296

It seems that copying back the context made Trio think that it was in the lifespan task permanently or something.

starlette/routing.py
```diff
@@ -647,6 +651,19 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+                try:
+                    if cvar.get() != self.user_ctx.get(cvar):
+                        cvar.set(self.user_ctx.get(cvar))
+                except LookupError:
+                    cvar.set(self.user_ctx.get(cvar))
```
adriangb: I'm pretty sure this sort of copying back the context can be used to allow things run by `run_in_threadpool` to behave like if they were being called directly (which would solve issues FastAPI has with sync dependencies)

starlette/context.py
```diff
@@ -0,0 +1,54 @@
+from typing import Any, Optional
+
+try:
+    from contextvars import Context, ContextVar, copy_context
+except ImportError:  # pragma: no cover
+    Context = object  # type: ignore
+    ContextVar = copy_context = None  # type: ignore
```
adriangb, line 7: This module can be cleaned up better. I was trying to encapsulate all of the `try: import contexvars` checks into one module

PR #1264, opened 2021-08-12
Fix BadSignature exception handling in SessionMiddleware

hanneskuettner:
The `SessionMiddleware` does not catch all possible exceptions of `TimestampSigner.unsign`, specifically there is an edge case where a `BadSignature` exception can be raised by the method if a totally bogus session cookie is provided that does not match the expected format.

This PR broadens the caught exceptions (since all the other exceptions inherit `BadSignature`) and fixes the issue.
The issue can be reproduced by running the new test without the fix applied.

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -3,7 +3,7 @@ import typing
 from base64 import b64decode, b64encode
 
 import itsdangerous
-from itsdangerous.exc import BadTimeSignature, SignatureExpired
+from itsdangerous.exc import BadSignature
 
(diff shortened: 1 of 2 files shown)


JayH5 (approved):
Seems reasonable, thanks 👍

gnat:
Great patch, thanks. Eagerly awaiting the release.

gnat:
Just wanted to say I've tested this in my project and it works well. No more non-recoverable Internal Server Errors when passing in malformed session cookies- just an elegant blank session.

Glad to see people stress testing exploitation vectors on Starlette to make it more robust!

PR #1266, opened 2021-08-14
Add content disposition type parameter to FileResponse

qu1ck:
disposition "attachment" causes browsers to download
the file. E.g. "inline" will will be attempted to be
displayed directly.


qu1ck:
Friendly ping.
I have a larger PR pending for support of 206 partial content responses that I want to send after this is merged.

starlette/responses.py
```diff
@@ -243,6 +243,7 @@ def __init__(
         filename: str = None,
         stat_result: os.stat_result = None,
         method: str = None,
+        content_disposition_type: str = "attachment",
```
Kludex, line 277: Only "inline" and "attachment" are allowed here, right?

Kludex (commented):
Documentation needs to be updated on this PR as well.

Kludex:
Flask 2.0 adds this possibility as well.

Reference: https://github.com/pallets/flask/blob/24aab7a08d0b3f40d9fc9e49706854ef8be7a05f/src/flask/helpers.py#L551-L554

starlette/responses.py
qu1ck: Correct

qu1ck:
I added the parameter description to the docs. There are a couple other parameters missing there but it's outside of the scope of the pull request, just FYI.

lovelydinosaur (approved):
I'm pretty okay with this yup.
It's a nice low-footprint change, so no big objections.

Issue #1099, opened 2020-11-23
Missing content-length header field in some responses


freol35241:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

The Content-Length header field is not set for cases where body is empty. In some cases this seems to not be according to the HTTP/1.1 spec and creates troubles when using for example traefik as a proxy in front of a starlette-based backend.

See this issue for details: https://github.com/traefik/traefik/issues/4456
Specifically, the final conclusion: https://github.com/traefik/traefik/issues/4456#issuecomment-593539169

### To reproduce

```python
>>> from starlette.responses import RedirectResponse
>>> RedirectResponse(url="/", status_code=302).headers
MutableHeaders({'location': '/'})
```
Header has not content-length field.

### Expected behavior

```python
>>> from starlette.responses import RedirectResponse
>>> RedirectResponse(url="/", status_code=302).headers
MutableHeaders({'content-length': '0', 'location': '/'})
```
I expect there to be a content-length field specifying zero length

### Actual behavior

```python
>>> from starlette.responses import RedirectResponse
>>> RedirectResponse(url="/", status_code=302).headers
MutableHeaders({'location': '/'})
```
Header has no content-length field.

### Debugging material

N/A

### Environment

- OS: Linux (WSL2)
- Python version: 3.8.2
- Starlette version: 0.13.6

### Additional context

This bug is probably also present for other type of responses (other status-codes). Last paragraph (in bold) from https://tools.ietf.org/html/rfc7230#section-3.3 is of interest:

> 3.3.  Message Body
> 
>    The message body (if any) of an HTTP message is used to carry the
>    payload body of that request or response.  The message body is
>    identical to the payload body unless a transfer coding has been
>    applied, as described in Section 3.3.1.
> 
>      message-body = *OCTET
> 
>    The rules for when a message body is allowed in a message differ for
>    requests and responses.
> 
>    The presence of a message body in a request is signaled by a
>    Content-Length or Transfer-Encoding header field.  Request message
>    framing is independent of method semantics, even if the method does
>    not define any use for a message body.
> 
>    The presence of a message body in a response depends on both the
>    request method to which it is responding and the response status code
>    (Section 3.1.2).  Responses to the HEAD request method (Section 4.3.2
>    of [RFC7231]) never include a message body because the associated
>    response header fields (e.g., Transfer-Encoding, Content-Length,
[...]

PR #1270, opened 2021-08-18
Implement EmptyResponse class

### LINKED ISSUE #1178: Add EmptyResponse class
HTTP responses with status codes 1xx, 204, 205, and 304 are special cases.  1xx, 204, 304 responses must not include a message body.  The empty message body for 205 responses can be handled in a few ways, but the only one that would make sense for Starlette would be to add a Content-Length header with value 0.

Currently it is not at all obvious to users how to implement these responses.  Starlette can make it explicit with the addition of an EmptyResponse class that does the right thing for each of the status codes cited above.

The "right thing" differs by status code. The key difference is 

* 1xx, 204: must not have a Content-Length header
* 205: should have Content-Length: 0
* 304: may have a Content-Length header if the value equals the length of the corresponding 200 response that could have been sent.

Thus, an EmptyResponse object should ensure no Content-Length header in the first case, ensure Content-Length: 0 in the second case, and pass along a Content-Length header if supplied in the third case; here it is the caller's responsibility to get the value right.

In all cases, an EmptyResponse object would send an 'http.response.body' event with body: b'', more_body: False so that the ASGI server knows to begin sending the response.

The addition of an EmptyResponse class should resolve at least one open issue (https://github.com/encode/starlette/issues/1099).

declaresub:
Closes #1178, #1099

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -188,7 +188,7 @@ Supplies an empty message body as the response.
 
 In particular, use an EmptyResponse object for 1xx, 204, 205, 304 responses as it sets or omits a Content-Length header as appropriate.
 
-Signature: `Response(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
+Signature: `EmptyResponse(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
 
 * `status_code` - An integer HTTP status code.
 * `headers` - A dictionary of strings.
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -182,29 +182,6 @@ async def app(scope, receive, send):
     await response(scope, receive, send)
 ```
 
-### EmptyResponse
-
-Supplies an empty message body as the response.  
-
-In particular, use an EmptyResponse object for 1xx, 204, 205, 304 responses as it sets or omits a Content-Length header as appropriate.
-
-Signature: `EmptyResponse(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
-
-* `status_code` - An integer HTTP status code.
-* `headers` - A dictionary of strings.
-* `background` - A BackgroundTask to be executed when the response is sent.
-
-
-```python
-from starlette.responses import EmptyResponse
-
-
-async def app(scope, receive, send):
-    assert scope['type'] == 'http'
-    response = EmptyResponse(status_code=204)
-    await response(scope, receive, send)
-```
-
 ## Third party middleware
 
 ### [SSEResponse(EventSourceResponse)](https://github.com/sysid/sse-starlette)
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -182,6 +182,29 @@ async def app(scope, receive, send):
     await response(scope, receive, send)
 ```
 
+### EmptyResponse
+
+Supplies an empty message body as the response.  
+
+In particular, use an EmptyResponse object for 1xx, 204, 205, 304 responses as it sets or omits a Content-Length header as appropriate.
+
+Signature: `Response(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
+
+* `status_code` - An integer HTTP status code.
+* `headers` - A dictionary of strings.
+* `background` - A BackgroundTask to be executed when the response is sent.
+
+
+```python
+from starlette.responses import EmptyResponse
+
+
+async def app(scope, receive, send):
+    assert scope['type'] == 'http'
+    response = EmptyResponse(status_code=204)
+    await response(scope, receive, send)
+```
+
 ## Third party middleware
 
 ### [SSEResponse(EventSourceResponse)](https://github.com/sysid/sse-starlette)
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -133,38 +133,48 @@ class Response:
             # Response is terminated after the status line.  So no headers and no body.
             # https://datatracker.ietf.org/doc/html/rfc7231#section-6.2
             raw_headers = []
-            body = b''
+            body = b""
         elif self.status_code == 204:
             # Response must not have a content-length header. See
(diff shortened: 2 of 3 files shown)


docs/responses.md
```diff
@@ -182,6 +182,29 @@ async def app(scope, receive, send):
     await response(scope, receive, send)
 ```
 
+### EmptyResponse
+
+Supplies an empty message body as the response.  
+
+In particular, use an EmptyResponse object for 1xx, 204, 205, 304 responses as it sets or omits a Content-Length header as appropriate.
+
+Signature: `Response(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
```
JayH5:
```suggestion
Signature: `EmptyResponse(status_code:int, headers: typing.Optional[typing.Dict[str, str]] = None, background: typing.Optional[BackgroundTask] = None)`
```

Alternatively, just take this line out. We don't have proper, generated API docs so trying to be consistent is difficult.

declaresub: I fixed the class name.  Following the superclass documentation seemed like the right thing to do.

lovelydinosaur:
I'm not convinced that we'd need a separate response class to deal with this - although maybe I'v not looked into it enough.
Don't recall Django, Flask, and others taking this approach. Can't we instead make sure that `Response` either includes or doesn't include the appropriate headers depending on the status code?

declaresub:
You certainly could take the approach of adding more special-case code to Response.  Since the status_code attribute can be set after initialization, I believe this approach would best be done in Response.__call__.  There you would check the value of self.status_code,  modify headers as appropriate just before sending, and send an empty body for the status codes that require it.

I would be willing to do this.

Kludex:
> I'm not convinced that we'd need a separate response class to deal with this - although maybe I'v not looked into it enough. Don't recall Django, Flask, and others taking this approach. Can't we instead make sure that `Response` either includes or doesn't include the appropriate headers depending on the status code?

Flask: [They do the same as what this PR proposes.](https://github.com/pallets/werkzeug/blob/3b79230cfd99c8e64d14d4abd655bc1d39398a87/src/werkzeug/wrappers/response.py#L473-L564)

Django: 
- 304: https://github.com/django/django/blob/bf7afe9c4e21f5fe5090c47b2b6ffc5a03a85815/django/http/response.py#L534-L545
- I can't find a special treatment for the other status code on Django, and also, see this [comment](https://code.djangoproject.com/ticket/3362).

starlette/responses.py
```diff
@@ -129,14 +129,64 @@ def delete_cookie(self, key: str, path: str = "/", domain: str = None) -> None:
...
+            raw_headers = []
+            body = b""
+        elif self.status_code == 204:
+            # Response must not have a content-length header. See
+            # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
+            # http spec does not appear to say whether or not there can be
+            # a content-type header. Some clients
+            # will attempt to parse the message body if there is a content-type
+            # header, so we ensure that there isn't one in
+            # the response.
```
Kludex, line 144: Why? Is it obvious?

Shouldn't we focus on what is defined?

declaresub:
I have revised the pull request to move handling of responses with status code 1xx, 204, 205, 304 into the Response class.

This is probably the better approach, as users can just set the status code, and the right thing should happen. Also it fits better with fastapi, and its use of JSONResponse by default.

Kludex:
I'm extremely happy about the changes here. :)

Thank you @declaresub ! :tada: 

This is an important PR, so I'll feel more comfortable having @tomchristie 's approval.

lovelydinosaur:
This is not at all clear to me.
For instance, while it's *related* to #1099, it doesn't appear to close it.

Implementation-wise I'd expect any "do we generate a Content-Length header or not" to be [handled in the `init_headers()` method](https://github.com/encode/starlette/blob/9d686a7125131fe026071d04e0d0e0a726e1afc5/starlette/responses.py#L59-LL84), but let's not start from that point.

Can we take this back to basics first...

Issue #1099 describes things at the API level that I'd like to see:

---

### Expected behavior

```python
>>> from starlette.responses import RedirectResponse
>>> RedirectResponse(url="/", status_code=302).headers
MutableHeaders({'content-length': '0', 'location': '/'})
```
I expect there to be a content-length field specifying zero length

### Actual behavior

```python
>>> from starlette.responses import RedirectResponse
>>> RedirectResponse(url="/", status_code=302).headers
MutableHeaders({'location': '/'})
```
Header has no content-length field.

---

The issue here isn't that we ought to be stripping the Content-Length header in several cases, but rather than we're *not* adding a correct `Content-Length: 0` header in cases where there is no body.

Which is what #1099 describes...

> The Content-Length header field is not set for cases where body is empty. In some cases this seems to not be according to the HTTP/1.1 spec and creates troubles when using for example traefik as a proxy in front of a starlette-based backend.

The sensible tack here would be to deal with *that* first.

I'd expect a pull request there to have a unit test at the API level of eg...

```python
def test_empty_response():
    response = Response()
    assert response.headers["Content-Length"] == "0"
```

And that the implementation to resolve this would be in `init_headers()`.

Once we have that PR we'd then want to take a look at refining it to deal with the `1xx`, `204`, and `304` cases.

declaresub:
I agree that #1099 is not closed here, but I don't think that fixing it in init_headers is the best approach.  Indeed, were I to do this, I would start by getting rid of that method.  Then I would move the invocation of render from __init__ into __call__, and set the content-* headers at that time.

declaresub:
And, at this point, I wonder whether the right thing to do would be to create a new issue that describes the need to set body and content-length appropriately for all combinations of request method and status code, then close the existing issues #1270, #1099, and maybe corresponding pull requests, with a reference to this new issue that describes the goal.

lovelydinosaur:
@declaresub - I'd like to close this PR and #1178 - an `EmptyResponse` class isn't the API we're looking for here.

#1099 is a valid issue, and properly described, so let's leave it open.

#1395 is the start of a PR addressing that issue.

lovelydinosaur:
Thanks for having spent some time looking into this, and for prompting the issue to get addressed.
An `EmptyResponse` class isn't the API I'd want to see here (also eg. not what we see in flask, django).

Closing in favour of #1397.

PR #1272, opened 2021-08-22
Add indent option to JSONResponse

Nebula1213:
New JSONResponse class:

https://pastecord.com/apemokopur.py

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -160,9 +160,15 @@ class JSONResponse(Response):
         headers: dict = None, 
         media_type: str = None, 
         background: BackgroundTask = None, 
-        indent: int = None) -> None:
+        indent: int = None
+    ) -> None:
         self.indent = indent
-        super().__init__(content=content, status_code=status_code, headers=headers, media_type=media_type, background=background)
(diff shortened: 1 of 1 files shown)


starlette/responses.py
```diff
@@ -153,12 +153,23 @@ class PlainTextResponse(Response):
...
     media_type = "application/json"
 
+    def __init__(
+        self, 
+        content: typing.Any = None, 
+        status_code: int = 200, 
+        headers: dict = None, 
+        media_type: str = None, 
+        background: BackgroundTask = None, 
+        indent: int = None) -> None:
```
aminalaee: I think as seen [here](https://github.com/encode/starlette/runs/3392828918) `black` will format this differently.
You can run `scripts/lint` locally.

Nebula1213: the newest commit should fix the formatting issue

Kludex:
What's the reasoning behind this? I don't see a good reason for this, could you enlighten me? :thinking:

Kludex:
As the author didn't reply in a long time, and there's no written reasoning behind this implementation, I'm closing this PR.

Issue #1275, opened 2021-08-28
Graphiql supports custom request headers


ssshier:
The graphiql version is too old and does not support request header settings, which is unfriendly to some authenticated interfaces.

Kludex:
Hey @ssshier, 

As you can see on https://www.starlette.io/graphql/, we've deprecated GraphQL. More details can be found [here](https://github.com/encode/starlette/issues/619).

That being said, I'm closing this issue.

PR #1277, opened 2021-08-28
Graphiql supports custom request headers (#1275)

ssshier:
1. update graphiql version.
2. add headers edit.


Kludex:
Thanks for the PR @ssshier ! 

As you can see on https://www.starlette.io/graphql/, we've deprecated GraphQL. More details can be found [here](https://github.com/encode/starlette/issues/619).

That being said, I'm closing this PR.

ssshier:
Sorry, this is my mistake. 

It's just that I encountered such a problem when using it. In this case, I will rewrite it in my own code.

Issue #685, opened 2019-10-22
Include routing information for middleware


cypai:
Is there a way to obtain which route the request is taking in middleware? Right now, the only available thing is `request.url.path`, which doesn't give me a nice name for metrics.

I see in https://github.com/steinnes/timing-asgi/blob/master/timing_asgi/integrations/starlette.py that it basically just loops through all the routes to obtain the matching one. Is there a better way to do it?

Looping in that manner also runs into an issue where if there are multiple matching routes, like
`/resource/default/1`
`/resource/1/1`
it can't find the right one since they both match.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/685">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/685/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/685/pledge.svg">
</picture>
</a>

bourdeau:
I got the same problem as @cypai with `AuthenticationMiddleware`. There is only `request.scope["path"]` and `request.scope["raw_path"]` available and I've to make quite some regex to handle the route properly.

Is it possible to build the Route() before *Middleware is called so one can get the route_name and other useful informations in the middleware? Maybe @cypai and I are both missing something?

lovelydinosaur:
So, the routing only occurs *after* the middleware handling.

I think options here are:

* We could potentially add some public API on the application so that it's easier to premptively resolve the request routing? Ideally we ensure that can be cached, so that the lookup only occurs once.
* We could add the ability to have some kinds of middleware run *after* the request routing. (Probably not super keen on that.)

bourdeau:
Thanks for your answer @tomchristie. I don't know Starlette enough to know which solution would be best. I know that in some other frameworks, there is usually some Events mechanisms with some Listeners attached to it.

If I can be of any help, let me know, we use Starlette in production in my company, so it's critical for us.

allan-simon:
@tomchristie  why the second option not something you're keen on ? (honest question, I have no opinion) 
but yes I've run into this issue because I was trying to put in a ContextVars the route name, so that my logger can get this information

[...]

PR #1286, opened 2021-09-13
feat: Add route-level middleware

### LINKED ISSUE #1174: How to use different middleware for different routes/path
### First check

* [x] I used the GitHub search to find a similar issue and didn't find it.
* [x] I searched the FastAPI documentation, with the integrated search.
* [x] I already searched in Google "How to X in FastAPI" and didn't find any information.

### Description

Is it possible to use different middleware for different routes/path?

### Additional context

In my case my need comes from CORS. But, I am sure there is other cases than CORS requirements that someone would need different middlewares for different paths.

* I want `myapi.com/path1` to allow origins of calls from `myfrontend.com`
* I want `myapi.com/path2` to allow origins of calls from anywhere ('*') since it is a public facing api.

I checked if it was possible to add a middleware at a router level and did not find any documentation about it.

### Code example

```python
from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware

app = FastAPI()

app.add_middleware(
    CORSMiddleware,
    allow_origins=['myfrontend.com'],
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)

# I want /path1 to allow only calls from 'myfrontend.com'
@app.get("/path1")
async def path1():
    return {"status": "alive"}

# I want /path2 to be a public facing api that can be accessed from anywhere ['*']
@app.get("/path2")
async def path2():
    return {"status": "alive"}

```

adriangb:
In case you reconsider https://github.com/encode/starlette/issues/685#issuecomment-550240999 @tomchristie 

I think this would be a useful feature. I often see middleware parameters that are "a regex pattern to match routes" because the middleware has to determine what routes to interact with before routing is done. [Here](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/fc8a02bcf8cf9fe985e03002c299642ddd53590b/instrumentation/opentelemetry-instrumentation-starlette/src/opentelemetry/instrumentation/starlette/__init__.py#L55) and [here](https://github.com/dmontagu/fastapi-utils/blob/af95ff4a8195caaa9edaa3dbd5b6eeb09691d9c7/fastapi_utils/timing.py#L25) are two examples. This also closes https://github.com/tiangolo/fastapi/issues/1174

Aside from being a pain to build some complex regex, there are performance implications: imagine (in the absurd scenario) that you have 100 endpoints and 100 middleware, but each middleware applies only to 1 endpoint. Now you have 99 middleware executing and doing regex matching when they could have just been bypassed to begin with if they executed after routing.

That said, re-using the existing Middleware construct for this use has it's cons. For example, it is obvious that the middleware cannot modify the path. I can't think of any other major limitations, but I there may be others.


alex-oleshkevich:
A more common use case of that is to apply a specific middleware to /api, /admin, and / scopes. Admin may require session, API wants token authentication, and landing page needs none of them.

alex-oleshkevich:
May I also ask you to add middleware to Mount and Host classes? That would be awesome!

adriangb:
> May I also ask you to add middleware to Mount and Host classes? That would be awesome!

I think that'd be cool, and should be doable, but maybe let's leave it for another PR, depending on the outcome of this one? I'd like to get an initial review / buy in from the Starlette team before making any more changes

lorthirk:
Sorry to jump in, but there's any news about this? It would be useful in my projects

adriangb:
> Sorry to jump in, but there's any news about this? It would be useful in my projects

It is waiting for review, there is no ETA. But, if you really needed to, you could just wrap the class in your app to add the functionality given that .app is a public attribute

lovelydinosaur:
Okay, ya know what - it's a pretty nice simple implementation, and even though I'm not *super* keen on route-based middlewares (just because *constraints are not a bad thing*) I think we could go with this.

Here's what I'd like to see in order to merge this...

* Let's get the same thing added to `Mount` as well. Having middleware on mounts *kinda* makes more sense to middleware on routes to me.
* Let's get docs included on the PR as well.

After that then I reckon it's a thumbs up from me. Good stuff!

adriangb:
Hey @tomchristie, thank you for looking at this!

I added the same implementation into `Mount` and repeated the test for that as well.

What are your thoughts on the other route types? I reckon this can also be added to WebSocketRoute and Host, but maybe it's better to wait until someone actually requests that feature.

I added a blurb in the docs, it's relatively short but I think it gets the point across.

docs/middleware.md
```diff
@@ -41,6 +41,31 @@ application would look like this:
 * Routing
 * Endpoint
 
+Middleware can also be added at the route level, in which case it will be executed after routing ocurrs:
```
```suggestion
Middleware can also be added at the route level, in which case it will be executed after routing occurs:
```

adriangb:
@tomchristie looks like I was able to implement your requested changes and get all tests passing.

I just wanted to highlight this question before we move forward with this:

> What are your thoughts on the other route types? I reckon this can also be added to WebSocketRoute and Host, but maybe it's better to wait until someone actually requests that feature.

Also, like we were discussing w/ @Kludex on Gitter, I'm open to being added as a maintainer so I can merge PRs after approval (given that I have several open across Starlette and httpx).

alex-oleshkevich:
Assuming the following pseudo-code I think there is a possible issue:
```python

class SomeMiddleware:
    def __init__(self, app):
        self.app = app
    async def __call_(self, scope, receive, send):
         # acquire some resource here like a database connection
         # release resource

def view():
    raise Exception()

app.add_route(Route('/', view, middleware=[Middleware(SomeMiddleware))
```

It seems, that when an exception is raised the code after `self.app` won't be called. That may cause resource/memory leaks.

adriangb:
Isn't that just how middleware works though?

First let me make your example self contained:

```python
from starlette.types import ASGIApp, Scope, Receive, Send


calls: list[str] = []


class SomeMiddleware:
    def __init__(self, app: ASGIApp):
        self.app = app
        calls.append("before")
        calls.append("after")


async def view(request: Request) -> None:
    raise Exception


app = Starlette(
    routes=[
        Route("/", view, middleware=[Middleware(SomeMiddleware)]),
    ]
)

client.get("/")
assert calls == ["before"]
```

Like you say, the code after `await self.app(scope, receive, send)` is never called.

But that's also how it works for app-level middleware:

```python
from starlette.types import ASGIApp, Scope, Receive, Send


calls: list[str] = []


class SomeMiddleware:
    def __init__(self, app: ASGIApp):
        self.app = app
        calls.append("before")
        calls.append("after")


async def view(request: Request) -> None:
    raise Exception


app = Starlette(
    routes=[
        Route("/", view),
    ],
    middleware=[Middleware(SomeMiddleware)],
)

client.get("/")
assert calls == ["before"]
```

In either case, if you want to ensure resource cleanup the solution is a `try:/finally:` block or using resources that provide a context manager\*, neither of which have anything to do with this PR as far as I can tell:

```python
from starlette.types import ASGIApp, Scope, Receive, Send


calls: list[str] = []


class SomeMiddleware:
    def __init__(self, app: ASGIApp):
        self.app = app
        calls.append("before")
        try:
        finally:
            calls.append("after")
        # could also actually catch the error here and do
        # await Response(500)(scope, receive, send)


async def view(request: Request) -> None:
    raise Exception


app = Starlette(
    routes=[
        Route("/", view, middleware=[Middleware(SomeMiddleware)]),
    ]
)

client.get("/")
assert calls == ["before", "after"]
```

\* you can also have a catch-all middleware, like Starlette's existing [ServerErrorMiddleware](https://github.com/encode/starlette/blob/master/starlette/middleware/errors.py), but that doesn't solve resource cleanup, it just handles uncaught errors.

alex-oleshkevich:
Well, let's discuss the following case:
1. you have enabled `SessionMiddleware` in `Mount` for `/admin`
2. you have a custom error handler for `ValidationError` exception type which is set globally
3. you navigate a view that should validate the request input and raise `ValidationError` if it is invalid. On that error, the form errors should be written into the session and the user should be redirected to the previous form. Forms fields now should be rendered along with error messages. We take these messages from the session.

Here is the problem: `ValidationError` gets propagated right to the `ExceptionMiddleware` leaving `SessionMiddleware` unfinished. The error handler writes form errors into the session (which is in `scope` because per-route middleware has set it) but these changes are lost.

When you set middleware globally, Starlette adds a new `ExceptionMiddleware` in front of the router so all exceptions are first handled by it and then calls back parent middleware. So our custom error handlers do not interrupt the flow and all middleware have a clean exit.

A workaround here is to explicitly copy-paste exception middleware with defined error handlers to every route it needs, which is not the perfect solution.

Alternatively, we can somehow implicitly add `ExceptionMiddleware` to the middleware stack in Route/Mount objects and copy error handlers from the app, making our application behave in the same way all time. But this looks like a hack.

UPD
This case is not artificial. Each app has forms and validations. Having a global error handler that removes that boilerplate is a cool idea.

adriangb:
I see, yes if there is an error handler installed at the application level and the error raised is caught by the error handler (which is the first/last thing in the middleware stack) then the _user_ middleware at the application will never see the error, it will just get a `Response` w/ whatever status code was set by `ExceptionMiddleware`. But in the case of the route level middleware in this PR, that same middleware would see the error (and have to handle it).

I think every middleware should handle exceptions. If it has resources to cleanup and would end up in a bad state if `await self.app(...)` raises an exception, then it should have a `try:/finally:` block or use a context manager to ensure the resources are cleaned up. It should not rely on some other middleware to catch exceptions. Not sure if @tomchristie or someone more familiar w/ the general spirit of generic ASGI middleware can comment on this. Looking at the implementation of `SessionMiddleware` indeed it does look like `send_wrapper` would not be run if there is an exception raised, so maybe my understanding of how ASGI middleware should be designed is flawed.

That said, I do see how this could be confusing to users if the same middleware behaves differently at the application level and the route level. But adding exception handling into each route would (1) introduce a lot of overhead (well, at least it would introduce 1 middleware) and (2) greatly expand the scope of this PR, so I'm not sure that's the best solution.

In any case, thank you for thinking of and bringing up the issue.

alex-oleshkevich:
`url_path_for` raises `NotMatchFound` for mounted router.

Steps:
1. give a name to your routes
```python
mounted_middleware_router = Router([
    Mount(
        ...
        name='demo',
    )
])
```

Then assert URL generation in test

```python
assert router.url_path_for('demo') == ''
```

Expected that it generates a path to route, actual result is raised exception.

adriangb:
Would you mind posting fully reproducible examples, or even better, an actual test that fails? That would make it a lot easier for me to look into vs. having to fill in the blanks / fix typos and syntax errors. Thanks

adriangb:
Alright @alex-oleshkevich, eb7d41d should resolve the issue you point out

adriangb:
Merged in master and ran formatters. Everything should be passing again now.

adriangb:
@tomchristie pinging for another review whenever you have a chance

adriangb:
Updated to master and fixed the failing tests

starlette/routing.py
```diff
@@ -340,16 +348,22 @@ def __init__(
         self.path = path.rstrip("/")
         if app is not None:
             self.app: ASGIApp = app
+            routes = getattr(self.app, "routes", [])
```
lovelydinosaur: Is this variable used?

adriangb: Ah good catch, it is not being used. It's leftover from a previous implementation.

starlette/routing.py
```diff
@@ -342,14 +350,21 @@ def __init__(
...
+        if middleware is not None:
+            for cls, options in reversed(middleware):
+                self.app = cls(app=self.app, **options)
+
     @property
     def routes(self) -> typing.List[BaseRoute]:
-        return getattr(self.app, "routes", [])
+        # we dynamically grab the routes so that if this is a Starlette router
+        # it can have routes added to it after it is mounted
+        return getattr(self._user_app, "routes", [])
```
adriangb: @tomchristie how would you feel about just grabbing a reference to `.routes` in `__init__` instead of doing a getattr here? I suppose that would break if the mounted router replaces its `.routes` with a new list, but otherwise it should be fine and it would avoid having to add this `_user_app` attribute.

adriangb: I think I like it more. I went ahead and implemented it that way: https://github.com/encode/starlette/pull/1286/commits/e9ff90328df6285b2ebf77ca5abf2c053d42add6

If you prefer it as it was doing a `getattr` in `self.routes`, I'm happy to revert it

lovelydinosaur (approved):
All seems pretty reasonable yup. I'm personally still not *that* wild about route-level middleware, but I'm happy to defer this one to the rest of the team. 🤣

starlette/routing.py
```diff
@@ -340,16 +350,22 @@ def __init__(
         self.path = path.rstrip("/")
         if app is not None:
             self.app: ASGIApp = app
+            self._routes = getattr(app, "routes", [])
         else:
             self.app = Router(routes=routes)
+            self._routes = getattr(self.app, "routes", [])
```
Kludex:
```suggestion
        else:
            self.app = Router(routes=routes)
        self._routes = getattr(self.app, "routes", [])
```

Kludex (commented):
There were two possible solutions proposed by Tom on the issue that motivated this PR.

> - We could potentially add some public API on the application so that it's easier to premptively resolve the request routing? Ideally we ensure that can be cached, so that the lookup only occurs once.
> - We could add the ability to have some kinds of middleware run after the request routing. (Probably not super keen on that.)

I took my time to think about the first solution, and how that would look like... I think what we can do is to add a parameter on the `Middleware` class:

```python
app = Starlette(
    middleware=[
        Middleware(PotatoMiddleware, path_prefix="/potato")  # or `path_prefixes`
    ]
)
```

Considering the above, we stick with the middleware concept being only at the app level.

Thoughts? @alex-oleshkevich @adriangb

adriangb:
>> We could potentially add some public API on the application so that it's easier to preemptively resolve the request routing?

> I think what we can do is to add a parameter on the `Middleware` class:
> 
> ```python
> app = Starlette(
>     middleware=[
>         Middleware(PotatoMiddleware, path_prefix="/potato")  # or `path_prefixes`
>     ]
> )
> ```

How would that be implemented? Is `Middleware` doing something special with `path_prefix`, or is it passing it to `PotatoMiddleware` like it does with all other keyword arguments? What if `PotatoMiddleware` expects a `path_prefix` argument? And when does routing run? Apologies if I'm just missing something, but I don't see how this could work. It also seems like this would end up in much of the same place we are now: middleware that accepts some `path_prefix_regex` parameter with regexes to match against (essentially re-inventing routing inside middleware).

adriangb:
FYI for those subscribed to this issue, I opened #1464 which I think solves this problem more elegantly

Kludex:
@adriangb Can we close this PR, and focus on #1464?

adriangb:
Yes

PR #1288, opened 2021-09-22
Use zero copy to send file if can

abersheeran:
https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send


starlette/responses.py
```diff
@@ -296,6 +296,22 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+            try:
+                await send(
+                    {
+                        "type": "http.response.zerocopysend",
+                        "file": fd,
+                        "more_body": False,
+                    }
+                )
+            finally:
+                os.close(fd)
```
```suggestion
                await anyio.to_thread.run_sync(os.close, fd)
```

starlette/responses.py
```diff
@@ -296,6 +296,22 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         )
         if self.send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
+        elif "http.response.zerocopysend" in scope["extensions"]:  # pragma: no cover
+            if os.name == "nt":
+                open_mode = os.O_RDONLY | os.O_BINARY
+            else:
+                open_mode = os.O_RDONLY
+            fd = os.open(self.path, open_mode)
```
```suggestion
            fd = await anyio.to_thread.run_sync(os.open, self.path, open_mode)
```

Kludex (changes requested):

starlette/responses.py
```diff
@@ -313,6 +313,22 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         )
         if self.send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
+        elif "http.response.zerocopysend" in scope["extensions"]:  # pragma: no cover
```
Kludex, line 330: Why should we not test it?

synodriver:
By the way, should we use [range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) and [if-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range) header? I saw it in [someones toolkit](https://github.com/abersheeran/baize/blob/5e84d69c2389aa46c81e5f4e96de54db4f0f1d3b/baize/asgi.py#L630-L640)

starlette/responses.py
abersheeran: I'm not sure how to test it. It seems too simple to just test if the result it returns is a dictionary. Do you have any suggestions?

abersheeran:
> By the way, should we use [range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) and [if-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range) header? I saw it in [someones toolkit](https://github.com/abersheeran/baize/blob/5e84d69c2389aa46c81e5f4e96de54db4f0f1d3b/baize/asgi.py#L630-L640)

https://github.com/encode/starlette/pull/1013

synodriver:
> > By the way, should we use [range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) and [if-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range) header? I saw it in [someones toolkit](https://github.com/abersheeran/baize/blob/5e84d69c2389aa46c81e5f4e96de54db4f0f1d3b/baize/asgi.py#L630-L640)
> 
> #1013

Than you for pointing out. But is that means ```zerocopysend``` have to do something in order to implement that? Like adding ```offset``` and ```count``` params.

abersheeran:
In addition, `baize.asgi.FileResponse` can work with Starlette. Just try the following code

```python
# main.py
from baize.asgi import FileResponse


async def homepage(request):
    return FileResponse("./main.py")

routes = [
    Route("/", endpoint=homepage)
]

```

abersheeran:
> > > By the way, should we use [range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) and [if-range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range) header? I saw it in [someones toolkit](https://github.com/abersheeran/baize/blob/5e84d69c2389aa46c81e5f4e96de54db4f0f1d3b/baize/asgi.py#L630-L640)
> > 
> > 
> > #1013
> 
> Than you for pointing out. But is that means `zerocopysend` have to do something in order to implement that? Like adding `offset` and `count` params.

Yes. These parameters need to be specified.

synodriver:
> In addition, `baize.asgi.FileResponse` can work with Starlette. Just try the following code
> 
> ```python
> # main.py
> from starlette.applications import Starlette
> from starlette.routing import Route
> from baize.asgi import FileResponse
> 
> 
> async def homepage(request):
>     return FileResponse("./main.py")
> 
> routes = [
>     Route("/", endpoint=homepage)
> ]
> 
> app = Starlette(debug=True, routes=routes)
> ```

Great feature!

abersheeran:
> > In addition, `baize.asgi.FileResponse` can work with Starlette. Just try the following code
> > ```python
> > # main.py
> > from starlette.applications import Starlette
> > from starlette.routing import Route
> > from baize.asgi import FileResponse
> > 
> > 
> > async def homepage(request):
> >     return FileResponse("./main.py")
> > 
> > routes = [
> >     Route("/", endpoint=homepage)
> > ]
> > 
> > app = Starlette(debug=True, routes=routes)
> > ```
> 
> Great feature!

Simply because they support the ASGI3 protocol. I was fascinated by this design in Starlette, so I implemented a similar part for WSGI.

synodriver (approved):

starlette/responses.py
```diff
@@ -327,6 +327,22 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+            try:
+                await send(
+                    {
+                        "type": "http.response.zerocopysend",
+                        "file": fd,
+                        "more_body": False,
+                    }
+                )
+            finally:
+                await anyio.to_thread.run_sync(os.close, fd)
```
lovelydinosaur, line 345: I'm... not comfortable with this.
Not at all obvious to me how the open/zerocopesend/close would interact.

I've got a bit of a preference here that we close this off for now as a "feature pause" until we're sufficiently on top of all existing issues that we can *thoroughly* review this.

Issue #1058, opened 2020-09-15
PathConvertor doesn't match new line characters


and-semakin:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`. (Checked on `starlette==0.13.8`).
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Let's imagine that I pass some text data as a path parameter and this data can generally contain any Unicode character. It turned out that it's impossible to pass the data with newline characters (`\n`) inside. Starlette decodes the URL somewhere before it tries to match routes, so it doesn't help if you try to pass this character in encoded form like `%0A`. It either doesn't match the route or cuts the new line character off (but it breaks the application logic). Seems like the issue is in [the regex](https://github.com/encode/starlette/blob/41218ac546a3c7a0f1982f2677177ce9a0ab3616/starlette/convertors.py#L30) used by the `PathConvertor`:

```python
class PathConvertor(Convertor):
    regex = ".*"
```

According [to many sources](https://riptutorial.com/regex/example/32238/why-doesn-t-dot-----match-the-newline-character----n---) on the internet, `.` (the dot character) matches any character except for `\n`. Should we allow it to match literally any character?

Possible solutions:

* replace `.` with `\s\S` — it's a popular trick to match ANY character including new lines:

```python
class PathConvertor(Convertor):
    regex = "[\s\S]*"
```

* put the `re.DOTALL` flag somewhere, for example [here](https://github.com/encode/starlette/blob/41218ac546a3c7a0f1982f2677177ce9a0ab3616/starlette/routing.py#L123):

```python
return re.compile(path_regex, flags=re.DOTALL), path_format, param_convertors
```

Notice that in both cases newline characters will appear in the matched data (it changes the current behavior but I'm not sure how big the impact will be).

I understand that:
* passing newline characters through the path is some kind of edge case and not what people usually do;
* some weird solutions are required to support this edge case and the profit is disputable;
* it could probably be better to restrict this data in the application so that it only contains characters from some allowed subset of Unicode characters, but the API is already established and it is out of the scope of this issue.

Anyway, I think that you should know about the issue. Does it worth to fix it? What do you think?

### To reproduce

```python

def test_starlette_route_with_new_line_characters() -> None:
    from starlette.routing import Route
    from starlette.middleware.wsgi import WSGIMiddleware

    # the second argument is not important, it can be anything
    route = Route("/{path:path}", WSGIMiddleware(lambda x: None))

    # everything works as expected without newline characters
[...]

PR #1290, opened 2021-09-24
Fix routing for pct-encoded newline char in path params

Pentusha:
#1058


Kludex:
Custom URL convertors have a public API now (Starlette 0.19.0). For that reason, I'll be closing this PR.

Thanks for the PR @Pentusha !

PR #1291, opened 2021-09-25
Support PyPy pipeline

Kludex:
Do we want this? I don't have an opinion about it. Tests are quite slow with it.

Feel free to close/discuss about it.


Kludex:
We can also run tests with `pypy` daily, or some variant.

Kludex:
I don't think it's worth it.

If someone sees this in the future and disagrees, please open a discussion or talk about it on our Gitter.

PR #1292, opened 2021-09-26
Update the Jinja2Templates() constructor to allow PathLike

tonybaloney:
The directory argument for `Jinja2Templates` is annotated as a `str`, which means a type checker fails if you pass a `pathlib.Path`, but Jinja2 accepts `PathLike[str]` (either a path string or a Path) for environment creation.

This PR updates the type annotation to allow both cases.

### DIFF
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -1,3 +1,4 @@
+from os import PathLike
 import typing
 
 from starlette.background import BackgroundTask
@@ -54,11 +55,11 @@ class Jinja2Templates:
     return templates.TemplateResponse("index.html", {"request": request})
     """
 
-    def __init__(self, directory: str) -> None:
+    def __init__(self, directory: PathLike[str]) -> None:
         assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
         self.env = self._create_env(directory)
 
-    def _create_env(self, directory: str) -> "jinja2.Environment":
+    def _create_env(self, directory: PathLike[str]) -> "jinja2.Environment":
         @pass_context
         def url_for(context: dict, name: str, **path_params: typing.Any) -> str:
             request = context["request"]
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -60,7 +60,7 @@ class Jinja2Templates:
         assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
         self.env = self._create_env(directory)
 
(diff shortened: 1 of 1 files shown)


starlette/templating.py
```diff
@@ -54,11 +56,11 @@ class Jinja2Templates:
     return templates.TemplateResponse("index.html", {"request": request})
     """
 
-    def __init__(self, directory: str) -> None:
+    def __init__(self, directory: Union[str, PathLike]) -> None:
```
```suggestion
    def __init__(self, directory: typing.Union[str, PathLike]) -> None:
```

starlette/templating.py
```diff
@@ -54,11 +56,11 @@ class Jinja2Templates:
...
         assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
         self.env = self._create_env(directory)
 
-    def _create_env(self, directory: str) -> "jinja2.Environment":
+    def _create_env(self, directory: Union[str, PathLike]) -> "jinja2.Environment":
```
```suggestion
    def _create_env(self, directory: typing.Union[str, PathLike]) -> "jinja2.Environment":
```

starlette/templating.py
```diff
@@ -1,4 +1,6 @@
+from os import PathLike
 import typing
+from typing import Union
```
```suggestion
import typing
from os import PathLike
```

Kludex (changes requested):
@tonybaloney Thanks for the PR! :tada: 

I also prefer to import each type from `typing`, but I would like to maintain consistency between modules. :kissing: 

Would you mind checking my suggestions? It also solves the CI.

tonybaloney:
Done! Thanks

starlette/templating.py
```diff
@@ -54,11 +55,11 @@ class Jinja2Templates:
     return templates.TemplateResponse("index.html", {"request": request})
     """
 
-    def __init__(self, directory: str) -> None:
+    def __init__(self, directory: typing.Union[str, PathLike]) -> None:
         assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
         self.env = self._create_env(directory)
 
-    def _create_env(self, directory: str) -> "jinja2.Environment":
+    def _create_env(self, directory: typing.Union[str, PathLike]) -> "jinja2.Environment":
```
```suggestion
    def _create_env(
        self, directory: typing.Union[str, PathLike]
    ) -> "jinja2.Environment":
```

Kludex (changes requested):

Kludex (approved):
Thanks @tonybaloney ! 🥳

PR #1198, opened 2021-06-11
🔥 Remove GraphQL support

Kludex:
This PR removes GraphQL entirely. I've assumed that we don't need to document the removal, just the CHANGELOG is enough (correct me otherwise).

The idea is to merge this PR before 1.0. There's no hurry on this, once we fulfill the other requirements, we can come back to this PR.

## Related issues
- Moving forward https://github.com/encode/starlette/issues/51
- Closes https://github.com/encode/starlette/issues/619

PR #1297, opened 2021-10-03
Add session argument for GraphQLApp() constructor

conao3:
* Add session argument for GraphQLApp() constructor to allow
specify sqlalchemy session.

---

see [graphene doc](https://docs.graphene-python.org/projects/sqlalchemy/en/latest/tips/#querying).

Now we use this method,

> Create a query for the models.
```python
Base = declarative_base()
Base.query = db_session.query_property()

class MyModel(Base):
    # ...

```

but below method cannot use.

> Set the db session when you do the execution:
```python
schema = graphene.Schema()
schema.execute(context_value={'session': session})
```

This patch allows to use latter method like this.
```python
schema = graphene.Schema(query=Query)
app = starlette.graphql.GraphQLApp(schema=schema, session={{your SA session}})
```

### DIFF
--- a/starlette/graphql.py
+++ b/starlette/graphql.py
@@ -33,6 +33,7 @@ class GraphQLApp:
         schema: "graphene.Schema",
         executor_class: type = None,
         graphiql: bool = True,
+        session: "sqlalchemy.orm.Session" = None,
     ) -> None:
         self.schema = schema
         self.graphiql = graphiql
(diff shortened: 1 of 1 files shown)


Kludex:
GraphQL's support in Starlette will be removed in the next release (0.17.0). Check: https://github.com/encode/starlette/pull/1198

PR #1302, opened 2021-10-07
Fix FileResponse not listening for http.disconnect

### LINKED ISSUE #1301: FileResponse Stream continues streaming after http.disconnect
### Checklist

### Describe the bug

While downloading a file provided via `FileResponse` the starlette application continues "sending" chunks even if the client aborts.
This is the same bug as I found reported and fixed for StreamResponse. References:
 - issue in Fastapi repo: https://github.com/tiangolo/fastapi/issues/1342
 - fixing PR in starlett repo: https://github.com/encode/starlette/pull/839

### To reproduce

```python
async def stream_file_response(request: Request) -> FileResponse:
    bg = BackgroundTask(...)
    return FileResponse(
        path=PATH, filename="example.png", background=bg
    )

###############################################################3

response = requests.get(
    f"http://{HOST}:{PORT}/file",
    stream=True,
)
response.close()
```

For details see following PR.

### Expected behavior

FileResponse is also listening for http.disconnect and aborts stream if client disconnects.

### Actual behavior

The application doesn't start with the backgroundtask right away, but continues sending all the cunks first until finished.

### Debugging material

See `tests/test_response_abort.py` in following PR.

### Environment

- OS: Linux Arch
- Python version: Python 3.9.7
- Starlette version: master, 0.15.0

### Additional context

n.a.

idotobi:
Fixes #1301

# Summary

 - implement aborting FileResponse stream on http.disconnect similar to
   StreamResponse
 - implement on_complete Background Tasks for
   StreamResponse/FileResponse
    - background: always executed (even if stream aborted)
    - on_complete: only executed if stream not abored by client
   I initially assumed that background would work as on_complete is now
   implemented. However as this is not the case, adding on_complete
   seemed like the best choice to ensure backwards compatibility.
 - implement testsuite for abort scenarios

# On_complete 
I'm aware that `on_complete` rather counts as a new feature, however it was very useful in writing the test and I would consider it a reasonable feature (or even default behaviour :thinking: ).
Consider it as a basis for discussion. If you prefer I'm happy to separate it out in a different PR.

# Further improvements

It might also make sense to implement FileResponse as a child class of Streaming Response, as they share quite some code.
If you agree I'd be happy to implement this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ pyyaml
 requests
 
 # Testing
+anyio
 autoflake
 black==20.8b1
 coverage>=5.3
@@ -19,6 +20,7 @@ types-PyYAML
 types-dataclasses
 pytest
 trio
+uvicorn
 
 # Documentation
 mkdocs
--- a/scripts/test
+++ b/scripts/test
@@ -11,7 +11,8 @@ if [ -z $GITHUB_ACTIONS ]; then
     scripts/check
 fi
 
-${PREFIX}coverage run -m pytest $@
+${PREFIX}coverage run --concurrency=multiprocessing -m pytest $@
+${PREFIX}coverage combine
 
 if [ -z $GITHUB_ACTIONS ]; then
     scripts/coverage
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -6,7 +6,6 @@ import stat
 import sys
 import typing
 from email.utils import formatdate
-from functools import partial
 from mimetypes import guess_type as mimetypes_guess_type
 from urllib.parse import quote
 
@@ -202,6 +201,7 @@ class StreamingResponse(Response):
         headers: dict = None,
         media_type: str = None,
         background: BackgroundTask = None,
+        on_complete: BackgroundTask = None,
     ) -> None:
         if isinstance(content, typing.AsyncIterable):
             self.body_iterator = content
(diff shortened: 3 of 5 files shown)


idotobi:
@Kludex : Thanks for approving the workflow. I updated the import sorting to fix the failing linting. 

Is there an easy way for me to get all checks passing in my fork? I don't want to bug you too much with my attempts to make the checks pass and keep the iteration cycles small :sweat_smile:

Kludex:
I don't think there is a way for that :thinking:

idotobi:
Ok. No worries, then I'll try to run the checks locally. Thanks for the quick reply :+1:

idotobi:
@Kludex : Should pass the tests now :crossed_fingers:

idotobi:
Ok, I think I would need some help to get the test to run consistently.

**The problem I'm trying to solve is the following:**
 I need to abort an ongoing streaming request while it's running.
It seems that only `requests` really supports this as both the `TestClient` as well as `httpx` seem to block until the full response is received, even if `stream=True` is set.

**My current approach is**:
I use two processes:
 - one for serving starlette via uvicorn
 - one for calling `requests`

However, this is not really pretty as it is very sensitive to timing (seems to not play well with azure pipelines) and it basically only tests starlette when served by uvicorn.

Does someone have an idea for a better approach? Did I simply use `httpx`/`TestClient` wrong?

Kludex:
@idotobi Thanks for the PR! :smile: 

This PR is huge, and it aims to address multiple things. It will be great if you can prepare the PR that solves #1301 first. I've already confirmed the issue, and the PR is more than welcome.

As for other things you mentioned on this PR, I think a discussion on Gitter or GitHub discussions is better before trying to jump into a PR for them. :)

idotobi:
> This PR is huge, and it aims to address multiple things. It will be great if you can prepare the PR that solves #1301 first. I've already confirmed the issue, and the PR is more than welcome.

@Kludex : Fair point. Will do so during the week.

idotobi:
Took a bit longer than a week :sweat_smile:  #1427

PR #1306, opened 2021-10-10
Zerocopysend extension support

synodriver:
According to [ASGI Spec](https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send), applications can use sendfile if supported by asgi server.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -313,18 +313,34 @@ class FileResponse(Response):
         )
         if self.send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
-        else:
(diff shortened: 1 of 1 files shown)


Kludex:
@synodriver Thanks for the PR, but I think we're going to go with https://github.com/encode/starlette/pull/1288

I guess you missed it 😅 

Do you mind reviewing it? :)

synodriver:
> @synodriver Thanks for the PR, but I think we're going to go with #1288
> 
> I guess you missed it 😅
> 
> Do you mind reviewing it? :)

Sorry, forget to search before that 😅

Issue #369, opened 2019-01-30
Finesse host-based routing.


lovelydinosaur:
Follow-on work from #363

* Needs documenting.
* The name of `request.path_params` isn't exactly correct anymore. Should use `request.host_params` here?

aminalaee:
I think we can close this now. Documented per #363

PR #1313, opened 2021-10-16
add host-based routing docs

WhiteApfel:
Yesterday I wanted to use host-based routing, saw the necessary methods in the Starlette class, but did not find any info in docs. After searching, I found #363 and decided to add at least a minimal mention of host-based routing to the documentation.

#369 opened almost three years ago and is still open 🤡

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -220,7 +220,7 @@ app = Router(routes=[
 
 ## Host-based routing
 
-If you want to use different routes for the same path based on the `Host` header
+If you want to use different routes for the same path based on the `Host` header.
 
 ```python
 site = Router()  # Use eg. `@site.route()` to configure this.
@@ -238,3 +238,8 @@ app.host('www.example.org', site)
 news_host = Host('news.example.org', news)
 app.router.routes.append(news_host)
 ```
+
+Note that port is removed from the `Host` header when matching.
+For example, `Host (host='example.org:3600', ...)` will not be processed 
+even if the `Host` header is `example.org:3600`. 
+Therefore, specify only the domain or IP address
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -187,6 +187,8 @@ For example, `Host (host='example.org:3600', ...)` will not be processed
 even if the `Host` header is `example.org:3600`. 
 Therefore, specify only the domain or IP address
 
+There are several ways to connect host-based routes to your application
+
 ```python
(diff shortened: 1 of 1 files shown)


docs/routing.md
```diff
@@ -217,3 +217,24 @@ app = Router(routes=[
     ])
 ])
 ```
+
+## Host-based routing
```
aminalaee: I think documenting host-based routing, and subdomains without documenting URL reversing and lookups is not complete:
```python
request.url_for("...", subdomain=...)
request.url_for("api:...")
...
```

aminalaee (changes requested):

docs/routing.md
WhiteApfel: @aminalaee, then it will be better to place `Host-based routing` between `Submounting routes` and `Reverse URL lookups` sections? Because information about `url_for` should be in the `Reverse URL lookups`

aminalaee: TBH, I've done this before but couldn't decide where the best place for that would be.
Maybe after `Reverse URL lookups`? Let's go with this and see if we can find any other places.

WhiteApfel: We in `Reverse URL lookups` describe what was above. No, of course we can put all the information about host-based in one section, but I don't know how correct it will be

aminalaee: Sorry I'm not sure if I understand correctly. In the current version of docs we do have examples for submounts but are there any examples of subdomains?
`request.url_for("endpoint", subdomain=...)`

WhiteApfel: No. They are not present in the current version. I want to add example for host-submounts, but I'm just deciding where to place the information.

aminalaee: I think having a section for subdomains after `Reverse lookup is fine`. Maybe later we can find a better place for it.

aminalaee (commented):
```python
...
routes = [
     Host('api.example.org', api)
 ]
 app.host('www.example.org', site)
 news_host = Host('news.example.org', news)
 app.router.routes.append(news_host)
```

I think this is much better. Do you think we may need a little explanation here?

WhiteApfel:
> There are several ways to connect host-based routes to your application

It'll do? I think that this does not make much sense, because the meaning from the upper sections is duplicated

aminalaee:
That makes sense.

aminalaee (approved):
Thank you for this.

PR #1314, opened 2021-10-16
Fix Staticfiles `404.html` in HTML mode

WhiteApfel:
In StaticFiles html-mode, when `404.html` and `index.html` files were missing, a FileNotFoundError exception was rised. As I understand it, a non-existent file was checked in the `os.stat` function

However, this is not the whole problem. Even if `404.html` exists, it might not be returned, because after checking for the existence of `index.html`, a basic 404 exception is returned, rather than returning a `404.html` file with the correct status code

Therefore, I moved the check for the existence of a `404.html` to the end, when index cannot be found. I also considered it correct to move the file absence exception to the `lookup_path` and, in case of absence, return an empty string and None, so as not to process Exception several times.

_Sorry for my bad english_

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -111,20 +111,6 @@ class StaticFiles:
             full_path, stat_result = await anyio.to_thread.run_sync(
                 self.lookup_path, path
             )
-        except (FileNotFoundError, NotADirectoryError):
-            if self.html:
-                # Check for '404.html' if we're in HTML mode.
-                full_path, stat_result = await anyio.to_thread.run_sync(
-                    self.lookup_path, "404.html"
-                )
-                if stat_result and stat.S_ISREG(stat_result.st_mode):
-                    return FileResponse(
-                        full_path,
-                        stat_result=stat_result,
-                        method=scope["method"],
-                        status_code=404,
-                    )
-            raise HTTPException(status_code=404)
         except PermissionError:
             raise HTTPException(status_code=401)
         except OSError:
@@ -149,6 +135,18 @@ class StaticFiles:
                     return RedirectResponse(url=url)
                 return self.file_response(full_path, stat_result, scope)
 
+        if self.html:
+            # Check for '404.html' if we're in HTML mode.
+            full_path, stat_result = await anyio.to_thread.run_sync(
+                self.lookup_path, "404.html"
+            )
+            if stat_result and stat.S_ISREG(stat_result.st_mode):
+                return FileResponse(
+                    full_path,
+                    stat_result=stat_result,
+                    method=scope["method"],
+                    status_code=404,
+                )
         raise HTTPException(status_code=404)
 
     def lookup_path(
@@ -161,7 +159,10 @@ class StaticFiles:
                 # Don't allow misbehaving clients to break out of the static files
                 # directory.
                 continue
-            return full_path, os.stat(full_path)
+            try:
+                return full_path, os.stat(full_path)
+            except (FileNotFoundError, NotADirectoryError):
+                continue
         return "", None
 
     def file_response(
--- a/tests/test_staticfiles.py
+++ b/tests/test_staticfiles.py
@@ -229,7 +229,7 @@ def test_staticfiles_304_with_last_modified_compare_last_req(
     assert response.content == b"<file content>"
 
 
-def test_staticfiles_html(tmpdir, test_client_factory):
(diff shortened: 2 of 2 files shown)


WhiteApfel:
I looked at the history of file changes. At 1db8b5b4e8c31da1d0836074625981f54a54c0d2 (#1220) has undergone "devastating" changes. For some reason, check of the `404.html` was moved up there. I read the story (#1123, #1124, #1220), but didn't find an explanation for this action

Obviously, we must first check `{path}/index.html`, and only then check `404.html`

starlette/staticfiles.py
```diff
@@ -111,20 +111,6 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
-                    self.lookup_path, "404.html"
-                )
-                if stat_result and stat.S_ISREG(stat_result.st_mode):
-                    return FileResponse(
-                        full_path,
-                        stat_result=stat_result,
-                        method=scope["method"],
-                        status_code=404,
-                    )
-            raise HTTPException(status_code=404)
```
WhiteApfel, line 127: Checking for `404.html` should be after checking for `index.html`

starlette/staticfiles.py
```diff
@@ -161,7 +159,10 @@ def lookup_path(
                 # Don't allow misbehaving clients to break out of the static files
                 # directory.
                 continue
-            return full_path, os.stat(full_path)
+            try:
+                return full_path, os.stat(full_path)
+            except (FileNotFoundError, NotADirectoryError):
+                continue
```
WhiteApfel, line 165: `os.stat` is only available for existing files and `"", None` works as no file

aminalaee (commented):
I think that was moved by mistake.
Maybe you can add tests for it so this won't happen again?

WhiteApfel:
Hmmm... @aminalaee, I don't understand the testing system well, but I have a problem

Tests fail due to HTTPException being called:

```
>       raise HTTPException(status_code=404)
E       starlette.exceptions.HTTPException

../starlette/staticfiles.py:150: HTTPException
```

I've seen this same construct in other tests. And no exception is raised there. I need help with this.

Even while writing the tests, I noticed that now when 404.html or index.html is not found, the '/' at the end of the URL is removed. 
I just took the liberty of removing these lines from my tests (b564f1fb04ac3898dbed7e94ad84d0238190b26d), because they are not target, however, they do not exclude the presence of "anomalous" behavior

aminalaee:
@WhiteApfel 

If I understand correctly, you need to write a test to catch an exception?
I think we're using py-test for that, something like this:

```python
with pytest.raises(RuntimeError) as exc_info:
        client.get("/example.txt")
```

This ensures calling the GET will raise the specified exception.

As for the removal of the tests, I'd rather we don't change them, that would lead to un-intented changes.

tests/test_staticfiles.py
```diff
@@ -262,6 +262,73 @@ def test_staticfiles_html(tmpdir, test_client_factory):
...
+        file.write("<h1>Hello</h1>")
+
+    app = StaticFiles(directory=tmpdir, html=True)
+    client = test_client_factory(app)
+
+    response = client.get("/dir/")
+    assert response.status_code == 200
+    assert response.text == "<h1>Hello</h1>"
+
+    response = client.get("/dir")
```
aminalaee, line 305: Am I missing something, or are you calling "/dir" twice?
It's the same in `test_staticfiles_html_without_index` test.

WhiteApfel:
@aminalaee 
> As for the removal of the tests, I'd rather we don't change them, that would lead to un-intented changes.

> https://github.com/encode/starlette/pull/1314#discussion_r730936586

```python
response = client.get("/dir/")
assert response.url == "http://testserver/dir/"

response = client.get("/dir")
assert response.url == "http://testserver/dir/"
```

Oh, I'm sorry. I made a typo during preliminary testing. Corrected. In some cases a '/' should appear and in others it should not. I got the opposite results due to a typo, but now everything is fine ✨

tests/test_staticfiles.py
WhiteApfel: I restore URL validation back. Yes, now (21edbc8db80af22bd6858b138c39265de436a3da, 5ce62f8cbc6ff3ec657ac45affa98fb1b4e0fc2e) different paths are checked because different responses are obtained depending on the files in directory.

WhiteApfel:
I think that's all. I don't plan to make any more changes

aminalaee (approved):
Thank you for this.

Issue #1175, opened 2021-04-28
Custom 500 or Exception handlers do not run through middleware like other handled exceptions.


danielknell:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### To reproduce

```python
from starlette.applications import Starlette
from starlette.middleware import Middleware
from starlette.middleware.cors import CORSMiddleware
from starlette.responses import JSONResponse

app = Starlette(routes=routes, middleware=middleware)

async def not_found(request, exc):
    return JSONResponse(content={"error": 404}, status_code=exc.status_code)

async def server_error(request, exc):
    return JSONResponse(content={"error": 500}, status_code=exc.status_code)

exception_handlers = {
    404: not_found,
    500: server_error
}

middleware = [
    Middleware(CORSMiddleware, allow_origins=['*'])
]

app = Starlette(routes=routes, middleware=middleware, exception_handlers=exception_handlers)
```

### Expected behavior

both 404 and 500 errors will respond with their relevant json response, and correct CORS headers

### Actual behavior

the 404 error will response with correct CORS headers but the 500 error will not.

### Debugging material

Because `Exception` and `500` are special cased in `build_middleware_stack` any error bubbles up to the outer most `ServerErrorMiddleware` which is outside the other middleware, instead of being handled by the `ExceptionMiddleware` and passing through the custom middleware.

see: https://github.com/encode/starlette/blob/master/starlette/applications.py#L74-L77

I can see four possible fixes, none of them perfect...

- handle custom 500 errors in `ExceptionMiddleware` and prevent `ServerErrorMiddleware` showing its helpful debug info when a custom handler is present.
- have `ServerErrorMiddleware` run the response of the error handler through the middleware stack again, and if that fails resort to existing behaviour, adding a bunch of complexity.
- move the debuging info to a separate middleware and put that in front of the ErrorMiddleware, which would then cause it to trigger for every exception, but would allow `ExceptionMiddleware` handle 500s as expected, and `ServerErrorMiddleware` becomes a last resort.
- update https://www.starlette.io/exceptions/ to explain 500 errors are handled differently, and to get the above behaviour something like the following work-around is required:

  ```py
  @app.middleware("http")
  def exception_middleware(request, call_next)
      try:
          return await call_next(request)
      except Exception:
          return JSONResponse(content={"error": 500}, status_code=500))
  ```

### Environment

- OS: N/A
- Python version: N/A
[...]

PR #1321, opened 2021-10-23
Fix issue when CORS header is missed on unhandled exception

### LINKED ISSUE #1116: Error are not reported using CORS middleware

uriyyo:
Fixes #1116


lovelydinosaur:
As per https://github.com/encode/starlette/issues/1175#issuecomment-1036343311 my suggestion is that we close this off as expected (and desired) behaviour.

(I *think* Django probably also has the same approach here?)

PR #1322, opened 2021-10-26
Fix port issue on host-based routing

### LINKED ISSUE #1317: 🐛 Different `Host` behavior for setting `host` argument and getting `Host` header
### Checklist

### Describe the bug
When adding a host-based routes, `host` value can accept the port and does not process it in any way. But `Host.matches()` cuts off the port number ([`routing.py:432`](https://github.com/encode/starlette/blob/master/starlette/routing.py#L432))

### Expected behavior
When adding, cut off the port, or сonsider it when matching.

### Actual behavior
```python
import uvicorn as uvicorn
from starlette.routing import Host, Router, Route

app = Starlette(routes=[
    Host('127.0.0.1:8889', app=Router(routes=[Route('/', lambda *a, **k: PlainTextResponse('Hello World'))])),
])

uvicorn.run(app, port=8889)
```

Breakpoint at [`routing.py:433`](https://github.com/encode/starlette/blob/master/starlette/routing.py#L433)
breakpoint variables

- Starlette version: 0.16.0

WhiteApfel:
`Host.host_regex` is only used to check "Host" header: ([`routing.py:432`](https://github.com/encode/starlette/blob/master/starlette/routing.py#L432))
```python
host = headers.get("host", "").split(":")[0]
match = self.host_regex.match(host)
```
However, when generating `Host.host_regex`, the port remained: ([`routing.py:149`](https://github.com/encode/starlette/blob/master/starlette/routing.py#L149))
```python
path_regex += re.escape(path[idx:]) + "$"
```

 I added one small fix. Now the port is also removed from `Host.host_regex`

**UPD:** Before this fix, the code from #1317 example returned `Not found`, and after `Hello World`


WhiteApfel:
If you accept this change, then we will need to update info in the documentation (I added a note in #1313). I can commit to this pull request, or create a new one. What should I do?

PrettyWood:
Hey thanks for that. Surely a unit test can be added!

parz3val (commented):
lgtm

aminalaee:
> If you accept this change, then we will need to update info in the documentation (I added a note in #1313). I can commit to this pull request, or create a new one. What should I do?

Yes, I think we can update the doc in the same PR since they are related.

WhiteApfel:
I think that in general the pull request is ready. But I'm not sure about the quality of the wording in the note

WhiteApfel:
These errors are not caused by code, is it?

aminalaee:
> These errors are not caused by code, is it?

Probably Github's issue:

`GitHub Actions has encountered an internal error when running your job.`

Kludex:
It looks like is running now. I've relaunch the pipeline.

aminalaee (approved):
LGTM. Thank you for this.

PR #1262, opened 2021-08-11
Prevent anyio.ExceptionGroup in error views under a BaseHTTPMiddleware

florimondmanca:
Closes #1255 

**Motivation**
#1157 switched to using `anyio` for async concurrency operations.

But now when a view raises an exception, say `MyExc`, and a `BaseHTTPMiddleware` is installed, calling the view would make Starlette raise an `anyio.ExceptionGroup`, rather than `MyExc` directly.

Previously, under `asyncio`, we were calling `task.result()` before raising `RuntimeError("No response sent.")`, so only `MyExc` was raised, as expected, see here: https://github.com/encode/starlette/blob/f5a08d019ade528ae3a855f7e52a5500b233cb0a/starlette/middleware/base.py#L45-L46

This PR switches back to that behavior by adding a `checkpoint` (equivalent to `asyncio.sleep(0)` — i.e. "allow switching tasks and checking for cancellation) before raising the "No response sent" exception.

cc @uSpike

PR #1328, opened 2021-11-02
Catch more general`BaseException` instead of `Exception`

mbergal:
`anyio.ExceptionGroup`, for example, is not an `Exception` but is a `BaseException`

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -157,7 +157,7 @@ class ServerErrorMiddleware:
 
         try:
             await self.app(scope, receive, _send)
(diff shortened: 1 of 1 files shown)


mbergal:
Might be not super relevant after https://github.com/encode/starlette/commit/6c9bc55af7d9d202c2bc1299c9a65e844640c197?diff=split

Kludex:
I think we don't want to merge this even if #1262 didn't land.

You can read more about it on https://www.python.org/dev/peps/pep-0352/

But thanks for the PR! :smile:

PR #1329, opened 2021-11-02
Update third-party-packages.md

Sobolev5:
Add Starlette+RabbitMQ RPC calls full working example.


Kludex:
Do we want another section for projects using Starlette, or should we follow the name of the section strictly i.e. "Third Party Packages"?

If the latter, then we should close this PR.

Kludex:
I think the section should be only for Third Party Packages, and the project mentioned on this PR doesn't fit in this category.

We're happy that you're able to build cool things with Starlette and share with us, but for the reason above, I'll close this PR.

Thanks for the interest! :)

PR #992, opened 2020-07-04
Add reason support for websocket.close

Chaostheorie:
Resolve #991.

PR #1339, opened 2021-11-22
 Add reason support for websocket.close

### LINKED ISSUE #991: [Feature] Reason for websocket.close
I've recently started actively developing with websockets in starlette. When closing a websocket you can provide a code e.g. 1000, 1008 … to tell the client the basic reason. But the websocket protocol allows the developer to also give UTF-8-encoded data as a complementary description/ reason. This would be useful in multiple cases such as mentioning the specific violitioan for code 1008 (POLICY VIOLIATION). [Websockets](https://websockets.readthedocs.io/en/stable/api.html?highlight=reason#websockets.protocol.WebSocketCommonProtocol.close) supports this too and their [implementation](https://websockets.readthedocs.io/en/stable/_modules/websockets/protocol.html#WebSocketCommonProtocol.close) could maybe help.

Specification: [https://tools.ietf.org/](https://tools.ietf.org/html/rfc6455#section-5.5.1)

Chaostheorie:
Recreate old changes

Carbon-Copy of #992 
Aims to resolve #991

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -151,7 +151,7 @@ class WebSocketClose:
         self.reason = reason
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if reason != "":
+        if self.reason is None:
             await send({"type": "websocket.close", "code": code})
         else:
             await send({"type": "websocket.close", "code": code, "reason": reason})
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -138,8 +138,8 @@ class WebSocket(HTTPConnection):
         else:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
-    async def close(self, code: int = 1000, reason: typing.Optional[str] = "") -> None:
-        if reason != "":
+    async def close(self, code: int = 1000, reason: str = None) -> None:
+        if reason is None:
             await self.send({"type": "websocket.close", "code": code})
         else:
             await self.send({"type": "websocket.close", "code": code, "reason": reason})
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -138,13 +138,20 @@ class WebSocket(HTTPConnection):
         else:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
-    async def close(self, code: int = 1000) -> None:
-        await self.send({"type": "websocket.close", "code": code})
+    async def close(self, code: int = 1000, reason: typing.Optional[str] = "") -> None:
+        if reason != "":
+            await self.send({"type": "websocket.close", "code": code})
+        else:
+            await self.send({"type": "websocket.close", "code": code, "reason": reason})
 
 
 class WebSocketClose:
-    def __init__(self, code: int = 1000) -> None:
+    def __init__(self, code: int = 1000, reason: typing.Optional[str] = "") -> None:
         self.code = code
+        self.reason = reason
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        await send({"type": "websocket.close", "code": self.code})
+        if reason != "":
+            await send({"type": "websocket.close", "code": code})
+        else:
+            await send({"type": "websocket.close", "code": code, "reason": reason})


michallowasrzechonek-silvair:
Re "it's not part of the spec", https://github.com/django/asgiref/commit/6469d0f3486dae0b6229e12d32ae6695600ed3ec

starlette/websockets.py
```diff
@@ -138,13 +138,20 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
         else:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
-    async def close(self, code: int = 1000) -> None:
-        await self.send({"type": "websocket.close", "code": code})
+    async def close(self, code: int = 1000, reason: typing.Optional[str] = "") -> None:
+        if reason != "":
+            await self.send({"type": "websocket.close", "code": code})
+        else:
+            await self.send({"type": "websocket.close", "code": code, "reason": reason})
```
```suggestion
    async def close(self, code: int = 1000, reason: str = None) -> None:
        if reason is None:
            await self.send({"type": "websocket.close", "code": code})
        else:
            await self.send({"type": "websocket.close", "code": code, "reason": reason})
```

starlette/websockets.py
```diff
@@ -138,13 +138,20 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
...
 
 
 class WebSocketClose:
-    def __init__(self, code: int = 1000) -> None:
+    def __init__(self, code: int = 1000, reason: typing.Optional[str] = "") -> None:
```
Kludex, line 149:
```suggestion
    def __init__(self, code: int = 1000, reason: str = None) -> None:
```

starlette/websockets.py
```diff
@@ -138,13 +138,20 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
...
         self.code = code
+        self.reason = reason
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        await send({"type": "websocket.close", "code": self.code})
+        if reason != "":
+            await send({"type": "websocket.close", "code": code})
+        else:
+            await send({"type": "websocket.close", "code": code, "reason": reason})
```
```suggestion
        if self.reason is None:
            await send({"type": "websocket.close", "code": code})
        else:
            await send({"type": "websocket.close", "code": code, "reason": reason})
```

Kludex (commented):
There are some approaches we can take with the "reason". On the "code" key, we choose to send `1000` by default, but that's not quite what is written in the specs. We could have choose to not send it, and the [server would assume `1000`](https://github.com/encode/uvicorn/blob/8b05af307e308e3e188fa2b4b0c4d47112321933/uvicorn/protocols/websockets/wsproto_impl.py#L299).

As for the "reason" key, we can do the same, and assume default being empty string. Or we can just do as the suggested snippets and only send the reason if a reason is added. 

I've just noticed an issue on uvicorn on the way [we retrieve the reason](https://github.com/encode/uvicorn/blob/8b05af307e308e3e188fa2b4b0c4d47112321933/uvicorn/protocols/websockets/wsproto_impl.py#L300), as we don't check if reason is None, which is allowed by the spec. I'll be opening a PR there. 

Reference for the new ASGI spec: https://asgi.readthedocs.io/en/latest/specs/www.html#close-send-event

Kludex:
> Re "it's not part of the spec", [django/asgiref@6469d0f](https://github.com/django/asgiref/commit/6469d0f3486dae0b6229e12d32ae6695600ed3ec)

@michallowasrzechonek-silvair I didn't get what you said. :kissing:

starlette/websockets.py
Chaostheorie: This seems like the proper way. I might add that we would need to change reasons type (`str`) to be `typing.Optional[str]` but this would be implicated anyway given the default argument.

starlette/websockets.py
Chaostheorie: Same as above. Both will be comitted.

Kludex:
Would you mind also adding tests for this? :pray:

Chaostheorie:
I will fix it tomorrow afternoon got some studying to do

Kludex:
@Chaostheorie Are you still interested on working on this?

Chaostheorie:
@Kludex Not really. Sorry to leave this PR hanging, will close it asap.

PR #1341, opened 2021-11-24
Fix `url_path_for` typehints

### LINKED ISSUE #1319: Inconsistent Typing Between request.url_for and application.url_path_for
[`request.url_for` takes `Any` for the `path_params`](https://github.com/encode/starlette/blob/217019b36303ea16a5869ad076d835e4fdc1cef3/starlette/requests.py#L168) whereas [`application.url_path_for` expects `path_params` to be `str`](https://github.com/encode/starlette/blob/537ab6afd110b79dca95f3c8ecc6980710b1de1c/starlette/applications.py#L107). Likewise the route / routers expect `path_params` to be `str`. See [here](https://github.com/encode/starlette/blob/537ab6afd110b79dca95f3c8ecc6980710b1de1c/starlette/routing.py#L159) and [here](https://github.com/encode/starlette/blob/537ab6afd110b79dca95f3c8ecc6980710b1de1c/starlette/routing.py#L584).

The type mismatch is not caught by the type checker as the `router` here is untyped:
https://github.com/encode/starlette/blob/217019b36303ea16a5869ad076d835e4fdc1cef3/starlette/requests.py#L169-L170

I think `Any` is probably more close to being the correct type as the `path_params` can be `UUID`, `int`, etc.

[The docs](https://www.starlette.io/routing/#reverse-url-lookups) would make it seems that the API for these two methods should have a consistent signature.

aminalaee:
Changes type hints for `url_path_for `. There's a mismatch between `url_for` and `url_path_for` as one accepts `Any` and the other one is `str`. They should both be `Any`.

Right now `url_path_for ` expects path_parameters to be string, which is not the case, and `url_for` could be `int`, `UUID`, `float`, etc.

Fixes #1319.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,3 +32,7 @@ filterwarnings=
 
 [coverage:run]
 source_pkgs = starlette, tests
+
+[coverage:report]
+exclude_lines =
+    if typing.TYPE_CHECKING:
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,3 +36,4 @@ source_pkgs = starlette, tests
 [coverage:report]
 exclude_lines =
     if typing.TYPE_CHECKING:
+    ...
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,4 +36,5 @@ source_pkgs = starlette, tests
 [coverage:report]
 exclude_lines =
     pragma: no cover
+    pragma: nocover
     if typing.TYPE_CHECKING:
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,5 +35,5 @@ source_pkgs = starlette, tests
 
 [coverage:report]
 exclude_lines =
+    pragma: no cover
     if typing.TYPE_CHECKING:
-    ...
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -11,7 +11,7 @@ from starlette.types import Message, Receive, Scope, Send
 
 try:
     from multipart.multipart import parse_options_header
-except ImportError:  # pragma: no cover
+except ImportError:  # pragma: nocover
     parse_options_header = None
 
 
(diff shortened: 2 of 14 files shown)


florimondmanca (approved):
From reviewing this and the associated issue, this seems legit to me!

starlette/requests.py
```diff
@@ -15,6 +15,10 @@
     parse_options_header = None
 
 
+if typing.TYPE_CHECKING:  # pragma: nocover
```
PrettyWood: IMO better to add directly in setup.cfg
```
[coverage:report]
exclude_lines =
    if typing.TYPE_CHECKING:
    @typing.overload
    ...
```
WDYT?

aminalaee: Yeah makes sense. We'll probably need it in other places.

aminalaee:
@PrettyWood @florimondmanca I think the `typing.Any` isn't worth much for the `path_params`.

Checking `starlette/convertors.py`:

```python
CONVERTOR_TYPES = {
    "str": StringConvertor(),
    "path": PathConvertor(),
    "int": IntegerConvertor(),
    "float": FloatConvertor(),
    "uuid": UUIDConvertor(),
}
``` 

we could have a `Union` of these to be more explicit. Any ideas?

setup.cfg
```diff
@@ -32,3 +32,9 @@ filterwarnings=
 
 [coverage:run]
 source_pkgs = starlette, tests
+
+[coverage:report]
+exclude_lines =
+    pragma: no cover
+    pragma: nocover
```
Kludex, line 39: Those 2 need to be included? :thinking:

aminalaee: yes, overriding `exclude_lines` takes full control over what's excluded https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html.

PR #1344, opened 2021-11-27
Unable to reproduce issue #1319

Kludex:


Kludex:
Pipeline passes successfully: https://github.com/encode/starlette/runs/4341295319

cancan101:
I believe this issue comes up when passing a non string to `application.url_path_for` even though non str (eg uuid) should be accepted.

PR #1345, opened 2021-11-27
Change default chunk size from 4Kb to 64Kb

### LINKED ISSUE #1100: Make FileResponse.chunk_size variable

Kludex:
- Closes https://github.com/encode/starlette/issues/1100

@tomchristie On that issue, you said that "httpx uses 64k read size", but I couldn't find that. Do you mind pointing me out what did you mean?

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -248,7 +248,7 @@ class StreamingResponse(Response):
 
 
 class FileResponse(Response):
-    chunk_size = 65536
+    chunk_size = 64 * 1024
 
     def __init__(
         self,
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -248,7 +248,7 @@ class StreamingResponse(Response):
 
 
 class FileResponse(Response):
-    chunk_size = 4096
+    chunk_size = 65536
 
     def __init__(
         self,


dalf:
The related issue in httpcore : https://github.com/encode/httpcore/issues/135
In the source code: https://github.com/encode/httpcore/blob/4f6d793e440ffeda31d7d064b434ad8083d42634/httpcore/_async/http11.py#L38

starlette/responses.py
```diff
@@ -248,7 +248,7 @@ async def wrap(func: typing.Callable[[], typing.Coroutine]) -> None:
 
 
 class FileResponse(Response):
-    chunk_size = 4096
+    chunk_size = 65536
```
```suggestion
    chunk_size = 64 * 1024
```

Kludex:
Thanks @dalf ! Much appreciated! :tada:

florimondmanca (approved):
Looking good! That's quite a similar change to the one we made in HTTPX indeed, and for the same reason (improving transfer speed by using larger chunks).

Kludex:
Thanks @florimondmanca ! 🎉

PR #1222, opened 2021-06-27
Allow for custom head method in HTTPEndpoint

theblazehen:
Fix for https://github.com/encode/starlette/issues/1221

PR #1346, opened 2021-11-27
Allow head method httpendpoint

### LINKED ISSUE #1221: Cannot define a custom HEAD handler in HTTPEndpoint
### Checklist

### Describe the bug

I am unable to set a custom head method for a HTTPEndpoint

### How to reproduce

Python:
```python

class TestHandler(HTTPEndpoint):
    async def head(self, request):
        print("Called head")
        return Response("head")

    async def get(self, request):
        print("Called get")
        return Response("get")

routes = [
    Route("/", TestHandler),
]

```
Then run uvicorn, and call `curl -XHEAD localhost:8000/` and `curl -XGET localhost:8000/`

You will see in uvicorn logs, both times the get handler gets called.
```
Called get
INFO:     127.0.0.1:53438 - "HEAD / HTTP/1.1" 200 OK
Called get
INFO:     127.0.0.1:53440 - "GET / HTTP/1.1" 200 OK
```

### Expected behavior

When calling the HEAD method, it should invoke the head() function

### Actual behavior

When calling the HEAD method, it invokes the get() function

### Environment

- OS: Linux
- Python version: Python 3.9.5
- Starlette version: 0.15.0

### Additional context

The issue is caused by the following code:
https://github.com/encode/starlette/blob/master/starlette/endpoints.py#L26

It should see if there is a head method, and if so, don't call the get method when head is requested

Eg. linked issues, or a description of what you were trying to achieve. -->

Kludex:
- Replaces #1222
- Closes #1221


starlette/endpoints.py
```diff
@@ -23,7 +23,12 @@ def __await__(self) -> typing.Generator:
 
     async def dispatch(self) -> None:
         request = Request(self.scope, receive=self.receive)
-        handler_name = "get" if request.method == "HEAD" else request.method.lower()
+        handler_name = (
+            "get"
+            if request.method == "HEAD" and not hasattr(self, "head")
```
aminalaee, line 28: Is `request.method` always capital `HEAD`?
Or should we consider `if request.method.lower() == "head"`?

Kludex: always uppercase https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope

aminalaee (approved):

Issue #1315, opened 2021-10-17
RunTimeError: got Future <Future pending> attached to a different loop when using custom loop in sync fixtures when upgrading from 0.14.2 to 0.15.0


sevaho:
### Checklist

- [X] The bug is reproducible against the latest release and/or `master`.
- [X] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Upgrading starlette>=0.15.0 breaks current testing strategy. The setup is mocking a [nats](https://nats.io) subscription by actually using the nats server.
The code works with starlette 0.14.2, upgradign to 0.15.0 gives `RunTumeError got Future <Future pending> attached to a different loop `. When upgrading to starlette 0.16.0 it gives TimeOut errors. I would love to keep tests sync.

### To reproduce

requirements.txt
```
starlette
requests
pytest
asyncio-nats-client
```

code

```python
from starlette.routing import Route
from starlette.testclient import TestClient
import pytest
from starlette.applications import Starlette
from starlette.responses import PlainTextResponse
import asyncio


from nats.aio.client import Client as NATS

"""
Test work with starlette 0.14.2
Error with starlette 0.15.0: RunTimeError: got Future <Future pending> attached to a different loop
Error with starlette 0.16.0: Nats timeout

The test is that the client code makes a nats request to a mocked nats service over nats itself.

Requirement a running nats server `docker run -d -p 4222:4222 nats:latest`
"""

HOST_NATS = "localhost:4222"


# =======================================================================
# CODE
# =======================================================================


def create_app():
    async def index(request):
        r = await request.app.state.nc.request("subject1", timeout=1, payload=b"PING")
        return PlainTextResponse(content=r.data.decode())

    async def setup() -> None:
        await app.state.nc.connect(HOST_NATS)
        print("Connected to nats")

    app = Starlette(debug=True, routes=[Route('/', index)], on_startup=[setup])
    app.state.nc: NATS = NATS()

    return app


app = create_app()


# =======================================================================
# MOCKS & TESTS
# =======================================================================


class NatsServiceMock:
    def __init__(self) -> None:
        self.nc: NATS = NATS()

    async def lifespan(self) -> None:
        await self.nc.connect(HOST_NATS)
        await self.nc.subscribe("subject1", cb=self.handle)

    async def handle(self, msg):
        await self.nc.publish(msg.reply, b"PONG")

    def __enter__(self):
        loop = asyncio.get_event_loop()
        loop.run_until_complete(self.lifespan())
        return self

[...]

PR #1347, opened 2021-11-28
Support customizing portal in TestClient

MatthewScholefield:
Consider a situation where we would like to create a second instance of `TestClient` within a started app:

```python
@pytest.fixture(scope='session')
def running_app() -> TestClient:
    with TestClient(app) as running_app:
        yield app

@pytest.fixture()
def user_client() -> TestClient:
    user_client = TestClient(app)
    user_client.headers['Authorization'] = 'Bearer abc123'
    return user_client

def test_auth(running_app: TestClient, user_client: TestClient):
    public_client = running_app
    assert public_client.get('/protected').status_code == 403
    assert user_client.get('/protected').status_code == 200
```

Previously, in `starlette<0.15.0`, this worked fine. However, since `starlette>=0.15.0` this causes mayhem when something attaches to the event loop on startup and then is used within a request. The specific reason this doesn't work is because two different event loops are created as described [here](https://github.com/encode/starlette/issues/1315#issuecomment-980784457) in #1315. This PR adds a new `portal` argument to `TestClient` which allows sharing the same `BlockingPortal` instance from anyio and transitively sharing the same event loop.

With this change, the example above just has to be slightly adapted as follows:
```python
@pytest.fixture()
def user_client(running_app) -> TestClient:
    user_client = TestClient(app, portal=running_app.portal)
    user_client.headers['Authorization'] = 'Bearer abc123'
    return user_client
```

Feel free to suggest alternate solutions. I figured I'd create the PR for discussion (since the code change is simple enough).

Also, note that from what I understand of Trio, this problem wouldn't happen there because it doesn't have the concept of unscoped task creation like asyncio does.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -400,7 +400,6 @@ class WebSocketTestSession:
 class TestClient(requests.Session):
     __test__ = False  # For pytest to not discover this up.
     task: "Future[None]"
-    portal: typing.Optional[anyio.abc.BlockingPortal] = None
 
     def __init__(
         self,
@@ -410,6 +409,7 @@ class TestClient(requests.Session):
         root_path: str = "",
         backend: str = "asyncio",
         backend_options: typing.Optional[typing.Dict[str, typing.Any]] = None,
+        portal: typing.Optional[anyio.abc.BlockingPortal] = None,
     ) -> None:
         super().__init__()
         self.async_backend = _AsyncBackend(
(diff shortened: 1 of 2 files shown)


aminalaee:
@MatthewScholefield Thank you for this, will you be able to add the required test?

MatthewScholefield:
Hi @aminalaee thanks for the ping. I've updated the test and it should now have 100% coverage.

lovelydinosaur:
> when something attaches to the event loop on startup and then is used within a request

It sounds to me like *this* is the actual issue here. `asyncio` exposes some primitives that you *really* don't want to be using.

I'd be up for working carefully through the issue, but I don't think `portal=...` is something we want to be exposing as public API here.

PR #1350, opened 2021-12-03
Add staticfiles packages with directory

### LINKED ISSUE #1265: StaticFiles support for directories other than "statics"
### Checklist

- [/] There are no similar issues or pull requests for this yet.
- [X - tried to get feedback but no replies] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

I want to be able to serve static files from arbitrary locations within a python package but I can't because currently it looks in a `statics` folder which is hardcoded.

I'd like to be able to pass as a parameter the path within the python package to look in for the static files.

I've considered changing the location of my packaged files to `statics` but this will have knock on effects across other systems, and `statics` itself is non-standard as far as I can tell.

aminalaee:
Closes #1265.

Adds ability to specify directory when using the `packages` argument in StaticFiles.

The final API is:

```python
routes=[
    ...
    Mount('/static', app=StaticFiles(packages=['bootstrap4', ('mypackage', 'dist')]), name="static"),
]
```

This will look for "statics" directory in "bootstrap4" package and "dist" directory in "mypackage".

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -40,7 +40,7 @@ class StaticFiles:
         self,
         *,
         directory: PathLike = None,
-        packages: typing.List[str] = None,
+        packages: typing.List[typing.Union[str, typing.Tuple[str, str]]] = None,
         html: bool = False,
         check_dir: bool = True,
     ) -> None:
@@ -49,41 +49,36 @@ class StaticFiles:
         self.all_directories = self.get_directories(directory, packages)
         self.html = html
         self.config_checked = False
-        if (
-            check_dir
-            and directory is not None
-            and not packages
-            and not os.path.isdir(directory)
-        ):
+        if check_dir and directory is not None and not os.path.isdir(directory):
             raise RuntimeError(f"Directory '{directory}' does not exist")
 
     def get_directories(
-        self, directory: PathLike = None, packages: typing.List[str] = None
+        self,
+        directory: PathLike = None,
+        packages: typing.List[typing.Union[str, typing.Tuple[str, str]]] = None,
     ) -> typing.List[PathLike]:
         """
         Given `directory` and `packages` arguments, return a list of all the
         directories that should be used for serving static files from.
-
-        It can take a single directory,
-        or a list of `packages` and look for default `statics` in each package,
-        or a list of `packages` with a `directory` to look for in each package.
         """
         directories = []
         if directory is not None:
             directories.append(directory)
-        else:
-            directory = "statics"
 
         for package in packages or []:
+            if isinstance(package, tuple):
+                package, statics_dir = package
+            else:
+                statics_dir = "statics"
             spec = importlib.util.find_spec(package)
             assert spec is not None, f"Package {package!r} could not be found."
             assert spec.origin is not None, f"Package {package!r} could not be found."
             package_directory = os.path.normpath(
-                os.path.join(spec.origin, "..", directory)
+                os.path.join(spec.origin, "..", statics_dir)
             )
             assert os.path.isdir(
                 package_directory
-            ), f"Directory {directory!r} in package {package!r} could not be found."
+            ), f"Directory '{statics_dir!r}' in package {package!r} could not be found."
             directories.append(package_directory)
 
         return directories
@@ -197,7 +192,7 @@ class StaticFiles:
         pointed at a directory, so that we can raise loud errors rather than
         just returning 404 responses.
         """
-        if self.directory is None or self.packages:
+        if self.directory is None:
             return
 
         try:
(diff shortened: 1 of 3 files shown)


florimondmanca:
Hi @aminalaee 

As I understand it, with this PR `StaticFiles(packages=["external"], directory="static")` would look for static assets in `/path/to/external/static`, correct?

But I don't think that's what we want. Currently, the behavior is that `StaticFiles(directory=..., packages=...)` adds `directory` _and_ one directory per package — we can see from from how `directories` is populated in the function at hand here.

I'm wondering if we should be considering this API instead:

```python
StaticFiles(
    directory=...,  # An independent directory that's unrelated to `packages` (as currently)
    packages=[
        "bootstrap4",  # Default, resolves to bootstrap4/statics
        ("whatevs", "dist"),  # Explicit subdir, resolves to "whatevs/dist"
    ],
)
```

And then do:

```python
for package in packages or []:
    try:
        package, statics_directory = package
    except ValueError:
        statics_directory = "statics"

    ...
```

aminalaee:
@florimondmanca  Yes you are right. I have mixed `directory` and `packages` and the API will be:

```python
StaticFiles(
    directory="statics",
    packages=["bootstrap"],
)
```
And this will look for "statics" directory inside the "bootstrap" package.

There's a point we didn't take into account, we don't allow sub directories.
Maybe we should consider this too, `directory` is independant as you said:

```python
StaticFiles(packages=["bootstrap.dir.statics"])
```

I'm happy to change this PR if we have a conclusion.

florimondmanca:
@aminalaee My API suggestion is on the same line than your second suggestion, yes — allowing to specify a per-package target sub-directory. I'd say let's do a single path item and avoid the problem of dealing with OS-dependent path representation for now, which is why I suggested a "string or 2-tuple" representation. WDYT?

aminalaee:
@florimondmanca I was thinking to handle the sub-directories in an OS-independant way before another issue comes up mentioning that.

I think that makes sense for now, I'll make `directory` independent and go with the tuple option.

starlette/staticfiles.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         self,
         *,
         directory: PathLike = None,
-        packages: typing.List[str] = None,
+        packages: typing.List[typing.Union[str, typing.Tuple[str, str]]] = None,
```
aminalaee, line 43: I tried using `@typing.overload` for here, but that doesn't seem to work on the initialisers.

starlette/staticfiles.py
```diff
@@ -64,17 +66,19 @@ def get_directories(
...
+            if isinstance(package, tuple):
+                package, statics_dir = package
+            else:
+                statics_dir = "statics"
             spec = importlib.util.find_spec(package)
             assert spec is not None, f"Package {package!r} could not be found."
-            assert (
-                spec.origin is not None
-            ), f"Directory 'statics' in package {package!r} could not be found."
+            assert spec.origin is not None, f"Package {package!r} could not be found."
```
aminalaee, line 75: I think this wasn't accurate before.

starlette/staticfiles.py
florimondmanca: We might be mixing 1-string and 2-tuple items in the same list, so I think this type definition is the most appropriate. :+1:

florimondmanca (approved):
Superb 💯

florimondmanca:
I left a comment on #1265 in case anyone has practical feedback on this, before we merge and include this in the next Starlette release.

paulmsmith:
@florimondmanca That looks like it would solve the issue myself and @bram2000 were having in [#1265](https://github.com/encode/starlette/issues/1265) 👍🏻

wanghaisheng:
@florimondmanca 
image

I have confused 
app.mount("/static", StaticFiles(directory="./static"), name="static")


python main.py is ok, when I bundle py to a exe., I even mapping a static folder in the root folder, how to specify this directory value is the best 

image

wanghaisheng:
if. I use this. ,still not found 

  File "/Users/wenke/github/tiktoka-studio-uploader-genius/src/app/fastapiserver.py", line 31, in <module>
    app.mount("/static", StaticFiles(directory="static"), name="static")
  File "/Users/wenke/github/tiktoka-studio-uploader-genius/.venv/lib/python3.9/site-packages/starlette/staticfiles.py", line 57, in __init__
    raise RuntimeError(f"Directory '{directory}' does not exist")
RuntimeError: Directory 'static' does not exist

aminalaee:
If I understand correctly you are shipping your package as an installable, so you would need to use the `StaticFiles(packages=["name_of_my_package"])` here.,

wanghaisheng:
if i understand correctly, there is 2 options for me here, 
option 1 :treat static as a folder in my own package,

src/app/static


you should use this, 

app.mount("/static", StaticFiles(packages=[('src.app','static')]), name="static")


the other option  is in the root folder of project.  
/static 

app.mount("/static", StaticFiles(directory="static"), name="static")



it seems  you can have directory and packages both there 

app.mount("/static", StaticFiles(directory="static",packages=[('src.app','static')]), name="static")

PR #1351, opened 2021-12-05
Share session data across mounted routes

### LINKED ISSUE #1261: Starlette 0.15.0 breaks SessionMiddleware by adding ASGI Path for Subroutes using Mount
### Checklist

### Describe the bug

As discussed with @mahmoudhossam on #1147, I am creating this issue.

Update 0.15.0 broke my session checking function to ensure users are logged in. I mount all the sub routes (users, etc..). The request.session has data in it when set, but when I redirect back to the start page then session is now null. I see no documentation on how to have the session across Mounts?

```
# Routes
routes = [
    Mount("/user", routes=
    [
        Route(
        ),
        Route(
            "/password-change",
            methods=["GET", "POST"],
        ),
    ],
    name="user",
    ),

# Session Checking
            logger.error(
            )

        else:
            )
            )

            if current == False:
                logger.error(
                )

            # update datetime of last use
            logger.info(
            )

    return check_login
```
### To reproduce

Create basic app, using greater than 0.14.2, using Mount add subroutes and then check the session cookies for the subroutes as there will be a new session cookie as it it now acting like a submounted app, instead of subroutes.
### Expected behavior

Mounts should not create a new session cookie, unless Mounting a sub application. See [Routing](https://www.starlette.io/routing/) and using Mounts in the documentation. It shows you need to add "app=" for the submounted app. If left out/none, it is just a submounted route and should not be treated as an submounted app.

Mount("/static", app=StaticFiles(directory="static"), name="static")

### Actual behavior

Using Mount it will create new session cookies.

### Environment

- OS:  Windows/Linux
- Python version:  3.8+
- Starlette version:  0.15.0 + created in release #1147 by issue #233
### Additional context

I would suggest checking if in Mount app=None as a way to determine whether to use ASGI path or "/".

florimondmanca:
Fixes #1261

The resolution here builds on @devsetgo's idea:

> I would suggest checking if in Mount app=None as a way to determine whether to use ASGI path or "/".

Indeed, if we `Mount(routes=...)`, the `routes` actually belong to the current app in the mount hierarchy, not to a fully separate app mounted on the sub-path, so session data on the current app should be shared with them.

Note that we otherwise keep the "use root_path for _(fully)_ sub-mounted apps" for #233.


devsetgo:
> Fixes #1261
> 
> The resolution here builds on @devsetgo's idea:
> 
> > I would suggest checking if in Mount app=None as a way to determine whether to use ASGI path or "/".
> 
> Indeed, if we `Mount(routes=...)`, the `routes` actually belong to the current app in the mount hierarchy, not to a fully separate app mounted on the sub-path, so session data on the current app should be shared with them.
> 
> Note that we otherwise keep the "use root_path for _(fully)_ sub-mounted apps" for #233.

This looks like a great solution... also, I now learned how to test the session cookie.

starlette/routing.py
```diff
@@ -339,6 +339,7 @@ def __init__(
             self.app: ASGIApp = app
         else:
             self.app = Router(routes=routes)
+        self._routes_given = routes is not None
```
Kludex: What's the purpose of `_routes_given`? :thinking:

florimondmanca: Oops, that's a remainder from when I was experimenting with how best to pass the "we were given `Mount(routes=...)`" to the sessions middleware. 😄 

I'm open to thoughts on whether `isinstance(self.app, Router)` fits that bill, though.

florimondmanca:
```suggestion
```

Kludex (dismissed):
Answering your comment above: in terms of styling, either the current or using the attribute you previously had, it's fine by me.

I would've used `session_path` as there's no concept of "session_root_path" yet, so we can choose the name. But whatever you prefer is fine. :)

florimondmanca:
Well hmm… Shouldn't we be using an even more explicit name, `session_cookie_path`?

Still, it's a bit odd to me that we are setting this info in the `Mount` component — this is introducing downward coupling. Ideally, the Session middleware (or any middleware) would pluck the info it needs, rather than `Mount` passing it, implying a session middleware should be present at all (oftentimes it might not, eg for APIs).

Thinking of relying on `"endpoint": self.app` in the child scope key instead. That should be unchanged through the middleware stack until we hit the route. Let's try…

Kludex:
Did you delete your comment? 🤔

florimondmanca:
@Kludex Yes 😅 I thought about an even more specific name, pushed it in a commit. Also I think we should use the attribute style, at least check on `routes` explicitly. I thought about what `Mount(..., app=Router(...))` should do and I _guess_ in that case we should definitely treat it as a separate thing, like we do for any `app=...`.

Kludex (approved):
LGTM! :)

Kludex:
@florimondmanca Is there something else you'd want to check here?

florimondmanca:
I _think_ this is an okay change — as in, it _does_ solve the use case. But the "special case" flavor of it still makes me think I'd like @tomchristie's opinion. Tom? 😄

0x1997:
Any updates?

lovelydinosaur:
Thanks @florimondmanca! Going to close this off in favour of #1512.

PR #1352, opened 2021-12-07
Make StaticFiles methods internal

aminalaee:
As far as I could see all `StaticFiles` methods should be internal, and shouldn't be used from outside.

This PR will change all existing methods to internal to reflect that.

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -86,21 +86,21 @@ class StaticFiles:
         assert scope["type"] == "http"
 
         if not self.config_checked:
-            await self.check_config()
+            await self._check_config()
             self.config_checked = True
 
-        path = self.get_path(scope)
-        response = await self.get_response(path, scope)
+        path = self._get_path(scope)
+        response = await self._get_response(path, scope)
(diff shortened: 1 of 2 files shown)


florimondmanca:
Dunno, it's fiddly. What if some people do rely on these methods today? The fact that our tests rely on these methods is interesting. Could we rearrange things to _not_ call these methods at all, meaning that low-level users _would_ have a way to not use these methods for common low-level use cases?

aminalaee:
@florimondmanca I think most of them are already private-ish requiring the request scope from `__call__`.
As for our own tests, it's already described why we can't use the test client in that test.

I can't think of any re-arrangements, I thought about modifiying `get_directories` as is so It's kept public, but I don't think it would be much useful as we can already access `staticfiles.all_directories`. Do you have anything in mind?

P.S. We can also mark them for deprecation so we can change it later.

lovelydinosaur:
I think I'd probably second @florimondmanca here. The pragmatic path here is probably just to leave things as they are.

aminalaee:
Cool, then I close this.

PR #1353, opened 2021-12-08
Add Mypy checks to tests

### LINKED ISSUE #1045: Have mypy check types on test code
As we do in other encode projects, we should run mypy against our test code to make sure the type annotations in the project work well when the project is actually used.

We do this not by adding type annotations to all the tests (as this is often not very useful for test code) but rather by using mypy's `check_untyped_defs` option. We usually have a mypy config like:

```ini
[mypy]
disallow_untyped_defs = True
ignore_missing_imports = True

[mypy-tests.*]
disallow_untyped_defs = False
check_untyped_defs = True
```

As of writing, enabling `check_untyped_defs` for the tests directory results in a lot of mypy errors which need to be fixed.

aminalaee:
Closes #1045.

Adds Mypy checks to the tests.

I'll add a few comments where it can be confusing to review this.


tests/test_datastructures.py
```diff
@@ -228,7 +226,7 @@ async def test_upload_file():
 
 
 def test_formdata():
-    upload = io.BytesIO(b"test")
+    upload = UploadFile("test")
     form = FormData([("a", "123"), ("a", "456"), ("b", upload)])
```
aminalaee, line 230: `FormData` needs a `str` or `UploadFile`.

tests/test_datastructures.py
```diff
@@ -339,7 +337,7 @@ def test_multidict():
     assert repr(q) == "MultiDict([('a', '123'), ('b', '456')])"
 
     q = MultiDict([("a", "123"), ("b", "456")])
-    q.update(None)
+    q.update(None)  # type: ignore
```
aminalaee: Not really sure what to do here. It was mentioned on the issue we might be able to remove this?

tests/test_routing.py
```diff
@@ -422,13 +422,13 @@ async def subdomain_app(scope, receive, send):
     await response(scope, receive, send)
 
 
-subdomain_app = Router(
+subdomain_router = Router(
```
aminalaee, line 462: The name clashes with the already defined `subdomain_app`.

tests/test_formparsers.py
```diff
@@ -36,16 +38,17 @@ async def app(scope, receive, send):
...
     request = Request(scope, receive)
     data = await request.form()
-    output = {}
+    output: typing.Dict[str, list] = {}
     for key, value in data.multi_items():
         if key not in output:
             output[key] = []
         if isinstance(value, UploadFile):
             content = await value.read()
+            content = content.decode() if isinstance(content, bytes) else content
```
aminalaee: Here `content` is a `Union[str,bytes]` and str won't have the decode method.

adriangb: Maybe #1312 is relevant?

aminalaee: Hmm..I think it might be. Do you think this can be a blocker?

adriangb: I think it would be nice to merge that first since you're making changes here to work around the problem. FYI I marked #1312 as breaking changes only because the use of str is in the public docs. I'm fairly certain there are zero people using it that way though, so really I don't think we should treat it as a break and change

That said this is just typing for tests, I don't think it's worth holding up if it's a lot of trouble.

tests/test_datastructures.py
florimondmanca: From https://github.com/encode/starlette/pull/345 it seems because `.update()` is allowed (see the default ` = None`). I'm actually surprised passing an explicit `None` value is a problem for mypy. But I _think_ the intended test case here was `q.update()`. So, maybe...?

```suggestion
    q.update()
```

aminalaee: I don't think so. It's not default `None`.
https://github.com/encode/starlette/blob/c8b9581e371968b4da42d0f2de01e4028d71ac3c/starlette/datastructures.py#L357-L365

florimondmanca: Oh that must have changed since #345 then (which _did_ default to `None`, hence `.update(None)` would have worked fine back then).

aminalaee: But yeah, I think the suggestion looks good.

florimondmanca: Looks like the change happened in https://github.com/encode/starlette/pull/355/files#diff-e939ae940ce35a0db1544bfea4ab853be43ba89ea33e3cac713aa0b19f4bb875L229-L233 which did not update the tests;

I think we can remove this little test case altogether, then?

aminalaee: Agreed.

tests/test_formparsers.py
aminalaee: Now that #1312 is merged, can you also review this? @adriangb @tomchristie

tests/middleware/test_cors.py
```diff
@@ -258,7 +258,7 @@ def test_cors_allow_all_methods(test_client_factory):
     )
 
     @app.route(
-        "/", methods=("delete", "get", "head", "options", "patch", "post", "put")
+        "/", methods=["delete", "get", "head", "options", "patch", "post", "put"]
```
adriangb, line 261: The fact that we had to edit this makes me think: should we relax the annotation to `typing.Sequence` and then cast it to a list if needed?

adriangb (approved):

tests/middleware/test_cors.py
aminalaee: Thanks for reviewing this.
Good point. I agree.
Let's wait for more feedback then.

aminalaee: Oh let me take that back. It will be out of scope for this PR. Let's raise it another one.

adriangb: Yep agreed that it's outside of the scope of this PR, it'd be nice to keep this to tests only

PR #1354, opened 2021-12-09
Add request-id middleware to third-party-middleware

sondrelg:
Hi!

We recently wrote a request-id middleware for our project at work, as we were not able to find another well tested and maintained single-purpose project. Was hoping to add the middleware to the list of third party middleware in the hope that it might benefit others.

Also took the liberty of sorting the middleware already in the list. If you prefer an unordered list, let me know and I can revert it back to its original state and append the new entry to the end of the list.

Thanks!


Kludex:
Have you seen [starlette-context](https://github.com/tomwojcik/starlette-context)? How does it differs from it?

sondrelg:
Yes! [starlette-context](https://github.com/tomwojcik/starlette-context) admittedly does much of the same things, and should probably be in this list as well. I'd say they overlap a little bit, but not completely. Personally I would categorize starlette-context as more of a general purpose context var middleware.

Here's the list of all its built-in plugins (which I'm sure you're aware of)

image

This is great, and _extremely_ flexible.

We actually initially adopted starlette-context, but found that it was missing a few components to get us started with request-id logging. In addition to setting up the middleware we **also** needed to roll our own log-filters and write code to integrate with celery and sentry. We ended up with a few hundred lines of supplementary code, which we'd need to duplicate for all our projects.

For our Django projects we use [django-guid](https://github.com/snok/django-guid) which has these things; so we decided to repurpose [django-guid](https://github.com/snok/django-guid) for ASGI.

tl;dr, the primary differences are:
- The middleware has a much narrower scope (does one thing),
- It comes with log filters included - users don't need to write any code to get started, 
- Documentation around request-IDs is a little bit more comprehensive, and focused on solving this single issue,
- The package has some extra request-id-specific support for sentry & celery (and I hope to add a mechanism for request ID transfer w/ arq soon)

Definitely understand why you bring it up though. Functionally they are very similar.

euri10:
> (and I hope to add a mechanism for request ID transfer w/ arq soon)

is there an issue that tracks it ? would be very interested in using it once done or at least follow the progress, will test it asap anyways without ! Looks great, thanks

sondrelg:
I opened an issue [here](https://github.com/snok/asgi-correlation-id/issues/17) now. If you or anyone else has ideas about how to associate request-ids with arq jobs in an async context, please feel free to chime in 👏

Kludex:
@sondrelg Thanks for the clarifications!

Kludex:
> Also took the liberty of sorting the middleware already in the list. If you prefer an unordered list, let me know and I can revert it back to its original state and append the new entry to the end of the list.

I think this is fine. Although we could have something "smarter", like running a script and inserting the middlewares in order, or writing on the section title that is alphabetically sorted so next PRs are aware of this, and we avoid resorting. I don't know...

Actually, is it sorted from older->new? :thinking:

sondrelg:
Yeah I think the list has been appended to the bottom for each addition, so sorted old -> new by PR date.

euri10 (approved):

Kludex:
Thanks :+1: 

I didn't merge it right away because of the sorting... :sweat_smile:

PR #1357, opened 2021-12-11
Remove Python 3.6

Kludex:
I don't think maintaining Python 3.6 is a burden for Starlette (I have a different opinion on Uvicorn tho). 

If we remove it on Uvicorn, should we also do it on Starlette?

This PR is to allow discussion having the changes that it implies in the same place. :)

Reference: https://pypistats.org/packages/starlette

Updated 14/01/22: `anyio` [3.4.0](https://anyio.readthedocs.io/en/stable/versionhistory.html) supports a feature to 3.7+.
Updated 27/03/22: `setuptools` and `jinja2` doesn't support Python 3.6 anymore.

Note to future me: remove `jinja2` import conditional about `context` function.


setup.py
```diff
@@ -4,7 +4,7 @@
 import os
 import re
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
```
Kludex: ups! :sweat_smile: 

```suggestion
from setuptools import setup, find_packages
```

Kludex:
Is that a visual bug on the pipeline? Python 3.6 is not running.

aminalaee:
> Is that a visual bug on the pipeline? Python 3.6 is not running.

Isn't it expected since you removed it in your branch? This check won't run in your PR.
I guess it's the same when you add a new Matrix, in that PR the new matrix won't be required since the changes aren't in master.

Kludex:
Ok... It makes sense... How do people merge PRs like this one, then? :thinking:

aminalaee:
> Ok... It makes sense... How do people merge PRs like this one, then? 🤔

As far as I remember the pipeline check won't be required, the approval will be enough. Might need double checking...

Kludex:
> > Ok... It makes sense... How do people merge PRs like this one, then? thinking
> 
> As far as I remember the pipeline check won't be required, the approval will be enough. Might need double checking...

Don't worry. I just open this for discussion initially. We'll see later. :)

aminalaee (dismissed):

Kludex:
Only @tomchristie can merge this, as the pipeline will never pass.

@tomchristie The reference about downloads per user is in the description of this PR, jfyk.

lovelydinosaur:
I've updated the settings so that 3.7-3.10 are required.

I don't have any strong feelings on this one. I suppose the sensible thing is mostly just that we update our packages consistently when we go for this. (Not for ay hard reasons, just because at least they're all on the same page then.)

What is FastAPI's current version support?

Do we have an outstanding PR for this change in uvicorn too?

aminalaee:
Sorry to jump in here, I don't know about FastAPI's plans for python 3.6 but this pypistats might be worth _considering_:
Screenshot 2022-01-10 at 10 46 20

Update: I'm not sure how accurate the numbers are but ignoring the recent peaks, there a few thousand downloads of Python3.6 happening every day.

lovelydinosaur:
Link to those FastAPI stats... https://pypistats.org/packages/fastapi

Kludex:
I've talked to @tiangolo about this, and we reached the conclusion that we can wait some time before losing support for Python 3.6, so people can migrate to a new Python version.

We can reevaluate the numbers some time from now.

tiangolo:
Yep! As @Kludex says, if possible, and if it's not being cumbersome to maintain, I would hold this for some time. :pray:

I assume there might already be "old" apps that can't migrate to 3.7 right away. If it gets annoying/cumbersome then yeah. But keeping support for 3.6 for a while might be appreciated by a few developers already stressed that their other libs don't support > 3.6 yet. :sweat_smile:

Kludex:
`anyio` 4.0 will drop Python 3.6, jfyk.

Kludex:
Trio dropped support for Python 3.6.

Kludex:
Not important on the dependency tree, but `coverage` dropped support as well.

Kludex:
`pytest` dropped support as well.

Kludex:
Daily downloads from PyPI Download Stats about Starlette:

```
3.8: ~95k
3.7: ~78k
3.9: ~44k
3.10: ~16k
3.6: ~10k
```
3.18% of our users use Python 3.6.

Ref.: https://pypistats.org/packages/starlette (date of the info above is 29 March 2022)

Based on this, I'd like to go with this idea.

euri10 (approved):

br3ndonland (approved):
Thank you for the thoughtful, data-driven approach to this update. :chart_with_upwards_trend: :eyes: :thumbsup:

adriangb (approved):

README.md
```diff
@@ -41,7 +41,7 @@ It is production-ready, and gives you the following:
 
 ## Requirements
 
-Python 3.6+
+Python 3.7+ (For Python 3.6 support, install version 0.17.1)
```
Kludex:
```suggestion
Python 3.7+ (For Python 3.6 support, install version 0.19.1)
```

docs/index.md
```diff
@@ -38,7 +38,7 @@ It is production-ready, and gives you the following:
 
 ## Requirements
 
-Python 3.6+
+Python 3.7+ (For Python 3.6 support, install version 0.17.1)
```
Kludex:
```suggestion
Python 3.7+ (For Python 3.6 support, install version 0.19.1)
```

PR #774, opened 2019-12-30
Fix config doc page examples (starlette==0.13.0)

elisong:
Fix two of doc examples (starlette==0.13.0):

- 'URL' (from  starlette.datastructures) object has no attribute 'database'.

- `from starlette.middleware.sessions ...` instead of `from starlette.middleware.session ...`

PR #1358, opened 2021-12-11
Fix snippet on config.md

Kludex:
Replaces #774

@aminalaee Could you take a look at this? I don't use neither `databases` nor the config module :sweat_smile:

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -130,7 +130,7 @@ application logic separated:
 ```python
 import databases
 from starlette.config import Config
-from starlette.datastructures import URL, Secret
+from starlette.datastructures import Secret
 
 config = Config(".env")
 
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -128,6 +128,7 @@ application logic separated:
 **myproject/settings.py**:
 
 ```python
+import databases
 from starlette.datastructures import URL, Secret
```
aminalaee:
```suggestion
from starlette.datastructures import Secret
```

aminalaee (changes requested):

docs/config.md
aminalaee: The URL is not used.

aminalaee (approved):
👍

Issue #1217, opened 2021-06-25
Use of asgiref typing


Kludex:
### Checklist

- [X] There are no similar issues or pull requests for this yet.
- [X] @JayH5 and @Kludex are positive about this idea, but it was not discussed on our Gitter channel.

### Is your feature related to a problem? Please describe.

On `uvicorn` we started a `mypy` compliance goal since some time ago, and we're about to accomplish it. On this road, we decided to use [asgiref typing](https://github.com/django/asgiref/blob/main/asgiref/typing.py) to annotate ASGI related types.

What I propose is that we do the same for `Starlette`. 

## Describe the solution you would like.

Remove the `types.py` module, and annotate everything that is needed with `asgiref.typing` types. 

There's no matching type to `Message`, as discussed on #1206 . So even if it would be more pleasant to our users, unfortunately it will be more troublesome to follow that path. 

## Describe alternatives you considered

Use `Message` as `Union[ASGISendCallable, ASGIReceiveCallable]` and solve the `mypy` issues. It will be more pleasant to ours users.

Kludex:
I guess the first step here would be to replace the `Message` type by two different message types, one for receive and another one for send.

aminalaee:
@Kludex if we can maybe split this into multiple part, I can help with it too.

Kludex:
The problem is that I'm not 100% sure, but I think what makes the issue hard to solve is the `Message` type. You'll see the mypy errors if you experiment some changes locally.

That being said... I believe the approach would be:
1. Split `Message` into `ASGISendEvent` and `ASGIReceiveEvent` - naming matching asgiref.
2. Rename the other types to asgiref (just renaming, not introduce the package yet).
3. Remove the types definition and import from asgiref - this should be straightforward based on step 2.

br3ndonland:
> The problem is that I'm not 100% sure, but I think what makes the issue hard to solve is the `Message` type. You'll see the mypy errors if you experiment some changes locally.
>
> That being said... I believe the approach would be:
>
> 1. Split `Message` into `ASGISendEvent` and `ASGIReceiveEvent` - naming matching asgiref.
> 2. Rename the other types to asgiref (just renaming, not introduce the package yet).
> 3. Remove the types definition and import from asgiref - this should be straightforward based on step 2.

I helped with type-annotating Uvicorn, and I'm happy to help here as well.

[...]

PR #1360, opened 2021-12-14
Fix `exception_handlers` type hints

### LINKED ISSUE #1142: mypy complains about `exception_handlers` dict with `HTTPException` key
### Checklist

Related, but not the same:
* #1034 
* #1139 
* #949

### To reproduce

Given an example.py: 

```python

def handle_http_exception(request: Request, exc: HTTPException) -> Response:
    return Response('foo')


exception_handlers = {
    HTTPException: handle_http_exception
}
app = Starlette(exception_handlers=exception_handlers)
```

### Expected behavior

Mypy does not present an error.

### Actual behavior

```console
$ mypy example.py
example.py:14: error: Argument "exception_handlers" to "Starlette" has incompatible type "Dict[Type[HTTPException], Callable[[Request, HTTPException], Response]]"; expected "Optional[Dict[Union[int, Type[Exception]], Callable[..., Any]]]"
```

### Debugging material

None

### Environment

- OS: Debian (`Linux ... 4.19.0-14-amd64 #1 SMP Debian 4.19.171-2 (2021-01-30) x86_64 GNU/Linux`)
- Python version: `Python 3.9.1`
- Starlette version: `starlette==0.14.2`

### Additional context

For propriety:

https://github.com/encode/starlette/blob/master/docs/exceptions.md#L42

https://github.com/encode/starlette/blob/ed73b97c9b8f4aa527eed88032c04ce9fe109a3c/starlette/applications.py#L44

I think this is a Starlette issue and not a mypy issue; apologies if I'm wrong here.

aminalaee:
Closes #1142 and replaces #1139.

Changes `exception_handlers` type hints:

```python
typing.Mapping[
    typing.Any, typing.Callable[[Request, Exception], Response]
]
```

I also changed the callable type hints, I think it makes sense to be more explicit, this will catch user-defined callable mismatches (?).

As mentioned in #1139 we can't use `typing.Mapping[typing.Union[int, ..], ...]` since dict keys are not covariant.<del>I think we could use `typing.TypedDict` but only after 3.8</del>.


starlette/exceptions.py
```diff
@@ -22,7 +22,12 @@ def __repr__(self) -> str:
 
 class ExceptionMiddleware:
     def __init__(
-        self, app: ASGIApp, handlers: dict = None, debug: bool = False
+        self,
+        app: ASGIApp,
+        handlers: typing.Mapping[
+            typing.Any, typing.Callable[[Request, Exception], Response]
+        ] = None,
```
k4black, line 29: It's quite a complicated type, should we move it to `starlette.types`? Like `ExceptionHandler` type

k4black:
Speaking about `typing.TypedDict` you can add python version check as a nice way to show that you'll migrate to dict when 3.7 will be deprecated

starlette/exceptions.py
Kludex: The current types on `types.py` are going to be removed at some point: https://github.com/encode/starlette/issues/1217

That file will only contain this type at some point. Is that a problem? :thinking:

k4black: Didn't notice the Issue,  
This mapping type just seems a bit complicated to me to be copied multiple times

aminalaee: One drawback I noticed in VScode, haven't really checked and there might be some workarounds, is that they don't show function signatures when using the aliases as you mention. Not sure if it happens on PyCharm and other tools.

Kludex (approved):

antonagestam:
@aminalaee @k4black FYI, I've opened a PR that improves on the typing added in this PR: https://github.com/encode/starlette/pull/1456

PR #1361, opened 2021-12-14
Additional headers for WS `accept` message.

matiuszka:
Hello,

Here is a proposal for improvement about supporting additional headers for the WebSocket `accept` messages.
As per disscuision in gitter:

> Hello,
I recently faced the problem with sending custom headers in WS accept messages. According to ASGI, it should be possible:
https://asgi.readthedocs.io/en/latest/specs/www.html#accept-send-event
I've created an issue and MR for Hypercorn to support that:
pgjones/hypercorn!66
although I am forced to use bare send for this.
Do you think that it can be supported by Starlette's API? If yes, let me know I can create PR.
Here is the proposal: matiuszka/starlette@a080bf5


Kludex:
It would be cool to have this supported by unicorn as well (although is not a prerequisite).

matiuszka:
@Kludex I will handle it soon.

matiuszka:
Here is PR for `uvicorn`: https://github.com/encode/uvicorn/pull/1293

tests/test_websockets.py
```diff
@@ -301,6 +301,20 @@ async def asgi(receive, send):
...
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=send)
+            await websocket.accept(headers=[(b"additional", b"header")])
+            await websocket.close()
+
+        return asgi
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/") as websocket:
+        websocket.additional_headers = [(b"additional", b"header")]
```
Kludex:
```suggestion
        assert websocket.additional_headers == [(b"additional", b"header")]
```

Kludex (changes requested):
Just a single comment. Thanks @matiuszka ! :smile:

starlette/testclient.py
```diff
@@ -298,6 +298,7 @@ def __init__(
         self.app = app
         self.scope = scope
         self.accepted_subprotocol = None
+        self.additional_headers = None
```
Kludex: Can we use `extra_headers` instead of `addional_headers` to match uvicorn?

docs/websockets.md
```diff
@@ -1,4 +1,3 @@
-
```
Kludex: Would you mind readding this line?

Kludex (changes requested):
Besides that, everything is cool!

Thanks @matiuszka ! :smile:

tests/test_websockets.py
```diff
@@ -301,6 +301,20 @@ async def asgi(receive, send):
...
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=send)
+            await websocket.accept(headers=[(b"additional", b"header")])
+            await websocket.close()
+
+        return asgi
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/") as websocket:
+        assert websocket.additional_headers == [(b"additional", b"header")]
```
Kludex:
```suggestion
        assert websocket.extra_headers == [(b"additional", b"header")]
```

Kludex (approved):

Kludex:
Thanks @matiuszka ! :smile:

starlette/testclient.py
```diff
@@ -315,6 +316,7 @@ def __enter__(self) -> "WebSocketTestSession":
             self.exit_stack.close()
             raise
         self.accepted_subprotocol = message.get("subprotocol", None)
+        self.extra_headers = message.get("headers", None)
```
Kludex, line 319: This should probably be `message.get("headers", [])` with #1422

aminalaee: I think this part is ok. When passing `None` you'll need to check `assert websocket.extra_headers is None` but with that change:

```python
def test_additional_headers(test_client_factory):
     def app(scope):
             await websocket.accept(headers=None)
             await websocket.close()

         return asgi

         assert websocket.extra_headers == []
```

Am I missing something?

Kludex: You're not. I don't have strong feelings on this. :+1:

PR #1363, opened 2021-12-14
Display missing route details.

alex-oleshkevich:
This PR improves error reporting for `NoMatchFound` exception.
image


Kludex:
For the record, this is what we have before this PR:

image

starlette/routing.py
```diff
@@ -31,6 +31,11 @@ class NoMatchFound(Exception):
     if no matching route exists.
     """
 
+    def __init__(self, name: str, path_params: typing.Dict[str, typing.Any]) -> None:
+        super().__init__(
+            f'No route exists for name "{name}" and params "{path_params}".'
```
Kludex: Should we be concerned about logging the `path_params` **values** here, and actually just log the **keys**?

alex-oleshkevich: I see it as a local dev tool so then more context it gives to you then it is easier to debug.

Kludex: My question is more related to the fact that `NoMatchFound` will log in stdout, even with `debug=False`.

alex-oleshkevich: makes sense. do you suggest showing only keys here?

Kludex: Yes, please.

lovelydinosaur:
This is a clear win, yup!

Do we think we need to be concerned about folks raising `NoMatchFound()` directly in their own codebases, and continue to support but deprecate that case? (I don't think we document it anywhere, so arguably it's private-enough-ish not to be too concerned about this?)

lovelydinosaur (approved):

PR #1364, opened 2021-12-14
Allow 'name' in keyworded args for url_for() and url_path_for()

### LINKED ISSUE #608: "url_for" signature prevents use of "name" in path arguments

Kludex:
- Replaces #611
- Closes #608

- Should I wrap into a `try..except` with a more specific error?

#### Missing
- [ ] Test
- [ ] Update docs

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -104,8 +104,8 @@ class Starlette:
         self._debug = value
         self.middleware_stack = self.build_middleware_stack()
 
-    def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
-        return self.router.url_path_for(name, **path_params)
+    def url_path_for(self, *args: str, **path_params: typing.Any) -> URLPath:
+        return self.router.url_path_for(*args, **path_params)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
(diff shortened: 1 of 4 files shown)


dansan:
I finally got to rebasing, so this is now a duplicate of https://github.com/encode/starlette/pull/611 ?

Kludex:
Let's continue there :)

Thanks!

PR #1366, opened 2021-12-15
fix: pass usedforsecurity to hashlib.md5 to fix error on FIPS-enabled systems

### LINKED ISSUE #1365: ETag checksum fails on FIPS-enabled systems when using MD5
### Checklist

### Describe the bug

The ETag checksum fails when using MD5. This is causing Starlette to not work at all under Red Hat Enterprise Linux when FIPS mode is enabled.

### Debugging material

Here's the exception that's thrown:

```
INFO:     10.42.1.7:34422 - "GET /app/static/foo.html HTTP/1.1" 500 Internal Server Error
  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 373, in run_asgi
  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 656, in __call__
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 408, in handle
  File "/usr/local/lib/python3.8/site-packages/starlette/staticfiles.py", line 97, in __call__
    response = await self.get_response(path, scope)
  File "/usr/local/lib/python3.8/site-packages/starlette/staticfiles.py", line 118, in get_response
    return self.file_response(full_path, stat_result, scope)
  File "/usr/local/lib/python3.8/site-packages/starlette/staticfiles.py", line 173, in file_response
    response = FileResponse(
  File "/usr/local/lib/python3.8/site-packages/starlette/responses.py", line 267, in __init__
  File "/usr/local/lib/python3.8/site-packages/starlette/responses.py", line 273, in set_stat_headers
    etag = hashlib.md5(etag_base.encode()).hexdigest()
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

### Environment

- OS: Red Hat Enterprise Linux 8 in FIPS mode
- Python version: 3.8.8
- Starlette version: 0.16.0

adriangb:
Closes #1365


starlette/_crypto.py
```diff
@@ -0,0 +1,14 @@
+import hashlib
+import inspect
+
+
+# TODO: Remove when dropping support for PY38. inspect.signature() is used to
+# detect whether the usedforsecurity argument is available as this fix may also
+# have been applied by downstream package maintainers to other versions in
+# their repositories.
+if "usedforsecurity" in inspect.signature(hashlib.md5).parameters:
```
adriangb: I decided to just check for the parameter directly here. Django has a bunch of utility functions around this, but we're only doing it once here so that seemed like overkill.

starlette/_crypto.py
```diff
@@ -0,0 +1,14 @@
...
+    def md5_hexdigest(data: bytes, *, usedforsecurity: bool = True) -> str:
+        return hashlib.md5(data, usedforsecurity=usedforsecurity).hexdigest()
```
adriangb: I decided to return the hexdigest directly because:
1. That's the only way Starlette is currently using it
2. It's easier to type the signature
3. It's a private implementation so we can easily change it in the future

starlette/responses.py
```diff
@@ -287,7 +287,7 @@ def set_stat_headers(self, stat_result: os.stat_result) -> None:
         content_length = str(stat_result.st_size)
         last_modified = formatdate(stat_result.st_mtime, usegmt=True)
         etag_base = str(stat_result.st_mtime) + "-" + str(stat_result.st_size)
-        etag = hashlib.md5(etag_base.encode()).hexdigest()
+        etag = md5_hexdigest(etag_base.encode(), usedforsecurity=False)
```
adriangb, line 290: Passing in the `usedforsecurity=False` so that it is obvious at the call site that we are _not_ using md5 for security

starlette/_crypto.py
```diff
@@ -0,0 +1,14 @@
+import hashlib
+import inspect
+
+
+# TODO: Remove when dropping support for PY38. inspect.signature() is used to
+# detect whether the usedforsecurity argument is available as this fix may also
+# have been applied by downstream package maintainers to other versions in
+# their repositories.
+if "usedforsecurity" in inspect.signature(hashlib.md5).parameters:
+    def md5_hexdigest(data: bytes, *, usedforsecurity: bool = True) -> str:
```
adriangb: Do we straight up want to raise an error if `usedforsecurity=True` in either of these implementations? Seems like overkill but it'd also be easy to implement.

starlette/responses.py
```diff
@@ -12,6 +11,7 @@
 
 import anyio
 
+from starlette._crypto import md5_hexdigest
```
adriangb: I made this a "private" module. I know there isn't much history of this in Starlette, but I figure this is as good as any other use case to start with.

adriangb (commented):
One thing I wasn't sure about: do we want a test that monkey patches `hashlib` to verify that we are passing `usedforsecurity=False`? Seems like overkill to me

starlette/_crypto.py
Kludex: It's fine.

starlette/responses.py
Kludex: I think it's fine. Although I think we should use `_compat.py` to be aligned with `httpx` and `httpcore`

adriangb: done. I just thought `_crypto.py` might get us some clout / stars

starlette/_crypto.py
Kludex: Just for my curiosity, is the `usedforsecurity` based exclusively on the Python version? Is there an environment with Python 3.8+ that will not have `usedforsecurity` parameter?

adriangb: I think the reason Django did that is that some enterprises might have backported the fix to an internal fork, or something like that. Figured we might as well follow what they did. But I'm happy to just check `sys.version_info`.

Since this is done once at import time, I don't think performance matters.

Kludex: I don't see how a _backport_ by checking the versions would be different from checking the parameter :thinking:
It looks like both are interchangeable, or am I missing something? Because if I'm not missing anything, then we could avoid the comment and just check the version :thinking:

starlette/responses.py
Kludex: Makes sense :+1:

starlette/_crypto.py
adriangb: I think the point is that the parameter _could_ be present in Python 3.7 if someone backported it internally. Invoking @vakwetu (author of the Django PR) in case they can clarify.

starlette/_crypto.py
Kludex: Why would we want to do that? On which scenario?

adriangb: We currently don't have any scenarios in which we'd want to pass `usedforsecurity=True` (and arguably md5 should never be used for security). Adding an error would just make it very explicit and add a barrier to misusing md5 for security in future changes.

adriangb: I think there are 2 scenarios this can help with:
1. Keeps us from shooting ourselves in the foot in the future once this PR is long forgotten.
1. Keeps users from misusing this function down the road (the module is private by name, but this is Python so someone could still use it).

starlette/_crypto.py
vakwetu: Yes, some distros have in fact backported this parameter.  Red Hat/Fedora/Centos has backported this to python 3.6 IIRC.

adriangb: Awesome, thank you for confirming!

starlette/_crypto.py
lovelydinosaur: No, let's just make the minimum possible change here.

starlette/_compat.py
```diff
@@ -0,0 +1,22 @@
+import hashlib
+
+# TODO: Remove when dropping support for PY38. a try/except is used to
+# detect whether the usedforsecurity argument is available as this fix may also
+# have been applied by downstream package maintainers to other versions in
+# their repositories.
```
lovelydinosaur: I think this pull request is just about right, yup! 👍

I think we could probably tweak this comment to be a bit more informative. Perhaps something like this?...

```python
# Compat wrapper to always include the `usedforsecurity=...` flag, which is only added from Python 3.9 onwards.
# We use this flag to indicate that we use `md5` hashes only for non-security cases (our ETag checksums).
# If we don't indicate that we're using MD5 for non-security related reasons, then attempting to use
# this function will raise an error when used environments which enable a strict "FIPs mode".
#
# See issue: https://github.com/encode/starlette/issues/1365
```

adriangb: incorporated, thanks!

lovelydinosaur (approved):

lovelydinosaur:
@adriangb Fantastic, thanks! I've invited you as a maintainer, so that you can merge this PR yourself.
You'll then be able to merge or approve other pull requests too. 👍

adriangb:
Worked perfectly, thank you Tom

PR #1368, opened 2021-12-15
test: use --show-error-codes for mypy

adriangb:
This is helpful in adding tactical `# type: ignore[error-code]`s


Kludex (approved):
We've done this on `uvicorn`, and it helps us to debug easily the mypy issues/add the "ignore code" properly.

LGTM! :tada: 

Thanks @adriangb !

PR #1372, opened 2021-12-16
Fix Mypy errors in routing and testclient

aminalaee:
Mypy 0.920 was released last night and causes failures in routing and test client.

https://mypy-lang.blogspot.com/2021/12/mypy-0920-released.html

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -151,7 +151,7 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
             for key, value in request.headers.items()
         ]
 
-        scope: typing.Dict[str, typing.Any] = {}
+        scope: typing.Dict[str, typing.Any]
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -151,6 +151,8 @@ def send(
             for key, value in request.headers.items()
         ]
 
+        scope: typing.Dict[str, typing.Any] = {}
```
Kludex: Does this work?
```suggestion
        scope: typing.Dict[str, typing.Any]
```

aminalaee: yup

Kludex (approved):
Thanks :)

Kludex:
We should still pin the dependencies and enable dependabot, like the other encode projects. (just as a note)

PR #1373, opened 2021-12-16
Pin requirements.txt dependencies

aminalaee:
Following [this](https://github.com/encode/starlette/pull/1372) PR, it'd make sense to pin dev dependencies to avoid this in the future.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,5 @@
 # Optionals
-itsdangerous==2.0.1
-jinja2==3.0.3
-python-multipart==0.0.5
-pyyaml==6.0
-requests==2.26.0
+-e .[full]
 
 # Testing
 autoflake==1.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,30 +1,30 @@
 # Optionals
-itsdangerous
-jinja2
-python-multipart
-pyyaml
-requests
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,30 +1,30 @@
...
-itsdangerous
-jinja2
-python-multipart
-pyyaml
-requests
+itsdangerous==2.0.1
+jinja2==3.0.3
+python-multipart==0.0.5
+pyyaml==6.0
+requests==2.26.0
```
Kludex: I guess we should do the same as this: https://github.com/encode/httpx/pull/1721#discussion_r660772826

Maybe we can use `-e .[full]`?

aminalaee: Oh right, we already have those.

Kludex (approved):
Thanks! :)

For the record, this follows the same strategy as `httpx` and `uvicorn`.
For the record2, dependabot is not in place yet.

aminalaee:
maybe @tomchristie can enable dependbot here too.

Kludex:
Just to have a background: on httpx and httpcore we use dependabot.yml on .github, but it looks like uvicorn have the GitHub app enabled (as a PR popped up once there, but I can't see the configuration to confirm).

PR #1374, opened 2021-12-17
♻️ Use asgiref.typing instead of starlette.types

### LINKED ISSUE #1217: Use of asgiref typing

br3ndonland:
## Description

Resolves encode/starlette#1217

This PR will replace usage of the custom Starlette types with standard ASGI types from the [asgiref](https://github.com/django/asgiref) package.

## Changes

Changes are based on the suggestions from @Kludex in https://github.com/encode/starlette/issues/1217#issuecomment-991688872.

- [x] 1. **Refactor `Message` type usage to `ASGISendEvent` and `ASGIReceiveEvent` to match asgiref**
- [x] 2. **Refactor usage of the other types from `starlette.types` to match asgiref**
  - [x] `ASGIApp` -> `ASGI3Application` _([Starlette 0.12](https://www.starlette.io/release-notes/#0120) upgraded to ASGI 3)_
  - [x] `Receive` -> `ASGIReceiveCallable`
  - [x] `Send` -> `ASGISendCallable`
  - [x] `Scope` -> `HTTPScope | LifespanScope | WebSocketScope`
- [x] 3. **Import types from asgiref**
  - [x] Add `asgiref>=3.4.0,<4` to `setup.py` _(version matches Uvicorn, but with upper bound added)_
  - [x] Update imports: `from starlette.types` -> `from asgiref.typing`
  - [x] Fix mypy errors after updating imports _(202 errors in 16 files)_
  - [x] Deprecate `starlette.types`
- [x] 4. **Fix test failures**
  - [x] `tests/test_endpoints.py`
  - [x] `tests/test_requests.py`
  - [x] `tests/middleware/test_cors.py`
  - [x] `tests/middleware/test_session.py`

### `Event` and `Scope` instances

Many of the type errors were resolved by using the asgiref `TypedDict` classes for event and scope instances, instead of plain dicts.

https://github.com/encode/starlette/blob/c3ba37c088feeca2aed18d72e39832f3038e5ce3/starlette/responses.py#L161-L170

### Type narrowing

There are some `assert` statements added for [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html).

https://github.com/encode/starlette/blob/c3ba37c088feeca2aed18d72e39832f3038e5ce3/starlette/endpoints.py#L116-L122

In these cases, the type checker thinks the object could be a broader type like `object`, and the `assert` helps ensure that the type is correct.

### `Scope` keys

There were a few `# type:ignore[index]` and `# type: ignore[typeddict-item]` comments needed when indexing into `Scope` using keys not present in the asgiref types:

- `scope["app"]`
- `scope["auth"]`
- `scope["path_params"]`
- `scope["router"]`
- `scope["session"]`
- `scope["state"]`
- `scope["user"]`

https://github.com/encode/starlette/blob/c3ba37c088feeca2aed18d72e39832f3038e5ce3/starlette/requests.py#L151-L156

In a future PR, we might consider setting up a `StarletteScope` datastructure that inherits from `TypedDict` or one of the asgiref `Scope` types to handle this.

## Related

- encode/uvicorn#1044


br3ndonland:
CI checks are passing, and the PR is now ready for review. Comments and suggestions welcome.

br3ndonland:
asgiref types added to tests

lovelydinosaur:
That's a well put together bit of work. Very neatly done. 😌

Having said that, I'm not sure about it yet.
To me it seems more obviously readable to me the current way around, and doesn't have the extra indirection.

Can someone describe what we're gaining by switching to `asgiref.typing`?

br3ndonland:
Thanks @tomchristie! :smile: It would be great to get some input from the other maintainers as well, as they're the ones who will be dealing with the types most often.

From my perspective, the use of `asgiref.typing` here is admittedly a bit of a mixed bag, but I think it's beneficial overall. The main benefit of `asgiref.typing` is not necessarily the names of the types, but the fields they contain.

Starlette's types are broad:

https://github.com/encode/starlette/blob/832bc856ec78abbd8b7d53f03c4c6f4d4dbaea37/starlette/types.py#L3-L4

In other words, `Scope` and `Message` are mappings, with unspecified keys, and values of `Any` type. It's not much different from typing those fields as just `dict`. It's left up to the developer to know which keys need to be in each dict, and type checking doesn't provide guarantees that they are passing data through the application correctly.

For example, when instantiating a `WebSocketEndpoint`:

https://github.com/encode/starlette/blob/832bc856ec78abbd8b7d53f03c4c6f4d4dbaea37/starlette/endpoints.py#L51-L59

We know `scope["type"]` should be `"websocket"`, but that's about it.

`asgiref.typing` datastructures are more specific. When using [`asgiref.typing.WebSocketScope`](https://github.com/django/asgiref/blob/3554e05def93e086690d19e9f38e593f1b5b7b61/asgiref/typing.py#L67-L80), we can see which keys should be in the dict, which types the values should be, and even the literal value needed for `scope["type"]`.

On the other hand, the greater specificity means type checking is more strict, hence the type narrowing and `type: ignore` comments added. And Starlette adds many fields to the scope, so as I suggested, I think we would still need some custom types here. It could be more effort to maintain, so I would understand if you didn't want to take on the additional maintenance burden.

Either way, I enjoyed a chance to work broadly across the Starlette code, and I would be happy to help again in the future.

Kludex:
It looks like we need to discuss this again, but I'm pretty much convinced to vendor `asgiref.typing` on `uvicorn`, and I think Starlette should follow the same path. But the review here would take some time.

Let's clean up the PRs we have around, and come back here after 0.19.1, and 0.20.0 are released. Check the [milestone](https://github.com/encode/starlette/milestones).

br3ndonland:
> It looks like we need to discuss this again, but I'm pretty much convinced to vendor `asgiref.typing` on uvicorn, and I think Starlette should follow the same path. But the review here would take some time.
>
> Let's clean up the PRs we have around, and come back here after 0.19.1, and 0.20.0 are released. Check the [milestone](https://github.com/encode/starlette/milestones).

Sounds good. I'll close this off for now.

Vendoring `asgiref.typing` would be reasonable. It might help resolve some of the compatibility issues, and we could also extend or modify the types for Starlette's scope extensions and other things (https://github.com/encode/starlette/discussions/1511). [Hypercorn basically does that](https://gitlab.com/pgjones/hypercorn/-/blob/main/src/hypercorn/typing.py).

We should make sure the other maintainers are on board though (https://github.com/encode/uvicorn/discussions/1209, https://github.com/encode/uvicorn/issues/1305#issuecomment-1036157361). I think there's still more discussion needed.

Kludex:
If you don't mind, I'd like to keep this open. 🙇 

There's a lot of work that was done here, and in case there's a positive feedback from the others, the only work here would be to change the imports, and vendorizing `asgiref.typing` - assuming no comments on the review.

Kludex:
I've been thinking about this...

I don't think that adding `asgiref` as a dependency would be a good thing for Starlette. I guess I'd go further than that, I'm not sure if there's a real benefit on adding a more ASGI type annotation. In any case, I think you were right @br3ndonland, we should discuss about it.

I'm going to close this, and turn the issue this PR refers to into a discussion. Let's see what others have to say - I'd also like to read your opinion.

PR #1377, opened 2021-12-17
Allow staticfiles to follow symlinks outside directory

### LINKED ISSUE #1083: StaticFiles middleware doesn't follow symlinks
### Checklist

### Describe the bug

The StaticFiles middleware is checking the `os.realpath` of a file and returning a 404 for symlinks that lead outside the static directory.

### To reproduce

1. create a minimal app with a staticfiles middleware
1. put a symlink in your static directory. the link's target must be above the static directory.
1. you'll get a 404

### Expected behavior

Support symlinks in static directory.

The use case for symlinks in static is to target frontend assets that are being generated in file-watch mode.

### Actual behavior

404.

### Debugging material

It's happening here:
https://github.com/encode/starlette/blob/b95acea973c20eea3e7cbbca42d09b1f5d4a3412/starlette/staticfiles.py#L147-L149

### Environment

- OS: linux
- Python version: 3.7.5
- Starlette version: 0.13.8

### Additional context

I'm happy to post a PR for this if useful, ideally adding a bool param to the StaticFiles middleware that allows symlinks.

aminalaee:
Fixes #1083.

Right now we don't allow StaticFiles to break outside of statics directory:

https://github.com/encode/starlette/blob/0aef1724cfafbe23f846979d427a5a173667f6b7/starlette/staticfiles.py#L162-L165

Since we already follow symlinks for directory and path, I think it makes sense to allow StaticFiles to break out of statics directory if it's following a symlink.


euri10:
isn't that behavior a security concern and would allow path traversal ?

aminalaee:
@euri10 For example in Nginx [docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#disable_symlinks)  this is the default behaviour it makes sense to break outside of the directory, only if following symlinks.
if I'm not mistaken that's also the case in Django, but I need to double check.

tests/test_staticfiles.py
```diff
@@ -441,3 +441,26 @@ def mock_timeout(*args, **kwargs):
     response = client.get("/example.txt")
     assert response.status_code == 500
     assert response.text == "Internal Server Error"
+
+
+def test_staticfiles_follows_symlinks_to_break_out_of_dir(tmpdir, test_client_factory):
```
euri10: what about we use the tmp_path fixture instead of tmpdir ? the former has a very similar api than the one pathlib provides and is imho more readable, opinions on that of course may vary so this is just a comment.

starlette/staticfiles.py
```diff
@@ -157,11 +157,13 @@ def lookup_path(
         self, path: str
     ) -> typing.Tuple[str, typing.Optional[os.stat_result]]:
         for directory in self.all_directories:
-            full_path = os.path.realpath(os.path.join(directory, path))
+            original_path = os.path.join(directory, path)
```
euri10: I haven't checked what the rest of the codebase uses, but I feel using the pathlib api is more modern and readable

euri10 (commented):
added a few comments mostly about os.path vs pathlib stuff, so not blocking on my side but I would see it as probably more readable

thoughts ?

starlette/staticfiles.py
aminalaee: Yeah makes sense. Just to be clear I would need to change a few other places in the staticfiles to use pathlib to be consistent, does that make sense?

starlette/staticfiles.py
```diff
@@ -154,20 +153,26 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
-            directory = os.path.realpath(directory)
-            if os.path.commonprefix([full_path, directory]) != directory:
+            original_path = Path(directory).joinpath(path)
+            full_path = original_path.resolve()
+            directory = Path(directory).resolve()
+            try:
+                full_path.relative_to(directory)
+                is_internal = True
+            except ValueError:
+                is_internal = False
```
aminalaee: Path provides `is_relative_to` method but only in 3.9 upwards [here](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to).

Checking the [source](https://github.com/python/cpython/blob/main/Lib/pathlib.py#L732) this is the implementation.

tests/test_staticfiles.py
```diff
@@ -166,8 +166,8 @@ def test_staticfiles_prevents_breaking_out_of_directory(tmpdir):
     directory = os.path.join(tmpdir, "foo")
     os.mkdir(directory)
 
-    path = os.path.join(tmpdir, "example.txt")
-    with open(path, "w") as file:
+    file_path = os.path.join(tmpdir, "example.txt")
```
aminalaee, line 169: Name `path` clashes in mypy since it has different type now.

starlette/staticfiles.py
aminalaee: The snippet can be simpler  but I think this is more readable. Maybe a personal preference.

starlette/staticfiles.py
```diff
@@ -73,11 +74,9 @@ def get_directories(
             spec = importlib.util.find_spec(package)
             assert spec is not None, f"Package {package!r} could not be found."
             assert spec.origin is not None, f"Package {package!r} could not be found."
-            package_directory = os.path.normpath(
-                os.path.join(spec.origin, "..", statics_dir)
-            )
-            assert os.path.isdir(
-                package_directory
+            package_directory = Path(spec.origin).joinpath("..", statics_dir).resolve()
```
euri10, line 81: reviews are fun, TIL about `joinpath`, I guess I would have written `(Path(spec.origin) /  "../statics_dir").resolve()` but I guess yours is cleaner

starlette/staticfiles.py
euri10: this looks good

euri10 (approved):
looks good to me

Kludex:
Please do not merge until 0.19.1 is released. This should be up on 0.20.0.

aminalaee:
@Kludex Can we merge this now?

Kludex:
correct me if I'm wrong, but this PR changes the API from `str` to `Path`, and allows staticfiles to follow symlinks outside directory. Right? - just for the release notes

starlette/staticfiles.py
```diff
@@ -158,20 +157,26 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
-            if os.path.commonprefix([full_path, directory]) != directory:
+            original_path = Path(directory).joinpath(path)
+            full_path = original_path.resolve()
+            directory = Path(directory).resolve()
+            try:
+                full_path.relative_to(directory)
+                is_internal = True
+            except ValueError:
+                is_internal = False
+            if not is_internal and not original_path.is_symlink():
```
graingert: Does is_symlink perform blocking IO?

graingert: Can os.lstat be used so the stat result can be reused if the path isn't a symlink?

aminalaee: > Does is_symlink perform blocking IO?

Not really sure about this.



> Can os.lstat be used so the stat result can be reused if the path isn't a symlink?

I think by that you mean on Line 171 we do `os.stat(full_path)` so you want to re-use that but here we check the symlink on the `original_path` before we resolve any symlinks, so they're different.
I might be missing something here....

aminalaee:
> correct me if I'm wrong, but this PR changes the API from `str` to `Path`, and allows staticfiles to follow symlinks outside directory. Right? - just for the release notes

@Kludex 
- About "following symlinks" the answer is yes.
- About "changing API from str to Path" it's a bit tricky. we are changing some methods like `get_response` and `lookup_path` but I'm not sure if they are the API here. The main entry to `Staticfiles` is not changed, but in some cases people might depend on the changed methods, which is not common, and it might break for them. So a note might be useful.

starlette/staticfiles.py
aminalaee: @graingert friendly reminder

graingert: You can reuse the lstat if it turns out the original_path and full_path are the same

aminalaee: Thanks for the review. Done. @graingert

starlette/staticfiles.py
```diff
@@ -158,20 +157,24 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
+            full_path = original_path.resolve()
+            directory = Path(directory).resolve()
             try:
-                return full_path, os.stat(full_path)
+                stat_result = os.lstat(original_path)
+                full_path.relative_to(directory)
+                return full_path, stat_result
+            except ValueError:
+                # Don't allow misbehaving clients to break out of the static files
+                # directory if not following symlinks.
```
Kludex: Can we remove the double negative here? It took me a long time to understand this.

```suggestion
                # Allow clients to break out of the static files directory if following symlinks.
```

tests/test_staticfiles.py
```diff
@@ -441,3 +441,28 @@ def mock_timeout(*args, **kwargs):
     response = client.get("/example.txt")
     assert response.status_code == 500
     assert response.text == "Internal Server Error"
+
+
+def test_staticfiles_follows_symlinks_to_break_out_of_dir(
```
Kludex, line 446: I've run this test without your changes, and this test still passes. I guess this is not expected...

starlette/staticfiles.py
```diff
@@ -158,20 +157,24 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
+                if not stat.S_ISLNK(stat_result.st_mode):
+                    continue
```
Kludex: This conditional is not doing anything, is it? 🤔

If it's true or false, it doesn't matter, it will still continue...

tests/test_staticfiles.py
Kludex: Don't we want the symlink file to be inside the directory, and the real file to be outside?

Kludex (changes requested):

starlette/staticfiles.py
aminalaee: Done.

tests/test_staticfiles.py
aminalaee: Yes you're right, the test should be the other way around.

starlette/staticfiles.py
aminalaee: Actually this came from a refactor requested to re-use the `lstat` result.
Now that I'm looking at the refactored correct version, I think we're not re-using it, it was better in the first place. Right now it is:

```python
try:
    stat_result = os.lstat(original_path)
     full_path.relative_to(directory)
     return full_path, stat_result
 except ValueError:
    # Allow clients to break out of the static files directory
    # if following symlinks.
    if stat.S_ISLNK(stat_result.st_mode):
         stat_result = os.lstat(full_path)
         return full_path, stat_result
```

So if `lstat` returns a symbolic directory, we have to run another `os.stat` on the real file anyway, I think we're complicating it and it's not worth it. What do you think?

starlette/staticfiles.py
```diff
@@ -158,20 +157,25 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
             try:
-                return full_path, os.stat(full_path)
+                stat_result = os.lstat(original_path)
+                full_path.relative_to(directory)
+                return full_path, stat_result
+            except ValueError:
+                # Allow clients to break out of the static files directory
+                # if following symlinks.
+                if stat.S_ISLNK(stat_result.st_mode):
+                    stat_result = os.lstat(full_path)
```
Kludex, line 174: The `stat_result` from `original_path` and `full_path` is the same, right? We should already have this value from before. Maybe I'm wrong here? 🤔 

We can also move the `stat_result = os.lstat(original_path)` outside (just before) the `try..except`, right?

starlette/staticfiles.py
aminalaee: @Kludex  I think this version which used the `Path` was better: https://github.com/encode/starlette/pull/1377/commits/3fb9da7e82ce00eb316593eba51f1d2f339c6314

starlette/staticfiles.py
Kludex: I think I'm wrong, and this implementation is just fine 👍

Kludex: Actually, I'm not sure... Please check this. As I think, what I'm saying is the same as this comment: https://github.com/encode/starlette/pull/1377#discussion_r874622960

starlette/staticfiles.py
Kludex: IMO, I think the current one is fine. It has less "negative conditionals", which makes the code easier to understand.

starlette/staticfiles.py
aminalaee: > The stat_result from original_path and full_path is the same, right?

I don't think so. For the `original_path` in this case will be `S_ISLNK` but when following the link, on the original file it should be `S_ISREG`. I might be wrong though, this is my understanding.

> We can also move the stat_result = os.lstat(original_path) outside (just before) the try..except, right?

I don't think we can, calling `full_path.relative_to(directory)` can cause ValueError, that's why it should be here.

Kludex: I didn't say to put the `relative_to` outside, just the `os.lstat`. But if my statement above is wrong, then this implementation is fine. 👍

starlette/staticfiles.py
```diff
@@ -158,20 +157,25 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
-                # Don't allow misbehaving clients to break out of the static files
-                # directory.
-                continue
+            original_path = Path(directory).joinpath(path)
+            full_path = original_path.resolve()
+            directory = Path(directory).resolve()
             try:
-                return full_path, os.stat(full_path)
+                stat_result = os.lstat(original_path)
+                full_path.relative_to(directory)
```
Kludex, line 168: We could also have used [`is_relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to), instead of depending on `ValueError`.

Or just...
```python
            original_path = Path(directory).joinpath(path)
            full_path = original_path.resolve()
            directory = Path(directory).resolve()
            stat_result = os.lstat(original_path)
            if full_path.is_relative_to(directory):
                return full_path, stat_result
            elif full_path.is_symlink():
                return full_path, os.lstat(full_path)
```

starlette/staticfiles.py
aminalaee: Oh sorry, calling os.lstat will also raise `FileNotFoundError, NotADirectoryError` exceptions in some cases, that's why it's there. @Kludex

starlette/staticfiles.py
aminalaee: @graingert I don't fully understand what you mean by them being the same, if one is a link to another, how can they be the same?

Kludex (approved):
LGTM! 👍

I've added a comment about an alternative solution, but use whatever you prefer.

starlette/staticfiles.py
Kludex: He's saying in case they are not.

starlette/staticfiles.py
aminalaee: I also tried this before, the `is_relative_to` was introduced in Python 3.9, so couldn't use it.

aminalaee: https://github.com/encode/starlette/pull/1377#discussion_r853867770

starlette/staticfiles.py
aminalaee: If they are not then we already have the correct lstat, so we're good.

PR #1378, opened 2021-12-17
Add overload in Config.__call__ for cast as Callable

flaeppe:
Adds an additional `@typing.overload` to `Config.__call__` that allows `cast` argument to be a `Callable`. As far as I could tell, having a `Callable` type is already supported at runtime.

Currently, one gets hit by an error like:

```
error: No overload variant of "__call__" of "Config" matches argument types "str", "Callable[[Any], int]"  [call-overload]
```

Here's an example usage of the new overload:

```python
from typing import Any

def b(value: Any) -> float:
    return float(value)

config = Config()

A = config("A", cast=lambda value: int(value))
reveal_type(A)  # Revealed type is "builtins.int*"

B = config("B", cast=b)
reveal_type(B)  # Revealed type is "builtins.float*"

C = config("C", b)
reveal_type(C)  # Revealed type is "builtins.float*"

D = config("D", cast=b, default=1.0)
reveal_type(D)  # Revealed type is "builtins.float"

# Errors with incompatible default= type
# Argument "cast" to "__call__" of "Config" has incompatible type "Callable[[Any], float]"; expected "Callable[[Any], int]"  [arg-type]
E = config("E", cast=b, default=1)
reveal_type(E)  # Revealed type is "builtins.int"
```


Kludex:
I've never done it, but is there a set of best practice to test stubs? Should we?

flaeppe:
Hm, I don't know about any best practices. Though I have seen a couple of approaches for testing, example: https://github.com/typeddjango/django-stubs. That approach is quite comprehensive.

Another one on the top of my head; include test files when running mypy, i.e. type check tests. Might help resolve some obvious typing issues.

A lot of times I've seen typing being handled as issues comes in, not testing it or anything like that. Basically as you do now.

flaeppe:
@Kludex Having a second look at the `@overload`s here. Wouldn't it be enough with the following 2? Or am I missing something?

```python
@typing.overload
def __call__(
    self, key: str, cast: typing.Type[T], default: T = ...
) -> T:  # pragma: no cover
    ...

@typing.overload
def __call__(
    self, key: str, cast: typing.Callable[[typing.Any], T] = ..., default: T = ...
) -> T:  # pragma: no cover
    ...
```

Kludex:
Yeah, there's probably something wrong with the current one, but `cast` and `default` are not necessarily used at the same time, so I think something else is missing. You can check more about it here: https://github.com/encode/starlette/pull/1097

Maybe the `cast: Type[T]` is unnecessary... As I `int`, `bool`, etc are callables. :thinking:

flaeppe:
> Maybe the `cast: Type[T]` is unnecessary... As I `int`, `bool`, etc are callables. 🤔

By the looks of https://github.com/python/mypy/issues/1670, it seems that `Type[T]` is a subtype of `Callable[[], T]`. Not sure if we should lean in to that

> ... but `cast` and `default` are not necessarily used at the same time, so I think something else is missing.

Would it make sense to type it so that `default` is `T | None`? It feels slightly off to support something like:

```python
reveal_type(config("VAL", cast=int, default=True))  # int | bool
```

(But perhaps I'm missing something. _I did, the config implementation is swallowing a bit more_)

And for the case where `cast` isn't an argument, the default type would be ~`str | None` (basically since the "default cast" is `str`~) that was incorrect, "default cast" is no cast, which leads to `Any`.

Summarised, this is the result ~I'm thinking~ I expected:

```python
reveal_type(config("A"))  # str | None
reveal_type(config("B", cast=int))  # int | None
reveal_type(config("C", default="default"))  # str
reveal_type(config("D", cast=int, default=4))  # int

reveal_type(config("E", default=15))  # Error -> requires default as str
reveal_type(config("F", cast=float, default=True))  # Error -> bool != float
```

__Edited types after looking more at the implementation:__

```python
reveal_type(config("A"))  # Any
reveal_type(config("B", cast=int))  # int | None (no cast is performed when value is None)
reveal_type(config("C", default="default"))  # Any | str  (this is just Any, really)
reveal_type(config("D", cast=int, default=4))  # int | None
reveal_type(config("E", default=15))  # Any
reveal_type(config("F", cast=float, default=True))  # float | None (float(True) is called when "F" isn't found)
```

---

An additional thing I just saw was that `cast` is also applied to `default`, which would lead to(when `"F"` isn't found):

```python
assert config("F", cast=float, default=True) == 1.0
```

flaeppe:
_I think_ by having looked at this again, that these 3 overloads align with the implementation:

```python
@typing.overload
def __call__(
    self, key: str, cast: typing.Callable[[typing.Any], T], default: typing.Any = ...
) -> typing.Optional[T]:
    ...

@typing.overload
def __call__(
    self, key: str, cast: typing.Type[T], default: typing.Any = ...
) -> typing.Optional[T]:
    ...

@typing.overload
def __call__(
    self, key: str, cast: None = ..., default: typing.Any = ...
) -> typing.Any:
    ...

def __call__(
    self, key: str, cast: typing.Callable = None, default: typing.Any = undefined
) -> typing.Any:
    return self.get(key, cast, default)
```

Kludex:
Now we have mypy on tests as well, so I guess we can add tests with the different `Config`s overload possibilities.

bfontaine:
> ```python
> @typing.overload
> def __call__(
>     self, key: str, cast: typing.Callable[[typing.Any], T], default: typing.Any = ...
> ) -> typing.Optional[T]:
>     ...
> ```

Hello, I don’t understand that one. Why is the return value `Optional[T]` and not `T`? The type of `cast` says it returns a value of type `T`, so the return value of `__call__` should be `T`. What am I missing?

flaeppe:
@bfontaine See here: https://github.com/encode/starlette/blob/424351cb231c67798a65c091b0b7d42790f5e444/starlette/config.py#L114-L115

lovelydinosaur (approved):

lovelydinosaur:
Great yup. Thank you!

PR #1379, opened 2021-12-17
Type background argument in responses as Optional

flaeppe:
Having `background` argument as `Optional` allows for conditionally pass a background task (`background=background if ... else None`) together with `no_implicit_optional=true` configured for mypy.

With the setup above, one is currently hit with:

```
error: Argument "background" to "Response" has incompatible type "Optional[BackgroundTask]"; expected "BackgroundTask"  [arg-type]
```

Here's a snippet for checking in on the adjusted type:

```python
import random

from starlette.responses import (
    FileResponse,
    RedirectResponse,
    Response,
    StreamingResponse,
)


def bg() -> None:
    return None


background = BackgroundTask(bg)

Response(background=background if random.choice([True, False]) else None)
Response(background=background)
Response(background=None)

RedirectResponse("", background=background if random.choice([True, False]) else None)
RedirectResponse("", background=background)
RedirectResponse("", background=None)

StreamingResponse("", background=background if random.choice([True, False]) else None)
StreamingResponse("", background=background)
StreamingResponse("", background=None)

FileResponse("", background=background if random.choice([True, False]) else None)
FileResponse("", background=background)
FileResponse("", background=None)

jinja = Jinja2Templates("")
jinja.TemplateResponse("", {}, background=background if random.choice([True, False]) else None)
jinja.TemplateResponse("", {}, background=background)
jinja.TemplateResponse("", {}, background=None)
```

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -40,7 +40,7 @@ class Response:
         status_code: int = 200,
         headers: dict = None,
         media_type: str = None,
-        background: BackgroundTask = None,
+        background: typing.Optional[BackgroundTask] = None,
     ) -> None:
         self.status_code = status_code
         if media_type is not None:
@@ -186,7 +186,7 @@ class RedirectResponse(Response):
         url: typing.Union[str, URL],
         status_code: int = 307,
         headers: dict = None,
-        background: BackgroundTask = None,
+        background: typing.Optional[BackgroundTask] = None,
     ) -> None:
         super().__init__(
             content=b"", status_code=status_code, headers=headers, background=background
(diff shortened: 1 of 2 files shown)


Kludex:
Thanks for the PR @flaeppe , but the type here is not wrong. It means that the type of `backgound` is `BackgroundTask`, but it shouldn't accept `None`.

That being said, I'm closing this PR. Also, jfyk, we are open to discuss this either on Gitter or on a Starlette issue if you still think this is necessary/wanted.

flaeppe:
Alright, gotcha 👍

PR #1382, opened 2021-12-22
feat: add headers to UploadFile

adriangb:
Alternative / subset of #1311.

This preserves the multipart field headers for file uploads. Currently only `content_type` is preserved.

As per https://github.com/encode/starlette/pull/1311#issuecomment-999752023, this is pretty much the same API Flask uses.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -423,14 +423,14 @@ class UploadFile:
         file: typing.IO = None,
         content_type: str = "",
         *,
-        raw_headers: typing.Optional[typing.List[typing.Tuple[bytes, bytes]]] = None,
+        headers: "typing.Optional[Headers]" = None,
     ) -> None:
         self.filename = filename
         self.content_type = content_type
         if file is None:
             file = tempfile.SpooledTemporaryFile(max_size=self.spool_max_size)
         self.file = file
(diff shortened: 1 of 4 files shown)


starlette/datastructures.py
```diff
@@ -415,15 +415,22 @@ class UploadFile:
...
+    headers: "Headers"
 
     def __init__(
-        self, filename: str, file: typing.IO = None, content_type: str = ""
+        self,
+        filename: str,
+        file: typing.IO = None,
+        content_type: str = "",
+        *,
+        raw_headers: typing.Optional[typing.List[typing.Tuple[bytes, bytes]]] = None,
```
lovelydinosaur: I think it'd make more sense to me if the API here was `headers: Headers`.

The `formparsers.py` module could be responsible for instantiating the headers from the raw items, right?

lovelydinosaur:
Yes looks good! Just the one comment here.

starlette/datastructures.py
adriangb: Yep great call, not sure why I wasn't doing that from the get go

adriangb:
Addressed, thank you! I also added a blurb to the docs.

lovelydinosaur (approved):
Fantastic - thanks!

PR #1383, opened 2021-12-23
Update type annotations for BackgroundTask and utils

uriyyo:


starlette/endpoints.py
```diff
@@ -34,7 +34,7 @@ async def dispatch(self) -> None:
         if is_async:
             response = await handler(request)
         else:
-            response = await run_in_threadpool(handler, request)
+            response = await run_in_threadpool(handler, request)  # type: ignore
```
Kludex: Maybe this above:
```python
        handler :typing.Callable[[Request], typing.Any] = getattr(self, handler_name, self.method_not_allowed)
```
or just:
```suggestion
            response = await run_in_threadpool(handler, request)  # type: ignore[arg-type]
```

uriyyo: Thanks, fixed

starlette/background.py
```diff
@@ -1,12 +1,16 @@
 import asyncio
 import typing
 
+import typing_extensions
```
Kludex: `typing_extensions` is only available for versions below Python 3.8 right now. I guess we'd want to change the setup.py?

uriyyo: Agree, fixed

Kludex: But now it needs to verify the version, otherwise it will raise `ModuleNotFoundError` on 3.10.

uriyyo: Agree, fixed again)

Kludex (approved):
Thanks @uriyyo ! :tada:

PR #1386, opened 2021-12-27
New exception page.

### LINKED ISSUE #30: Include more information in `DebugMiddleware`.

alex-oleshkevich:
This PR updates debugger UI.

Features:
* code preview
* local variables for snippet
* request/session/headers/etc details
* beautify UI
* responsible

UI is influenced by the [Phoenix Framework](https://www.phoenixframework.org/) error page

localhost_7001_alex

### DIFF
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -32,9 +32,10 @@ body {
     margin: 0;
     color: var(--gray);
     background: var(--gray-lightest);
-    font-family: Inter var, ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
-        Helvetica Neue, Arial, Noto Sans, sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol,
-        Noto Color Emoji;
+    font-family: Inter var, ui-sans-serif, system-ui,
+        -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
+        Helvetica Neue, Arial, Noto Sans, sans-serif, Apple Color Emoji,
+        Segoe UI Emoji, Segoe UI Symbol, Noto Color Emoji;
 }
 
 .text-muted {
@@ -170,7 +171,7 @@ summary {
 
 dl {
     display: grid;
-    grid-template-columns: repeat(5, minmax(0, 1fr));
+    grid-template-columns: repeat(12, minmax(0, 1fr));
     align-items: center;
     margin: 0;
     padding: 4px;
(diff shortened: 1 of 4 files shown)


Kludex:
Beautiful. 👍

Kludex:
Closes #30

alex-oleshkevich:
Here is a full page screenshot with all sections open.

fullpage

alex-oleshkevich:
This handler may raise an exception if one of locals is `QuerystringParser`.
The bug itself is here https://github.com/andrew-d/python-multipart/issues/29, seems fixed but no new version has been released for a long time.

starlette/middleware/errors.py
```diff
@@ -1,127 +1,442 @@
...
+def mask_secrets(key: str, value: str) -> str:
+    key = key.lower()
+    if any(
+        [
+            "key" in key,
+            "token" in key,
+            "password" in key,
+            "secret" in key,
+        ]
+    ):
```
aminalaee, line 435:
```suggestion
    if any(k in key for k in ["key", "token", "password", "secret"]):
```
Is this less readable? What do you think?

starlette/middleware/errors.py
```diff
@@ -1,127 +1,442 @@
...
+        return "*" * 8
```
aminalaee, line 436: I haven't really checked any other projects, but do you think it would be useful to mask a part of the value? For example masking first half of the string? Or at least mask the length of the value, as I think this might be confusing.

aminalaee (commented):
This is nice.

lovelydinosaur:
This looks fantastic.
Having said that it's also an *awful* lot of extra code to pull into the core package and maintain.

Presumably you'd also be able to package this up as a third party package, and document how to install it as a custom exception handler, right? (Or are there any blockers there that I've not figured out?)

A good rule of thumb for proposals to Starlette should probably be "if it can be a third party package, it should be".

That's worked really well for REST framework, and it'd work really well for Starlette too - help make sure we can just keep it small and well-scoped, but still allow folks to build fantastic stuff on top of it.

alex-oleshkevich:
> 

Yes, I have the same thought.

alex-oleshkevich:
Available as https://github.com/alex-oleshkevich/starception

lovelydinosaur:
Fantastic stuff.
Want to issue a pull request linking to it from https://www.starlette.io/exceptions/ and https://www.starlette.io/third-party-packages/ ?

PR #1387, opened 2021-12-27
Allow Session scoped cookies.

alex-oleshkevich:
This PR changes the behavior of SessionMiddleware.
If the user sets `max_age` to None or 0 then a cookie will be Session-only and will be destroyed after closing the browser.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -98,7 +98,7 @@ The following arguments are supported:
 
 * `secret_key` - Should be a random string.
 * `session_cookie` - Defaults to "session".
-* `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to None or 0 seconds then the session will be cleared on tab close.
+* `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to `None` then the cookie will last as long as the browser session.
 * `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
 * `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to `False`.
 
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -98,7 +98,7 @@ The following arguments are supported:
 
 * `secret_key` - Should be a random string.
 * `session_cookie` - Defaults to "session".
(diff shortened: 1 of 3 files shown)


docs/middleware.md
```diff
@@ -98,7 +98,7 @@ The following arguments are supported:
 
 * `secret_key` - Should be a random string.
 * `session_cookie` - Defaults to "session".
-* `max_age` - Session expiry time in seconds. Defaults to 2 weeks.
+* `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to None or 0 seconds then the session will be cleared on tab close.
```
Kludex: Should we really include the zero here? :thinking: 

I've rephrased a little to make the sentence more clear, does it make sense?
```suggestion
* `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to `None` then the cookie will last as long as the browser session.
```

Kludex (changes requested):
`max_age` type on the `__init__` should be `Optional[int]` as it can receive `None` now.

tests/middleware/test_session.py
```diff
@@ -125,3 +125,20 @@ def test_invalid_session_cookie(test_client_factory):
...
+    client = test_client_factory(app)
+
+    response = client.post("/update_session", json={"some": "data"})
+    assert response.json() == {"session": {"some": "data"}}
+
+    # check cookie max-age
+    set_cookie = response.headers["set-cookie"]
+    assert "Max-Age" not in set_cookie
+
+    client.cookies.clear_session_cookies()
```
Kludex, line 146: Note to myself: use `client.cookies.clear` instead of `client.cookies.clear_session_cookies` on https://github.com/encode/starlette/pull/1376

Kludex (approved):
Thanks @alex-oleshkevich :)

tests/middleware/test_session.py
alex-oleshkevich: why? afaik this method will keep all other cookies except session-scoped.

Kludex: Oh right, yeah... `httpx` doesn't have `clear_session_cookies` :disappointed:

PR #1388, opened 2021-12-28
Add `WebSocketRoute` docs in `Routing`

### LINKED ISSUE #764: WebSocketRoute undocumented
The "routing" docs page doesn't mention that `Route()` is for HTTP only and `WebSocketRoute()` is required for websocket endpoints. I found `WebSocketRoute` through examples on other pages but it woudl be good if it was fully described on the "routing" page.

aminalaee:
Closes #764 .

Currently we have two examples of `WebSocketRoute` in applications and endpoints, but they don't document the `WebSocketRoute`.

I've included the `endpoint` argument in the docs and pointed general features like Path Parameters, Reverse URL Lookups and etc. to the Route docs.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -1,3 +1,5 @@
+## HTTP Routing
+
 Starlette has a simple but capable request routing system. A routing table
 is defined as a list of routes, and passed when instantiating the application.
 
@@ -26,8 +28,7 @@ The `endpoint` argument can be one of:
 
 * A regular function or async function, which accepts a single `request`
 argument and which should return a response.
-* A class that implements the ASGI interface, such as Starlette's [class based
-views](endpoints.md).
+* A class that implements the ASGI interface, such as Starlette's [HTTPEndpoint](endpoints.md).
 
 ## Path Parameters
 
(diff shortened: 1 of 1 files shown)


docs/routing.md
```diff
@@ -262,3 +263,42 @@ app = Router(routes=[
...
+    WebSocketRoute("/{name}", endpoint=websocket_user),
+]
+
+app = Starlette(routes=routes)
+```
+
+The `endpoint` argument can be one of:
+
+* An async function, which accepts a single `websocket` argument.
+* A class that implements the ASGI interface, such as Starlette's [WebSocketEndpoint](endpoints.md).
```
Kludex: Can't we redirect specifically to the `WebSocketEndpoint` section in the `endpoints.md` page?

docs/routing.md
```diff
@@ -262,3 +263,42 @@ app = Router(routes=[
...
+    WebSocketRoute("/", endpoint=websocket_index),
+    WebSocketRoute("/{name}", endpoint=websocket_user),
+]
+
+app = Starlette(routes=routes)
+```
+
+The `endpoint` argument can be one of:
+
+* An async function, which accepts a single `websocket` argument.
```
Kludex, line 303: We don't mention the `WebSocket` class on this page, should we?

docs/routing.md
```diff
@@ -262,3 +263,42 @@ app = Router(routes=[
...
 ])
 ```
+
+## WebSocket Routing
+
+When working with WebSocket endpoints, you should use `WebSocketRoute`
+instead of the usual `Route`.
+
+Path parameters, and reverse URL lookups for `WebSocketRoute` work the the same
+as HTTP `Route`, which can be found in the [Route](#path-parameters) documentation.
```
Kludex: Looks weird talking about it like if it was in another location, but in reality, it's actually above :thinking:

Can we rephrase this last part (after the last comma)? Something like "like mentioned in the section [Path parameters](#path-parameters) above", or something like that?

If you disagree, feel free to close this thread :+1:

aminalaee: I agree with making it obvious they're on the same page.
Another point, since we're linking multiple things (path params, url lookups, ..), what do you think instead we link it to the `HTTP Routing` section? It's not really accurate to point to path parameters.

docs/routing.md
aminalaee: I didn't know we could do it 😄  Changed it.

docs/routing.md
aminalaee: I think it would make sense. We don't do this for HTTP endpoints that accept a `Request` either. I think they can be helpful.

docs/routing.md
Kludex: Sure :+1:

docs/routing.md
Kludex: I didn't know how to do it either... hahaha

docs/routing.md
aminalaee: After giving it more thought I think there are lots of other places we can specify technical details, probably it can be a better option to add docstrings or something.
I'll post this into gitter to get some feedback.

Kludex: So... You want to wait something before merging this?

aminalaee: Already tried that, I don't think we can do that. IMO this should be good for now.

Kludex (approved):

PR #1397, opened 2022-01-07
Don't set  headers for responses with 1xx, 204 and 304 status code

### LINKED ISSUE #1178: Add EmptyResponse class

### LINKED ISSUE #1282: JSONResponse Wrong Content-Length Value
### Describe the bug

`JSONResponse` populates the response header with a **non-zero** `Content-Length` whenever the response object is instantiated **without any content**.

### Checklist

### To reproduce

```python
from starlette.datastructures import MutableHeaders

r1 = JSONResponse('')
r1.headers

r2 = JSONResponse()
r2.headers
```

### Expected behavior

```python
( r1.headers == r2.headers
  and
  r1.headers == MutableHeaders({'content-length': '0', 'content-type': 'application/json'}) )
```

### Actual behavior

```python
( r1.headers != r2.headers
  and
  r1.headers == MutableHeaders({'content-length': '2', 'content-type': 'application/json'})
  and
  r2.headers == MutableHeaders({'content-length': '4', 'content-type': 'application/json'}))
```

### Debugging material

All that is needed is the currently stable release of `Starlette`, and the trusty `terminal`.

### Environment

- OS: same on all of Linux/Windows/macOS
- Python version: 3.7.x
- Starlette version: 0.14.x

### Additional context

I was trying to adhere to the HTTP ref spec when constructing responses to the OPTIONS and HEAD methods.

Kludex:
Closes #1178 (?)
Closes #1282

### DIFF
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -1,5 +1,4 @@
 import os
-from typing import Callable
 
 import anyio
 import pytest
@@ -334,6 +333,13 @@ def test_empty_response(test_client_factory):
     assert response.headers["content-length"] == "0"
 
 
+def test_empty_204_response(test_client_factory):
+    app = Response(status_code=204)
+    client: TestClient = test_client_factory(app)
+    response = client.get("/")
+    assert "content-length" not in response.headers
+
+
 def test_non_empty_response(test_client_factory):
     app = Response(content="hi")
     client: TestClient = test_client_factory(app)
@@ -367,20 +373,3 @@ def test_streaming_response_known_size(test_client_factory):
     client: TestClient = test_client_factory(app)
     response = client.get("/")
     assert response.headers["content-length"] == "10"
-
-
-@pytest.mark.parametrize("status_code", (100, 204, 304))
-def test_no_content_length(
-    status_code: int, test_client_factory: Callable[..., TestClient]
-):
-    app = Response(content=b"non_empty", status_code=status_code)
-    client = test_client_factory(app)
-    response = client.get("/")
-    assert "content-length" not in response.headers
-
-
-def test_json_response_no_content_length(test_client_factory):
-    app = JSONResponse("null", status_code=204)
-    client = test_client_factory(app)
-    response = client.get("/")
-    assert "content-length" not in response.headers
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -1,4 +1,5 @@
 import os
+from typing import Callable
 
 import anyio
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_responses.py
```diff
@@ -366,3 +367,20 @@ def test_streaming_response_known_size(test_client_factory):
...
+    client = test_client_factory(app)
+    response = client.get("/")
+    assert "content-length" not in response.headers
+
+
+def test_json_response_no_content_length(test_client_factory):
+    app = JSONResponse("null", status_code=204)
+    client = test_client_factory(app)
+    response = client.get("/")
+    assert "content-length" not in response.headers
```
lovelydinosaur: I'd simplify what we test here. Instead of this, I'd suggest we just add the following case, after the existing `test_empty_response` case...

```python
def test_empty_204_response(test_client_factory):
    app = Response(status_code=204)
    client: TestClient = test_client_factory(app)
    response = client.get("/")
    assert "content-length" not in response.headers
```

This helps make the difference between the 200 and the 204 case clearer. Also it's a valid, sensible, thing to do, which these two cases aren't. (they *are* including a body, but setting status codes that shouldn't have one.)

TBBle:
Looking at this change and prompted by https://github.com/encode/starlette/pull/1395#issuecomment-1006582075, what happens if you still pass a body in, i.e. `Response(status_code=204, content="hi")` in the test?

My skimming of the relevant [ASGI 3 spec](https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event) suggests the ASGI server may add `Transfer-Encoding: chunked`, which then becomes RFC-invalid again, as the presence of either of `Content-Length` or `Transfer-Encoding` signals a body, which is disallowed for 204 and related responses.

Perhaps the implicit expectation is that the ASGI server will use "may" here and recognise response codes for which `Transfer-Encoding` should not be added (and also implicitly drop the body data, perhaps)? Although that seems like a gap in the spec that it's not explicit, and a gap in behaviour that the same thing is not done for `Content-Length`, as the same HTTP spec restrictions apply.

I apologise if I'm asking a question with an obvious/known answer already. I'm new to the ASGI ecosystem, and may be misunderstanding the division of responsibility or otherwise.

lovelydinosaur:
> what happens if you still pass a body in, i.e. Response(status_code=204, content="hi") in the test?

Probably you've just got yourself an RFC-invalid response.

We *could* potentially guard against those kinds of cases, and I'd be happy for us to discuss that as a follow-up issue if anyone's sufficiently motivated to do so.

The most important thing for *this* pull request tho, is to make sure we're doing the right thing for the sensible cases.

lovelydinosaur (approved):
Looks great, yup.

Observations prompted by the PR...

I think our naming of `populate_content_length` / `populate_content_type` in the `init_headers` method isn't really ideal at this point, and I'd *also* consider us switching it to a plain function rather than a stateful method (because I think it's easier to understand the behaviour then).

Not something to address here, but perhaps worth consideration.

PR #1398, opened 2022-01-07
add support for datetime,date,time converters

Mng-dev-ai:
These converters are useful when we want to capture a URL that contains a Date,Date-time or Time.
 e.g. /users/created_at/2020-01-01


Kludex:
More information is needed.

aminalaee:
I think it's a bit tricky as these types could have different formattings. What do you think?

Maybe a generic RegexConverter would cover all these.

Mng-dev-ai:
@aminalaee You're right but I think it would be better if the both are supported because regex maybe hard for some people to use.

Kludex:
@Mng-dev-ai Thanks for the PR!

I don't think this is the right path to follow on Starlette, but I do think you should be able to create custom converters. I'm going to work on the "register custom converters" PR (follow https://github.com/encode/starlette/pull/1437).

That being said, I'm going to close this PR.

PR #1399, opened 2022-01-08
Add Python 3.11-dev to the pipeline

Kludex:


Kludex:
Pipeline fails due to `greenlet`, but the problem was already solved on master. We need to wait for the next release.

Watch releases on: https://github.com/python-greenlet/greenlet/tags
The `greenlet` version of today is 1.1.2.

Reference: https://github.com/python-greenlet/greenlet/issues/273

aminalaee:
Looks like we need to pin the development version. From the greenlet PR [here](https://github.com/python-greenlet/greenlet/blob/5ed467e5cb34651cc013c286158d0b6d7ff0a26a/src/greenlet/greenlet_cpython_compat.hpp#L52
)

```
Greenlet won't compile on anything older than Python 3.11 alpha 4 (see
https://bugs.python.org/issue46090). Summary of breaking internal changes:
- Python 3.11 alpha 1 changed how frame objects are represented internally.
  - https://github.com/python/cpython/pull/30122
- Python 3.11 alpha 3 changed how recursion limits are stored.
  - https://github.com/python/cpython/pull/29524
- Python 3.11 alpha 4 changed how exception state is stored. It also includes a
  change to help greenlet save and restore the interpreter frame "data stack".
  - https://github.com/python/cpython/pull/30122
  - https://github.com/python/cpython/pull/30234
  ```

lovelydinosaur:
Any idea which of our dependencies ends up pulling in greenlet?

Kludex:
databases -> sqlalchemy -> greenlet

lovelydinosaur:
Since there's no obvious movement on a `greenlets` release *right now*, let's just bump this off until 3.11 beta and have another look then.

https://www.python.org/dev/peps/pep-0664/#schedule

PR #1400, opened 2022-01-08
Set `request` in template context

### LINKED ISSUE #482: Rendering templates without explicitly passing request
As present, Starlette's templating requires that the response variable be passed explicitly. This requires `TemplateResponse('index.html', {'request': request})` to be written everywhere, thus I have been playing around with the idea of performing the render in the middleware where the request would be available, thus avoiding the need of passing it explicitly.

To do this, I tried adding a subclass of `Response` like so:

```
class TemplateResponse(Response):
    def __init__(self, template, context={}, status_code=200, headers={}, media_type='text/html', background=None):
        self.template = jinja2_env.get_template(template)
        self.context = context
        headers['X-Templated-Response'] = '1'
        super().__init__('', status_code, headers, media_type, background)
```

... and then render in the middleware and return the "real" response like so:

```
class TemplatingMiddleware(BaseHTTPMiddleware):
        if response.headers.get('X-Templated-Response'):
            del response.headers['X-Templated-Response']
            content = response.template.render({
                **response.context,
                'config': self.app.config,
                'request': request
            })
            response = Response(content, response.status_code, response.headers, response.media_type)
        return response
```

However, it appears that all responses get converted to a `StreamingResponse` object preventing me from achieving this goal.

Is there a way to preserve the original "TemplateResponse" object; or is there another way of achieving this?

aminalaee:
Closes #482 according to [this](https://github.com/encode/starlette/issues/482#issuecomment-484138702).

Sets up request for template context automatically. So instead of this:

```python
async def homepage(request):
    return templates.TemplateResponse('index.html', {'request': request})
```

We can do:

```python
async def homepage(request):
    return templates.TemplateResponse('index.html')
```

So the `context` argument to `templates.TemplateResponse` is not required anymore.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -24,7 +24,7 @@ routes = [
 app = Starlette(debug=True, routes=routes)
 ```
 
-The Jinja2 template context will automatically include a `url_for` function,
+The Jinja2 template context will automatically include `request` and a `url_for` function,
 so we can correctly hyperlink to other pages within the application.
 
 For example, we can link to static files from within our HTML templates:
@@ -58,6 +58,7 @@ def test_homepage():
     response = client.get("/")
     assert response.status_code == 200
     assert response.template.name == 'index.html'
+    assert "request" in response.context
 ```
 
 ## Asynchronous template rendering
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -14,7 +14,7 @@ from starlette.staticfiles import StaticFiles
 templates = Jinja2Templates(directory='templates')
 
 async def homepage(request):
-    return templates.TemplateResponse('index.html', {'request': request})
+    return templates.TemplateResponse('index.html')
 
 routes = [
     Route('/', endpoint=homepage),
@@ -23,9 +23,6 @@ routes = [
 
 app = Starlette(debug=True, routes=routes)
 ```
-
-Note that the incoming `request` instance must be included as part of the
-template context.
 
 The Jinja2 template context will automatically include a `url_for` function,
 so we can correctly hyperlink to other pages within the application.
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -58,7 +58,6 @@ def test_homepage():
     response = client.get("/")
     assert response.status_code == 200
     assert response.template.name == 'index.html'
-    assert "request" in response.context
 ```
 
 ## Asynchronous template rendering
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -24,7 +24,7 @@ routes = [
 app = Starlette(debug=True, routes=routes)
 ```
 
-The Jinja2 template context will automatically include `request` and a `url_for` function,
+The Jinja2 template context will automatically include a `request` instance, and a `url_for` function,
 so we can correctly hyperlink to other pages within the application.
 
 For example, we can link to static files from within our HTML templates:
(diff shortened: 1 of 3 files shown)


starlette/templating.py
```diff
@@ -23,20 +24,26 @@ class _TemplateResponse(Response):
...
         self.template = template
-        self.context = context
-        content = template.render(context)
-        super().__init__(content, status_code, headers, media_type, background)
+        self.context = context or {}
+        super().__init__(b"", status_code, headers, media_type, background)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if "request" not in self.context:
+            self.context["request"] = Request(scope, receive, send)
```
aminalaee, line 39: Now that the request is set internally, I propose to rename it to `_request` so it doesn't interfere with user's `context`.

lovelydinosaur: I'm not sure I understand the motivation of having `"_request"` in the template context automatically.

Using...

```python
        if "request" not in self.context:
            self.context["request"] = Request(scope, receive, send)
```

seems pretty sensible to me. Can you talk me through the motivation of changing that?

aminalaee: If we do something like:

```python
@app.route("/")
async def homepage(request):
    return templates.TemplateResponse("index.html", {"request": "example"})
```

Then the `url_for` will throw and exception as it's always expecting the `request` to be Starlette  `Request` object. We're somehow making the `request` keyword unavailable to template context.
Does that make sense?

aminalaee: I'm not entirely sure we need the `if` check anymore. If we set it internally, I can't think of any reason why the user request would be different from the one created by TemplateResponse. Unless I'm missing a point here.

lovelydinosaur: I think a reasonable API would be that we document that `TemplateResponse` includes `"request"` by default.

> We're somehow making the request keyword unavailable to template context.

That's okay with me. It's such a semantically obvious name, that I think a user including `"request"` and setting it to something *other* that a `Request` instance is causing trouble for themselves.

(And users *can* still override it, if they set it explicitly)

aminalaee: Great, thanks for the feedback.

docs/templates.md
```diff
@@ -24,7 +24,7 @@ routes = [
 app = Starlette(debug=True, routes=routes)
 ```
 
-The Jinja2 template context will automatically include a `url_for` function,
+The Jinja2 template context will automatically include `request` and a `url_for` function,
```
lovelydinosaur: Perhaps...

> a `"request"` instance, and a `"url_for"` function

?

aminalaee: Makes sense.

lovelydinosaur:
Okay, I've been looking through this for review.
I hadn't appreciated a couple of things until just now...

* I think this change will end up setting `Content-Length: 0` on template responses. (Which would be broken.)
* Switching `TemplateResponse` so that it renders at the point it is called, rather than on `__init__` is a bit of a behaviour changing thing for use to do. If `response.body` is inspected directly in the endpoint, it'll be `b""`.

I'm not sure what the upshot of those ought to be for us. One option is that we just decide this wasn't the right thing for us to do, and drop those issues. Another would be to make sure we set the Content-Length header properly, and then just be okay with the other behaviour changes.

?

aminalaee:
I have to admit I didn't know about the first issue. I thought we just had to live with the second one.

I think unless we find a clean way to achieve the desired outcome, if possible avoid `request` in context but return proper response and have stable behaviour for `Response`, we should drop this and leave it as is.

Update: I'll spend more time on it and see if that's possible.

lovelydinosaur:
I guess let's just drop this for now, and close the related issues as "we've decided not to do this". (?)
It'd be very nice if it was less awkward. But it's not, so.

Issue #1214, opened 2021-06-24
Support multiple template directories for large projects


nickleman:
I would like to be able to have multiple directories included in the Jinja2Template response. This allows for well structured applications with sub-application separation.  For example in flask the individual blueprints can register their own template directory and the templates in that directory can also reference/include templates from the parent project on up. This enables having base templates with the look and feel of the site at the parent level and then the page content in the sub-project.

There is an implementation out there (https://github.com/accent-starlette/starlette-core/blob/master/starlette_core/templating.py) but I'd like to see it built in. Especially since the referenced implementation isn't in PyPI.

alex-oleshkevich:
I'd rather add an option to pass jinja environment object to the `Jinja2Template` constructor so we could have full control over it.

nickleman:
That works just fine for me. Maybe get the environment from the parent instance and pass it to the constructor?

aminalaee:
@alex-oleshkevich Wouldn't it be easier to just pass `loader` to the `Jinja2Template` so the caller can implement the loader but keep the `env` in `Jinja2Templates` itself?

nickleman:
Or pass the "parent" ninja template instance top the child one along with the new directory.

alex-oleshkevich:
Loader is only one of multiple options that Jinja's Environment class expects. There are also extensions, caching, escaping, and many other options you may want to customize time to time. See https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Environment

Personally, I'd prefer to have a fine grained control over the environment class. As a shortcut, we can also accept template dirs or loader instance. This is discussable.

We can also get env class with `get_env` method and configure it ad-hoc, but, IMO, this is not a nice API.

aminalaee:
I agree with the `Environment` argument approach.

Persoanally I'd prefer the loader argument to the list of templates shortcut because leaving the loader implementation to the caller would be cleaner.

graingert:
fyi there's https://accent-starlette.github.io/starlette-core/templating/ that provides a loader kwarg

aminalaee:
We can achieve this with #1401:

```python
loader = jinja2.FileSystemLoader(directory)

templates = Jinja2Templates(
        directory=directory, loader=loader
)
```

Any feedback would be appreciated

nickleman:
[...]

PR #1401, opened 2022-01-09
Allow `env_options` in ``Jinja2Templates`

### LINKED ISSUE #427: templateing: jinja2: pass kwargs for environment
I think it would be good to pass something like `env_kwargs` via https://github.com/blueyed/starlette/blob/24c135de71ac56a73f7f797258115941579155bf/starlette/templating.py#L51-L53.

While you can change the env afterwards, it would allow Jinja2 to validate e.g. `enable_async`, and call `load_extensions` etc.

aminalaee:
Closes #427 

Allows `Jinja2Templates` to set Jinja2 `Environment` options [here](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Environment).

For example:

```python
templates = Jinja2Templates(
        directory=directory, autoescape=False, auto_reload=True, enable_async=True
)
```

Or:

```python
loader = jinja2.FileSystemLoader(directory)

templates = Jinja2Templates(
        directory=directory, loader=loader
)
```

Since `directory` is still a required argument, it will be passed in this case, but the loader passed will take precedence over the `directory` and the loader created by Starlette. I can't think of any way to handle this right now, until maybe we can consider another API for Jinja2Templates later. This way we can deal with issues like #1214 .

- [x] Tests
- [x] Docs

### DIFF
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -55,12 +55,14 @@ class Jinja2Templates:
     return templates.TemplateResponse("index.html", {"request": request})
     """
 
-    def __init__(self, directory: typing.Union[str, PathLike]) -> None:
+    def __init__(
+        self, directory: typing.Union[str, PathLike], **env_options: typing.Any
+    ) -> None:
         assert jinja2 is not None, "jinja2 must be installed to use Jinja2Templates"
-        self.env = self._create_env(directory)
+        self.env = self._create_env(directory, **env_options)
 
     def _create_env(
-        self, directory: typing.Union[str, PathLike]
+        self, directory: typing.Union[str, PathLike], **env_options: typing.Any
     ) -> "jinja2.Environment":
         @pass_context
         def url_for(context: dict, name: str, **path_params: typing.Any) -> str:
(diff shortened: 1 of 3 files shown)


starlette/templating.py
```diff
@@ -55,20 +55,25 @@ class Jinja2Templates:
...
+        self, directory: typing.Union[str, PathLike], **env_options: typing.Any
     ) -> "jinja2.Environment":
         @pass_context
         def url_for(context: dict, name: str, **path_params: typing.Any) -> str:
             request = context["request"]
             return request.url_for(name, **path_params)
 
         loader = jinja2.FileSystemLoader(directory)
-        env = jinja2.Environment(loader=loader, autoescape=True)
+        env_options.setdefault("loader", loader)
```
aminalaee, line 73: We can decide if we want to allow loader to be passed in env_options or not. The point is if we add it it makes `directory` reduntant but allows user full control over the loader.

lovelydinosaur (approved):
Neat. 👍

aminalaee:
The documentation was not done yet. Can you check again?

lovelydinosaur:
Yup - I oughta have noted that. 😅
Docs look good.

PR #1402, opened 2022-01-10
Check if `status_code` is `None` on `Response`

Kludex:
Either here or https://github.com/tiangolo/fastapi/pull/4411

It makes more sense to have this on FastAPI, but the `RedirectResponse` makes it difficult to be there...

EDIT: Deciding if this is going to be solved by Starlette or FastAPI should be done before the next release.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -74,7 +74,11 @@ class Response:
         if (
             body is not None
             and populate_content_length
-            and not (self.status_code < 200 or self.status_code in (204, 304))
+            and not (
+                self.status_code
+                and self.status_code < 200
+                or self.status_code in (204, 304)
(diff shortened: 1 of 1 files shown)


Kludex:
If we chose this path instead of the FastAPI one, then we need to change the type annotation to `Optional[int]` (?).

lovelydinosaur:
This doesn't make sense to me. It's annotated as an `int`, it defaults to an int if no value is provided.
Setting it to `None` in the constructor isn't something we support.
(Also it will end up with an invalid ASGI message if you do that.)

Kludex:
My concern was with FastAPI, as it will break the logic flow, but I have the same thoughts.

PR #1403, opened 2022-01-11
feat: fully type annotate datastructures.py

adriangb:
This adds annotations to `datastructes.py`.
Mainly this is making the mapping-like classes generic.
This means that `FormData`, `Headers`, etc. can be properly type annotated.


adriangb:
I'm not going to do it in this PR, but it would also be good to make the arguments to `ImmutableMultiDict.get` keyword only arguments like they are in the base `Mapping` classes, at least from a type annotation perspective (so static type checkers would break, but runtime would be the same, making this is mostly non-breaking change)

starlette/requests.py
```diff
@@ -136,7 +136,7 @@ def cookies(self) -> typing.Dict[str, str]:
     @property
     def client(self) -> Address:
         host, port = self.scope.get("client") or (None, None)
-        return Address(host=host, port=port)
+        return Address(host=host, port=port)  # type: ignore[arg-type]
```
Kludex: Hmmm... This is actually not ASGI compliant: https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope

> client (Iterable[Unicode string, int]) – A two-item iterable of [host, port], where host is the remote host’s IPv4 or IPv6 address, and port is the remote port as an integer. Optional; if missing defaults to None.

starlette/datastructures.py
```diff
@@ -1,19 +1,35 @@
...
+    host: typing.Optional[str]
+    port: typing.Optional[int]
 
 
 class URL:
     def __init__(
-        self, url: str = "", scope: Scope = None, **components: typing.Any
+        self,
+        url: str = "",
+        scope: typing.Optional[Scope] = None,
```
Kludex, line 31:
```suggestion
        scope: Scope = None,
```

starlette/datastructures.py
```diff
@@ -215,6 +231,14 @@ def __init__(self, value: typing.Union[str, typing.Sequence[str]]):
     def __len__(self) -> int:
         return len(self._items)
 
+    @typing.overload
+    def __getitem__(self, index: int) -> str:
+        ...  # pragma: no cover
+
+    @typing.overload
+    def __getitem__(self, index: slice) -> typing.Sequence[str]:
+        ...  # pragma: no cover
```
Kludex, line 240: If this PR doesn't get merged, you can also create a separate PR for this. :+1:

starlette/datastructures.py
```diff
@@ -246,7 +270,7 @@ def __init__(
         if kwargs:
             value = (
                 ImmutableMultiDict(value).multi_items()
-                + ImmutableMultiDict(kwargs).multi_items()
+                + ImmutableMultiDict(kwargs).multi_items()  # type: ignore[operator]
```
Kludex, line 274: Why is not a valid operation?

starlette/datastructures.py
```diff
@@ -266,33 +290,43 @@ def __init__(
...
+    def get(
+        self, key: _KT, default: typing.Optional[typing.Union[_VT_co, _T]] = ...
+    ) -> typing.Union[_VT_co, _T]:
+        ...  # pragma: no cover
+
+    def get(self, key: _KT, default: typing.Any = None) -> typing.Any:
         if key in self._dict:
-            return self._dict[key]
+            return typing.cast(_VT_co, self._dict[key])
         return default
```
Kludex: Not sure if I fully understand the reason of what is written here. :thinking:

starlette/requests.py
adriangb: How is this not ASGI compliant? The NamedTuple _is_ a tw-item utterable of [host, port]. It's the same thing we had before. In fact, I don't think we even need the `# type: ignore`. I just removed it in 417a3d7 and now there are 0 changes in this file.

starlette/datastructures.py
adriangb: Could you explain this change request to me please, or reference some documentation on the subject? I understand that there is both a practical and semantic difference, but as far as I understand using `Scope | None` is now preferred if the default value is `None`. For example, see https://stackoverflow.com/questions/62732402/can-i-omit-optional-if-i-set-default-to-none

starlette/datastructures.py
adriangb: Yep may have to split this up

starlette/datastructures.py
adriangb: because the type of `value` cannot be inferred from the expression `value = args[0] if args or []`

starlette/datastructures.py
adriangb: This is saying:

- If you do not provide a default, the return value is going to be of type `_VT_co` (so a value of the dictionary) (or an error)
- If you do provide a default value, the return value is going to be either `_VT_co` (if the key is found) or the type of your default value (if the key is not found)

adriangb: I removed the `typing.cast` by giving `ImmutableMultiDict._dict` a type annotation

starlette/datastructures.py
adriangb: this could be fixed with a bit of juggling of how this is handled, but I wanted to do this with the least amount of runtime changes necessary, so adding a `type: ignore` for now and maybe doing a rework in a future PR seems like the best bet

starlette/requests.py
Kludex: Because `Address` is not supposed to accept (`None`, `None`), client should actually be `Optional[Address]`.

My bad anyway, I should have mentioned that it's not because of your PR... I just noticed that it's currently wrong.

adriangb: let's fix it in another PR then 😄

adriangb:
Closing in favor of:

- https://github.com/encode/starlette/pull/1447
- https://github.com/encode/starlette/pull/1448
- https://github.com/encode/starlette/pull/1449

Once / if #1449 gets merged, I'll make follow ups for `MultiDict`, `Headers` and `FormData` (which all depend on `ImmutableMultiDict`)

starlette/requests.py
adriangb: Opened #1462

PR #1404, opened 2022-01-11
feat: add type annotation to Request.path_params

adriangb:


starlette/requests.py
```diff
@@ -59,7 +58,7 @@ class ClientDisconnect(Exception):
     pass
 
 
-class HTTPConnection(Mapping):
+class HTTPConnection(typing.Mapping[str, typing.Any]):
```
aminalaee: Isn't `typing.Mapping` dropped in favor of `collections.abc.Mapping`?
I actually expected this to fail in Python 3.10 😄 Am I missing something?

aminalaee: Reference [here](https://docs.python.org/3/library/typing.html#typing.Mapping) and PEP 585 https://www.python.org/dev/peps/pep-0585/

adriangb: `collections.Mapping` doesn't work with generics on < 3.9, that functionality was introduced to enable the changes proposed in PEP 585.

My plan was to use `typing.Mapping` until Starlette's minimum version is 3.9, then switch over to `collections.abc.Mapping`.

PEP 585 says:
> The deprecated functionality will be removed from the typing module in the first Python version released 5 years after the release of Python 3.9.0.

So I think this should work fine.

aminalaee: > My plan was to use typing.Mapping until Starlette's minimum version is 3.9, then switch over to collections.abc.Mapping

You're right. Do you think it's worth reflecting that in the code?
I mean some conditional import based on version so we can see when this switch can happen. There's still a few years until that time (Python 3.9 being minimum version).

I'll leave it for more feedback 👍

adriangb: Yeah I was thinking about that too. Not sure how I feel about the dynamically setting the base class 🤔, but it may be the best option.

In any case, I realized we don't _need_ this for this PR. So I reverted the change and will resolve this thread 😄

lovelydinosaur (approved):
Okay. 🤷‍♂️

PR #1409, opened 2022-01-13
Remove count of available convertors

Mng-dev-ai:
This PR fix count of available convertors which are five not four as declared in docs.

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -39,7 +39,7 @@ Route('/users/{username}', user)
 ```
 By default this will capture characters up to the end of the path or the next `/`.
 
-You can use convertors to modify what is captured. Four convertors are available:
+You can use convertors to modify what is captured. Five convertors are available:
 
 * `str` returns a string, and is the default.
 * `int` returns a Python integer.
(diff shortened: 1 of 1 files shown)


docs/routing.md
```diff
@@ -39,7 +39,7 @@ Route('/users/{username}', user)
 ```
 By default this will capture characters up to the end of the path or the next `/`.
 
-You can use convertors to modify what is captured. Four convertors are available:
+You can use convertors to modify what is captured. Five convertors are available:
```
aminalaee: Maybe remove that number completely so we won't face this again?
I mean the list is displayed right here, what's the importance of counting it?

aminalaee (commented):
Thanks for the PR.

docs/routing.md
Mng-dev-ai: Yes, That's would be much better.
Do you wanna me do it ?

Kludex: If you're willing, it would be cool. 😁

Mng-dev-ai: Done.

Kludex (approved):

PR #1412, opened 2022-01-14
Use typing `NoReturn`

aminalaee:
In a few places where we always raise an exception the type hint should be `NoReturn`.

Any feedback would be appreciated.


Kludex (approved):

starlette/datastructures.py
```diff
@@ -163,7 +163,7 @@ class URLPath(str):
 
     def __new__(cls, path: str, protocol: str = "", host: str = "") -> "URLPath":
         assert protocol in ("http", "websocket", "")
-        return str.__new__(cls, path)  # type: ignore
+        return str.__new__(cls, path)
```
aminalaee, line 166: This ignore and the next one are not used.

PR #1415, opened 2022-01-15
Sort third-party packages and add `starlette-wtf`

aminalaee:
According to discussion comment [here](https://github.com/encode/starlette/discussions/1411#discussioncomment-1968621) adds starlette-wtf package to the docs.


aminalaee:
I can see the third-party packages can be improved. We can sort the packages alphabetically, the same as Third Party Middlewares, and move the middleware to Third Party Middlewares.

Maybe even include #1414 here as we'll move a few things around.

aminalaee:
Right now we have `Starlette Context`, `Starlette Cramjam` and `Starlette OAuth2 API` in third party packages but they're middleware and should be moved to Third Party Middlewares.

docs/third-party-packages.md
```diff
@@ -3,7 +3,6 @@ Starlette has a rapidly growing community of developers, building tools that int
 
 Here are some of those third party packages:
 
-
```
adriangb, line 6: Should we maybe be using a markdown linter / formatter to control newlines and stuff like that?

adriangb (approved):

docs/third-party-packages.md
aminalaee: I think it would be great if we find one, otherwise we should just do it in PR review.

lovelydinosaur: +1 for just doing it in PR review.
Not a big deal either ways, and it's not worth the extra tooling.

Issue #814, opened 2020-01-29
How to use request.url_for in mounted apps


jussiarpalahti:
I'm trying to compose apps from smaller Starlette apps. I can't find a way to make mounted apps that work regardless of app that is mounting them when using url_for.

In code below the mounted app can't use its *internal* names. It has to be somehow made aware that it's now running under namespace set by root app. Otherwise route cannot be found. With correct name hierarchy reversing URL does work.

```python
from starlette.applications import Starlette
from starlette.responses import PlainTextResponse
from starlette.routing import Route, Mount

async def sub(request):
    return PlainTextResponse(f"Subpage {request.url_for('sub')}") # doesn't work
#  return PlainTextResponse(f"Subpage {request.url_for('under:sub')}") # works

subroutes = [
    Route("/", endpoint=sub, name="sub")
]

subapp = Starlette(routes=subroutes)

async def homepage(request):
    return PlainTextResponse("Homepage")

routes = [
    Mount("/under", subapp, name="under"),
    Route("/", endpoint=homepage, name="root")
]

app = Starlette(routes=routes)
```

sm-Fifteen:
Are you trying to get `/under/` (the canonical path, possibly with the app's root_path prepended to it) or `/` (the app-local path) here?

I remember trying something similar under Starlette 12.12 when reporting tiangolo/fastapi#829 and it works mostly fine if I convert it to vanilla Starlette using the old syntax.

```py
from starlette.applications import Starlette
from starlette.responses import JSONResponse
from starlette.routing import Route, Mount, Request

app = Starlette()

@app.route("/app", ['GET'])
def read_main():
    return JSONResponse({"message": "Hello World from main app"})

subapp = Starlette()

@subapp.route("/sub", ['GET'])
def read_sub(request: Request):
    return JSONResponse({
        #"root_path": request.scope['root_path'],
        #"raw_path": request.scope['raw_path'],
        #"path": request.scope['path'],
        "req_url_for": request.url_for("read_sub"),
        "app_url_for": app.url_path_for("read_sub"),
        "subapp_url_for": subapp.url_path_for("read_sub"),
    })

app.mount("/subapp", subapp)
```

```json
{
"req_url_for":"http://127.0.0.1:8000/subapp/sub",
"app_url_for":"/subapp/sub",
"subapp_url_for":"/sub"
}
```

[...]

PR #1416, opened 2022-01-15
Let Request.url_for lookup routes in mounted app first

plankthom:
Proposed fix for https://github.com/encode/starlette/issues/814

This modifies the lookup of routes by name to first look in the routes of the mounted app before resolving routes in the root app.

#### Motivation
As mentioned by @jussiarpalahti in  by https://github.com/encode/starlette/issues/814#issuecomment-593104008, this lookup algorithm makes more sense if the mounted sub-application is maintained as an independent module. 

In that case you would not expect the sub-application to know under what name (if any) it is mounted. 

With this feature, you can even mount the same application twice under different routes, without any need to configure the mount prefix as an application property.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -95,14 +95,23 @@ class HTTPConnection(Mapping):
         return self._url
 
     @property
-    def base_url(self) -> URL:
-        if not hasattr(self, "_base_url"):
+    def app_root_base_url(self) -> URL:
+        if not hasattr(self, "_app_root_base_url"):
             base_url_scope = dict(self.scope)
             base_url_scope["path"] = "/"
             base_url_scope["query_string"] = b""
             base_url_scope["root_path"] = base_url_scope.get(
                 "app_root_path", base_url_scope.get("root_path", "")
             )
+            self._app_root_base_url = URL(scope=base_url_scope)
+        return self._app_root_base_url
+
+    @property
+    def base_url(self) -> URL:
+        if not hasattr(self, "_base_url"):
+            base_url_scope = dict(self.scope)
+            base_url_scope["path"] = "/"
+            base_url_scope["query_string"] = b""
             self._base_url = URL(scope=base_url_scope)
         return self._base_url
 
@@ -170,9 +179,16 @@ class HTTPConnection(Mapping):
         return self._state
 
     def url_for(self, name: str, **path_params: typing.Any) -> str:
-        router: Router = self.scope["router"]
-        url_path = router.url_path_for(name, **path_params)
-        return url_path.make_absolute_url(base_url=self.base_url)
+        app_root_router = self.scope["router"]
+        app_router = getattr(self.get('app', {}), 'router', None)
+        if app_router and app_router != app_root_router:
+            try: 
+                url_path = app_router.url_path_for(name, **path_params)
+                return url_path.make_absolute_url(base_url=self.base_url)
+            except Exception:
+                pass
+        url_path = app_root_router.url_path_for(name, **path_params)
+        return url_path.make_absolute_url(base_url=self.app_root_base_url)
 
 
 async def empty_receive() -> typing.NoReturn:
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -517,6 +517,40 @@ def test_url_for_with_root_path(test_client_factory):
     }
 
 
+
+def echo_urls_for(*keys):
+    def echo_urls(request):
+        return JSONResponse(
+            {
+                k: request.url_for(k)
+                for k in keys
+            }
+        )
+    return echo_urls
+
+def test_url_with_sub_app(test_client_factory):
+    sub_app = Starlette(
+        routes=[Route("/",  echo_urls_for('subapp_index'), name="subapp_index", methods=["GET"])
+    ])
+    app = Starlette(
+        routes=[
+            Route("/", echo_urls_for('index', 'submount:subapp_index'), name="index", methods=["GET"]),
+            Mount('/submount', app=sub_app, name='submount')
+        ])
+    
+    client = test_client_factory(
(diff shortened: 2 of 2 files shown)


aminalaee:
@plankthom Thank you for the PR. Can you please explain a bit about this?

As far as I could see this was not really a bug, this was the intended behaviour.

For doing reverse URL lookup on sub-mounted apps you'd need `request.url_for('sub:<name>')` and calling `request.url_for(<name>)` should not work.

plankthom:
@aminalaee : I`ve added some motivation in the PR description.

A FastApi example
`test/trinkets.py`
```python
from fastapi import FastAPI, Request

app = FastAPI(title='Trinkets')
trinkets = [ {'name': 'nice'}, {'name': 'colorful'} ]

@app.get('/')
def info(request: Request):
    return {
        '_links: {'trinkets': {'href':  request.url_for('list_trinkets')}}
    }

@app.get('/trinket')
def list_trinkets(request: Request):
    return [
        {
            **trinket,
            '_links': {'self': {'href': request.url_for('get_trinket', trinket_name=trinket['name'])}}
        }
        for trinket in trinkets
    ]


@app.get('/trinket/{trinket_name}')
def get_trinket(trinket_name: str, request: Request):
    return next(t for t in trinkets if t['name'] == trinket_name)
```
`test/main.py`
```python
import fastapi
import trinket

app = fastapi.FastAPI(title='Trinkets and Treasures')

@app.get('/')
def info(request: fastapi.Request):
    return { 
        'trinkets_v1' :  request.url_for('v1:list_trinkets'),
        'trinkets_v2' :  request.url_for('v2:list_trinkets')
    }

app.mount('/api/v1', app=trinket.app, name= 'v1')
app.mount('/api/v2', app=trinket.app, name= 'v2')
```

`curl http://localhost:8000 | jq`
```json
{
  "trinkets_v1": "http://localhost:8000/api/v1/trinket",
  "trinkets_v2": "http://localhost:8000/api/v2/trinket"
}
```
`curl curl http://localhost:8000/api/v2/trinket`
```json
[
  {
    "name": "nice",
    "_links": {
      "self": {
        "href": "http://localhost:8000/api/v2/trinket/nice"
      }
    }
  },
  {
    "name": "colorful",
    "_links": {
      "self": {
        "href": "http://localhost:8000/api/v2/trinket/colorful"
      }
    }
  }
]
```

plankthom:
Note that current fastapi treats these sub-apps independently in their documentation (e.g. documented with the local paths):

`http://localhost:8000/api/v1/docs`
image

(the current [fastapi docs on that topic](https://fastapi.tiangolo.com/advanced/sub-applications/?h=mount#check-the-automatic-api-docs) lag a bit behind)

jussiarpalahti:
Hi. Original issue author here. Thank you @plankthom for looking into this.

My original use case was to make several small, reusable ASGI apps and combine them into larger app instances. Their URLs and URL namespace could differ depending on needs of the main app. There shouldn't be a need for any of the sub-apps to know where they reside, if ASGI framework can manage their URLs. Examples of such apps I had was authorization, and pub/sub messaging.

plankthom:
@aminalaee did previous comments clarify the intent of this PR? If so would you consider reviewing it?

alex-oleshkevich:
@Kludex I want this too. This is smth I suffer too. 

Here is my example: https://github.com/alex-oleshkevich/ohmyadmin/blob/master/ohmyadmin/resources.py#L485
App can have many Resource classes mounted under different paths. And it is impossible to generate a valid URL path to the route within the mounted resource by name. I had to wrap it in a hackish ` name=self.url_name('edit')` to add a unique prefix.

See
```python
class Resource(Router):
    def __init__():
        super().__init__(routes=[
			Route('/', self.index_view, name='index'),
			Route('/edit', self.edit_view, name='edit'),
       ])

app = Starlette(routes=[
	Mount('/app/products', Resource()),
	Mount('/app/users', Resource()),
	Mount('...', Resource()),
])
```
There is no easy way to generate a URL path for `edit` from `index` and vice versa:
```python
def index_view(self, request):
	return self.request.url_path(???) # 'edit' is a global name
```
You have to prefix route names to make them unique:
```python
Route('/', self.index_view, name='products_index'),
Route('/', self.index_view, name='users_index'),
```
And do this for all. This is error-prone and too verbose.

If you decide to namespace URL in Mount everything worsens. You cannot generate URL to `/app/users` without knowing the namespace name defined by the user (is it 'users' in this example).

```python
Mount('/app/users', Resource(), name='users'),

# in view code of 3rd-party library
def index_view(self, request):
	return self.request.url_path(namespace???:routename???) # 'edit' is a global name
```

I think we can do something like this:
```python
self.request.url_path('.edit')
self.request.url_path('.index')
```

If route name starts with dot, then we look up the current ASGI app first (inner router for example) and then iterate parents"
```python
self.request.url_path('.edit') # works, returns `/api/products/edit`
self.request.url_path('.index') # works, returns `/api/products`

self.request.url_path('home') # works, returns `/`
self.request.url_path('index') # works, returns url for `index` route name, note it is not prefixed with dot.
```

The Flask does it in the same way with blueprints - https://flask.palletsprojects.com/en/2.2.x/blueprints/#building-urls

plankthom:
@Kludex @alex-oleshkevich I can update this PR with the proposal (use `.` prefixed paths to have resolution local to the inner router) ... This will indeed improve backward compatibility, although IMO there is very little risk at incompatibility in the current proposal  (other than that previously broken lookups would work now).

Kludex:
Can someone reply [this](https://github.com/encode/starlette/issues/814#issuecomment-1236103161)? Then we can proceed...

Kludex:
Closing it as stale since no one replied my last question.

Happy to reopen. 👍

PR #1417, opened 2022-01-15
Add reason to WebSocket closure

### LINKED ISSUE #991: [Feature] Reason for websocket.close

aminalaee:
Closes #991.

According to the ASGI spec [here](https://asgi.readthedocs.io/en/latest/specs/www.html#close-send-event) `reason` can be included in `websocket.close` which is an optional string.

<del>I think what @Kludex mentioned [here](https://github.com/encode/starlette/pull/1339#pullrequestreview-812914693) makes sense:</del>

<del>- We can include `code` in `websocket.close` only when it's specified in Starlette, and don't include it when not set, as per the ASGI spec it is optional int and already handled by Uvicorn [here](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/websockets/websockets_impl.py#L261) and Daphne [here](https://github.com/django/daphne/blob/main/daphne/ws_protocol.py#L262) and Hypercorn [here](https://gitlab.com/pgjones/hypercorn/-/blob/main/src/hypercorn/protocol/ws_stream.py#L293). </del>
- We will include `reason` in `websocket.close` only when it's specified as per the ASGI spec it's optional and will be handled by ASGI servers.

Any feedback would be appreciated.


tests/test_websockets.py
```diff
@@ -391,3 +391,39 @@ async def mock_send(message):
...
+        assert exc.value.reason == "Closing"
+
+
+def test_websocket_close_reason_manual(test_client_factory) -> None:
+    def app(scope):
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=send)
+            await websocket.accept()
+
+            websocket_close = WebSocketClose(code=1001, reason="Closing")
```
aminalaee: This test is different from the above test, as that calls `close` on websocket but this one will await `WebSocketClose` manually. (For coverage)

docs/websockets.md
```diff
@@ -75,7 +75,7 @@ Use `websocket.receive_json(data, mode="binary")` to receive JSON over binary da
 
 ### Closing the connection
 
-* `await websocket.close(code=1000)`
+* `await websocket.close(code=1000, reason="Normal Closure")`
```
Kludex:
```suggestion
* `await websocket.close(code=1000, reason=None)`
```

starlette/websockets.py
```diff
@@ -13,8 +13,9 @@ class WebSocketState(enum.Enum):
 
 
 class WebSocketDisconnect(Exception):
-    def __init__(self, code: int = 1000) -> None:
+    def __init__(self, code: int = None, reason: str = None) -> None:
```
```suggestion
    def __init__(self, code: int = 1000, reason: str = None) -> None:
```

starlette/websockets.py
```diff
@@ -144,13 +145,28 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
         else:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
-    async def close(self, code: int = 1000) -> None:
-        await self.send({"type": "websocket.close", "code": code})
+    async def close(self, code: int = None, reason: str = None) -> None:
+        message: dict = {"type": "websocket.close"}
+
+        if code is not None:
+            message["code"] = code
```
```suggestion
        message = {"type": "websocket.close", "code": code}

```

starlette/websockets.py
```diff
@@ -144,13 +145,28 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
...
+        if reason is not None:
+            message["reason"] = reason
+
+        await self.send(message)
 
 
 class WebSocketClose:
-    def __init__(self, code: int = 1000) -> None:
+    def __init__(self, code: int = None, reason: str = None) -> None:
```
```suggestion
    def __init__(self, code: int = 1000, reason: str = None) -> None:
```

starlette/websockets.py
```diff
@@ -144,13 +145,28 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
...
         self.code = code
+        self.reason = reason
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        await send({"type": "websocket.close", "code": self.code})
+        message: dict = {"type": "websocket.close"}
+
+        if self.code is not None:
+            message["code"] = self.code
```
```suggestion
        message = {"type": "websocket.close", "code": code}

```

Kludex (commented):
I understand the default value change on `code`, but that is another issue, and it's also considered a breaking change.

starlette/websockets.py
```diff
@@ -144,13 +145,22 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
         else:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
-    async def close(self, code: int = 1000) -> None:
-        await self.send({"type": "websocket.close", "code": code})
+    async def close(self, code: int = 1000, reason: str = None) -> None:
+        if reason is None:
+            await self.send({"type": "websocket.close", "code": code})
+        else:
+            await self.send({"type": "websocket.close", "code": code, "reason": reason})
```
Kludex: Should we do this?

```suggestion
    await WebSocketClose(code, reason)
```

tests/test_websockets.py
Kludex: This doesn't look cool. For this case, I guess we'd want to test the `WebSocketClose` class standalone, but see my comment above.

starlette/websockets.py
aminalaee: That makes sense but it should be something like:

```python
websocket_close = WebSocketClose(code, reason)
await websocket_close(self.scope, self.receive, self.send)
```

aminalaee: What do you think about this?

Kludex: Hmm, I think it's better to ignore my comment. :+1: 

The way it is already, it's easier to understand.

tests/test_websockets.py
Kludex: But this should probably be tested elsewhere?

aminalaee: I don't think we need the standalone anymore, thanks to your suggestion. I say let's drop it when there's no value in it.

aminalaee: The only place `WebSocketClose` is used is in the `routing` where we close the connection. And in those cases we rely on the defaults. So the case with `reason` is not covered.

starlette/websockets.py
aminalaee: Actually when I first did this I thought that maybe we should get rid of `WebSocketClose`. I know it's out of scope and backwards incompatible.

But what you suggest is even better, it will make use of `WebSocketClose`, as I think it was meant to.
If you don't have any strong feelings about it I say let's wait for someone else's feedback here.

starlette/websockets.py
```diff
@@ -146,10 +146,7 @@ async def send_json(self, data: typing.Any, mode: str = "text") -> None:
             await self.send({"type": "websocket.send", "bytes": text.encode("utf-8")})
 
     async def close(self, code: int = 1000, reason: str = None) -> None:
-        if reason is None:
-            await self.send({"type": "websocket.close", "code": code})
-        else:
-            await self.send({"type": "websocket.close", "code": code, "reason": reason})
+        await self.send({"type": "websocket.close", "code": code, "reason": reason})
```
Kludex: image

Kludex: `reason = reason or ""`

starlette/testclient.py
```diff
@@ -352,7 +352,7 @@ async def _asgi_send(self, message: Message) -> None:
 
     def _raise_on_close(self, message: Message) -> None:
         if message["type"] == "websocket.close":
-            raise WebSocketDisconnect(message.get("code", 1000))
+            raise WebSocketDisconnect(message.get("code", 1000), message.get("reason"))
```
Kludex: I think this should default to empty string, but at the end it doesn't make a difference. Your call.
```suggestion
            raise WebSocketDisconnect(message.get("code", 1000), message.get("reason", ""))
```

Kludex (approved):
Added a comment, but your call.

Everything is cool :+1: Thanks! :tada:

starlette/testclient.py
aminalaee: I noticed that earlier and changed `WebsocketDisconnect`:

```python
class WebSocketDisconnect(Exception):
     def __init__(self, code: int = 1000, reason: str = None) -> None:
         self.code = code
         self.reason = reason or ""
```

I think this will do that, right?

Kludex: Yeah, but the raise on the line that I've commented sends either the reason or None. In theory, `WebSocketDisconnect` doesn't receive `None` (looking at the annotation).

Kludex: The annotation is right.

aminalaee: Sorry I want to ask this out of curiosity, not for this PR.I get your point that in this internal code we will fix this.
But I guess both Python and MyPy (strict) will be ok about passing `None` to `reason`, right?

Kludex: If you mean the `strict` flag, then no. See below:

```bash
❯ mypy --strict main.py      
main.py:1: error: Incompatible default for argument "b" (default has type "None", argument has type "str")
main.py:4: error: Argument 1 to "a" has incompatible type "None"; expected "str"
Found 2 errors in 1 file (checked 1 source file)
(base) 
~/Development via 🐍 v3.8.8 on ☁️  marcelotryle@gmail.com 
❯ cat main.py                                                          
def a(b: str = None) -> None:
    return None

a(None)
```

Kludex: (we don't use the strict flag on starlette, jfyk)

aminalaee: But this will be ok:

```python
import typing

def a(b: typing.Optional[str] = None) -> None:
    return None

a(None)
```

I'll fix this as it's not probably related and we can discuss later. Thank you.

PR #1420, opened 2022-01-16
Bunch of improvements on authentication documentation

mkniewallner:
`AuthenticationBackend.authenticate` signature was updated [here](https://github.com/encode/starlette/pull/335/files#diff-44db7dbf84f50ab50eaa07be93238de53ac2d2363189fd6eef7f1de5b62752dfR71) to handle both HTTP and websocket protocols, but the documentation still uses `request`.

Also removing `UnauthenticatedUser` import in `AuthenticationBackend` example, as it is not used, and adding some missing imports in a few examples.

Side-note: The documention mixes single and double quotes usage for strings. Should a consistent style be used? Maybe https://github.com/asottile/blacken-docs could be helpful?

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -135,6 +135,10 @@ For class-based endpoints, you should wrap the decorator
 around a method on the class.
 
 ```python
+from starlette.authentication import requires
+from starlette.endpoints import HTTPEndpoint
+
+
 class Dashboard(HTTPEndpoint):
     @requires("authenticated")
     async def get(self, request):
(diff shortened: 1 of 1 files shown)


Kludex (approved):

Kludex:
> Side-note: The documention mixes single and double quotes usage for strings. Should a consistent style be used? Maybe https://github.com/asottile/blacken-docs could be helpful?

I recommend a GitHub discussion or introducing the idea on our Gitter server.

Thanks for the PR! :tada:

PR #1422, opened 2022-01-20
Default WebSocket accept message headers to an empty list

Klavionik:
Recently, #1361 introduced a possibility to pass a set of extra headers into the `accept` method. If no headers are passed, the default value is `None`. This value then gets sent to the protocol server.

This causes Uvicorn to raise an exception: `TypeError: NoneType object is not iterable` when using websockets and `TypeError: can only concatenate list (not "NoneType") to list` when using wsproto.

This PR fixes this, setting the `headers` argument to an empty list if it is not passed into the method. This is recommended by the ASGI spec (https://asgi.readthedocs.io/en/latest/specs/www.html#accept-send-event).

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -74,8 +74,7 @@ class WebSocket(HTTPConnection):
         subprotocol: str = None,
         headers: typing.Iterable[typing.Tuple[bytes, bytes]] = None,
     ) -> None:
-        if headers is None:
-            headers = []
+        headers = headers or []
 
         if self.client_state == WebSocketState.CONNECTING:
             # If we haven't yet seen the 'connect' message, then wait for it first.
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -74,6 +74,9 @@ class WebSocket(HTTPConnection):
         subprotocol: str = None,
         headers: typing.Iterable[typing.Tuple[bytes, bytes]] = None,
     ) -> None:
+        if headers is None:
+            headers = []
+
         if self.client_state == WebSocketState.CONNECTING:
             # If we haven't yet seen the 'connect' message, then wait for it first.
             await self.receive()
--- a/tests/test_websockets.py
+++ b/tests/test_websockets.py
@@ -316,27 +316,17 @@ def test_additional_headers(test_client_factory):
 
 
 def test_no_additional_headers(test_client_factory):
-    expected_headers = []
-
-    def wrap(send):
-        async def _send(message):
-            if "headers" in message:
-                actual_headers = message.get("headers")
-                assert actual_headers == expected_headers
-            return await send(message)
-        return _send
-
     def app(scope):
         async def asgi(receive, send):
-            websocket = WebSocket(scope, receive=receive, send=wrap(send))
+            websocket = WebSocket(scope, receive=receive, send=send)
             await websocket.accept(headers=None)
             await websocket.close()
 
         return asgi
 
     client = test_client_factory(app)
-    with client.websocket_connect("/"):
-        pass  # pragma: nocover
+    with client.websocket_connect("/") as websocket:
+        assert websocket.extra_headers == []
 
 
 def test_websocket_exception(test_client_factory):
--- a/tests/test_websockets.py
+++ b/tests/test_websockets.py
@@ -319,7 +319,7 @@ def test_no_additional_headers(test_client_factory):
     def app(scope):
         async def asgi(receive, send):
             websocket = WebSocket(scope, receive=receive, send=send)
-            await websocket.accept(headers=None)
+            await websocket.accept()
             await websocket.close()
 
         return asgi
--- a/tests/test_websockets.py
+++ b/tests/test_websockets.py
@@ -315,6 +315,30 @@ def test_additional_headers(test_client_factory):
         assert websocket.extra_headers == [(b"additional", b"header")]
 
 
+def test_no_additional_headers(test_client_factory):
+    expected_headers = []
+
+    def wrap(send):
+        async def _send(message):
+            if "headers" in message:
+                actual_headers = message.get("headers")
+                assert actual_headers == expected_headers
+            return await send(message)
+        return _send
+
+    def app(scope):
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=wrap(send))
+            await websocket.accept(headers=None)
+            await websocket.close()
+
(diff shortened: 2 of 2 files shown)


aminalaee:
Thank you for this. I guess adding/editing the test would be valuable here so this is safe for future changes.
Maybe in the `test_additional_headers` test?

Klavionik:
I've added a test, but I'm really not sure how adequate it is to do it that way. :) Just let me know if it's not.

tests/test_websockets.py
```diff
@@ -315,6 +315,30 @@ async def asgi(receive, send):
...
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=wrap(send))
+            await websocket.accept(headers=None)
+            await websocket.close()
+
+        return asgi
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/"):
+        pass  # pragma: nocover
```
aminalaee:
```suggestion
def test_no_additional_headers(test_client_factory):
    def app(scope):
        async def asgi(receive, send):
            websocket = WebSocket(scope, receive=receive, send=send)
            await websocket.accept(headers=None)
            await websocket.close()

        return asgi

    client = test_client_factory(app)
    with client.websocket_connect("/") as websocket:
        assert websocket.extra_headers == []
```

aminalaee: I think this will do that. I thought about using pytest.mark.parametize to handle all cases in one test but because of nested list and tuples it wasn't much readable.

Kludex:
My interpretation of the ASGI spec is that the server will interpret the absence of `headers` as an empty array, not the application. If my understanding is correct, I think we should send the message without `headers` in this case.

aminalaee:
Yes, I agree that's what it says. Wouldn't that be a breaking change? I mean that's like the case with `code`.

Kludex:
No, because we didn't release it yet. This is a feature introduced a few days ago.

aminalaee:
Oh that makes sense, I thought it was from the last release 😆

starlette/websockets.py
```diff
@@ -74,6 +74,9 @@ async def accept(
         subprotocol: str = None,
         headers: typing.Iterable[typing.Tuple[bytes, bytes]] = None,
     ) -> None:
+        if headers is None:
```
aminalaee, line 77: As suggested by Kludex please remove the `headers` from message when it is None.

Kludex:
Should we also change the `TestClient.extra_headers` default value as well? https://github.com/encode/starlette/pull/1361#discussion_r789552938

Klavionik:
@Kludex I see your point, but I'm not really sure about it, to be honest. 

My humble reflection on this topic:

why should an application care about how exactly the server will interpret the absence of headers - be it an empty list or anything else? Why should the spec tell us about that? I'm leaning towards thinking that this is an instruction for application developers so that server developers would know what to expect. And in accordance to that, right now Uvicorn expects headers to be something iterable.

Anyway, if so, shouldn't we also remove the `subprotocol` too, when it is None? It's a similar case, I guess.

Kludex:
> And in accordance to that, right now Uvicorn expects headers to be something iterable.

Uvicorn only expects the headers to be iterable if **present**.

> Anyway, if so, shouldn't we also remove the `subprotocol` too, when it is None? It's a similar case, I guess.

But yeah, you have a point. Ok, I'm fine with this :+1:

starlette/websockets.py
```diff
@@ -74,6 +74,9 @@ async def accept(
...
+            headers = []
```
```suggestion
        headers = headers or []
```

matiuszka:
@Klavionik Thanks for fixing that. I completely forgot about that.

aminalaee:
> Anyway, if so, shouldn't we also remove the `subprotocol` too, when it is None? It's a similar case, I guess.

I think we should pick one approach and stick with it.
In some cases we do send the optional values <del>even if they are None</del>, like `subprotocol` and `code`.
In some other places like the `reason` we don't send the value if it's None (still under review).

Kludex:
> > Anyway, if so, shouldn't we also remove the `subprotocol` too, when it is None? It's a similar case, I guess.
> 
> I think we should pick one approach and stick with it. In some cases we do send the optional values even if they are None, like `subprotocol` and `code`. In some other places like the `reason` we don't send the value if it's None (still under review).

We send None when it's allowed by the ASGI spec, but yeah. I agree. We should probably adapt the `reason` PR to stick with the others. :thinking:

starlette/websockets.py
Klavionik: Love this idiom, thanks! Sometimes I'm just too shy about using it as it's not that explicit, as `if ... is None`.

tests/test_websockets.py
Klavionik: Yes, I agree that it's better to create a dedicated test case. Thanks for the suggestion, it's much more clear.

Klavionik:
I accepted all the suggestions (first time using it, hope I did it right).

> @Klavionik Thanks for fixing that. I completely forgot about that.

Happy to help! I was actually tinkering with BlackSheep, trying to add support for WebSocket connections to it, when I discovered this. I was referencing Starlette and did the same thing with the `accept` method. Luckily I was testing it using the latest Uvicorn release - that's how I saw the server crashing on the `accept` call.

So maybe we should consider adding some kind of check to Uvicorn? Like `assert isinstance(headers, typing.Iterable), "The 'headers' argument must be an iterable."`? Just to make sure it will raise a more clear exception if someone tries to do the same thing.

tests/test_websockets.py
```diff
@@ -315,6 +315,20 @@ async def asgi(receive, send):
         assert websocket.extra_headers == [(b"additional", b"header")]
 
 
+def test_no_additional_headers(test_client_factory):
+    def app(scope):
+        async def asgi(receive, send):
+            websocket = WebSocket(scope, receive=receive, send=send)
+            await websocket.accept(headers=None)
```
Kludex: I'm not sure about this, because the `headers` parameter doesn't accept `None` as argument. The annotation there is `typing.Iterable[typing.Tuple[bytes, bytes]]`. We can do this instead:

```suggestion
            await websocket.accept()
```

Klavionik: I rather agree. I was worried that it would not be clear, what exactly we are testing.

Kludex (approved):

Kludex:
Thanks @Klavionik ! 🎉

PR #1423, opened 2022-01-21
Adjust type of `exception_handlers` to allow async callable

michaeloliverx:
In #1360 the type for `exception_handlers` was updated but it now only accounts for callables that return a `Response`. We can have sync and async handlers here so this change should fix that.


Kludex (approved):

aminalaee:
Great 👍

PR #1425, opened 2022-01-21
Move `exception_handlers` type to starlette.types

michaeloliverx:
I don't know if these are appropriate changes so I opened this in a draft.

I think using a Callback protocol [^callback-protocol] to define the exception handler functions is easier to read. 

I moved the type defintion to `starlette.types`. In my own applications I often construct a mapping(s) of exception handlers and pass it into the app like this:

```python

exception_handlers = {}

app = Starlette(exception_handlers=exception_handlers)

```

Now I can type it like this:

```python
from starlette.types import ExceptionHandlers

exception_handlers: ExceptionHandlers = {}

app = Starlette(exception_handlers=exception_handlers)
```

[^callback-protocol]: https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols


michaeloliverx:
@aminalaee @Kludex What do you think of this?

aminalaee:
As far as I know the approach is to _try_ to remove the current types.py file in the future, and the type hints are kept as simple as possible.
You can also discuss this on gitter https://gitter.im/encode/community

adriangb:
> You can also discuss this on gitter https://gitter.im/encode/community

If you can still find it, it would be great to copy the discussion to this issue. As far as I know, there's not way to "permalink" to a message or thread in Gitter.

PR #1426, opened 2022-01-22
use Mapping type for Response.headers

jouve:
Hi,

I have read your guideline and I think this fit the "clearly identified bug"

the `headers` arg of  `Response` (or it sub classes) is annotated as  `dict`.

It is *only* used as the argument of `init_headers(self, headers: typing.Mapping[str, str] = None) -> None:`

which accepts `typing.Mapping[str, str]`, so the PR replaces the `dict` type declaration by `typing.Mapping[str, str]`.

ref:
https://github.com/encode/starlette/blob/master/starlette/responses.py#L59

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -38,7 +38,7 @@ class Response:
         self,
         content: typing.Any = None,
         status_code: int = 200,
-        headers: dict = None,
+        headers: typing.Mapping[str, str] = None,
         media_type: str = None,
         background: BackgroundTask = None,
     ) -> None:
@@ -199,7 +199,7 @@ class RedirectResponse(Response):
         self,
         url: typing.Union[str, URL],
         status_code: int = 307,
-        headers: dict = None,
+        headers: typing.Mapping[str, str] = None,
         background: BackgroundTask = None,
     ) -> None:
         super().__init__(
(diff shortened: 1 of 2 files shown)


starlette/datastructures.py
```diff
@@ -631,7 +631,7 @@ def setdefault(self, key: str, value: str) -> str:
         self._list.append((set_key, set_value))
         return value
 
-    def update(self, other: dict) -> None:
+    def update(self, other: typing.Mapping[str, str]) -> None:
```
adriangb: Is this required for this PR? I'd think long term we want do implement this the same as in httpx:
https://github.com/encode/httpx/blob/321d4aa5097fe7f24cdfed7191c44de589294780/httpx/_models.py#L965-L970

jouve: this one was for consistency, but it's not part of  `Response` constructor, I can remove it.

I agree with you long term plan, but it's much more involved than this fix :)

for the funny story,  I noticed this issue by passing a  `httpx.Response.header` (which is a  `typing.MultableMapping`)  to `starlette.Response(header=...)`

adriangb: > this one was for consistency, but it's not part of Response constructor, I can remove it.

yeah since it's not directly related I would punt it until a future PR, but that's just my personal opinion, not a blocker for me

> for the funny story, I noticed this issue by passing a httpx.Response.header (which is a typing.MultableMapping) to starlette.Response(header=...)

hmm @tomchristie seems like something we should (eventually) fix

adriangb (approved):

starlette/datastructures.py
jouve: removed :)

the PR now changes only `Response` & its subclasses

Kludex (approved):

PR #1427, opened 2022-01-22
Add listener for disconnect to FileResponse

### LINKED ISSUE #1301: FileResponse Stream continues streaming after http.disconnect

idotobi:
This fixes #1301.

This makes `FileResponse` behave the same way as `StreamResponse`,
namely aborting sending data once a `http.disconnect` is received from the client.

Note that this behaviour is currently not tested in `StreamResponse`.
I found one way to test this behaviour. However, it would require to add `httpx` as a testing requirement hence it is not part of the PR yet.
`FileResponse` is even more tricky to test without having a way to mark a `FileResponse` as finished (would require a new property).


idotobi:
Note: The code for `__call__` and `listen_for_disconnect` are implemented exactly as in `StreamResponse`. It might be worthwhile to make `FileResponse` inherit from `StreamResponse` in the future as they are quite similar in nature.

Kludex:
Interesting. We don't have a specific test for `StramingResponse` as well, for the same case.

Kludex:
> I found one way to test this behaviour. However, it would require to add httpx as a testing requirement hence it is not part of the PR yet.

@idotobi How?

lovelydinosaur:
Hiya @idotobi - thank you for the neatly tackled bit of work here.

I'd like to close this off with the justification as outlined on https://github.com/encode/starlette/issues/1301

There's always a complexity trade-off in these kinds of cases, and my feeling is that Starlette ought to lean very strongly towards as-simple-as-possible-wherever-possible.

*(Something we need to get better at as a team is making sure we're fielding incoming requests like this, so that you're not sending time on neatly done pull requests like this one, and then having them closed off. We're working on that.)*

idotobi:
@Kludex : It was a rather messy prototype, where I had to remove some asserts within the httpx testclient, so I wouldn't recommend it for production usage.

The idea was basically to exchange the `receive` callback in the app definition 

```py
        async def receive() -> dict:
            return {"type": "http.disconnect"}
```

This way the server would abort the connection itself, if it listened to `http.disconnect`.
...As I said. Not pretty but sadly there is to Testclient I found that was able to send disconnect while waiting on the response out-of-the-box.

idotobi:
@tomchristie : No worries. I learned a bit on the way, so it's no time wasted.

PR #1428, opened 2022-01-23
Aman/brotli middleware

A1man1:
checking BrotliMiddlware as like gzipMiddleware

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ venv*/
 .python-version
 build/
 dist/
+env/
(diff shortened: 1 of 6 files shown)


Kludex:
Thanks for the PR @A1man1 !

Following our strategy to have a minimal package, I think this should not be included on Starlette itself.

Instead, we can add [brotli-asgi](https://github.com/fullonic/brotli-asgi) to the middleware list on our documentation.

In any case, if you or anyone else still think this should be included, you're welcome to start a discussion on our Gitter or create an issue.

A1man1:
thanks , I will read all documentation. I will trying to enhance instead create without see anything.

PR #1431, opened 2022-01-25
Make `content` argument required to `JSONResponse`

aminalaee:
Making `content` argument required in `JSONResponse`:

- `Response` will work as before defaulting to `b""` when `content` is `None`
- `JSONResponse()` will fail as `content` is required
- `JSONResponse({})` will work as expected
- `JSONResponse(None)` will be rendered as `null`


tests/test_responses.py
```diff
@@ -327,7 +327,7 @@ def test_head_method(test_client_factory):
 
 
 def test_empty_response(test_client_factory):
-    app = Response()
+    app = Response(status_code=200)
```
aminalaee, line 330: It might be worth considering to assert either `content` or `status_code` provided, both shouldn't be empty.

lovelydinosaur:
Okay - neat starting point.

We'll need to expand on the *actual* motivation here, and the desired behavioural changes, but lemme first jump in with some review points.

starlette/responses.py
```diff
@@ -33,25 +33,31 @@ def guess_type(
 class Response:
     media_type = None
     charset = "utf-8"
+    default_status_code = 200
+    default_response = b""
```
lovelydinosaur: Let's not have these attributes - it's not a bit of behaviour that we'd want subclasses to modify.

starlette/responses.py
```diff
@@ -33,25 +33,31 @@ def guess_type(
...
     ) -> None:
-        self.status_code = status_code
         if media_type is not None:
             self.media_type = media_type
         self.background = background
-        self.body = self.render(content)
+
+        if content is None:
+            self.body = self.default_response
+            self.status_code = status_code
```
lovelydinosaur:
```suggestion
            self.body = b""
            self.status_code = status_code or 204
```

starlette/responses.py
```diff
@@ -33,25 +33,31 @@ def guess_type(
...
+        else:
+            self.body = self.render(content)
+            self.status_code = status_code or self.default_status_code
```
lovelydinosaur:
```suggestion
            self.body = self.render(content)
            self.status_code = status_code or 200
```

starlette/responses.py
```diff
@@ -74,7 +80,11 @@ def init_headers(self, headers: typing.Mapping[str, str] = None) -> None:
         if (
             body is not None
             and populate_content_length
-            and not (self.status_code < 200 or self.status_code in (204, 304))
+            and not (
+                self.status_code
+                and self.status_code < 200
+                or self.status_code in (204, 304)
+            )
```
lovelydinosaur: This line shouldn't need to change.

starlette/responses.py
aminalaee: You're right, the `default_status_code` should be removed. But need the `default_response` to be modified by `JSONResponse`. Maybe that indicates an issue here but not calling render on json will create a non-valid jSON response b"".

starlette/responses.py
```diff
@@ -173,6 +183,7 @@ class PlainTextResponse(Response):
 
 class JSONResponse(Response):
     media_type = "application/json"
+    default_response = b"{}"
```
lovelydinosaur: Hrm... depends on what behaviour we want `return JSONResponse()` to have.

I'd *expect* we actually want that case to return a 204 with no content.

(Use a `return JSONResponse({})` if you wanted an empty dict serialized.)

tests/test_responses.py
```diff
@@ -327,7 +327,7 @@ def test_head_method(test_client_factory):
...
     client: TestClient = test_client_factory(app)
     response = client.get("/")
     assert response.headers["content-length"] == "0"
```
lovelydinosaur: I'd expect to see this behaviour...

```suggestion
    app = Response()
    client: TestClient = test_client_factory(app)
    response = client.get("/")
    assert response.headers["content-length"] == "0"
    assert response.status_code == 204
```

tests/test_responses.py
```diff
@@ -38,12 +38,12 @@ async def app(scope, receive, send):
...
 def test_json_none_response(test_client_factory):
     async def app(scope, receive, send):
-        response = JSONResponse(None)
+        response = JSONResponse(status_code=200)
         await response(scope, receive, send)
 
     client = test_client_factory(app)
     response = client.get("/")
-    assert response.json() is None
+    assert response.json() == {}
```
lovelydinosaur:
```suggestion
    assert response.content == b""
```

tests/test_responses.py
```diff
@@ -38,12 +38,12 @@ async def app(scope, receive, send):
 
 def test_json_none_response(test_client_factory):
     async def app(scope, receive, send):
-        response = JSONResponse(None)
+        response = JSONResponse(status_code=200)
```
lovelydinosaur:
```suggestion
        response = JSONResponse()
```

lovelydinosaur (commented):
Okay, so *my* expectations around the changes we'd want to see are:

* The status code becomes "defaults to either 200 or 204" rather than "defaults to 200".
* Omitting the `content` or passing `content=None` defaults to "don't render anything, body is `b''`" rather than "render `None` and return that as the body".

We've probably also got a follow up tweak that we'd want to *not* set any content-type in that case, but let's take this step by step.

lovelydinosaur:
Okay, see whatcya think about that.

Kludex:
> The status code becomes "defaults to either 200 or 204" rather than "defaults to 200".

What about 307?

lovelydinosaur:
> What about 307?

`RedirectResponse` continues to use 307 as the default in it's `__init__` signature, and life is happy and good.

aminalaee:
@Kludex I think the `RedirectResponse` passes the `status_code=307` to `Response` so it's overriden.

aminalaee:
This makes sense now. Thank you for the feedback. I'll take a look at the tests later today.

tests/test_responses.py
```diff
@@ -330,7 +330,8 @@ def test_empty_response(test_client_factory):
     app = Response()
     client: TestClient = test_client_factory(app)
     response = client.get("/")
-    assert response.headers["content-length"] == "0"
+    assert response.status_code == 204
+    assert "content-length" not in response.headers
```
aminalaee: @Kludex Does that `content-length` make sense? Or should we include `content-length` in empty response with 204 status code?

t1waz:
hi guys, maybe You should consider my point of view:
I think 200 status code is better as default than 204. 204 is something more just than No Content:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

from my expierence 204 should also be used in cases when request was process successfully on some entity. It's inform that it was process and do not need to reload or something (look link above)

200 is more general, more intuitive, for empty responses it's like standard. It is also good when default status code will remain same, not to create confusion for someone new :).

I think it's good idea to stay with is there is not content body should be b''.

Let me now, what You guys think

tests/test_responses.py
Kludex: Yes. https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2

>  A server MUST NOT send a Content-Length header field in any response
   with a status code of 1xx (Informational) or 204 (No Content).  A
   server MUST NOT send a Content-Length header field in any 2xx
   (Successful) response to a CONNECT request (Section 4.3.6 of
   [RFC7231]).

aminalaee: Great thanks, just wanted to double check since you were recently in this area.

Kludex: It's good to have a reference everywhere, so we can come back in the future and see the reason. Makes things faster anyway :P

Kludex:
@aminalaee a humbly remind to add the motivation of this PR on the description :pray:

Kludex:
Flask's `Response` class defaults to 200, jfyk:

ref.: https://github.com/pallets/werkzeug/blob/b611af2f5bf5f447d1d535b45f7c19309b0dc48d/src/werkzeug/sansio/response.py#L92

lovelydinosaur:
Hrm. I'm just not sure here. We're actually bundling up two different but related concerns...

1. If `content=None` do we still want to call `render()` or do we just `b""` it?
2. For empty content do we want to default to 204.

It's possible we should consider those separately.

I'm maybe not *that* fussed about (2) - it probably ends up being slightly unexpected for some users.

However (1) *probably?* sounds like a good idea. Practically the behavioural change it makes *within Starlette itself* is only for the `JSONResponse()` and `JSONResponse(None)` cases, where we'll render `b""` instead of `null`. Which... I think is probably better. Returning `null` is just. Well.. let's not help folks to do that.

Having said that, we'd *also* want to adapt `init_headers` to *not* include a content type header on those cases. Because we're not including any content. So "JSONResponse()" is a bit of a big ol' fib, if it doesn't *actually* have any content.

Forgetting about (2) for now, I think options are...

* Leave things as they are. `JSONResponse()` can render `null`. S'all good. Move along folks.
* Don't render anything when `content=None`. `return JSONResponse()` is actually the same as `return Response()`.
* Make the content argument *mandatory* on `JSONResponse(...)`. (Optional: disallow None)

After reading that through, I think the last one is a pretty neat plan.

Eg...

* `Response()` - An empty response - no content type.
* `JSONResponse()` - Raises an error - missing mandatory argument.
* `JSONResponse(...)` - Exactly what you'd expect.
* `JSONResponse(None)` - Little bit of a silly thing to do, but renders `null`.

I quite like that approach overall. Doesn't allow me to do things that's ambiguous. Has the behaviour I'd expect in all other cases.

It's a breaking change, but it's only a breaking change for users doing something a bit wrong and marginally broken.

We'd describe it to users in this way...

> "If you're currently returning `JSONResponse()` you now need to either return `Response()` (which is an empty response), or `JSONResponse(None)` (which is a JSON response, rendering `null` as the content)"

lovelydinosaur:
This seems better to me, right?

Also - I don't think we really need to deal with auto-defaulting 204 for `Response()` - let's just leave it up to the user. It might very often not be what they want.

This change on it's own is actually a neat enough bit of guard-railing.

lovelydinosaur (approved):
Makes a lot of sense to me, yup.

PR #1432, opened 2022-01-25
Add test case for `HTTPException` when status code is 204

Kludex:


starlette/exceptions.py
```diff
@@ -99,5 +99,5 @@ async def sender(message: Message) -> None:
 
     def http_exception(self, request: Request, exc: HTTPException) -> Response:
         if exc.status_code in {204, 304}:
-            return Response(b"", status_code=exc.status_code)
+            return Response(status_code=exc.status_code)
```
Kludex, line 102: This is internally converted to `b""` anyway.

adriangb (approved):

PR #1434, opened 2022-01-26
Add "Allow" headers to 405 Method Not Allowed responses

Kludex:
This PR was inspired based on a [FastAPI issue](https://github.com/tiangolo/fastapi/issues/4478) reported by @hellocoldworld.

As we can see on the [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5), responses with 405 status code **MUST** provide the `allow` headers. Starlette doesn't honor that.

What this PR introduces is a potential solution, which I'd like to discuss - to understand if it's the best solution. Changes here:
- `ExceptionMiddleware` adds `allow` headers when `HTTPException.status_code == 405`. (I'm not sure if this makes sense yet).
- Add `allow` headers when `PlainResponse.status_code == 405`.

There's still a case that is not being handled on this PR:
- When we have AnyResponse(status_code=405), it doesn't add the `allow` headers. It should, right? 

Important resources:
- [Click here to see](https://github.com/django/django/blob/a93a1ba347be9744e8492dd8288f18b3831caa02/django/http/response.py#L560-L573) Django way of handling 405 allow headers.
- [Click here to see](https://github.com/pallets/werkzeug/blob/b611af2f5bf5f447d1d535b45f7c19309b0dc48d/src/werkzeug/exceptions.py#L351-L384) Flask way of handling 405 allow headers.

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -43,9 +43,17 @@ class HTTPEndpoint:
         # If we're running inside a starlette application then raise an
         # exception, so that the configurable exception handler can deal with
         # returning the response. For plain ASGI apps, just return the response.
+        allowed_methods = filter(
+            lambda method: getattr(self, method.lower(), None) is not None,
+            ["GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
+        )
         if "app" in self.scope:
             raise HTTPException(status_code=405)
-        return PlainTextResponse("Method Not Allowed", status_code=405)
+        return PlainTextResponse(
+            "Method Not Allowed",
+            status_code=405,
+            headers={"allow": ", ".join(allowed_methods)},
+        )
 
 
 class WebSocketEndpoint:
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -98,6 +98,16 @@ class ExceptionMiddleware:
             await response(scope, receive, sender)
 
     def http_exception(self, request: Request, exc: HTTPException) -> Response:
+        headers = {}
+        if exc.status_code == 405:
+            router = request.scope["router"]
+            for route in router.routes:
+                match, _ = route.matches(request.scope)
+                methods = getattr(route, "methods", [])
+                if methods and match != 0:  # not Match.NONE
+                    headers = {"allow": ", ".join(methods)}
         if exc.status_code in {204, 304}:
             return Response(b"", status_code=exc.status_code)
-        return PlainTextResponse(exc.detail, status_code=exc.status_code)
+        return PlainTextResponse(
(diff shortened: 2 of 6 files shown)


adriangb:
Could we instead add a `headers` parameter to HTTPException and thus handle this in Route? I don't particularly like iterating through routes from within the middleware.

For context, FastAPI already has a subclass that does this, so that PR would also upstream their customization: https://github.com/tiangolo/fastapi/blob/291180bf2d8c39e84860c2426b1d58b6c80f6fef/fastapi/exceptions.py#L13

Kludex:
The code will look faaaaar more clear with that. I've created #1435, as it's another issue.

But maybe none of those two PRs are needed, as we need to solve the question I asked above:

> When we have AnyResponse(status_code=405), it doesn't add the allow headers. It should, right?

Should we add a `NotAllowedResponse` class that accepts `allowed_methods`? :thinking:

lovelydinosaur:
> Should we add a `xyz` class

Default answer to this is nope, thanks. 😅

Let's not extend the API of Starlette *at all* if we can help it.

starlette/endpoints.py
```diff
@@ -43,9 +43,17 @@ async def method_not_allowed(self, request: Request) -> Response:
...
+            lambda method: getattr(self, method.lower(), None) is not None,
+            ["GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
+        )
         if "app" in self.scope:
             raise HTTPException(status_code=405)
-        return PlainTextResponse("Method Not Allowed", status_code=405)
+        return PlainTextResponse(
+            "Method Not Allowed",
+            status_code=405,
+            headers={"allow": ", ".join(allowed_methods)},
```
lovelydinosaur, line 55: I guess we'd may as well use standard HTTP casing here - `Allow`.

(Eg. I checked https://github.com/encode/starlette/blob/master/starlette/middleware/gzip.py to see what we use internally elsewhere)

adriangb:
> But maybe none of those two PRs are needed, as we need to solve the question I asked above:
> 
> > When we have AnyResponse(status_code=405), it doesn't add the allow headers. It should, right?
> 
> Should we add a `NotAllowedResponse` class that accepts `allowed_methods`? 🤔

My thought was that once #1435 gets merged we'd just use that feature in [Route.handle](https://github.com/encode/starlette/blob/81bd557de69c1a780d81e94c1e6ee08009377052/starlette/routing.py#L251-L259) and be done with this.

I don't think we need to automagically enforce this for any response with `status_code=405`. I think that if Starlette itself does this when it returns a 405, that's good enough.

Kludex:
Replaced by #1436

PR #1435, opened 2022-01-26
Add `headers` parameter to `HTTPException`

Kludex:
Idea from @adriangb : https://github.com/encode/starlette/pull/1434#issuecomment-1021789683

The PR https://github.com/encode/starlette/pull/1434 created a use-case for `HTTPException.headers`. FastAPI already makes use of this, as mentioned by Adrian on https://github.com/tiangolo/fastapi/blob/291180bf2d8c39e84860c2426b1d58b6c80f6fef/fastapi/exceptions.py#L13 .

PR open for discussion. :)

### DIFF
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -9,11 +9,14 @@ from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 
 class HTTPException(Exception):
-    def __init__(self, status_code: int, detail: str = None) -> None:
+    def __init__(
+        self, status_code: int, detail: str = None, headers: dict = None
+    ) -> None:
         if detail is None:
             detail = http.HTTPStatus(status_code).phrase
         self.status_code = status_code
         self.detail = detail
+        self.headers = headers
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
@@ -99,5 +102,7 @@ class ExceptionMiddleware:
 
     def http_exception(self, request: Request, exc: HTTPException) -> Response:
         if exc.status_code in {204, 304}:
-            return Response(status_code=exc.status_code)
-        return PlainTextResponse(exc.detail, status_code=exc.status_code)
+            return Response(status_code=exc.status_code, headers=exc.headers)
+        return PlainTextResponse(
+            exc.detail, status_code=exc.status_code, headers=exc.headers
+        )
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -4,6 +4,8 @@ how you return responses when errors or handled exceptions occur.
 
 ```python
 from starlette.applications import Starlette
+from starlette.exceptions import HTTPException
+from starlette.requests import Request
 from starlette.responses import HTMLResponse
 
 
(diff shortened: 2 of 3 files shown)


lovelydinosaur:
Seems pretty reasonable to me yeah.

lovelydinosaur:
Needs a lil' docs update on https://www.starlette.io/exceptions/ - then I think it's ready to go.

starlette/exceptions.py
```diff
@@ -9,11 +9,14 @@
 
 
 class HTTPException(Exception):
-    def __init__(self, status_code: int, detail: str = None) -> None:
+    def __init__(
+        self, status_code: int, detail: str = None, headers: dict = None
```
adriangb, line 13:
```suggestion
        self, status_code: int, detail: str = None, headers: typing.Optional[typing.Mapping[str, str]] = None
```

`dict` == `Dict[Any, Any]`, so let's start with the most restrictive version here and we can expand it later. Plus `Response` accepts typing.Mapping[str, str]` ([here](https://github.com/encode/starlette/blob/81bd557de69c1a780d81e94c1e6ee08009377052/starlette/responses.py#L59)) so it makes sense to match that.

adriangb (approved):
This looks great! We should just test it against FastAPI first to at least know if it will be breaking for them. I don't think it will be.

starlette/exceptions.py
aminalaee: I don't think the `typing.Optional` is needed in the current situation.

Kludex: I've implemented with `typing.Mapping` initially, but when the type is more restrictive, there's a mypy issue, and `Response` accepts `headers: dict = None`, so I cannot add the right type here until `Response` is fixed.

adriangb: Ah okay, we should fix Response.
Optional isn't needed, but it's the right thing to do. Otherwise, MyPy strict will complain (see https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling)

aminalaee: I get that. I just wanted to say if we want to go for mypy strict, there are a lot of other changes needed.

For example in the same line `detail: str = None` needs to be `detail: typing.Optional[str] = None`. I just wanted to point that out. I've seen both cases in the code anyway.

adriangb: Oh yeah, a lot of other changes would be needed. But since using Optional doesn't hurt anything I just go with that as a default choice.

Kludex: Conceptually, it does hurt. :thinking: 

Using `Optional[<type>]`, you're saying that the parameter accepts `None`, with `<type>`, you're saying that it doesn't accept `None`... even if the default value is `None`.

adriangb:
> This looks great! We should just test it against FastAPI first to at least know if it will be breaking for them. I don't think it will be.

Well I tried, a bunch of other things are broken. But just eyeballing the change, I don't see how it would break FastAPI. So I think we're good to go on this!

starlette/exceptions.py
adriangb: Agreed, but I also think that having a default value that's not accepted isn't very intuitive, I'd rather use `_NOT_SET = object()` or whatever. Anyhow, this is getting off topic, I think my request was answered in https://github.com/encode/starlette/pull/1435#discussion_r792699586 and it's not worth holding up this PR for typing related discussion

PR #1436, opened 2022-01-26
add Allow header to 405 responses as required by RFC 7231

adriangb:
To be applied on top of #1435, replaces #1434

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -17,10 +17,11 @@ class HTTPEndpoint:
         self.scope = scope
         self.receive = receive
         self.send = send
-        self._allowed_methods = filter(
-            lambda method: getattr(self, method.lower(), None) is not None,
-            ["GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
-        )
+        self._allowed_methods = [
+            method
+            for method in ("GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS")
+            if getattr(self, method.lower(), None) is not None
+        ]
 
     def __await__(self) -> typing.Generator:
         return self.dispatch().__await__()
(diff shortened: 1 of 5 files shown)


starlette/endpoints.py
```diff
@@ -17,6 +17,10 @@ def __init__(self, scope: Scope, receive: Receive, send: Send) -> None:
         self.scope = scope
         self.receive = receive
         self.send = send
+        self._allowed_methods = filter(
+            lambda method: getattr(self, method.lower(), None) is not None,
+            ["GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
+        )
```
lovelydinosaur: I *guess* I probably find a list comprehension style easier to understand...

```python
self._allowed_methods = [
    method for method in HTTP_METHODS
    if getattr(self, method.lower(), None) is not None
]
```

Butttttt it's just a matter of preference isn't it. Possibly useful feedback(?), but not a blocker to me.

lovelydinosaur (approved):
Fab, yup. One bit of feedback, but up to you.

starlette/endpoints.py
```diff
@@ -17,10 +17,11 @@ def __init__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
         self.send = send
-        self._allowed_methods = filter(
-            lambda method: getattr(self, method.lower(), None) is not None,
-            ["GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS"],
-        )
+        self._allowed_methods = [
+            method
+            for method in ("GET", "HEAD", "POST", "PUT", "PATCH", "DELETE", "OPTIONS")
+            if getattr(self, method.lower(), None) is not None
+        ]
```
lovelydinosaur, line 24: Super. 🌟

starlette/endpoints.py
adriangb: yep makes sense, thanks for the suggestion!

PR #1437, opened 2022-01-27
Add custom URL convertor register

### LINKED ISSUE #231: Add user-defined converters to routing

Kludex:
## Custom URL Convertor

Currently, we don't have a way to create URL convertors, meaning that our users can only match the default types. Those are:
https://github.com/encode/starlette/blob/2b54f427613ebd1a6ccb2031aa97364e2d5070a5/starlette/convertors.py#L75-L81

To be fair, there's a way: adding a `key, value` pair on the `CONVERTOR_TYPES` dictionary. That's not available on Starlette's public API, or documented.

The idea of this PR is to document and introduce a public API that can be used to create custom URL convertors.

- Closes https://github.com/encode/starlette/issues/231
- Reattempts https://github.com/encode/starlette/pull/458

## How does Django and Flask handle this?

Both of them have a way to register custom URL convertors. You can find the way Django makes it possible [here](https://docs.djangoproject.com/en/4.0/topics/http/urls/#registering-custom-path-converters), and for Flask [here](https://uniwebsidad.com/libros/explore-flask/chapter-6/url-converters).

## Other

Recently, there was a PR to introduce time related convertors on https://github.com/encode/starlette/pull/1398, and I think we should follow the path this PR proposes here, and not creating more convertors on Starlette itself.

## Missing

- [x] docs (Do we want a single page with this, or it should be a section on the Routing page?)

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -55,6 +55,31 @@ Route('/floating-point/{number:float}', floating_point)
 Route('/uploaded/{rest_of_path:path}', uploaded)
 ```
 
+If you need a different converter that is not defined, you can create your own.
+See below an example on how to create a `datetime` convertor, and how to register it:
+
+```python
+from starlette.convertors import Convertor, register_url_convertor
+
+
+class DateTimeConvertor(Convertor):
+    regex = "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(.[0-9]+)?"
+
+    def convert(self, value: str) -> dt.datetime:
+        return dt.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S")
+
+    def to_string(self, value: dt.datetime) -> str:
+        return value.strftime("%Y-%m-%dT%H:%M:%S")
+
+register_url_convertor("datetime", DateTimeConvertor())
+```
+
+After registering it, you'll be able to use it as:
+
+```python
+Route('/history/{date:datetime}', history)
+```
+
 Path parameters are made available in the request, as the `request.path_params`
 dictionary.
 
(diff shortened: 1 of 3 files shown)


tests/test_convertors.py
```diff
@@ -0,0 +1,99 @@
+import datetime as dt
```
Kludex: I've created multiple convertors here with the purpose of reproducing https://github.com/encode/starlette/pull/1398. I can also simplify the test suite if wanted. Just let me know.

aminalaee: I agree with simplifying the tests, as this can be extended for other things like `UUID` and there's not really much difference. This is really good.

lovelydinosaur:
Well this is nice and simple isn't it? Seems worthwhile.
Would need [some brief docs](https://www.starlette.io/routing/#path-parameters).

aminalaee:
@Kludex I guess this PR can close this #1058 with a reference/example, what do you think?

Kludex:
> @Kludex I guess this PR can close this #1058 with a reference/example, what do you think?

Yeah, this PR closes that issue as well. I've used another convertor as an example on the docs, but we can answer that issue with "how to register/change the `path` convertor now".

docs/routing.md
```diff
@@ -55,6 +55,31 @@ Route('/floating-point/{number:float}', floating_point)
...
+
+```python
+from starlette.convertors import Convertor, register_url_convertor
+
+
+class DateTimeConvertor(Convertor):
+    regex = "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}(.[0-9]+)?"
+
+    def convert(self, value: str) -> dt.datetime:
+        return dt.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S")
```
aminalaee: Do we want to fix `dt` import? I mean it's probably clear but will lead to another `Fix Typo ..` PR 😄

Kludex: I've simplified it, and added the import statement. Good catch! :)

aminalaee (approved):

PR #1263, opened 2021-08-11
Add `WebSocketException` and support for WS handlers

Kludex:
- Replaces #527 

## Missing

- [x] Address what @tomchristie said on https://github.com/encode/starlette/pull/527#discussion_r343613065:
image

- [x] Check how this PR works interacting with the server.

PR #1440, opened 2022-01-27
Update exception handler signature to reflect HTTPConnection

aminalaee:
Right now the docs and examples show the following signature for handlers:

```python
async def not_found(request: Request, exc):
    return Response(content=HTML_404_PAGE, status_code=exc.status_code)
```

Even though this should be possible too:

```python
async def handler(conn: HTTPConnection, exc):
    return Response(content=HTML_404_PAGE, status_code=exc.status_code)
```

Still in progress to find all occurrences that should be update.


aminalaee:
I was expecting that the `ExceptionMiddleware` should reflect this too, but I think because of this line the WebSocket exceptions are not processed?

https://github.com/encode/starlette/blob/2b54f427613ebd1a6ccb2031aa97364e2d5070a5/starlette/exceptions.py#L64-L68

adriangb:
What is the point of accepting an `HTTPConnection`? Error handlers don't work with WebSockets (unless #527 or similar gets merged), so in practice it's always going to be a `Request`, which is an `HTTPConnection` with additional info, right?

aminalaee:
> Error handlers don't work with WebSockets

Well I know they don't work. That's exactly what my last comment is pointing to. I didn't see @Kludex had taken the lead on that in #1263

PR #1443, opened 2022-01-29
Deprecate `concurrency.run_until_first_complete`

Kludex:
Discussion about this can be found: https://github.com/encode/starlette/pull/1157#discussion_r652914582


adriangb (approved):

lovelydinosaur:
Did a quick grep to make sure we're not using this anymore.
Good stuff, yup.

Issue #734, opened 2019-11-25
Endpoint class decorators


azmeuk:
I need to implement a decorator for an endpoint, and for some technical reasons, I would like to implement it with a class. I also would like my decorator to not need parenthesis when called. (i.e. prefer `@my_decorator` over `@my_decorator()`).

Here is a simple implementation:
```python
from starlette.responses import JSONResponse
from starlette.applications import Starlette
from starlette.routing import Route


class foobar:
    def __init__(self, func):
        self.func = func

    def __call__(self, *args, **kwargs):
        return self.func(*args, **kwargs)


@foobar
def my_endpoint(request):
    return JSONResponse({"response": "my_endpoint"})


app = Starlette(
    debug=True,
    routes=[
        Route('/my_endpoint', my_endpoint),
    ],
)
```

But accessing to `/my_endpoint` produces this traceback:

```python
Traceback (most recent call last):
  File "~/myproject/env/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 385, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "~/myproject/env/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__
    return await self.app(scope, receive, send)
  File "~/myproject/env/lib/python3.8/site-packages/starlette/applications.py", line 102, in __call__
    await self.middleware_stack(scope, receive, send)
  File "~/myproject/env/lib/python3.8/site-packages/starlette/middleware/errors.py", line 178, in __call__
    raise exc from None
  File "~/myproject/env/lib/python3.8/site-packages/starlette/middleware/errors.py", line 156, in __call__
    await self.app(scope, receive, _send)
  File "~/myproject/env/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "~/myproject/env/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "~/myproject/env/lib/python3.8/site-packages/starlette/routing.py", line 551, in __call__
    await route.handle(scope, receive, send)
  File "~/myproject/env/lib/python3.8/site-packages/starlette/routing.py", line 228, in handle
    await self.app(scope, receive, send)
  File "./app.py", line 11, in __call__
    return self.func(*args, **kwargs)
TypeError: my_endpoint() takes 1 positional argument but 3 were given
```

[...]

PR #1444, opened 2022-01-29
Improve detection of async callables

### LINKED ISSUE #886: objects defining async __call__ not run/awaited when used in on_startup

Kludex:
- Closes #886
- Related to #1256

References:
- https://github.com/django/asgiref/issues/251
- https://github.com/django/asgiref/blob/3554e05def93e086690d19e9f38e593f1b5b7b61/asgiref/sync.py#L28-L40
- https://gitlab.com/pgjones/quart/-/blob/main/src/quart/utils.py#L102-128


starlette/_utils.py
```diff
@@ -0,0 +1,12 @@
+import asyncio
+import functools
+import typing
+
+
+def iscoroutinefunction(obj: typing.Any) -> bool:
+    while isinstance(obj, functools.partial):
+        obj = obj.func
+
+    return asyncio.iscoroutinefunction(obj) or (
```
bibajz, line 10: Hi @Kludex, would it make sense to use [`inspect.iscoroutinefunction`](https://docs.python.org/3/library/inspect.html?highlight=inspect#inspect.iscoroutinefunction) here?

I understand why we cannot just use the said function alone, since `async def __call__(...)` is not captured by it:

```
>>> import inspect
>>> class A:
...     async def __call__(self):
...             ...
... 
>>> inspect.iscoroutinefunction(A())
False
```

I have looked at the code  of [`asyncio.iscoroutinefunction`](https://github.com/python/cpython/blob/3.10/Lib/asyncio/coroutines.py#L164) and it only additionally check for the deprecated `@coroutine` decorator.

Cheers,
Libor

tests/test__utils.py
```diff
@@ -0,0 +1,70 @@
...
+        ...  # pragma: no cover
+
+    def func():
+        ...  # pragma: no cover
+
+    assert iscoroutinefunction(async_func)
+    assert not iscoroutinefunction(func)
+
+
+def test_async_partial():
```
bibajz, line 17: Would it also make sense to test a nested partial to check the `while` loop?

```python
def test_async_nested_partial():
    async def async_func(a, b):
        ...

    def func(a, b):
        return partial(async_func, a=a)

    nested_partial = functools.partial(func, 1)
    assert iscoroutinefunction(nested_partial)
```

bibajz (commented):
Nice!

Quick check with grep reveals that `asyncio` is now mentioned only in `starlette/_utils` and tests ;)

tests/test__utils.py
Kludex: Good idea! Thanks! :)

Kludex: The proposed snippet is not supported, because `func` is sync.

I'll add a test for nested one, tho.

bibajz: Ah, got it, my bad with the snippet, but you got the gist of it and corrected ;)

bibajz:
By the way, the unwrapping pattern of `while isinstance(obj, functools.partial)`, I have run a grep on the main branch and

```
$ rg 'isinstance\(\w+, (partial|functools.partia)l\)'
starlette/routing.py
46:    while isinstance(obj, functools.partial):
202:        while isinstance(endpoint_handler, functools.partial):
283:        while isinstance(endpoint_handler, functools.partial):
```

Of the cases found, the one relevant here is line 46 in ['iscoroutinefunction_or_partial'](https://github.com/encode/starlette/blob/master/starlette/routing.py#L41) function.

Does it make sense to use your newly defined function instead of it?

Edit: Looking at it further, wouldn't the replacement of `iscoroutinefunction_or_partial` by your checker actually solve this issue #734 ? Especially this one: https://github.com/encode/starlette/issues/734#issuecomment-612033518

^ @Kludex

Kludex:
> Edit: Looking at it further, wouldn't the replacement of iscoroutinefunction_or_partial by your checker actually solve this issue #734 ? Especially this one: #734 (comment)

Doesn't look like :disappointed: 

> Of the cases found, the one relevant here is line 46 in 'iscoroutinefunction_or_partial' function.
> Does it make sense to use your newly defined function instead of it?

One of the goals for this PR was to remove/deprecate that, and I forgot! :laughing:
Thanks!

Kludex:
> > Edit: Looking at it further, wouldn't the replacement of iscoroutinefunction_or_partial by your checker actually solve this issue #734 ? Especially this one: #734 (comment)
> 
> Doesn't look like disappointed

Hmmm.. Now I understand why `ismethod` is [included on asgiref](https://github.com/django/asgiref/blob/3554e05def93e086690d19e9f38e593f1b5b7b61/asgiref/sync.py#L28-L40).

It looks like it solves that comment, yes. But I'll have to do something like:

```python
def find_function(obj: typing.Any) -> typing.Callable:
    while inspect.ismethod(obj):
        obj = obj.__func__

    while isinstance(obj, functools.partial):
        obj = obj.func

    return obj


def iscoroutinefunction(obj: typing.Any) -> bool:
    obj = find_function(obj)

    return asyncio.iscoroutinefunction(obj) or (
        callable(obj) and asyncio.iscoroutinefunction(obj.__call__)
    )
```

And then use the `find_function` on the `WebSocketRoute` and `Route`.

bibajz:
And not even like this? I have reproduced the behaviour from your branch and it works - 

as in, `/a` prints `A` and `/b`  prints `B` in browser

```
from starlette import applications, responses, routing                                                                                                                                        
    
    
class Ping:    
    def __init__(self, message):    
        self._message = message    
    
    async def __call__(self, scope, request, response):    
        plain_text_resp = responses.PlainTextResponse(self._message)    
        return await plain_text_resp(scope, request, response)    
      
      
app = applications.Starlette(      
    routes=[      
        routing.Route("/a", Ping("A"), methods=["GET"]),      
        routing.Route("/b", Ping("B"), methods=["GET"]),      
    ]      
) 
```

Or am I mixing different concepts here?

Kludex:
> And not even like this? I have reproduced the behaviour from your branch and it works -
> 
> as in, `/a` prints `A` and `/b` prints `B` in browser
> 
> ```
> from starlette import applications, responses, routing                                                                                                                                        
>     
>     
> class Ping:    
>     def __init__(self, message):    
>         self._message = message    
>     
>     async def __call__(self, scope, request, response):    
>         plain_text_resp = responses.PlainTextResponse(self._message)    
>         return await plain_text_resp(scope, request, response)    
>       
>       
> app = applications.Starlette(      
>     routes=[      
>         routing.Route("/a", Ping("A"), methods=["GET"]),      
>         routing.Route("/b", Ping("B"), methods=["GET"]),      
>     ]      
> ) 
> ```
> 
> Or am I mixing different concepts here?

Yeah, you're mixing things. That's not the same as the example provided on the comment you mentioned above. Your example (on the quoted of this message) should never be supported...

bibajz:
Ad https://github.com/encode/starlette/pull/1444#issuecomment-1024944358 : Yeah, I figured. xD

Anyway, as you pointed to asgiref - the correct way is then

1. `unmethod`
2. `unpartial`
3. `iscoroutinefunction` is that so?

Kludex:
Actually, I'm going to let the state of this PR as is.

My previous statement about the `ismethod` was wrong.

Kludex:
Thanks for the review @bibajz ! :)

bibajz:
Ad https://github.com/encode/starlette/pull/1444#issuecomment-1024949340

Yeah, it looks like distinguishing coroutine/normal functions is pretty tricky, so good idea to leave for another day.

Glad to help, I really dig Starlette and other stuff from `encode` ecosystem, so if you can point me to some issue that would be worth to look at here, let me know!

Cheers
Libor

Kludex:
We need to organize ourselves a little better. It's hard for me to tell you where the help is needed... You can check the issues and PRs and see if you can help on them. I'm available on our Gitter if you need something.

bibajz:
Thanks, I will look around and pick something. :)

starlette/_utils.py
adriangb: You might want to handle functools.wraps as well (you can check obj.__wrapped__)

starlette/_utils.py
```diff
@@ -0,0 +1,12 @@
+import asyncio
+import functools
+import typing
+
+
+def iscoroutinefunction(obj: typing.Any) -> bool:
+    while isinstance(obj, functools.partial):
```
adriangb, line 7: You might want to handle functools.wraps as well (you can check `obj.__wrapped__`)

adriangb (approved):

lovelydinosaur:
What's the simplest possible example of an issue that this resolves from a user perspective?

starlette/_utils.py
Kludex: What case does it cover?

adriangb: Something like:

```python
db = object()

def inject_db(func):
    def wrapper(req):
        return func(req, db)
    return wrapper

@inject_db
async def endpoint(request, db):
    ...
```

frerikandriessen:
@tomchristie I ran into this issue while working on a custom route_handler in fastapi. I'm guessing the exact specifics of what and why are not that interesting to you, but this fix would actually be used (at least by me). I could of course solve it by not using a class, or wrapping the class in a async function, but I think in the spirit of duck-typing, an async callable class instance should be handled the same way as the normal async functions.

starlette/_utils.py
```diff
@@ -0,0 +1,12 @@
+import asyncio
+import functools
+import typing
+
+
+def iscoroutinefunction(obj: typing.Any) -> bool:
```
florimondmanca: Coming from reviewing #1644 

This is a nit, but should we call this `is_async_callable`?

AFAICT, having "something we can call and `await`" is all we care about when checking for apps or endpoints.

"Coroutine function" is a well-defined word in the Python glossary: a "function which returns a coroutine object", defined with "`async def`".

https://docs.python.org/3/glossary.html#term-coroutine-function

Also, the current naming makes it look as a compatibility shim on `asyncio.iscoroutinefunction` which focuses on checking just the definition above^, whereas we do some more checks, such as looking for `__call__`.

Kludex: > This is a nit, but should we call this is_async_callable?

Sure. 👍 I'll adapt later today. Thanks for the review. 🙏

starlette/_utils.py
Kludex: I'll not be doing this on this PR, as there's quite a big discussion on #734, and it didn't attract much attention on those years... But doing that check does solve #734.

Kludex:
> What's the simplest possible example of an issue that this resolves from a user perspective?

As a user, I'll be able to create an endpoint with _async callable objects_, _async callable methods_, and `functool.partial`:
```python
class Foo:
    async def __call__(self, request):
        return Response(...)
        
class Foo:
    async def method(self, request):
        return Response(...)
```
This also includes creating startup/shutdown events, background tasks and exception handlers as such. See more on [the test](https://github.com/encode/starlette/pull/1444/files#diff-fb71c2cd5efe5b9f0631089b43c379aeed343cf5371504e2c473133de222949e).

As a user, I don't see any reason for the framework to restrict my choices here.

abersheeran (approved):
Good job!

starlette/_utils.py
adriangb: No problem, we can easily do it in a followup

Kludex:
Thanks everybody! 🎉

PR #1446, opened 2022-01-29
`TestClient` timeout simulates `http.disconnect`

### LINKED ISSUE #1247: Disconnect event not received when using TestClient
### Checklist

### Describe the bug

When using the TestClient provided by the starlette framework, `http.disconnect` are not received. This is a problem if trying to test streaming.

### To reproduce

Implement simple app:
```python
    await send({
        'type': 'http.response.start',
        'status': 200,
    })
    while True:
        message = await receive()
        print('message', message)
        if message['type'] == 'http.disconnect':
            break
```

Write test using provided TestClient, which closes the connection during "processing":
```python
from bug import app

client = TestClient(app)

def test_client():
    response = client.post('/', stream=True)
    print('response.status', response.status)
    response.close()
```

### Expected behavior

The server behaviour should be the same if using TestClient or Uvicorn. If I run the application provided above with Uvicorn:

```bash
uvicorn bug:app
```

And write a test using the normal `requests` as follows:
```python
import requests

def test_server():
    response = requests.post('http://localhost:8000/', stream=True)
    print('response.status', response.status)
    response.close()
```

The `http.disconnect` event is received. This should be the same for the `TestClient`

### Actual behavior

No `http.disconnect` event is received by the app when using `TestClient`.

### Environment

- OS: Linux
- Python version: 3.8.10
- Starlette version: 0.16.0

Kludex:
- Supersedes #751
- Closes #1109
- Closes #1247

There's an alternative `timeout` _meaning_ on #1109, but I think we should proceed with this idea, as I can already see three usages:
- Test `StreamingResponse` properly
- Test #1427
- Test #1441

Missing:
- [ ] docs
- [ ] Replace `requests` timeout exception by starlette native one

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -12,6 +12,7 @@ import typing
 from concurrent.futures import Future
 from urllib.parse import unquote, urljoin, urlsplit
 
+import anyio
 import anyio.abc
 import requests
 from anyio.streams.stapled import StapledObjectStream
@@ -190,9 +191,17 @@ class _ASGIAdapter(requests.adapters.HTTPAdapter):
         request_complete = False
         response_started = False
         response_complete: anyio.Event
+        timeout_called = False
         raw_kwargs: typing.Dict[str, typing.Any] = {"body": io.BytesIO()}
         template = None
         context = None
+
+        def do_timeout() -> None:
+            nonlocal timeout_called, response_complete, response_started
+            timeout_called = True
+            if request_complete:
+                response_started = True
+                response_complete.set()
 
         async def receive() -> Message:
             nonlocal request_complete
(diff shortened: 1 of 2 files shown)


lovelydinosaur:
My honest opinion on this is that we pretty much  shouldn't be changing TestClient *at all*, until #1376 is in.

(I'm also not convinced by supporting timeout on the TestClient. It's okay for these things to have some constraints. It's not something that eg. Django's TestClient has, it's not something that the `httpx.WSGITransport` or `httpx.ASGITransport` cases support, and I'd be absolutely okay with Starlette not supporting it directly either.)

Kludex:
Would you have a recommendation on how to test the tests mentioned on the description? i.e. when we have an `http.disconnect` on the application.

lovelydinosaur:
For now, I'd just suggest "The TestClient doesn't support timeouts. If you need to test this kind of behaviour you'll need to do so using ASGI messaging directly".

Kludex:
Thanks

PR #1448, opened 2022-01-29
use explicit Optional[...] for parameters with None as a default in datastructures.py

adriangb:
See https://www.python.org/dev/peps/pep-0484/#union-types

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -497,7 +497,7 @@ class Headers(typing.Mapping[str, str]):
         self,
         headers: typing.Optional[typing.Mapping[str, str]] = None,
         raw: typing.Optional[typing.List[typing.Tuple[bytes, bytes]]] = None,
-        scope: typing.Optional[typing.Mapping[str, str]] = None,
+        scope: typing.Optional[typing.Mapping[str, typing.Any]] = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


Kludex:
> See [https://www.python.org/dev/peps/pep-0484/\#id29](https://www.python.org/dev/peps/pep-0484/%5C#id29)

The link is "not found".

adriangb:
Weird, it got mangled, it should be https://www.python.org/dev/peps/pep-0484/#union-types

adriangb:
Specifically:

> A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:

```python
def handle_employee(e: Employee = None): ...
```

> This would have been treated as equivalent to:

```python
def handle_employee(e: Optional[Employee] = None) -> None: ...
```

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

lovelydinosaur (approved):
Sure. Don't know what the motivation is behind...

> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

But happy enough either ways. 😌

Kludex (approved):

adriangb:
There's probably discussions on the topic in BPO and such. Personally my motivation is just to make sure the project works as well as possible with type checkers.

PR #1449, opened 2022-01-29
add generics type parameters to ImmutableMultiDict

adriangb:
Slice of #1403

The goal here is to make these data structures more reusable and fix some currently under specified type annotations that make type checkers unhappy downstream


lovelydinosaur:
In the trade-off between readability vs. type-strictness I'm personally happy enough to stick with the under-defined "Any" cases, because I just find it so awkwardly complicated to read through the annotations. (And because `Any` in these cases isn't incorrect, it's just under-specifying.)

I don't have strong objections to it, but just gonna pitch that one out there.

adriangb:
Yeah, I see that side of this as well.

What for me makes this worth it is that:

1. These data structures have a lot of downstream consumers (both inside of Starlette and in anything that depends on Starlette
2. We're unlikely to change these classes much, they're pretty stable.
3. We also gain some readability, for example looking at `class FormData(ImmutableMultiDict[str, Union[str, UploadFile])` you'll be able to tell that it's a mapping of string field names to strings or UploadFiles without looking at the docs (and IDEs will also give you this info)

So we do give up some readability in Starlette's codebase, but we will be _improving_ legibility for downstream classes and consumers of the codebase.

But that's just my thought process / personal bias, I'll understand if others are not of the same opinion.

adriangb:
As a motivating example of what this can do for users of Starlette and downstream packages:

```python

async def func(req: Request) -> str:
    form = await req.form()
    file_or_string = form["field"]
    if isinstance(file_or_string, str):
        return file_or_string
    return (await file_or_string.read()).decode()

```

Running mypy on this:

```    
test.py:9: error: Statement is unreachable  [unreachable]
```

Because mypy thinks that `file_or_string` is always a string.

And from an IDE's perspective:

image

I get no auto-complete for `.read()`, thus no auto-complete for `.decode()` as well as several false positives like a redundant `isinstance` check and returning an unknown type.

On the other hand, this is what it looks like on #1403 (which this current PR is working towards):

image

IDE examples are on VSCode + Pylance (which is what [a large percentage of devs use nowadays](https://insights.stackoverflow.com/survey/2021#section-most-popular-technologies-integrated-development-environment))

lovelydinosaur:
Always appreciate seeing a good user-centred "this is why we might choose to make this change". 😎

starlette/datastructures.py
```diff
@@ -238,13 +243,15 @@ def __str__(self) -> str:
...
+
     def __init__(
         self,
         *args: typing.Union[
-            "ImmutableMultiDict",
-            typing.Mapping,
-            typing.List[typing.Tuple[typing.Any, typing.Any]],
+            "ImmutableMultiDict[_KT, _VT_co]",
+            typing.Mapping[_KT, _VT_co],
+            typing.Sequence[typing.Tuple[_KT, _VT_co]],
```
Kludex: Just for reference about the "why": https://mypy.readthedocs.io/en/stable/common_issues.html#invariance-vs-covariance

starlette/datastructures.py
```diff
@@ -254,7 +261,7 @@ def __init__(
         if kwargs:
             value = (
                 ImmutableMultiDict(value).multi_items()
-                + ImmutableMultiDict(kwargs).multi_items()
+                + ImmutableMultiDict(kwargs).multi_items()  # type: ignore[operator]
```
Kludex, line 266: We don't know the returned value of the operator `+` when there's a sum of two `ImmutableMultiDict`. That's why we ignore this error here.

Kludex (approved):
This PR is clear to me, and after I've start using a more strict type checking environment, I do feel @adriangb 's pain.

I'm not sure if `_T`, `_KT` and `_VT_co` are the best names for those, even if this convention is spreadly used... Maybe `DefaultType`, `KeyType` and `ValueType` can improve the readability, and @tomchristie will see it with better eyes (?).

adriangb:
> Maybe DefaultType, KeyType and ValueType can improve the readability

I'm 👍 on this, happy to make the change or accept suggestions from either of you.

adriangb:
Rename done

starlette/datastructures.py
```diff
@@ -238,32 +246,34 @@ def __str__(self) -> str:
...
+
     def __init__(
         self,
         *args: typing.Union[
-            "ImmutableMultiDict",
-            typing.Mapping,
-            typing.List[typing.Tuple[typing.Any, typing.Any]],
+            "ImmutableMultiDict[_KeyType, _ValueType]",
+            typing.Mapping[_KeyType, _ValueType],
+            typing.Sequence[typing.Tuple[_KeyType, _ValueType]],
```
abersheeran: image

Maybe we use `typing.Iterable` is better? Just like `dict`.

starlette/datastructures.py
```diff
@@ -274,33 +284,45 @@ def __init__(
...
+
+    @typing.overload
+    def get(
+        self,
+        key: _KeyType,
+        default: typing.Optional[typing.Union[_ValueType, _DefaultType]] = ...,
+    ) -> typing.Union[_ValueType, _DefaultType]:
+        ...  # pragma: no cover
+
+    def get(self, key: _KeyType, default: typing.Any = None) -> typing.Any:
```
abersheeran: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

I'm not quite sure if there are other benefits to implementing mixin methods like `.get`. `MutableMapping` already implements them. And they are already of the correct type.

image

starlette/datastructures.py
adriangb: yep

starlette/datastructures.py
adriangb: So we can just delete `get` here! Great point. This makes this a PR that (maybe) changes behavior instead of just typing, but I think that's worth it 😄

tests/test_datastructures.py
```diff
@@ -220,7 +220,9 @@ def test_url_blank_params():
     assert "abc" in q
     assert "def" in q
     assert "b" in q
-    assert len(q.get("abc")) == 0
+    val = q.get("abc")
+    assert val is not None
+    assert len(val) == 0
```
adriangb, line 225: This is just to make mypy happy

starlette/datastructures.py
```diff
@@ -382,7 +382,7 @@ def update(
         self._dict.update(value)
 
 
-class QueryParams(ImmutableMultiDict):
+class QueryParams(ImmutableMultiDict[str, str]):
```
adriangb, line 386: Just to make MyPy happy, previously it was picking up `str` from `ImmutableMultiDict` but now that `ImmutableMultiDict` is generic it was complaining about an unknown type.

starlette/datastructures.py
```diff
@@ -13,6 +13,14 @@ class Address(typing.NamedTuple):
     port: int
 
 
+_DefaultType = typing.TypeVar("_DefaultType")
+# Mapping values are covariant, keys are invariant
+_KeyType = typing.TypeVar("_KeyType")  # key type
+_ValueType = typing.TypeVar(
+    "_ValueType", covariant=True
+)  # value type for covariant containers
+
```
florimondmanca, line 21: (Nit) These comments seem a tad redundant with the code and the comment above?

```suggestion
_KeyType = typing.TypeVar("_KeyType")
_ValueType = typing.TypeVar("_ValueType", covariant=True)
```

Plus, any strong opinion against a more concise `_KT, _VT` naming?

I also note that the [`typing` docs](https://docs.python.org/3/library/typing.html) tend to hint covariance in the typevar name everywhere. So, we could use `_KT, _VT_co` to provide an even more consitent UX? _And_ link to the typing docs as a motivation for what we're doing here.

```suggestion
# Follow generics conventions listed here:
# https://docs.python.org/3/library/typing.html#typing.Mapping
# But underscore-prefixed so they don't get exposed as public API.
_KT = typing.TypeVar("_KT")
_VT_co = typing.TypeVar("_VT_co", covariant=True)
```

starlette/datastructures.py
```diff
@@ -13,6 +13,14 @@ class Address(typing.NamedTuple):
     port: int
 
 
+_DefaultType = typing.TypeVar("_DefaultType")
```
florimondmanca: Looks like this is unused?

```suggestion
```

florimondmanca (commented):
I do like this as well. I had already found myself in cases when the default typing on `Form` required me to do unecessary `assert` or `isinstance` checks to benefit from typing.

starlette/datastructures.py
florimondmanca: Also means we now get `QueryParams(...).get(...)` to be a `str`, rather than `Any`. Which is neat and correct, right?

starlette/datastructures.py
Kludex: This PR was using `_KT` and `_VT_co` before my comment here: https://github.com/encode/starlette/pull/1449#pullrequestreview-981163983 😅

I thought it will be more friendly to @tomchristie's eyes, but if more members prefer the convention used in the python docs, then I guess it's fine...

florimondmanca: Ha! I see. :) Didn't read up the full thread. Well, no strong opinions on my side, we can keep things like this.

florimondmanca: I might still suggest dropping some of the redundant comments:

```python
_KeyType = typing.TypeVar("_KeyType")
_ValueType = typing.TypeVar("_ValueType", covariant=True)
```

starlette/datastructures.py
adriangb: Yup usage just got removed 5 min ago 😅

starlette/datastructures.py
adriangb: I think the `_VT_co` comes from the fact that in those modules there's also a `_VT`: values in `Mapping` are covariant, values in `Dict` are invariant.

So in this case, _right now_ we could just use `_ValueType` as we are. But once we add types to `MultiDict` (the subclass) I think we'll need an invariant value type. So I'll rename this to `_CovariantValueType`.

And yes, the comments are now redundant.

adriangb: I landed on:

```python
_KeyType = typing.TypeVar("_KeyType")
# Mapping keys are invariant but their values are covariant since
# you can only read them
# that is, you can't do `Mapping[str, Animal]()["fido"] = Dog()`
_CovariantValueType = typing.TypeVar("_CovariantValueType", covariant=True)
```

Does the example helps illustrate why the values are covariant?

starlette/datastructures.py
adriangb: Yep! There's a lot of downstream work from this (like actually fixing `Form`), but I wanted to keep this PR as narrowly scoped as possible. Hopefully we can approve and merge those downstream PRs quicker once the basic structure for stronger typing is added via this PR.

florimondmanca (approved):

starlette/datastructures.py
adriangb: Thank again for this, great catch!

PR #1451, opened 2022-01-29
Make base convertors.Convertor generic, annotate its subclasses

bibajz:
Small follow-up of #1437 , making the Convertor generic in its `convert`/`to_string`, thus getting rid of `typing.Any` and also annotating the `regex` as a `ClassVar`.


Kludex (approved):
@bibajz Thanks for the PR! :)

PR #1452, opened 2022-01-30
Raise exceptions that happened on mounted apps using `BaseHTTPMiddleware`

### LINKED ISSUE #1433: Raising Exceptions in sub-applications routes
### Checklist

### Describe the bug

Let's start with this PR: #1262

It's about preventing raise `anyio.ExceptionGroup` in views under a `BaseHTTPMiddleware`. PR resolve that problem with nonlocal variable that stores our exception. But in the case of sub-applications, it does not work. 

As I can see (fyi I am not good at asyncio), in the case below, we reach and read a response before we raise an exception and store it to our nonlocal variable:

fragment of `BaseHTTPMiddleware.__call__`
```python

    async def coro() -> None:
        nonlocal app_exc

        async with send_stream:
            try:
                task = await self.app(scope, request.receive, send_stream.send)
            except Exception as exc:
                app_exc = exc

    task_group.start_soon(coro)

    try:
        message = await recv_stream.receive()
    except anyio.EndOfStream:
        if app_exc is not None:
            raise app_exc
    
    ...
    response = StreamingResponse(
        status_code=message["status"], content=body_stream()
    )
    return response
```

in this moment: `except anyio.EndOfStream:` exception still no raised.

```python
import httpx
import pytest
from fastapi import FastAPI, APIRouter


class SomeError(Exception):
    pass


class SomeMiddleware(BaseHTTPMiddleware):
    async def dispatch(
    ) -> Response:

# Drop (or use not BaseHTTPMiddleware based) middleware and test works fine
app = FastAPI(middleware=[Middleware(SomeMiddleware), ])


async def simple_route(request: Request):
    raise SomeError


another_router = APIRouter(
    routes=[Route('/simple-route/', simple_route, methods=['GET'])]
)
sub_app = FastAPI()
sub_app.include_router(another_router)
app.router.mount(f'/api', sub_app)


@pytest.mark.asyncio
async def test_simple_route():
        with pytest.raises(SomeError):
            await client.get("http://testserver/api/simple-route/")

```

### Expected behavior

An exception was raised and caught by pytest exception

### Actual behavior

An exception wasn't raised

### Debugging material

_No response_

### Environment

macOS Monterey 12.0.1, starlette 0.17.1, Python 3.9.9

### Additional context

_No response_

Kludex:
Closes #1433

The problem here is that the mounted app is able to send `http.response.start` (and the following `http.response.body`), but we check if an exception happened only if there was an `anyio.EndOfStream` i.e. only if messages were not sent.

The solution here just raises the exception at the end of the middleware.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -63,10 +63,9 @@ class BaseHTTPMiddleware:
             request = Request(scope, receive=receive)
             response = await self.dispatch_func(request, call_next)
             await response(scope, receive, send)
+            if app_exc is not None:
+                raise app_exc
             task_group.cancel_scope.cancel()
-
-        if app_exc is not None:
-            raise app_exc
 
     async def dispatch(
         self, request: Request, call_next: RequestResponseEndpoint
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -22,8 +22,9 @@ class BaseHTTPMiddleware:
             await self.app(scope, receive, send)
             return
 
+        app_exc: typing.Optional[Exception] = None
+
         async def call_next(request: Request) -> Response:
-            app_exc: typing.Optional[Exception] = None
             send_stream, recv_stream = anyio.create_memory_object_stream()
 
             async def coro() -> None:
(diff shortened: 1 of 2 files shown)


Kludex:
Replaced by #1459

PR #1453, opened 2022-01-30
minimize event loop round trips in routing

adriangb:
This change minimizes roundtrips to the event loop during routing and by modifying `BaseRoute` allows implementers of the interface more freedom in how the implement functions returning awaitables.

To work this would need https://github.com/agronholm/anyio/pull/416#pullrequestreview-867085900

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
     install_requires=[
-        "anyio @ git+https://github.com/adriangb/anyio.git@allow-awaitable-in-to-thread",
+        "anyio>=3.0.0,<4",
         "typing_extensions; python_version < '3.10'",
         "contextlib2 >= 21.6.0; python_version < '3.7'",
     ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -38,7 +38,7 @@ def get_long_description():
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
     install_requires=[
-        "anyio>=3.0.0,<4",
+        "anyio @ git+https://github.com/adriangb/anyio.git@allow-awaitable-in-to-thread",
```
adriangb: Just to demo that this works

adriangb:
I'm going to close this in favor of making several smaller PRs in increasing order of straightforwardness:

1. Modify TestClient to accept non-coroutine ASGI apps (#1458 )
2. Modify BaseRoute to use `def __call__(...) -> Awaitable[None]` so that it's subclasses can make use of this optimization.
3. Modify the routes (`Route`, `Mount`) to make use of this optimization
4. Modify Router to make use of this optimization

Issue #1101, opened 2020-11-24
Allow regex patterns for route path


ar45:
### Checklist

- [x] There are no similar issues or pull requests for this yet.
- [ ] I discussed this idea on the [community chat](https://gitter.im/encode/community) and feedback is positive.

### Is your feature related to a problem? Please describe.

I want to be able to define generic routes which match more specific criteria. Creating a `Converter` for each instance just to be able to use a different regex seems a bit too much bloat. 

One example, consider I would like my route to match optional suffix
 `/post/{id}/attachments/{attachment_id}.suffix?`
so `/post/1/attachments/22.mp4` and `/post/1/attachments/22` both match but `/post/1/attachments/bogus` should not

Another use case is to be able to control case sensitivity on some urls.
There is infinite power that comes with being able to provide regex patterns.

## Describe alternatives you considered

Alternatively, creating a custom converter for each pattern, but still does not provide control over regex flags. 
Creating Custom converters for each pattern 
```python
class AttachmentConverter(StringConverter):
    regex = '\d+(?:\.mp4|\.mp3)?'
```

## Describe the solution you would like.
What I ended up doing is patching `compile_path` to accept a `RegexPath`.

```python
@router.get(RegexPath(rf'/(?P<greet>hello|hi)/(?P<count>{IntegerConvertor.regex})/(?P<id_>{UUIDConvertor.regex})'))
async def greet(count: int, greet: str, limit: int, request: Request):
   ...

# limit is an integer but must be greater than 100.
@router.get(RegexPath(r'/(?P<greet>hello|hi)/(?P<count>\d+)/(?P<limit>1[0-9][0-9]+)'))
async def greet(count: int, greet: str, limit: int, request: Request):


# Optionally supply Converters otherwise it defaults to `StringConverter`
@router.get(RegexPath(r'/(?P<greet>hello|hi)/(?P<count>\d+)/(?P<limit>1[0-9][0-9]+)', path_converters={'limit': IntegerConverter()}))




# Passing Compiled regex with flags
regex = re.compile(r'/(?P<greet>hello|hi)/(?P<count>\d+)/(?P<limit>1[0-9][0-9]+)', re.IGNORECASE)
@router.get(RegexPath(regex, path_converters={'limit': IntegerConverter()}))
```

ar45:
Bellow is an implementation that `router.url_for_path(...)` kinda works, the problem is that `url_for_path` does not check the parameters match the regex. which effects the current parameter based paths as well.

```python
[...]

PR #1455, opened 2022-01-30
Add-support-for-converter-arguments

Mng-dev-ai:
Add support for converter arguments which will be used to allow regex patterns for route path as requested here #1101 
### How does Django and Flask handle this?
Flask do the same thing here as described here https://werkzeug.palletsprojects.com/en/2.0.x/routing/#rule-format 

### How adding support for converter arguments related to allow regex patterns for route path ?
By simply create a regex converter which will takes the first argument as our new regex as declared in test_converters and also here https://stackoverflow.com/a/5872904/10799351

### DIFF
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -8,7 +8,7 @@ T = typing.TypeVar("T")
 class Convertor(typing.Generic[T]):
     regex: typing.ClassVar[str] = ""
 
-    def __init__(self, *args) -> None:
+    def __init__(self, *args: typing.Any) -> None:
         pass
 
     def convert(self, value: str) -> T:
@@ -86,5 +86,5 @@ CONVERTOR_TYPES = {
 }
 
 
-def register_url_convertor(key: str, convertor: Convertor) -> None:
+def register_url_convertor(key: str, convertor: typing.Any) -> None:
     CONVERTOR_TYPES[key] = convertor
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -8,6 +8,9 @@ T = typing.TypeVar("T")
 class Convertor(typing.Generic[T]):
     regex: typing.ClassVar[str] = ""
 
+    def __init__(self, *args) -> typing.NoReturn:
+        pass
+
     def convert(self, value: str) -> T:
         raise NotImplementedError()  # pragma: no cover
 
@@ -75,11 +78,11 @@ class UUIDConvertor(Convertor):
 
 
 CONVERTOR_TYPES = {
-    "str": StringConvertor(),
-    "path": PathConvertor(),
-    "int": IntegerConvertor(),
-    "float": FloatConvertor(),
-    "uuid": UUIDConvertor(),
+    "str": StringConvertor,
+    "path": PathConvertor,
+    "int": IntegerConvertor,
+    "float": FloatConvertor,
+    "uuid": UUIDConvertor,
 }
 
 
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -8,7 +8,7 @@ T = typing.TypeVar("T")
 class Convertor(typing.Generic[T]):
     regex: typing.ClassVar[str] = ""
 
-    def __init__(self, *args) -> typing.NoReturn:
+    def __init__(self, *args) -> None:
         pass
 
     def convert(self, value: str) -> T:
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -6,9 +6,9 @@ T = typing.TypeVar("T")
 
 
 class Convertor(typing.Generic[T]):
-    regex: typing.ClassVar[str] = ""
+    regex: str = ""
 
-    def __init__(self, *args: typing.Any) -> None:
+    def __init__(self, *args: str) -> None:
         pass
 
     def convert(self, value: str) -> T:
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -9,7 +9,7 @@ class Convertor(typing.Generic[T]):
     regex: str = ""
 
     def __init__(self, *args: str) -> None:
-        pass
+        self.args = args
 
     def convert(self, value: str) -> T:
         raise NotImplementedError()  # pragma: no cover
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -102,7 +102,7 @@ def replace_params(
 # Match parameters in URL paths, eg. '{param}', and '{param:int}'
 PARAM_REGEX = re.compile(
     "{([a-zA-Z_][a-zA-Z0-9_]*)(:[a-zA-Z_][a-zA-Z0-9_]*)?"
-    "((?:(?:\\(.*\\))|(?:\\[.*\\]))*)}"
+    "(?:\\((.*?)\\))?}"  # Match any number of characters in parentheses
 )
 
 
@@ -130,7 +130,7 @@ def compile_path(
             convertor_type in CONVERTOR_TYPES
         ), f"Unknown path convertor '{convertor_type}'"
         if args:
-            normalized_args = args.strip("()").split(",")
+            normalized_args = args.split(",")
             convertor = CONVERTOR_TYPES[convertor_type](*normalized_args)
         else:
             convertor = CONVERTOR_TYPES[convertor_type]()
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -129,11 +129,11 @@ def compile_path(
         assert (
             convertor_type in CONVERTOR_TYPES
         ), f"Unknown path convertor '{convertor_type}'"
-        if args:
+        if args:  # pragma: no cover
             # split args by comma except in {}
             normalized_args = re.split(r"\,(?![^\{]*\})", args)
             convertor = CONVERTOR_TYPES[convertor_type](*normalized_args)
-        else:
+        else:  # pragma: no cover
             convertor = CONVERTOR_TYPES[convertor_type]()
         path_regex += re.escape(path[idx : match.start()])
         path_regex += f"(?P<{param_name}>{convertor.regex})"
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -101,7 +101,8 @@ def replace_params(
 
 # Match parameters in URL paths, eg. '{param}', and '{param:int}'
 PARAM_REGEX = re.compile(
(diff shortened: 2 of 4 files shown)


starlette/convertors.py
```diff
@@ -8,6 +8,9 @@
 class Convertor(typing.Generic[T]):
     regex: typing.ClassVar[str] = ""
 
+    def __init__(self, *args) -> typing.NoReturn:
```
aminalaee: I guess `typing.NoReturn` is not really accurate it returns `None` right?

```suggestion
    def __init__(self, *args) -> None:
```

Mng-dev-ai: Yes, You’re right I missed it

starlette/convertors.py
```diff
@@ -8,6 +8,9 @@
 class Convertor(typing.Generic[T]):
     regex: typing.ClassVar[str] = ""
```
bibajz: Hi! Since I authored https://github.com/encode/starlette/commit/84820a25f2237c7c8fe7911f4fea2394f02994e8 yesterday, there was no need to set `regex` during instantiation, so I set it as a `ClassVar`.

Now, it is no longer the case, so you need to put it back to `regex: str = ""`

starlette/convertors.py
bibajz: Also, all the arguments are strings, since they are defined in the route, so

```
def __init__(self, *args: str) -> None:
```

starlette/routing.py
```diff
@@ -121,12 +123,12 @@ def compile_path(
     idx = 0
     param_convertors = {}
     for match in PARAM_REGEX.finditer(path):
-        param_name, convertor_type = match.groups("str")
+        param_name, convertor_type, args = match.groups("str")
```
bibajz: If no parentheses are found, `args` is equal to empty string `""`.

If parentheses are found, `args` is equal to the whole parenthesized string (!).

To reproduce (it's just copy-pasted `compile_path` function with pprints inside)

```python
from pprint import pprint

PARAM_REGEX = re.compile(
    "{([a-zA-Z_][a-zA-Z0-9_]*)(:[a-zA-Z_][a-zA-Z0-9_]*)?((?:(?:\\(.*\\))|(?:\\[.*\\]))*)}"
)

def compile_path(path):
    path_regex = "^"
    path_format = ""
    duplicated_params = set()
    idx = 0
    param_convertors = {}
    for match in PARAM_REGEX.finditer(path):
        param_name, convertor_type, args = match.groups("str")
        convertor_type = convertor_type.lstrip(":")
        pprint([param_name, convertor_type, args])
        pprint([type(param_name), type(convertor_type), type(args)])
```

```
>>> compile_path("/{name:str}")
 ['name', 'str', '']
[<class 'str'>, <class 'str'>, <class 'str'>]

>>> compile_path("/{foo:bar([0-1]{2},[2-3]{3})}")
['foo', 'bar', '([0-1]{2},[2-3]{3})']
[<class 'str'>, <class 'str'>, <class 'str'>]
```

Therefore, you would need something like this:

```python
if args:
    # Strip the `()` and split by `,`
    normalized_args = args.strip("()").split(",")
    convertor = CONVERTOR_TYPES[convertor_type](*normalized_args)
else:
    convertor = CONVERTOR_TYPES[convertor_type]
```

---

This would also mean that `CONVERTOR_TYPES` is of type `dict[str, Union[Convertor, Type[Convertor]]]`, because I think the builtin convertors can already be instantiated singletons and also begs the question, whether should the newly added initializer of `Convertor` only accept one argument and that of type `list[str]`, but that is a minor detail compared to the above mentioned points.

cc @aminalaee

bibajz (changes requested):

tests/test_convertors.py
```diff
@@ -27,8 +27,8 @@ def to_string(self, value: datetime) -> str:
 
 
 class RegexConvertor(Convertor):
-    def __init__(self, *args):
-        self.regex = args[0]
+    def __init__(self, *args: typing.Any) -> None:
+        self.__class__.regex = args[0]
```
bibajz: The way to suppress mypy error is here: https://github.com/encode/starlette/pull/1455#discussion_r795218143

starlette/routing.py
bibajz: In other words, the functionality is now broken, but since testing was done on a `RegexConvertor` with one initializer argument, it could not be detected so -  `"/regex/{param:regex([0-9]{3}-[0-9]{3}-[0-9]{4})}"` and the initialization looked like:

```
inst = RegexConvertor("([0-9]{3}-[0-9]{3}-[0-9]{4})")
```

And therefore, `inst.regex = "([0-9]{3}-[0-9]{3}-[0-9]{4})"` - notice the parentheses, which however still matched the test case!

```
>>> re.match("abc", "abc")
<re.Match object; span=(0, 3), match='abc'>
>>> re.match("(abc)", "abc")
<re.Match object; span=(0, 3), match='abc'>
```

Mng-dev-ai: Thank you @bibajz I checked it and I confirmed the bug and fixed it.

starlette/routing.py
```diff
@@ -100,7 +100,10 @@ def replace_params(
 
 
 # Match parameters in URL paths, eg. '{param}', and '{param:int}'
-PARAM_REGEX = re.compile("{([a-zA-Z_][a-zA-Z0-9_]*)(:[a-zA-Z_][a-zA-Z0-9_]*)?}")
+PARAM_REGEX = re.compile(
+    "{([a-zA-Z_][a-zA-Z0-9_]*)(:[a-zA-Z_][a-zA-Z0-9_]*)?"
+    "((?:(?:\\(.*\\))|(?:\\[.*\\]))*)}"
```
lovelydinosaur: This regex is *wwwaaaayyyy* to complex for us to have in without comments explaining it.

Kludex:
I'd like to use named parameters instead of positional ones :thinking:

starlette/routing.py
Mng-dev-ai: I tried to make it more simple but it doesn’t work

Kludex: He meant to add a comment on the code `# This regex matches...`, I guess.

Mng-dev-ai: I understand maybe I over complicate the pattern I will check it again

Mng-dev-ai: I made it much simpler can you please check it again ?

starlette/routing.py
```diff
@@ -121,13 +124,17 @@ def compile_path(
...
     for match in PARAM_REGEX.finditer(path):
-        param_name, convertor_type = match.groups("str")
+        param_name, convertor_type, args = match.groups("str")
         convertor_type = convertor_type.lstrip(":")
         assert (
             convertor_type in CONVERTOR_TYPES
         ), f"Unknown path convertor '{convertor_type}'"
-        convertor = CONVERTOR_TYPES[convertor_type]
-
+        if args:  # pragma: no cover
```
bibajz: Hi @Mng-dev-ai,

instead of `pragma: no cover` the `if-else`, you could add a separate test for the `compile_path` function altogether, since it is actually a key compoment of routing.

starlette/convertors.py
```diff
@@ -75,13 +78,13 @@ def to_string(self, value: uuid.UUID) -> str:
 
 
 CONVERTOR_TYPES = {
```
bibajz, line 80: What about the `CONVERTOR_TYPES` a dictionary of already instantiated convertors - those, that do not need customization at runtime and those, that need, so:

```python
CONVERTOR_TYPES: typing.Dict[str, typing.Union[Convertor, Type[Convertor]]] = {
    "path": PathConvertor(),
    "int": IntegerConvertor(),
    "float": FloatConvertor(),
    "uuid": UUIDConvertor(),
}
```

and also:

```python

def register_url_convertor(key: str, convertor: typing.Union[Convertor, Type[Convertor]]) -> None:
    CONVERTOR_TYPES[key] = convertor
```

therefore, you would not need instantion in the `else` branch of `convert_path`, if no `args` are present

Kludex: I think only the `convertor: Type[Convertor]` on the `register_url_convertor` is needed here.

bibajz: If you were to keep the instantion of the converter in the `else` branch of `if args:`, then yes, only `convertor: Type[Convertor]` is sufficient.

But as per my proposal to remove the instantion and have long-lived instantiated singletons in `CONVERTOR_TYPES`, I think you need the `typing.Union[Convertor, typing.Type[Convertor]]`

starlette/routing.py
Mng-dev-ai: Hi @bibajz,
I'm not sure if it should be added to testing or not.
what do you think @Kludex ?

bibajz: Hello, it seems to me that you misunderstood the meaning of `# pragma: no cover`. It is usually found during conditional imports or interface declaration, not within a core functionality. You can try to grep for its usages -

```
$ grep -rHn --after-context=3 --before-context=3 'pragma: \w*\W*cover$' starlette/
...
```

and as an example, see here:

```
starlette/requests.py-12-try:
starlette/requests.py-13-    from multipart.multipart import parse_options_header
starlette/requests.py:14:except ImportError:  # pragma: nocover
starlette/requests.py-15-    parse_options_header = None
```

As you can see, a conditional import.

Therefore I see it as a misuse of this `pragma` and a sign that tests specifically for `compile_path` need to be added.
Also, your previous implementation was incorrect, which also adds to the necessity of tests.

Kludex: Well... You already added a test that covers this, haven't you?

Mng-dev-ai: @bibajz Yes, I think i misunderstood the meaning of #  pragma: no cover
@Kludex I meant for the compile path part which seems I have to add a test for it too as @bibajz declared above

Kludex:
After thinking a lot about this, I've decided to close this PR.

As an alternative for what this PR proposes, the user can create multiple convertors that suit its need - even if it's not that convenient, it doesn't look like a common use case.

Thanks @Mng-dev-ai for the initiative here. It's much appreciated. Sorry for taking long to take a decision here!

This is not a definite "no", but it doesn't look like we have a strong case to have this into Starlette for now. If people still think that this is a cool feature, and we have more people interested, we can discuss this once again. At that time, please create a discussion.

PR #1458, opened 2022-01-30
support non-coroutine asgi apps in TestClient

adriangb:
This adds support for testing ASGI apps like `starlette.endpoints.HTTPEndpoint` via TestClient

This is a piece of #1453

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -1,7 +1,5 @@
-import asyncio
 import contextlib
 import http
-import inspect
 import io
 import json
 import math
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -81,26 +79,22 @@ def _get_reason_phrase(status_code: int) -> str:
...
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        instance = self.app(scope)
-        await instance(receive, send)
+        try:
+            asgi3app = typing.cast(ASGI3App, self.app)
+            aw = asgi3app(scope, receive, send)
+        except TypeError:
+            asgi2app = typing.cast(ASGI2App, self.app)
+            aw = asgi2app(scope)(receive, send)
+        await aw
```
lovelydinosaur, line 97: Do we need to support ASGI 2.0 anymore?
Perhaps we could look at dropping that to keep things more obvious?

(I don't see any great resource in the ASGI docs on the 2 -> 3 changes?)

Kludex: here: https://asgi.readthedocs.io/en/latest/specs/main.html?highlight=2.0#legacy-applications

adriangb: Based on that it seems reasonable for TestClient to drop support for ASGI 2.0, I'm happy to drop it here or in another prequel/sequel PR if there is consensus. As is this is a backwards compatible change.

lovelydinosaur: I think dropping ASGI 2 first would be pretty reasonable, yup.

adriangb:
I opened #1473 to remove ASGI2 support.

I also think that #1453 is an interesting experiment, but is likely not worth moving forward with.

Once #1473 gets merged and eliminates the issue false detection of non-coroutine ASGI 3 apps as ASGI 2.0 apps (whaat this PR was addressing) there is still a tiny issue, but it's on the AnyIO side and would be fixed by https://github.com/agronholm/anyio/pull/416.

Given that no one has complained about this thus far, I'm going to close this PR and if someone _does_ want to use TestClient with `HTTPEndpoint` directly, they'll just have to wait until https://github.com/agronholm/anyio/pull/416 gets merged.

PR #1459, opened 2022-01-31
Prevent BaseHTTPMiddleware from hiding errors of StreamingResponse

### LINKED ISSUE #1433: Raising Exceptions in sub-applications routes

o-fedorov:
The issue is observed when:

*  an application uses a BaseHTTPMiddleware subclass;
*  it provides StreamingResponse;
*  there is an exception raised inside the generator wrapped with StreamingResponse.

Prior to v0.17.0 the exception was raised and the traceback was printed to the server logs. Now it is silently ignored.

Edit by @Kludex : Closes #1433

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -47,8 +47,6 @@ class BaseHTTPMiddleware:
             assert message["type"] == "http.response.start"
 
             async def body_stream() -> typing.AsyncGenerator[bytes, None]:
-                nonlocal app_exc
-
                 async with recv_stream:
                     async for message in recv_stream:
                         assert message["type"] == "http.response.body"
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -47,10 +47,15 @@ class BaseHTTPMiddleware:
             assert message["type"] == "http.response.start"
 
             async def body_stream() -> typing.AsyncGenerator[bytes, None]:
+                nonlocal app_exc
+
                 async with recv_stream:
                     async for message in recv_stream:
                         assert message["type"] == "http.response.body"
                         yield message.get("body", b"")
+
+                if app_exc is not None:
+                    raise app_exc
 
             response = StreamingResponse(
                 status_code=message["status"], content=body_stream()
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -173,6 +173,7 @@ def test_fully_evaluated_response(test_client_factory):
     response = client.get("/does_not_exist")
     assert response.text == "Custom"
 
+
 def test_exception_on_mounted_apps(test_client_factory):
     sub_app = Starlette(routes=[Route("/", exc)])
     app.mount("/sub", sub_app)
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -3,7 +3,7 @@ import pytest
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.base import BaseHTTPMiddleware
-from starlette.responses import PlainTextResponse
+from starlette.responses import PlainTextResponse, StreamingResponse
 from starlette.routing import Route
 
 
@@ -26,6 +26,16 @@ def homepage(request):
 @app.route("/exc")
 def exc(request):
     raise Exception("Exc")
+
+
+@app.route("/exc-stream")
+def exc_stream(request):
+    return StreamingResponse(_generate_faulty_stream())
+
+
+def _generate_faulty_stream():
+    yield b"Ok"
+    raise Exception("Faulty Stream")
 
 
 @app.route("/no-response")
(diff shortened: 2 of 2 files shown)


starlette/middleware/base.py
```diff
@@ -47,11 +47,16 @@ async def coro() -> None:
             assert message["type"] == "http.response.start"
 
             async def body_stream() -> typing.AsyncGenerator[bytes, None]:
+                nonlocal app_exc
```
Kludex: This shouldn't be necessary, right?

Kludex (commented):
Alternative to #1452

Would you mind adding this test?
```python
def test_exception_on_mounted_apps(test_client_factory):
    sub_app = Starlette(routes=[Route("/", exc)])
    app.mount("/sub", sub_app)

    with pytest.raises(Exception) as ctx:
        client.get("/sub/")
    assert str(ctx.value) == "Exc"
```

starlette/middleware/base.py
o-fedorov: You are right, there is no assignment to `app_exc`. Removed.

tests/middleware/test_base.py
```diff
@@ -158,3 +172,12 @@ async def dispatch(self, request, call_next):
     client = test_client_factory(app)
     response = client.get("/does_not_exist")
     assert response.text == "Custom"
+
```
Kludex, line 175:
```suggestion


```

o-fedorov: Thank you, already did it locally

o-fedorov:
> Alternative to #1452
> 
> Would you mind adding this test?
> 
> ```python
> def test_exception_on_mounted_apps(test_client_factory):
>     sub_app = Starlette(routes=[Route("/", exc)])
>     app.mount("/sub", sub_app)
> 
>     client = test_client_factory(app)
>     with pytest.raises(Exception) as ctx:
>         client.get("/sub/")
>     assert str(ctx.value) == "Exc"
> ```

The test is added

Kludex:
I don't think this work if the `Response` object from the middleware is ignored, as below:

```python

class CustomHeaderMiddleware(BaseHTTPMiddleware):
        await call_next(request)
        return Response("Do you see an exception now?")

middleware = [
    Middleware(CustomHeaderMiddleware)
]

```
Can you confirm? Using the same `StreamingResponse` endpoint, you created on your test with this middleware.

#1452 should be able to solve this.

o-fedorov:
Sorry, both this approach and #1452 seems to not solving the issue  if the Response object from the middleware is ignored

Kludex (approved):
> Sorry, both this approach and #1452 seems to not solving the issue if the Response object from the middleware is ignored

My bad... That exception is not even raised! :laughing:

In any case, I like this PR more than the one I created because it brings the _exception conditional_ closer to where it can be retrieved.

Kludex:
More context about this PR can be found on #1452 and #1433.

The issue here is that we check if the endpoint (or app) raised an exception only if `http.response.start` is not sent, but `StreamingResponse` (and mounted apps) pass that conditional, and we can have an exception after that point.

Kludex:
Thanks for the PR @o-fedorov ! :)

PR #1460, opened 2022-01-31
Bump anyio requirement from 3.x to 3.x, 4.x

Kludex:
Checking if `anyio` 4.0 dev branch works with Starlette.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     package_data={"starlette": ["py.typed"]},
     include_package_data=True,
     install_requires=[
-        "anyio @ git+https://github.com/agronholm/anyio.git#4.0-dev",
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Given that this passes, should we switch it up?
Would `"anyio>=3,<=4"` do what we'd want here? (Any 3.x, 4.x version)

Kludex:
No, that will not get `4.x`. `anyio>=3.4,<5.0` should do the job.

lovelydinosaur:
Ah yeah. Shall we switch over to that in this pull request then?

Kludex:
I was waiting the anyio release, but yeah...

lovelydinosaur (approved):

PR #1463, opened 2022-02-01
Using debug mode in ExceptionMiddleware

aminalaee:
A proposal to make use of `debug` mode in the ExceptionMiddleware to process with the Starlette `Debugger` for better debug information.

Right now for example we have this even if debug is set True:
`Method not allowed`

But with this change:
Screenshot 2022-01-28 at 14 27 26

The traceback should probably not be shown but the debugger can be useful when we have a more informative debug page showing variables, routes, path params and etc.

**Still a work in progress**


lovelydinosaur:
Hiya Amin,

That's pretty neat!

I think that there's some good work we could do here *at some point*, but rather than repurposing `ExceptionMiddleware` I'd want routing nopes to instead display the routes that's configured and help the user identify why the request didn't match them.

I personally think we probably need a bit of a "feature freeze" mindset for Starlette and Uvicorn at the moment - there's an awful lot of code churn going on, and *personally* I think we should just leave this is at the moment.

aminalaee:
Hey Tom,

Yes that's what we've discussed before. That makes sense. I'm on the side to keep things as they are. There's a lot going on right now (including PRs from myself 😆)

Thanks for the feedback.

lovelydinosaur:
💛

PR #1470, opened 2022-02-01
Make router decorators gentle towards types

antonagestam:
What does this do? These decorators currently destroy the signatures of the functions they're applied to. We remedy this by using a typevar, which makes the signature of the decorators say the equivalent of "I take any callable, and I will return the exact type that I'm given".

With this example:

```python

app = Starlette()


@app.route("/")
def handle_request(request: Request) -> Response:
    return Response()

reveal_type(handle_request)
```

The current behavior is:

```
demo.py:12: note: Revealed type is "Any"
```

But with this PR applied:

```
demo.py:12: note: Revealed type is "def (request: starlette.requests.Request) -> starlette.responses.Response"
```

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -11,6 +11,9 @@ from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 
+_C = typing.TypeVar("_C", bound=typing.Callable)
+
+
 class Starlette:
     """
     Creates an application instance.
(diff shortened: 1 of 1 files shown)


adriangb:
One the one hand I think this is a pretty reasonable typing improvement.
On the other hand:

```
 # The following usages are now discouraged in favour of configuration
 #  during Starlette.__init__(...)
```

Kludex:
I think we should be strong on decisions: **if** we want to remove it at some point, let's deprecate it first...

I really want this, but if we don't want to estimulate the usage of those decorators, to me it doesn't make sense to make it "more correct" type wise.

In any case, I'll be closing this PR just because of the _in-code_ comment we have, and because I'm trying to reduce the amount of stale PRs we have.

Thanks for the PR @antonagestam , and sorry for taking long to take a decision here!

PR #1473, opened 2022-02-04
remove support for ASGI2 apps from TestClient

adriangb:
https://github.com/encode/starlette/pull/1458#discussion_r799331851

This ended up being a wee bit more involved than I was hoping for, mainly because _we_ were using ASGI 2.0 apps in a lot of our tests. But once it's all said and done, this is a nice cleanup that removes a good bit of library and test code.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -1,7 +1,5 @@
-import asyncio
 import contextlib
 import http
-import inspect
 import io
 import json
 import math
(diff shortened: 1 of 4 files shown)


lovelydinosaur:
Ohhh... very nice. 😌
Yup I'd not realised how much we were still using that internally.

adriangb:
So with #1476 merged, all that is left is to remove the test for ASGI 2 support (in `test_testclient.py`) and remove the support in `testclient.py` itself). I can't do the former without the latter because coverage would drop substantially, and I can't do the latter because it would create a lot of merge conflicts for #1376. So I'm going to close this PR and leave that part of the change to be done in #1376 or a small PR after that gets merged. FYI @Kludex

PR #1475, opened 2022-02-04
Require typing-extensions>=3.10

### LINKED ISSUE #1474: Minimum typing-extensions version is not specified
### Checklist

### Describe the bug

Starlette has a bare `typing-extensions` requirement, with no minimum version. But using older versions of `typing-extensions` will cause an error.

```
pip install typing-extensions~=3.7.0
pip install starlette
python -c "from starlette.applications import Starlette"
```

### Expected behavior

Starlette should specify the minimum version of `typing-extensions` it requires, which would cause the `pip install starlette` command to upgrade `typing-extensions` to the required version (3.10).

### Actual behavior

Starlette doesn't install the required `typing-extensions` version, then the import fails.

### Debugging material

```
$ python -c "from starlette.applications import Starlette"
  File "<string>", line 1, in <module>
  File ".../lib/python3.8/site-packages/starlette/applications.py", line 3, in <module>
  File ".../lib/python3.8/site-packages/starlette/datastructures.py", line 8, in <module>
    from starlette.concurrency import run_in_threadpool
  File ".../lib/python3.8/site-packages/starlette/concurrency.py", line 10, in <module>
ImportError: cannot import name 'ParamSpec' from 'typing_extensions' (.../lib/python3.8/site-packages/typing_extensions.py)
```

### Environment

- Ubuntu 20.04, Python 3.8, Starlette 0.18

### Additional context

_No response_

drasmuss:
Fixes https://github.com/encode/starlette/issues/1474

I manually verified that tests pass with `typing-extensions==3.10.0`. If we wanted to be really rigorous we could add a build to CI that specifically installs 3.10.0, in order to detect if any code is added that requires `typing-extensions>3.10` in the future. But I thought that might be overkill in this case.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "anyio>=3.4.0,<4",
-        "typing_extensions; python_version < '3.10'",
+        "typing_extensions>=3.10.0; python_version < '3.10'",
         "contextlib2 >= 21.6.0; python_version < '3.7'",
     ],
     extras_require={


setup.py
```diff
@@ -39,7 +39,7 @@ def get_long_description():
     include_package_data=True,
     install_requires=[
         "anyio>=3.4.0,<4",
-        "typing_extensions; python_version < '3.10'",
+        "typing_extensions>=3.10.0; python_version < '3.10'",
```
Kludex, line 42: Using the style of the one below :sweat_smile: 
```suggestion
        "typing_extensions >= 3.10.0; python_version < '3.10'",
```

Kludex (approved):

setup.py
drasmuss: Yeah I wondered about that, but the one above uses this style `anyio>=3.4.0` 😄

lovelydinosaur: Gonna just resolve this conversation given that we don't have a consistent style *either* way.
Would accept a PR bringing all the cases into line one way or the other. 🙃

PR #1476, opened 2022-02-07
Convert ASGI 2.0 apps in tests to ASGI 3

adriangb:
This is a subset of #1473 that avoids merge conflicts with #1376.
We can then just remove ASGI 2.0 support in #1376.

cc @tomchristie  and @Kludex

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -99,6 +99,9 @@ def test_handled_exc_after_response(test_client_factory, client):
 
 
 def test_force_500_response(test_client_factory):
+    # use a sentinal variable to make sure we actually
+    # make it into the endpoint and don't get a 500
+    # from an incorrect ASGI app signature or something
     called = False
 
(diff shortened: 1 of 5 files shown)


tests/test_exceptions.py
```diff
@@ -99,11 +99,19 @@ def test_handled_exc_after_response(test_client_factory, client):
...
+    async def app(scope, receive, send):
+        nonlocal called
+        called = True
         raise RuntimeError()
 
     force_500_client = test_client_factory(app, raise_server_exceptions=False)
     response = force_500_client.get("/")
+    assert called
     assert response.status_code == 500
     assert response.text == ""
```
adriangb, line 116: I feel like this entire test should be moved to `test_testclient.py`: it's not testing any of Starlette itself.

lovelydinosaur (approved):
Fantastic, yes.
Good thinking on splitting this one out into an independent PR.
Much nicer now. 👍

PR #1478, opened 2022-02-08
Support for Websocket Denial Response

### LINKED ISSUE #1362: Wrong unauthorized WebSocket status code
### Checklist

### Describe the bug

I have a protected websocket endpoint, with the decorator `@requires('authenticated', status_code=401)`.
If the authentication check fails, the response code from the websocket is `403 Forbidden`, but it should be `401 Unauthorized`.

### To reproduce

Here an example to reproduce the use case.

```python
import uvicorn
from starlette.authentication import AuthenticationBackend, AuthenticationError, AuthCredentials, SimpleUser, requires


        auth = request.headers.get("Authorization", None)
        if auth != 'sample_key':
            raise AuthenticationError('Invalid auth credentials')

app = Starlette(
    middleware=[
        Middleware(AuthenticationMiddleware, backend=BasicAuthBackend()),
    ],
)


@app.websocket_route("/websocket")
@requires("authenticated", status_code=401)
async def websocket(websocket: WebSocket):
    await websocket.accept()
    await websocket.send_json({'message': 'Connected and authenticated!'})
    await websocket.close()


if __name__ == '__main__':
    uvicorn.run("websocket:app", host="0.0.0.0", port=10123)
```

Schermata 2021-12-14 alle 13 22 55

### Expected behavior

The websocket endpoint should return 401, according to the status_code specified in the `@requires` decorator.

### Actual behavior

The websocket endpoint should return 403, even if I specify the desired status_code in the `@requires` decorator.

### Environment

- OS: Linux
- Python version: 3.8
- Starlette version: 0.17.1

paulo-raca:
When a websocket endpoint throws an exception (e.g., because of invalid authentication), we currently end with either with a generic 403 response (triggered by "websocket.close" message) or a generic HTTP500 response (if there is an unhandled error)

Instead, it is better to return useful HTTP error responses using the [Websocket Denial Response](https://asgi.readthedocs.io/en/latest/extensions.html#websocket-denial-response) ASGI extension, if it is available

This PR makes several changes in order to support this extension:
- Added `WebsocketDenialResponse` class to return a HTTP Response to a websocket request (if the extension is supported, otherwise just closes with code 403 as before)
- `AuthenticationMiddleware` and `@requires` have been updated (Fixes #1362)
- `HTTPSRedirectMiddleware` has been updated
    - Improved the URL mapping code
    - Added an argument to specify the port mappings (e.g., if the app is running on ports 8080/8443 instead of 80/443)
- `ExceptionMiddleware` and `ServerErrorMiddleware` have been updated to take advantage of it
    - Both classes have been refactored to share code, and a new `BaseExceptionMiddleware` superclass was added
    - `ExceptionTypeOrStatusCode` and `ExceptionHandler` types were added to simplify typing annotations
    - Error handlers now take a `HTTPConnection` instead of a `Request`


paulo-raca:
Here is a simple demo (Based on FastAPI):

(It needs to be executed with hypercorn, as it is the only ASGI server currently supporting this extension)

```
app = FastAPI(debug=True)

@app.websocket("/ws")
async def websocket_endpoint(ws: WebSocket, err: int = None):
    if err == 500:
        raise Exception("Internal Server Error")
    elif err is not None:
        raise HTTPException(status_code=err)
    await ws.accept()
    await ws.send_text("Hello")
```

### No error:
```
$python -m websockets "ws://localhost:8000/ws"
Connected to ws://localhost:8000/ws.
< Hello
Connection closed: 1000 (OK).
```

### With custom HTTP code:
```
$ python -m websockets "ws://localhost:8000/ws?err=404"
Failed to connect to ws://localhost:8000/ws?err=404: server rejected WebSocket connection: HTTP 404.
```

Response body:
```{"detail":"Not Found"}```

### With uncaught exception
```
$ python -m websockets "ws://localhost:8000/ws?err=500"
Failed to connect to ws://localhost:8000/ws?err=500: server rejected WebSocket connection: HTTP 500.
```
Response body:
```
  [...]
Exception: Internal Server Error
```

paulo-raca:
Superseeded by #1315

Kludex:
Superseded by https://github.com/encode/starlette/pull/1535 actually.

Issue #1481, opened 2022-02-08
Drop usage of routing decorators in test cases.


lovelydinosaur:
For quite a while now we've no longer documented the decorator-style for routing.

*We document this style of routing configuration...*

```python
def homepage(request):
    ...

routes = [
    Route('/', homepage),
]

app = Starlette(routes=routes)
```

*We no longer document this...*

```python
app = Starlette()

@app.route("/")
def homepage(request):
    ...
```

However several of our test cases still continue to use the old style, which is a bit inconsistent.

The same also applies to `app.add_middleware` vs. `app = Starlette(..., middleware=...)`.

We could do with a bit of clean-up here.

* Drop routing decorators in test cases.
* Drop `.add_middleware` in test cases.
* Add docstrings to the now undocumented API - "We no longer document this decorator style API, and its usage is discouraged"

For splitting this into individual chunks I'd probably suggest...

* A pull request for the docstrings change.
* Several pull requests for the test case change. Each pull request should only affect one file.

adriangb:
Wow @aminalaee amazing work 🥳 ! I think #1505 is the last one (I found it by deleting the deprecated decorators and running tests)

aminalaee:
@adriangb Thank you :)
Yes I think that's probably the last one. It will need a final check.

adriangb:
@tomchristie I think we're done here :)

PR #1485, opened 2022-02-09
Remove routing decorators in `test_routing.py`

aminalaee:
Subset of #1481 drops usage of routing decorators in tests/test_routing.py .

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -54,6 +54,45 @@ class PartialRoutes:
         await websocket.close()
 
 
+def func_homepage(request):
+    return Response("Hello, world!", media_type="text/plain")
+
+
+def contact(request):
+    return Response("Hello, POST!", media_type="text/plain")
+
+
+def int_convertor(request):
+    number = request.path_params["param"]
+    return JSONResponse({"int": number})
+
+
+def float_convertor(request):
+    num = request.path_params["param"]
+    return JSONResponse({"float": num})
+
+
+def path_convertor(request):
+    path = request.path_params["param"]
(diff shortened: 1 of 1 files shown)


tests/test_routing.py
```diff
@@ -54,86 +54,91 @@ async def async_ws_endpoint(cls, websocket: WebSocket):
...
+            endpoint=path_with_parentheses,
+            name="path-with-parentheses",
+        ),
+        WebSocketRoute("/ws", endpoint=websocket_endpoint),
+        # WebSocketRoute("/ws/{room}", endpoint=websocket_params),
+    ]
+)
+
+
 @app.websocket_route("/ws/{room}")
```
aminalaee: Need this for coverage in `websocket_route`.

lovelydinosaur: Gotcha. No surprises there.

A sensible approach would be to *first* issue a pull request adding the "We no longer document this decorator style API, and its usage is discouraged" docstrings noted in https://github.com/encode/starlette/issues/1481 and also adding `pragma: nocover` to those decorator methods.

lovelydinosaur: Okay - you ought to be able to add the last one in as well, now.

lovelydinosaur (changes requested):
Once `websocket_params` is also in, this looks good.

tests/test_routing.py
aminalaee: Thanks, done.

lovelydinosaur (approved):

lovelydinosaur:
Fab 🌟

PR #1486, opened 2022-02-09
Add docstrings to discourage routing decorator style API

aminalaee:
Subset of #1481 to add docstrings to discourage usage of decorator style API.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -180,7 +180,7 @@ class Starlette:
         """
         We no longer document this decorator style API, and its usage is discouraged.
         Instead you should use the following approach:
- 
+
         routes = [
             Route(path, endpoint=..., ...),
             ...
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -121,10 +121,6 @@ class Starlette:
     # The following usages are now discouraged in favour of configuration
     # during Starlette.__init__(...)
     def on_event(self, event_type: str) -> typing.Callable:  # pragma: nocover
-        """
-        We no longer document this decorator style API, and its usage is discouraged.
-        """
-
         return self.router.on_event(event_type)
 
     def mount(self, path: str, app: ASGIApp, name: str = None) -> None:
@@ -168,10 +164,6 @@ class Starlette:
     def exception_handler(
         self, exc_class_or_status_code: typing.Union[int, typing.Type[Exception]]
     ) -> typing.Callable:  # pragma: nocover
-        """
-        We no longer document this decorator style API, and its usage is discouraged.
-        """
-
         def decorator(func: typing.Callable) -> typing.Callable:
             self.add_exception_handler(exc_class_or_status_code, func)
             return func
@@ -187,6 +179,14 @@ class Starlette:
     ) -> typing.Callable:  # pragma: nocover
         """
         We no longer document this decorator style API, and its usage is discouraged.
+        Instead you should use the following approach:
+ 
+        routes = [
+            Route(path, endpoint=..., ...),
+            ...
+        ]
+
+        app = Starlette(routes=routes)
         """
 
         def decorator(func: typing.Callable) -> typing.Callable:
@@ -206,6 +206,14 @@ class Starlette:
     ) -> typing.Callable:  # pragma: nocover
         """
         We no longer document this decorator style API, and its usage is discouraged.
+        Instead you should use the following approach:
+
+        routes = [
+            WebSocketReoute(path, endpoint=..., ...),
+            ...
+        ]
+
+        app = Starlette(routes=routes)
         """
 
         def decorator(func: typing.Callable) -> typing.Callable:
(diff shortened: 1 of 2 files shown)


aminalaee:
Need to add no coverage.

Kludex:
If we are discouraging the usage, why not deprecating? But why is it discouraged again? :sweat_smile: 

FastAPI does recommend the decorator way of doing things, specially `on_event` and `exception_handler`.

aminalaee:
I agree that a deprecation warning can be useful here. At least it can get better attention 😄 
But as you said that will probably be causing trouble for FastAPI, right?

lovelydinosaur:
> If we are discouraging the usage, why not deprecating?

It's the route towards a very gentle deprecation path.
We've not been documenting it for ages.
Next we make sure the docstrings also discourage it.
Sometime further along the line we can pending deprecation it.
Then we can deprecate it.

> But why is it discouraged again?

Because constraints are a good thing, and having *one* way to do routing with Starlette, rather than *two* is neater if you want to build a consistent, coherent ecosystem.

The other part of that question would be "well then, why declared-up-front rather than decorator-annotated". Which is a bit of a design sensibility issue. This style fits the design intent of Starlette better. It's overall simpler under the hood. It puts all the routing visibly into one place. It moves users towards an immutable(ish) app instance rather than mutating the routing state based on what module are imported.

> I agree that a deprecation warning can be useful here.

We don't need to step that far ahead just yet.

> FastAPI does recommend the decorator way of doing things, specially on_event and exception_handler.

Which is great - absolutely happy and in favour of frameworks that *use* Starlette using our building blocks to take an alternate design approach.

I don't *think* that FastAPI *uses* the Starlette decorators (I think it uses our routing building blocks within it's own routing decorators methods) - so I don't think we have a concern there. But we *would* want to make a full assessment of any downstream dependancies using the decorator styles if/when we decide to bump 'em up to "pending deprecation"

starlette/applications.py
```diff
@@ -118,9 +118,11 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
         await self.middleware_stack(scope, receive, send)
 
-    # The following usages are now discouraged in favour of configuration
-    #  during Starlette.__init__(...)
```
lovelydinosaur, line 122: Ah, I'd forgotten that we *did* already have a comment pointing in this direction.

That's great. Let's keep this comment, but now *also* add the docstrings.

aminalaee: I've added that back but the diff is about removing a space which I think is correct now, right?

Kludex:
> I don't think that FastAPI uses the Starlette decorators (I think it uses our routing building blocks within it's own routing decorators methods) - so I don't think we have a concern there. But we would want to make a full assessment of any downstream dependancies using the decorator styles if/when we decide to bump 'em up to "pending deprecation"

FastAPI doesn't use `route` and `websocket_route` directly, it uses `on_event` and `exception_handler`.

The problem here is that the editor will show the docstrings when using those two methods with FastAPI. It's not really a problem, but on FastaPI side, we'll need to create our own `on_event` and `exception_handler` methods.

starlette/applications.py
```diff
@@ -176,7 +184,11 @@ def route(
         methods: typing.List[str] = None,
         name: str = None,
         include_in_schema: bool = True,
-    ) -> typing.Callable:
+    ) -> typing.Callable:  # pragma: nocover
+        """
+        We no longer document this decorator style API, and its usage is discouraged.
+        """
```
lovelydinosaur, line 190: I guess it would make sense for each of these docstrings to show the alternate approach in each case.

```python
routes = [
    Route(path, endpoint=..., ...),
    ...
]

```

lovelydinosaur:
> [...] it uses on_event and exception_handler.
>
> The problem here is that the editor will show the docstrings when using those two methods with FastAPI.

Okay. We're not making any behavioural changes here (yet or for a while) so we're fairly safe, but that's an interesting point re. the docstrings.

We could either omit the docstrings on those two for now, or else include them, and make sure that FastAPI subclasses and overrides 'em. I don't mind either way, but perhaps let's not give their team any more work at the moment.

starlette/routing.py
```diff
@@ -745,7 +757,21 @@ def decorator(func: typing.Callable) -> typing.Callable:
...
-    def websocket_route(self, path: str, name: str = None) -> typing.Callable:
+    def websocket_route(
+        self, path: str, name: str = None
+    ) -> typing.Callable:  # pragma: nocover
+        """
+        We no longer document this decorator style API, and its usage is discouraged.
+        Instead you should use the following approach:
+
+        routes = [
+            WebSocketReoute(path, endpoint=..., ...),
```
lovelydinosaur:
```suggestion
            WebSocketRoute(path, endpoint=..., ...),
```

lovelydinosaur:
> FastAPI doesn't use route and websocket_route directly, it uses on_event and exception_handler.

On the `Router` class or on the `Starlette` application class?

Any chance you could you point me to it?

Kludex:
By "use" I mean it uses the Starlette one directly i.e. inherits the methods.

lovelydinosaur:
Okay - let's role with this for now.
There's several other methods that we *could also* be adding docstrings for, but this is a start.

lovelydinosaur (approved):

PR #1489, opened 2022-02-09
Remove routing decorators in `test_application.py`

aminalaee:
Subset of #1481 drops usage of routing decorators in tests/test_application.py .

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -179,7 +179,7 @@ class Starlette:
         methods: typing.List[str] = None,
         name: str = None,
         include_in_schema: bool = True,
-    ) -> None:
+    ) -> None:  # pragma: no cover
         self.router.add_route(
             path, route, methods=methods, name=name, include_in_schema=include_in_schema
         )
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -107,14 +107,7 @@ class Starlette:
         return self._debug
 
     @debug.setter
-    def debug(self, value: bool) -> None:  # pragma: no cover
-        """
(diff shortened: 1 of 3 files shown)


aminalaee:
@tomchristie I've left this to fail on purpose because we're avoiding `add_middleware` method and `host` and `mount` decorators which will lack coverage. Should we treat these the same way and `no cover` them or leave them as they are and clean in another round of PRs?

lovelydinosaur:
We should also `no pragma` those, yup.

docstrings on `host` and `mount` would make sense.

You can either bundle all that into this PR, or treat it separately. It probably doesn't matter that much.

aminalaee:
Based on changes here, I'll double check if there are any `mount` or `host` or any other deprecated usages in the files already merged. Might need another PR to clean all of them.

starlette/applications.py
```diff
@@ -107,7 +107,14 @@ def debug(self) -> bool:
...
     @debug.setter
-    def debug(self, value: bool) -> None:
+    def debug(self, value: bool) -> None:  # pragma: no cover
+        """
+        We no longer document this API, and its usage is discouraged.
+        Instead you should use the following approach:
+
+        app = Starlette(debug=True)
+        """
+
```
lovelydinosaur: Not obvious that we want to change this. It's not in the "The following usages are now discouraged..." block. Can we keep it as-is, and discuss any changes there separately?

```suggestion
    def debug(self, value: bool) -> None:
```

aminalaee: Reverted this change for later discussion.

lovelydinosaur (approved):

PR #1493, opened 2022-02-10
Remove deprecated decorators from test_base

adriangb:


tests/middleware/test_base.py
```diff
@@ -82,20 +82,21 @@ def test_custom_middleware(test_client_factory):
 
 
 def test_middleware_decorator(test_client_factory):
```
aminalaee: Do we need to keep this test?
As the name suggests I think we're not really testing what it says. I tried to remove that in #1499 but I'll close it in favour of this.

lovelydinosaur: > Do we need to keep this test?

I don't think so, no.

adriangb: removed 😄

tests/middleware/test_base.py
```diff
@@ -80,31 +80,6 @@ def test_custom_middleware(test_client_factory):
...
-    client = test_client_factory(app)
-    response = client.get("/")
-    assert response.text == "OK"
-
-    response = client.get("/homepage")
-    assert response.text == "Homepage"
-    assert response.headers["Custom"] == "Example"
-
-
 def test_state_data_across_multiple_middlewares(test_client_factory):
```
aminalaee, line 84:
```suggestion

def test_state_data_across_multiple_middlewares(test_client_factory):
```

aminalaee:
I was expecting when this would happen 😄 The methods deprecated removed need the `no cover`.

adriangb:
Yup added the no covers

aminalaee (approved):

PR #1499, opened 2022-02-10
Remove middleware decorator from test_base.py and test_authentication.py

aminalaee:


tests/middleware/test_base.py
```diff
@@ -81,30 +81,6 @@ def test_custom_middleware(test_client_factory):
         assert text == "Hello, world!"
 
 
-def test_middleware_decorator(test_client_factory):
```
aminalaee, line 84: @adriangb Maybe you need this change in #1493 so we can close this one in favour of that?

PR #1505, opened 2022-02-11
remove deprecated decorator from test_staticfiles.py

adriangb:


starlette/applications.py
```diff
@@ -157,7 +157,23 @@ def host(
 
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(
+        self, middleware_class: type, **options: typing.Any
+    ) -> None:  # pragma: no cover
+        """
```
aminalaee: I'll have to double check if we can add this doctoring.
As mentioned by @Kludex some of these methods are directly exposed through FastAPI and it was decided to not add more work for them now.
So we only add the docstrings to the ones that aren't exposed directly.

adriangb: Oh okay, that makes sense, I can remove it. Sorry if I missed the earlier convo.

aminalaee: No worries. Thank you.

aminalaee (approved):
👍

Bala2105 (commented):
Hi! I'm reviewing this as a new contributor. The change correctly replaces the @app.middleware decorator with the explicit Middleware class and passes it into the starlette instance. The logic remains the same, and it follows the project's goal of removing deprecated syntax. Looks good to me!

PR #1512, opened 2022-02-14
SessionMiddleware uses an explicit path=..., instead of defaulting to the ASGI 'root_path'.

### LINKED ISSUE #1261: Starlette 0.15.0 breaks SessionMiddleware by adding ASGI Path for Subroutes using Mount

lovelydinosaur:
Closes https://github.com/encode/starlette/issues/1261

This pull request reverts https://github.com/encode/starlette/pull/1147.

I'd obviously made the wrong call on that.

The intent was "okay, let's limit the path against which session cookies are applied. Our server might be submounted to a given path, and we shouldn't cookie to / in that case." - Which is reasonable enough idea if it only applied to where the main app was mounted, but as it happens actually just breaks all sorts of cases for us. Because we allow for submounted apps too.

Instead we support a configurable path="/" on SessionMiddleware. This doesn't magically pick up any root path that the server might be mounted to, but *does* allow users to configure the path if they need too.

Also refs https://github.com/encode/starlette/issues/233, #1351.

Prompted by [this discussion](https://github.com/encode/starlette/discussions/1511#discussioncomment-2172604).

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -17,6 +17,7 @@ class SessionMiddleware:
         secret_key: typing.Union[str, Secret],
         session_cookie: str = "session",
         max_age: typing.Optional[int] = 14 * 24 * 60 * 60,  # 14 days, in seconds
+        path: str = "/",
         same_site: str = "lax",
         https_only: bool = False,
     ) -> None:
@@ -24,6 +25,7 @@ class SessionMiddleware:
         self.signer = itsdangerous.TimestampSigner(str(secret_key))
         self.session_cookie = session_cookie
         self.max_age = max_age
+        self.path = path
         self.security_flags = "httponly; samesite=" + same_site
         if https_only:  # Secure flag can be used with HTTPS only
             self.security_flags += "; secure"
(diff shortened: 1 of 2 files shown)


aminalaee (approved):
👍

PR #1519, opened 2022-02-15
Cache request 'body' and 'stream_consumed' in the ASGI scope

### LINKED ISSUE #493: Question about Request.Body()

### LINKED ISSUE #495: Consume request body in middleware is problematic

### LINKED ISSUE #847: Middleware Request parse Hangs forever

lovelydinosaur:
Cache the request `body` and `stream_consumed` in the ASGI scope, rather than on the request instance.

This preserves the existing behaviours when calling `.body()` or `.stream()` multiple times, but extends it to also persisting when new Request instances are created. (Eg. both within some middleware, as well as within an endpoint.)

Similar to #944 but should also handle raising `RuntimeError("Stream consumed")` if user code attempts to stream the request body more than once.

Edit by @Kludex:
- Closes #495
- Closes #847
- Closes #493


adriangb:
This looks nice. I feel like the implementation is easier to grok than #944.

And it's all implementation details so it we can always change it in the future 👍 (for example, I think it may make sense to cache other request state into the asgi scope)

Testing wise, I think we need:
1. A test like [this one to](https://github.com/adriangb/starlette/blob/a4bd353c41d1b647d506a518155100d573c70021/tests/test_requests.py) to verify the bug/issue is fixed
2. A couple tests to verify the current behavior w.r.t. consuming the request body and errors.
3. Maybe tests to verify interaction with middleware. I know this is an edge case, but I think that with this implementation if a middleware consumes the body and then another middleware replaces `receive` the behavior would be different than what we currently have. Might be worth documenting the change with a test.

starlette/requests.py
```diff
@@ -194,9 +194,19 @@ def __init__(
...
+    def _get_request_state(self, name: str, default: typing.Any = None) -> typing.Any:
+        return self.scope.get("extensions", {}).get("starlette", {}).get(name, default)
+
+    def _set_request_state(self, name: str, value: typing.Any) -> None:
+        if "extensions" not in self.scope:
+            self.scope["extensions"] = {"starlette": {name: value}}
+        elif "starlette" not in self.scope["extensions"]:
+            self.scope["extensions"]["starlette"] = {name: value}
+        else:
+            self.scope["extensions"]["starlette"][name] = value
```
adriangb, line 221: Nice, liking the generic functions

adriangb (approved):

adriangb:
This is a test I think is worth at least thinking about:

```python
def test_request_middleware_replaces_receive(test_client_factory):
    """Test the behavior of middleware consuming the request body"""

        # a middleware consumes the body
        await Request(scope, receive, send).body()
        # another middleware replaces receive

         async def new_receive() -> Message:
            return {"type": "http.request", "body": b"data"}

         # the endpoint creates a new Request that uses wrapped_receive
        data = await Request(scope, new_receive, send).body()
        await Response(data)(scope, receive, send)

    resp = client.get("/", content=b"original data")
    assert resp.status_code == 200, resp.content
    assert resp.content == b"data"
```

It passes on master but not on this branch.

retrry:
Any movement with this PR as the #944 was closed?

lovelydinosaur:
So...

I'm not necessarily convinced that this is a road that we should go down.

Starlette is built around "everything is just ASGI". The relevant part here is that middleware is just ASGI middleware.
There's pros and cons to that. The pros are that it helps to promote reusable components, the cons are that it's sometimes a more fiddly abstraction to understand.

My feeling here is that rather than attempting to kludge our middleware into a `request`/`response` style we ought to instead just provide better documentation about the limitations.

In particular:

* Good documentation around how to write plain ASGI middleware.
* Documentation around how to write ASGI middleware that inspects the messages being passed through the `receive` channel.
* As identified in https://github.com/encode/starlette/issues/1029, nudge folks away from `BaseHTTPMiddleware`. It was a neat idea, but there's actually a bunch of different issues with the approach. (#472, #847, #1438)
* Document the constraint the requests using `BaseHTTPMiddleware` or in the exception handlers *cannot* read the request body. And make sure that it's raising a clear and helpful exception if this is attempted.

I get that this is a cookie that a bunch of folks want. But it has some pretty unhealthy ingredients in it, and I'm not sure I'm comfortable selling it.

retrry:
Okay, maybe you have some examples where I could see middleware that inspects or uses request body?

lovelydinosaur:
@retrry Let me pitch this the other way around... if somebody provides one of the concrete use-cases that they're running into, we could then use that to discuss if/how you'd be able to meet that use-case.

I think that tends to be a more productive tactic, since it helps ground the conversion in an actual end-user-facing issue.

retrry:
So my concrete use-case would be that I need to log all POST requests bodies. POST request bodies contain queries, which I need to log for debugging and accounting purposes.

lovelydinosaur:
Okay.

ASGI middleware *is* a little complex, so I need to start from basics in figuring out an example here.

First up I went and reminded myself how Starlette does it's class-based ASGI middleware, which I did by taking a look [at this particular case](https://github.com/encode/starlette/blob/master/starlette/middleware/authentication.py). Because I knew it was a fairly simple one.

Next let's start with an empty example to work from.

```python
class EmptyMiddleware:
    def __init__(
        self,
        app: ASGIApp,
    ) -> None:
        self.app = app

        if scope["type"]  != "http":
            return

        # Do stuff with the HTTP case...
```

Now. There's two different options.

First up, we could ingest the entire request body and log it at the point the middleware runs...

```python
class BodyLoggingMiddleware:
    def __init__(
        self,
        app: ASGIApp,
    ) -> None:
        self.app = app

        if scope["type"]  != "http":
            return

        # Ingest all body messages from the ASGI `receive` callable.
        messages = []
        more_body = True
        while more_body:
            message = await receive()
            messages.append(message)
            more_body = message.get("more_body", False)

        # Log the request
        body = b''.join([message.get("body", b"") for message in messages])
        print(scope["method"], body)

        # Dispatch to the ASGI callable
        async def wrapped_receive():
            # First up we want to return any messages we've stashed.
            if messages:
                return messages.pop(0)

            # Once that's done we can just await any other messages.
            return await receive()

        await self.app(scope, wrapped_receive, send)
```

Note that I'm also having to refer to https://asgi.readthedocs.io/en/latest/specs/www.html to remind myself how the ASGI HTTP events are structured.

Alternatively we could structure the middleware so that the body is *not* read entirely into memory, but instead log it only if/when it is accessed by the application.

```python
class BodyLoggingMiddleware:
    def __init__(
        self,
        app: ASGIApp,
    ) -> None:
        self.app = app

        if scope["type"]  != "http":
            return

        async def wrapped_receive():
            message = await receive()
            more_body = message.get("more_body", False)
            if body:
                # Log each individual chunk as it is ingested.
                print(scope["method"], body, more_body)
            return message

        await self.app(scope, wrapped_receive, send)
```

So.

(And yup, these are a bit fiddly for developers to figure out.)

*Disclaimer: I haven't tested these, let me know if you use either of these patterns. Feedback and alternate implementations are most welcome.*

adriangb:
That generally looks good to me. I went ahead and pushed the second implementation a bit further, intending to handle cases where the app errors or the client disconnects: https://gist.github.com/adriangb/2f003410b05783924d3d6bf3ff18ad6f

Obviously it is a business-level decision if you want to log all of the body even if the client errored, but at lest this shows how you might do that if you wanted to.

retrry:
Hey, thanks for the code - it works and we have been using it for our logging implementation.

I have one additional question: is it possible to access response in the same middleware? Our use case in other project is: we were logging body (with patched starlette with this MR: https://github.com/encode/starlette/pull/944) and logging response headers in the same log message. Can I do the same in async middleware or it is not possible and I should split log messages to separate messages?

Edit: I think the example you wrote should be added to starlette docs or repo 🙂

gnat:
Been closely following this issue for a long time through it's incarnations; have seen people come and go.

From a "treating Starlette as low level ASGI plumbing" perspective, @tomchristie yes  agreed this is the ideal solution, and it is elegant.

That said, the original problem is Starlette, by default, will visually "hang forever" if you access the request body more than 1 time by default. Not expected or intuitive default behaviour out of the box (#495, #771, #493, #847, and FastAPI had many issues parenting into https://github.com/tiangolo/fastapi/issues/394 until they added what is essentially #944 to core).

It takes everyone by surprise, and still does.

Early adopters of Starlette such as myself, @retrry and those in #944 already knew to pre-cache the body, but this is too much friction for newcomers (and just annoying to keep in our brains as regular users.. another thing to explicitly remind newcomers of).

As early adopters we know and accept this going in. We're some of your best UX feedback because we're dogfooding your project from a perspective the core devs may not have anymore.

For someone not already heavily invested in Starlette, the friction and cognitive load of understanding how the low level plumbing all fits together, is too much for us to expect: Starlette is already a HIGHLY innovative, brilliant framework-- but "gotchas" roadblocking people to get expected behaviour need to be as small as possible (ideally non-existent), or people will just choose something else.

gnat:
I strongly feel we need to entertain a merge on this or #944 if for no other reason that:

1. This is an inevitable stumbling block nearly every Starlette user will run into.
2. The current resolution path is to search issues on GitHub. :eyes:  

Even if the perfect documentation exists on the official site, the hangup will still be perceived as a "bug" every newbie will run into and we'll need to refer them to the manual for.

Reminder: The reason we do not have a flood of issues from FastAPI users is because they patched this with something similar to #944 in 2019 (see  https://github.com/tiangolo/fastapi/pull/589 and tiangolo/fastapi#394 for their big thread).

Early adopters of Starlette such as myself already know to work around this by storing the request body, but I could see newcomers seriously re-assessing whether they want to write pure Starlette or choose something else if they hit a snagging issue like this early on.

A "will not fix" resolution is risky to the future growth of the framework.

@tomchristie  Suggestion: Perhaps shipping a tiny, performant, built-in middleware for this instead of modifying the core (make it common like the SessionMiddleware). I'd vastly prefer something like this in core but this may satisfy all parties.

Kludex (approved):
@tomchristie I think this solution is neat. We are not making much of a sacrifice here, we are just expanding the functionality.

From a user's point of view, this is a huge improvement.

Maybe we can move the `_get_request_state` and `_set_request_state` to another module, as we had the talk about moving the Starlette specific extensions to `["extensions"]["starlette"][<extension>]`.

Kludex:
@tomchristie I've added this PR to next milestone release. Please block me if you don't wish to have this in, and if it's the case, perhaps we should take a decision to not have this in until 1.0. We can go back to this after we get the stable release.

If we are not blocked, I'll merge this before the next release (which will be 0.21.0).

adriangb:
I've become convinced that this is not the right approach. Take a look at this example:

```python
from typing import Awaitable, Callable


CallNext = Callable[[Request], Awaitable[Response]]


class LogRequestBodySize(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next: CallNext) -> Response:
        print(len(await request.body()))


def replace_body_middleware(app: ASGIApp) -> ASGIApp:
    async def wrapped_app(scope: Scope, receive: Receive, send: Send) -> None:
        async def wrapped_rcv() -> Message:
            msg = await receive()
            msg["body"] += b"foo-"
            return msg

        await app(scope, wrapped_rcv, send)

    return wrapped_app


async def endpoint(request: Request) -> Response:
    assert (await request.body()).startswith(b"foo-")
    return Response()


app: ASGIApp = Starlette(routes=[Route("/", endpoint)])
app = replace_body_middleware(app)
app = LogRequestBodySize(app)
```

When you hit the `/` endpoint you'll get a 500 response because the body read in the endpoint is not the body 

I know that `replace_body_middleware` seems contrived, but (1) it could be something like uncompressing an incoming request and (2) it's a perfectly valid ASGI middleware.

Kludex:
What's the right approach?

adriangb:
I think I'd prefer something like #1692. This doesn't solve the issue for things that use `Request` outside of `BaseHTTPMiddleware`, but I think it solves the issue for things that use it inside of `BaseHTTPMiddleware` which is most of the cases where users encounter this.

I'm also open to saying this is a wontfix and move towards something like #1691

gnat:
> but I think it solves the issue for things that use it inside of BaseHTTPMiddleware which is most of the cases where users encounter this.

The necessity has been discussed at length previously- it's certainly not just middleware. Using POST data anywhere in a request/response is extremely common for websites. See Toms starting post.

The ideal DX is being able to call `post = await request.form()` or `body = await request.body()` any time and not worrying about consuming it--- or storing it in the request state, or storing it yourself in an intermediate variable somewhere.

Just to name a few off the top of my head: Sanic, FastAPI, vanilla PHP, etc. all do this. Not having this is just extra manual busywork people should not have to get hung up on.

adriangb:
Where else would you want to call `Request.body()`? Like you say it already works in FastAPI endpoints and dependencies (because the same Request instance is used everywhere).

Kludex:
- Closing this in favor of #1692

@adriangb can you confirm that the issues this PR closes, are also closed on #1692? 🙏

Kludex:
#1692 doesn't solve the issues this PR solves.

adriangb:
Mind being more specific about which issue? Also JFYK [this change is going to wreak all sorts of havoc](https://github.com/encode/starlette/pull/1519/files#r1089400760)

gnat:
>Also JFYK this method is going to wreak all sorts of havoc

How? What's wrong with this implementation?

Reminder:
* FastAPI already patches Starlette to cache the request body.
* Sanic, Starlette's closest equivalent, also caches the request body.
* The entire PHP ecosystem, also caches the request body.

tests/test_requests.py
```diff
@@ -493,3 +493,65 @@ async def app(scope, receive, send):
...
+        async for chunk in request2.stream():
+            chunks += chunk
+
+        response = JSONResponse({"body": body.decode(), "stream": chunks.decode()})
+        await response(scope, receive, send)
+
+    client = test_client_factory(app)
+
+    response = client.post("/", data="abc")
+    assert response.json() == {"body": "abc", "stream": "abc"}
```
adriangb, line 557:
```suggestion
    assert response.json() == {"body": "abc", "stream": "abc"}


def test_request_body_then_replace_body(test_client_factory):
    import zlib

    data = "Hello, world!"
    gzipped_payload = zlib.compress(data.encode())

    def ungzip_middleware(app):
        async def wrapped_app(scope, receive, send):
            dec = zlib.decompressobj()  # offset 16 to skip the header
            buffer = bytearray()
            async def wrapped_rcv():
                more_body = True
                while more_body:
                    msg = await receive()
                    more_body = msg.get("more_body", False)
                    buffer.extend(msg["body"])
                    if len(buffer) > 16 or not more_body:
                        msg["body"] = dec.decompress(buffer)
                        yield msg
                        buffer.clear()
            await app(scope, wrapped_rcv().__anext__, send)
        return wrapped_app

    def log_request_body_middleware(app):
        async def wrapped_app(scope, receive, send):
            await Request(scope, receive).body()
            await app(scope, receive, send)
        return wrapped_app

    async def app(scope, receive, send):
        request = Request(scope, receive)
        body = await request.body()

        response = JSONResponse({"body": body.decode()})
        await response(scope, receive, send)

    app = log_request_body_middleware(ungzip_middleware(app))

    client = test_client_factory(app)

    response = client.post("/", data=gzipped_payload)
    assert response.json() == {"body": data}

```

@gnat this is what I'm referring to

`ungzip_middleware` is a perfectly valid ASGI middleware. It could even be inside of a mounted ASGI app or something like that. If we implement this PR we completely break this sort of thing and it breaks in a very non-intuitive way. We're essentially dealing with [cache invalidation](https://martinfowler.com/bliki/TwoHardThings.html)

alex-oleshkevich (approved):

enono78:
@tomchristie @Kludex 
It would be great to get this improvement in Starlette to be able to read safely the request body within a middleware.
I was wondering if this change is planned to be applied also on previous Starlette versions (like 0.19.1)?
I will be happy to discuss also some other potential issues that have been solved (or not) on Starlette 0.19.1

Kludex:
> I was wondering if this change is planned to be applied also on previous Starlette versions (like 0.19.1)?

No.

Kludex:
Given this [comment](https://github.com/encode/starlette/pull/1519#issuecomment-1236166180), we can't accept this PR.

@adriangb worked on replacements for this:
- https://github.com/encode/starlette/pull/2026
- https://github.com/encode/starlette/pull/1692

Let's continue on them.

PR #1521, opened 2022-02-15
add failing test

adriangb:


tests/test_requests.py
```diff
@@ -550,3 +550,24 @@ async def app(scope, receive, send):
...
+            return {"type": "http.request", "body": b"data"}
+
+        # the endpoint creates a new Request that uses wrapped_receive
+        data = await Request(scope, new_receive, send).body()
+        await Response(data)(scope, receive, send)
+
+    client = test_client_factory(app)
+    resp = client.get("/", content=b"original data")
+    assert resp.status_code == 200, resp.content
+    assert resp.content == b"data"
```
adriangb, line 573: @tomchristie this test passes on master but no longer passes on the `keep-request-state` branch. I'd say this is probably "unsupported" behavior and maybe it's not even worth including as a test, but I wanted to call it out.

PR #1522, opened 2022-02-17
Mark type for optional parameter as `Optional`.

garyd203:
This makes it compatible with strict type checking for subclasses.


lovelydinosaur:
We could consider doing this, yup.
But if we do, then let's do it via a pull request that updates it everywhere in the codebase, rather than just in one particular method.
Consistency wins.

PR #1528, opened 2022-02-23
Proof of concept for detecting stream double consumption

adriangb:
As per https://github.com/encode/starlette/pull/1519#issuecomment-1047723040 instead of "fixing" this issue we may want to want to write better documentation and warn/raise exceptions.

This is a proof of concept for _how_ we might go about detecting problematic behaviors or raising errors.

I'm not sure that raising a runtime error here is exactly the only thing we want to do.
We could:

1. Raise an error if the middleware tries to consume the request body.
2. Warn if the middleware tries to consume the request body.
3. Warn if the middleware consumed the request body _and then_ the endpoint tries to do this again.
4. Detect if the middleware consumed all of or only part of the body stream.
5. Insert a new ASGI receive message with `b""` so that the endpoint doesn't fail and instead interprets this as "nothing left in the body"

There are obviously many variations to this, but I think the mechanisms I'm proposing here could apply to all of them.


adriangb:
An alternative to all of this might be to wrap the stream with something that always returns `b""` instead of hanging

PR #1534, opened 2022-03-03
Make responses generic

cglacet:
I think this might be usefull to allow (static) typechecking route responses.

For now responses content type is `typing.Any`, instead we could use generics. 
The following doesn't yield any error:

```python
from typing import TypedDict

class Coordinates(TypedDict):
    latitude: float
    longitude: float

async def valid_route(request: Request) -> JSONResponse[Coordinates]: 
    return JSONResponse({'latitude': 1, 'longitude': 2})
```

While this would raise a type error (`"dict[str, int | str]" is incompatible with "Coordinates"`): 

```python
async def invalid_route(request: Request) -> JSONResponse[Coordinates]: 
    return JSONResponse({'latitude': 1, 'longitude': 'douze'})
```

I know many people are using FastAPI with pydantic, but I needed to have this so I though maybe I'm not alone. 

Maybe the same could be done for responses `.json()`?

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -30,13 +30,16 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
-class Response:
+T = typing.TypeVar("T")
+
+
+class Response(typing.Generic[T]):
     media_type = None
     charset = "utf-8"
 
     def __init__(
         self,
-        content: typing.Any = None,
+        content: T = None,
         status_code: int = 200,
         headers: typing.Mapping[str, str] = None,
         media_type: str = None,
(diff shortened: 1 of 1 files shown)


Kludex:
@euri10 If I'm not mistaken, you were interested on something related to this... But I'm not sure...

Kludex:
@cglacet Thanks for the PR. I think this content is relevant, but I think a discussion about it should be created first. If you're still interested in this, would you mind creating one?

Closing this so we can have a discussion.

Kludex:
When https://github.com/python/mypy/issues/14851 gets in, we can implement it in Starlette.

PR #1349, opened 2021-11-29
Raise a more obvious exception when form data is missing "boundary".

toxadx:
The best strategy is to let KeyError raise when parsing request with missing boundary parameter, so end-user can process KeyError exception in their code.

Closes #1125
Related to #1303

PR #1544, opened 2022-03-10
Add custom exception to handle missing boundary error

### LINKED ISSUE #1542: Raise custom Starlette Exception on missing "boundary"
We should probably catch that and raise something else from Starlette itself :thinking:

_Originally posted by @Kludex in https://github.com/encode/starlette/issues/1349#issuecomment-1028902007_

Check comment for context.

opeco17:
Closes #1542

## Outline
As per [#1542](https://github.com/encode/starlette/issues/1542), I've added a custom exception (`MissingBoundaryException`) to handle boundary parameter related error

## Background
When form data is missing boundary parameter, the following line causes the KeyError
https://github.com/encode/starlette/blob/master/starlette/formparsers.py#L162

In the following discussion, it has been decided to use custom exception instead of using KeyError directly
https://github.com/encode/starlette/pull/1349#issuecomment-1028902007

I've added a custom exception (`MissingBoundaryException`) in this PR for that

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -154,12 +154,17 @@ class MultiPartParser:
         self.messages.append(message)
 
     async def parse(self) -> FormData:
+        from starlette.exceptions import MissingBoundaryException
+
         # Parse the Content-Type header to get the multipart boundary.
         content_type, params = parse_options_header(self.headers["Content-Type"])
         charset = params.get(b"charset", "utf-8")
         if type(charset) == bytes:
             charset = charset.decode("latin-1")
-        boundary = params[b"boundary"]
+        try:
+            boundary = params[b"boundary"]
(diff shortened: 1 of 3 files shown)


starlette/formparsers.py
```diff
@@ -154,12 +154,17 @@ def on_end(self) -> None:
         self.messages.append(message)
 
     async def parse(self) -> FormData:
+        from starlette.exceptions import MissingBoundaryException
```
opeco17, line 157: To avoid a circular import error, I've added an import line in the function instead of top-level in the module

Kludex:

Here's what I've found out:

Kludex:
As we are not following what the initial issue proposed, I'm closing this PR.

Sorry for wasting your time @opeco17.

PR #1549, opened 2022-03-20
Add `typing.Optional` on missing annotations

Kludex:
There was a previous comment by @tomchristie about adding `typing.Optional` where is missing in a single PR.

I can't find the comment.

Note: I need to add `no-implicit-optional` on mypy configuration.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,11 @@ max-line-length = 88
 [mypy]
 disallow_untyped_defs = True
 ignore_missing_imports = True
+no_implicit_optional = True
 show_error_codes = True
+
+[mypy-starlette.testclient]
+no_implicit_optional = False
 
 [mypy-tests.*]
 disallow_untyped_defs = False
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -43,11 +43,14 @@ class Starlette:
         debug: bool = False,
         routes: typing.Optional[typing.Sequence[BaseRoute]] = None,
         middleware: typing.Optional[typing.Sequence[Middleware]] = None,
-        exception_handlers: typing.Mapping[
-            typing.Any,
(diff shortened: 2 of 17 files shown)


odiseo0:
I have a question why use 
`
import typing
`
instead of
`
from typing import Optional, Sequence, Callable
`
Is there a code style guide for contributing in Starlette?

Kludex:
> I have a question why use 
> `
> import typing
> `
> instead of
> `
> from typing import Optional, Sequence, Callable
> `
> Is there a code style guide for contributing in Starlette?

It's just to comply with the current pattern you see around in the code.

odiseo0:
> > I have a question why use
> > `import typing`
> > instead of
> > `from typing import Optional, Sequence, Callable`
> > Is there a code style guide for contributing in Starlette?
> 
> It's just to comply with the current pattern you see around in the code.

Understood, thanks

adriangb (approved):

Kludex:
I've added `no_implicit_optional`, so we don't have a future regression on this subject. I've also ignored this rule on `testclient.py`, as we are _thinking_ about changing the implementation there.

@adriangb Does your approval stands after my latest changes?

setup.cfg
```diff
@@ -5,8 +5,12 @@ max-line-length = 88
 [mypy]
 disallow_untyped_defs = True
 ignore_missing_imports = True
+no_implicit_optional = True
 show_error_codes = True
 
+[mypy-starlette.testclient]
+no_implicit_optional = False
```
adriangb, line 12: 👍 good call

adriangb (approved):
Yep I think this is good. I gave this a scroll through, I didn't see anything wrong so if MyPy is happy, I'm happy. LGTM!

PR #1554, opened 2022-03-26
Ignore mypy issue on `jinja2.contextfunction` function 

Kludex:
`jinja2` released a [new version](https://jinja.palletsprojects.com/en/3.1.x/changes/), and as a consequence our pipeline broke.

EDIT: we need to pin `jinja2` on our pipeline because `mkdocs` doesn't work with the latest `jinja2`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ pytest==7.0.1
 trio==0.19.0
 
 # Documentation
-mkdocs==1.2.3
+mkdocs==1.2.4
 mkdocs-material==8.1.3
 mkautodoc==0.1.0
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,9 @@
 # Optionals
 -e .[full]
 
+# temporary pin until mkdocs release
+jinja2==3.0.3
+
 # Testing
 autoflake==1.4
 black==22.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,6 @@
 # Optionals
 -e .[full]
 
-# temporary pin until mkdocs release
-jinja2==3.0.3
-
 # Testing
 autoflake==1.4
 black==22.1.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,9 @@
 # Optionals
 -e .[full]
 
+# temporary pin until mkdocs release
+jinja2==3.0.3
+
```
```suggestion
```

starlette/templating.py
```diff
@@ -12,7 +12,7 @@
     if hasattr(jinja2, "pass_context"):
         pass_context = jinja2.pass_context
     else:  # pragma: nocover
-        pass_context = jinja2.contextfunction
+        pass_context = jinja2.contextfunction  # type: ignore[attr-defined]
```
adriangb, line 15: Could you clarify why this is necessary? Is Jinja2 untyped / the types are wrong? I'm going to run this locally also, going to take me a sec tho

Kludex: The attribute is not defined on the latest `jinja2` (which is the one installed for Python 3.7+).

The conditional there is because Python 3.6 will have a previous `jinja2` version installed.

adriangb: Oh I see. It's just missing in newer versions of Jinja. And so MyPy complains that it's missing. Seems like a reasonable ignore.

adriangb (approved):

Kludex:
Thanks @adriangb :)

starlette/templating.py
graingert: Can you use sys.version_info instead?

Kludex: I guess we could... I've added a note on the PR that removes Python 3.6 (https://github.com/encode/starlette/pull/1357) to remove this logic anyway.

adriangb: I thought about that but it seems like it could lead to issues because the coupling between the Python version and Jinja2 version is coming from our requirements file, so it's pretty much an implementation detail. If there were a `sys.version_info` but for jinja that _is_ what we would want to do

PR #1556, opened 2022-03-27
add note on type ignore in templating.py

adriangb:
I was a bit confused in #1554  at first so can't hurt to have, but if it's obvious to you then just decline this please


Kludex:
Hmm... Can't we just update the comment above about the functions? Instead of "it will be removed", we can use "it was removed, hence..."

adriangb:
Sure sure, that works as well, I just wanted to clarify specifically why the type ignore is needed

Kludex (approved):

Kludex:
Note for reference: this logic/note will be removed when Starlette drops Python 3.6.

PR #1580, opened 2022-04-07
Deprecate `WS_1004_NO_STATUS_RCVD` and `WS_1005_ABNORMAL_CLOSURE`

### LINKED ISSUE #1578: BUG: Incorrect WebSocket status code 1005 description, missing status code 1006
### Discussed in https://github.com/encode/starlette/discussions/1577

<sup>Originally posted by **vbsd** April  7, 2022</sup>
It seems that there is a mistake in starlette's WebSocket [status code list](https://github.com/encode/starlette/blob/0.19.0/starlette/status.py#L83).

Status code 1005 has the name `1005_ABNORMAL_CLOSURE` and 1006 is missing altogether. However, [Mozilla MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code) indicate that 1005 actually means "no status received". It is status code 1006 that means "abnormal closure".

It looks like what should have been
```
WS_1005_NO_STATUS_RECEIVED = 1005
WS_1006_ABNORMAL_CLOSURE = 1006
```
got mashed together and became just one line:
```
WS_1006_NO_STATUS_RECEIVED = 1005
```

Making starlette backwards-incompatible over this wouldn't make much sense, but it seems reasonable to at least add the two correct lines and document `WS_1005_ABNORMAL_CLOSURE` as deprecated/wrong.</div>

Kludex:
Closes #1578

This PR has the same idea as https://github.com/django/asgiref/pull/272/.

The idea here is to deprecate two constants using the [PEP 562](https://peps.python.org/pep-0562/). Unfortunately I need to use a backport for Python 3.6 - but I guess we'd want to release this for Python 3.6 before removing the Python 3.6 support.

NOTE: After this PR gets in, a patch release is needed.


aminalaee:
@Kludex Is this worth the effort as Python 3.6 is already EOL and there's plan to remove support in Starlette too?
I mean as you said we can hold it after Python 3.6 is removed.

aminalaee (commented):
@Kludex Is this worth the effort as Python 3.6 is already EOL and there's plan to remove support in Starlette too?
I mean as you said we can hold it after Python 3.6 is removed.

Kludex:
> I mean as you said we can hold it after Python 3.6 is removed.

I didn't say that, at least I didn't mean it. I mean that we should fix it before stop supporting Python 3.6, so at least the last version has it right.

euri10 (approved):
looks good to me, thanks @Kludex

adriangb:
So the idea here is to cut 1 release with those code then immediately remove it right?

So:
1. Merge this PR. Cut a patch release w/ Python 3.6 support.
2. New PR that deletes the back port code.
3. Cut a minor release dropping Python 3.6 support.

Am I getting this right?

Kludex:
Yes.

Kludex:
The backport removal will be on the PR that drops Python 3.6.

PR #1581, opened 2022-04-10
use any() instead of forloop

Mng-dev-ai:
Simplify forloop using any() and all() methods


Kludex:
Thanks for the PR!

But... PRs that sacrifice readability or are not clear winners regarding readability or performance are likely to be closed.

PR #1590, opened 2022-04-15
Annotate `samesite` parameter on `set_cookie`

Kludex:


euri10 (commented):
probably weird question but shouldn't it be `Lax` `Strict` and `None` or we lowercase everything and it doesnt make a difference ?

Kludex:
Good question. We lower case anyway on `set_cookie`, but not on the `SessionMiddleware`. As per https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cookie-same-site-00#section-4.1 , both are accepted.

Maybe the right type would be `Literal["lax", "Lax", "none", "None", "strict", "Strict"]`. At this point, is it better to let `str`? Maybe just documenting the values somewhere in the docs?

adriangb:
I think this is good. It helps users but also developers so we can know what values are accepted just by looking at it.

I think we should just type it all as lowercase, even if at runtime we accept both.

Side note: would it make sense to have a _typing.py module for this sort of thing?

adriangb (approved):

Kludex:
> Side note: would it make sense to have a _typing.py module for this sort of thing?

Not sure... Just for the type added here is enough? 🤔 Maybe... Is it the only type that could be added there right now? 🤔 

In any case, I guess is not a blocker for this PR. I'll wait to see if someone has any other comment about this PR, and merge it before 0.20.0.

adriangb:
Just for this it doesn't make sense, but using that argument it would never make sense. I think there are other shared things like this scattered around the codebase? And maybe some of the stuff in typing.py and datastructructures.py could go in there? Just a thought...

PR #1593, opened 2022-04-18
Add 400 response on MultiParser exceptions

### LINKED ISSUE #1303: multipart/form-data subpart without name causes internal error
### Checklist

### Describe the bug

If attempting to read a multipart/form-data content type from a request and one of the subparts' Content-Disposition headers doesn't contain a "name" parameter, the form parser throws a KeyError.

### To reproduce

multipart-buglette.py :
```python


async def handler(request):
    return JSONResponse({k: v for k, v in (await request.form()).items()})


app = Starlette(
    debug=True,
    routes=[
        Route("/", handler, methods=["POST"]),
    ],
)
```
Run with:
```sh
uvicorn multipart-buglette:app
```
OK response:
```sh
% curl -F 'foo=bar' localhost:8000
{"foo":"bar"}
```
Error response:
```sh
% curl -F '=bar' localhost:8000
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 656, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 259, in handle
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 61, in app
  File "/tmp/github/./multipart-buglette.py", line 7, in handler
    return JSONResponse({k: v for k, v in (await request.form()).items()})
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/requests.py", line 247, in form
    self._form = await multipart_parser.parse()
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/formparsers.py", line 212, in parse
    field_name = _user_safe_decode(options[b"name"], charset)
KeyError: b'name'
```

### Expected behavior

[RFC 7578 (Returning Values from Forms: multipart/form-data)](https://www.rfc-editor.org/rfc/rfc7578#section-4.2) does state that:
> The Content-Disposition header field MUST also contain an additional parameter of "name"

So, one might expect a parsing ValueError subtype rejecting the multipart/form-data, similar to getting a `json.decoder.JSONDecodeError` when attempting to read bad JSON.

Or, one might be more lenient and map the lack of "name" to the empty string "".

### Actual behavior

An internal KeyError exception is thrown for the missing "name" parameter.

### Debugging material

Please refer to the "to reproduce" section.

### Environment

- OS: Linux
- Python version: 3.9.7
- Starlette version: master @ 3282d2364119c205adcd13b4e1db8a53af2202c7

### Additional context

Unfortunately, other more egregiously malformed multipart/form-data doesn't fare much better in raising an easily identifiable parsing error; e.g.

```sh
% curl -H 'Content-Type: multipart/form-data' -d 'bogus' localhost:8000
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 656, in __call__
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 259, in handle
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/routing.py", line 61, in app
  File "/tmp/github/./multipart-buglette.py", line 7, in handler
    return JSONResponse({k: v for k, v in (await request.form()).items()})
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/requests.py", line 247, in form
    self._form = await multipart_parser.parse()
  File "/tmp/github/venv/lib/python3.9/site-packages/starlette/formparsers.py", line 177, in parse
    parser = multipart.MultipartParser(boundary, callbacks)
  File "/tmp/github/venv/lib/python3.9/site-packages/multipart/multipart.py", line 1042, in __init__
    self.boundary = b'\r\n--' + boundary
TypeError: can't concat NoneType to bytes
```
...though in this case it is an issue in the underlying `python-multipart` package and not `starlette` itself.

### LINKED ISSUE #1588: Send 400 on missing `boundary`

Here's what I've found out:

_Originally posted by @Kludex in https://github.com/encode/starlette/issues/1544#issuecomment-1080197920_

Kludex:
- Closes #1588
- Closes #1303

Let's call this solution 1.

Why I don't like solution 1:
- It adds a new exception to the `ExceptionMiddleware`, which is weird.

Possible solution 2:
- Use `HTTPException` with import inside the `MultiPartParser` to avoid circular import.

Possible solution 3:
- Create an `http_exception.py` file to hold only `HTTPException`, and import it on `exceptions.py` and `formparsers.py`.
- This solution avoids the antipatterns from previous solutions.

I'm exposing this draft as a self note. Solution 3 looks like the best one.


Kludex:
Solution 2 and 3 are problematic because we need the information that we are running on a Starlette application or not (so we either raise HTTPException or return a PlainTextResponse), and we can't retrieve that there. 😞 See #1601 for code example.

Possible solution 4:
- Add an exception handler on the application level to `MultiParserException`.

Solution 4 and 1 are similar...

EDIT: And now I realize that solution 4 and 1 have the same problem as solution 2 and 3. 🥲 
EDIT2: Based on this information, I think we need to raise a clear exception on formparser i.e. except `TypeError` and raise a custom error e.g. `MultiPartException` with a clear message, and catch that exception on `Starlette` itself to return a 400. This will have the following behavior:
- On `app = Request(scope, receive, send); await request.form()` usage, it will raise `MultiPartException`.
- On `app = Starlette()` and further usage of `.form()`, it will return a response with 400.

Kludex:
Close this in favor of #1617, but the monologue here is important for that one.

PR #1597, opened 2022-04-21
Add verbose return type hint to HTTPConnection.json method

fisher60:
The original return type hint here, `typing.Any`, is too vague. A more accurate and specific type hint can be added. Using `typing.Dict[str, typing.Any]` is more clear as this response should always be JSON, therefore `typing.Any` is misleading.

Initially raised as discussion #1596

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -236,7 +236,7 @@ class Request(HTTPConnection):
             self._body = b"".join(chunks)
         return self._body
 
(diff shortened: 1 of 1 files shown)


Kludex:
As shortly mentioned on https://github.com/encode/starlette/discussions/1596#discussioncomment-2609643, this PR makes the wrong assumption.

Thanks for the interested tho!

PR #1601, opened 2022-04-22
Add 400 response on MultiParser exceptions [2]

Kludex:
Related to #1593


starlette/requests.py
```diff
@@ -250,8 +252,15 @@ async def form(self) -> FormData:
...
-                self._form = await multipart_parser.parse()
+                try:
+                    multipart_parser = MultiPartParser(self.headers, self.stream())
+                    self._form = await multipart_parser.parse()
+                except MultiPartException as exc:
+                    if "app" in self.scope:
+                        raise HTTPException(status_code=400, detail=exc.message)
+                    else:
+                        response = PlainTextResponse("Not Found", status_code=404)
+                        return await response(self.scope, self._receive, self._send)
```
Kludex, line 263: Having this logic here is not cool... :/

PR #1617, opened 2022-05-03
Add 400 response when `boundary` is missing 

### LINKED ISSUE #1588: Send 400 on missing `boundary`

Kludex:
- Closes #1588
- Related to #1303 (I've not included it on this PR for now...)
- Replaces #1593 (monologue about possible solutions from there is important)

Behavior of this PR:
- Custom exception in case `"app" not in scope`.
- `HTTPException(status_code=400)` in case `"app" in scope`.

Changes:
- Move `HTTPException` to `http_exception.py` to avoid circular import.
- Add custom exception `MultiPartException` on missing `boundary`.
- Handle `MultiPartException` on `form()` with the behavior mentioned above.

Alternatives/thoughts related to this PR:
- Check #1593 - my thoughts are listed there in a timely manner...
- The handling of `MultiPartException` can also be on an upper level:
https://github.com/encode/starlette/blob/830f3486537916bae6b46948ff922adc14a22b7c/starlette/routing.py#L250-L261
This alternative would avoid creating the `http_exception.py`... 🤔


starlette/http_exception.py
```diff
@@ -0,0 +1,20 @@
...
+    ) -> None:
+        if detail is None:
+            detail = http.HTTPStatus(status_code).phrase
+        self.status_code = status_code
+        self.detail = detail
+        self.headers = headers
+
+    def __repr__(self) -> str:
+        class_name = self.__class__.__name__
+        return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
```
adriangb: Why do we need to move this?

starlette/formparsers.py
```diff
@@ -38,6 +38,12 @@ def _user_safe_decode(src: bytes, codec: str) -> str:
         return src.decode("latin-1")
 
 
+class MultiPartException(Exception):
+    def __init__(self, message: str, *args: typing.Any) -> None:
```
adriangb: Why do we need to define `__init__` here? Are we using `*args`?

starlette/formparsers.py
```diff
@@ -38,6 +38,12 @@ def _user_safe_decode(src: bytes, codec: str) -> str:
...
+        self.message = message
+        super().__init__(message, *args)
```
Kludex:
```suggestion
   ...
```

starlette/http_exception.py
Kludex: Circular import. `exceptions -> requests`; `requests -> formparsers`; `formparsers -> exceptions`

adriangb: Gotcha. I was always confused as to why there was a middleware in the exceptions module. Could we move the middleware instead?

Kludex: hmm... I didn't think about it. But it makes sense. 👍

starlette/formparsers.py
```diff
@@ -38,6 +38,10 @@ def _user_safe_decode(src: bytes, codec: str) -> str:
         return src.decode("latin-1")
 
 
+class MultiPartException(Exception):
+   ...
```
Kludex:
```suggestion
    ...
```

starlette/http_exception.py
adriangb: might be nice to "deprecate" it from this module as well even if we re-export it in the short term

Kludex: Ok. That's probably the reason I didn't thought about it. With the relocation of `HTTPException`, the import remains here, so there's no need to deprecate it. I'd have to introduce the `__dir__` and `__getattr__` on this file as well.

starlette/formparsers.py
Kludex: No, only the message. I've removed when you mentioned this, but I need to put it back. Conceptually, `MultiPartException` is not an `HTTPException`, so it needs to inherit from `Exception`. The message is necessary here because there will be another part of the code that will raise this error in a future PR (the "name" mentioned on the description).

starlette/http_exception.py
adriangb: I think that's worth it right? I really do feel like the middleware should be in `starlette.middleware.exceptions`

starlette/formparsers.py
adriangb: I still don't understand why it needs to be here, standard behavior for exceptions is that the first argument is a string which corresponds to the exception message. Is this just for when `"app"` is not in scope? If so users can just parse the exceptions, I don't think it's a use case we should be putting effort into supporting.

starlette/requests.py
```diff
@@ -250,8 +251,13 @@ async def form(self) -> FormData:
...
             content_type, options = parse_options_header(content_type_header)
             if content_type == b"multipart/form-data":
-                multipart_parser = MultiPartParser(self.headers, self.stream())
-                self._form = await multipart_parser.parse()
+                try:
+                    multipart_parser = MultiPartParser(self.headers, self.stream())
+                    self._form = await multipart_parser.parse()
+                except MultiPartException as exc:
+                    if "app" in self.scope:
+                        raise HTTPException(status_code=400, detail=exc.message)
```
Kludex, line 259: I don't want to use `exc.args[0]` here. Do you have a better solution? @adriangb

starlette/formparsers.py
Kludex: Let's continue on the other thread.

starlette/requests.py
Kludex: I can have two different exceptions as well, and don't use `exc.message`...

adriangb: Ah I see where this was going now! Sorry for not catching on. Seeing this I think that `message` is okay, but 2 exceptions are also a good idea, I think there's precedent for both. I'll leave it up to you.

starlette/http_exception.py
Kludex: Not sure 😅

starlette/exceptions.py
```diff
@@ -26,86 +23,22 @@ def __repr__(self) -> str:
...
-        if scope["type"] != "http":
-            await self.app(scope, receive, send)
-            return
+__deprecated__ = "ExceptionMiddleware"
 
-        response_started = False
 
-        async def sender(message: Message) -> None:
-            nonlocal response_started
+def __getattr__(name: str) -> typing.Any:  # pragma: no cover
```
Kludex, line 29: I need to add a test for this.

starlette/http_exception.py
adriangb: Fair enough, it's your call now that you've tried it both ways!

tests/test_formparsers.py
```diff
@@ -390,10 +394,19 @@ def test_user_safe_decode_ignores_wrong_charset():
     assert result == "abc"
 
 
-def test_missing_boundary_parameter(test_client_factory):
+@pytest.mark.parametrize(
+    "app,expectation",
+    [
+        (app, pytest.raises(MultiPartException)),
+        (Starlette(routes=[Mount("/", app=app)]), does_not_raise()),
```
adriangb, line 401: The use of `does_not_raise()` here is super nifty!

starlette/exceptions.py
adriangb, line 29: What's the plan to remove all of this deprecation code? I get why we need it, but it'd be nice to know when we're going to remove it so that:

1. We can put it in the code for our knowledge down the road.
2. We can put it in the error message.

I suppose this depends on #1623 , but I think we should make an executive decision (maybe @tomchristie ?) on when / if the deprecation shim is ever going to be removed.

Personally, I feel like 3 minor releases or 1 year, whichever comes first, sounds reasonable?

adriangb (approved):

adriangb (approved):
This is an interesting change.

Overall the change (returning a proper exception to the client instead of a 500) makes absolute sense.

Unfortunately you ended up needing to do a bunch of surgery to the codebase to make it happen because raising an `HTTPexception` from within a `Request` requires that `Request` reference `HTTPException`, but `ExceptionMiddleware` was in the same module as `HTTPException`, which caused circular references. This makes the diff super ugly and huge relative to the small straightforward feature.

But I think that's okay: we are paying the price to fix an issue that already existed in our codebase that might have caused us issues down the road some other time. And personally I think moving the middleware into `middlewares/` really cleans things up nicely.

I'm approving this, I do however want 1 more approver and some consensus on the deprecation policy because this is a breaking change for our downstream dependencies.

starlette/exceptions.py
Kludex: I was more into 6 months and 3 minors... But any number is good for me... I'd just like to have a number 👍 But let's mention those stuff on that discussion.

For here, there's no plan yet. The plan should be defined on that discussion. If this is merged, it's just common sense i.e. in some arbitrary time we just remove it...

Kludex:
> [...] I do however want 1 more approver [...]

Yep 👍 

Thanks for the review! 🙏

starlette/exceptions.py
lovelydinosaur: Didn't realise you could do this at a module level. 😎

starlette/middleware/exceptions.py
```diff
@@ -0,0 +1,93 @@
...
+import typing
+
+from starlette.concurrency import run_in_threadpool
+from starlette.exceptions import HTTPException
+from starlette.requests import Request
+from starlette.responses import PlainTextResponse, Response
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+
+class ExceptionMiddleware:
```
lovelydinosaur, line 11: 👍 Yup - much better to have this here.

Have confirmed to myself that the class here and the class when it was in `exceptions.py` are identical.

starlette/requests.py
```diff
@@ -250,8 +251,13 @@ async def form(self) -> FormData:
             content_type_header = self.headers.get("Content-Type")
             content_type, options = parse_options_header(content_type_header)
             if content_type == b"multipart/form-data":
-                multipart_parser = MultiPartParser(self.headers, self.stream())
-                self._form = await multipart_parser.parse()
+                try:
+                    multipart_parser = MultiPartParser(self.headers, self.stream())
+                    self._form = await multipart_parser.parse()
+                except MultiPartException as exc:
+                    if "app" in self.scope:
```
lovelydinosaur, line 258: Do we need to have the switch here? Can we just *always* coerce to an `HTTPException` here instead? Are we already using this pattern, and if so, where?

tests/test_formparsers.py
```diff
@@ -390,10 +394,19 @@ def test_user_safe_decode_ignores_wrong_charset():
...
+    ],
+)
+def test_missing_boundary_parameter(
+    app, expectation, test_client_factory: typing.Callable[..., TestClient]
+) -> None:
     client = test_client_factory(app)
-    with pytest.raises(KeyError, match="boundary"):
-        client.post(
+    with expectation:
```
lovelydinosaur, line 408: I'd *personally* probably nudge towards the simpler `pytest.raises(KeyError, match="boundary")` case here, just because it's more obviously readable to me.

The parameterised `expectation` is neat, but also more complex to understand.

Just my personal perspective tho, happy to go either way.

lovelydinosaur (approved):
Some questions which could be addressed, but they're not blockers so will leave this in your hands at this point @Kludex.

Really nice switch around with moving the middleware. 👍

starlette/requests.py
adriangb: https://github.com/encode/starlette/blob/master/starlette/routing.py#L253-L258

Kludex: Yep, we use the pattern in the code, as @adriangb pointed out.

Always coerce to an `HTTPException` is an option. I opted to not do it because we only use `HTTPException`s when we are in `Starlette` (as application) e.g. `"app" in self.scope`.

tests/test_formparsers.py
Kludex: The idea was to avoid the creation of two very similar tests. Do you think is clear if I create two tests instead?

starlette/requests.py
adriangb: Realistically, I think we could remove that pattern everywhere. IMO it's not worth supporting (nice) usage of Starlette's Request object outside of a Starlette app. If someone wants to do that, they can handle the HTTPException.

adriangb:
I think your way or making 1 test is the cleanest I've seen, but if it's
only 2 tests and not 10 I also think making it more explicit would be nice.

tests/test_formparsers.py
adriangb: I think your way or combining an exception and non-exception test into 1 is the cleanest I've seen for that pattern, but if it's
only 2 tests and not 10 I also think making it more explicit would be nice

tests/test_formparsers.py
Kludex: Thanks! 🙇

starlette/requests.py
Kludex: >  If someone wants to do that, they can handle the HTTPException.

You have a point, but I think that's another discussion.

I really don't think it's a burden to maintain that, but I wonder if someone really uses that code outside `Starlette` (app)... 🤔

Kludex:
I'm going to merge this.

Some things left unanswered:
- Deprecation policy, how do we proceed?
- Should we check if someone is using parts of Starlette without the Starlette application itself? i.e. `if "app" in scope:`.

PR #1625, opened 2022-05-05
Secret: add a __bool__ implementation

rcoup:
Allows using a Secrets in a conditional statement to check for empty/`None` without needing to cast it to a `str`.

Before:

```python
if str(settings.A_SECRET):  # without the cast it's always True
  # do stuff
```

After:

```python
if settings.A_SECRET:
  # do stuff
```


Kludex (approved):

Kludex:
Thanks for the PR @rcoup !

PR #1630, opened 2022-05-09
Document unsuported `params` on `TestClient.websocket_connect`

### LINKED ISSUE #1187: Test client drops WebSocket query parameters
### Checklist

### Describe the bug

When using the method `websocket_connect` of `TestClient` to connect to a WebSocket, query parameters passed through the `params` argument are dropped.

It works if query parameters are hard-coded directly in the URL.
### To reproduce

```py


    assert scope['type'] == 'websocket'
    await websocket.accept()
    await websocket.send_json(dict(websocket.query_params))
    await websocket.close()


def test_app():
    client = TestClient(app)
    with client.websocket_connect('/', params={"foo": "bar"}) as websocket:
        data = websocket.receive_json()
        assert data == {"foo": "bar"}  # Fails
```

### Expected behavior

The query parameters should be kept when passing them through the `params` argument.

### Actual behavior

The query parameters are dropped.

### Debugging material

The issue probably comes from the [prepare_url method in requests](https://github.com/psf/requests/blob/f6d43b03fbb9a1e75ed63a9aa15738a8fce99b50/requests/models.py#L373-L378). Indeed, if it detects a scheme different than `http`, it stops the URL preparation and thus drops the parameters.

The solution could be to encode them manually in `websocket_connect`, at the beginning:

https://github.com/encode/starlette/blob/d26f44b4aeaec63ae8ce0a4f34dea7439da48012/starlette/testclient.py#L434-L437

### Environment

- OS: macOS 11.3.1
- Python version: 3.7.10
- Starlette version: 0.14.2
- Requests version: 2.25.1

### Additional context

Since it works when directly putting them in the URL string, it's not a big deal; but I thought it could be worth documenting it since it puzzled me for quite some time.

Kludex:
- Closes #1187

Alternative solution:
- Include the `params` into the URL.

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -97,6 +97,15 @@ May raise `starlette.websockets.WebSocketDisconnect` if the application does not
 
 `websocket_connect()` must be used as a context manager (in a `with` block).
 
+!!! note
+    The `params` argument is not supported by `websocket_connect`. If you need to pass query arguments, hard code it
+    directly in the URL.
+
+    ```python
+    with client.websocket_connect('/path?foo=bar') as websocket:
+        ...
+    ```
+
 #### Sending data
 
 * `.send_text(data)` - Send the given text to the application.


lovelydinosaur (approved):
Thank you  @Kludex. 🙏🏼

This is a nice easy-to-approve documentation fix. 😊

PR #1631, opened 2022-05-10
Allow cross-origin requests by path regex

phuong:
### Issue
- Could not be able to allow cross-origin requests based on a path pattern, no matter where's the origin domain.

### Updated
- Added `allow_path_regex` param, CORSmiddleware will allow past the check on paths that match with defined param.
- Updated the docs for the new param.
- Add test case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ venv*/
 .python-version
 build/
 dist/
+.idea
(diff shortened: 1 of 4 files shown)


Kludex:
We need to add the contribution guidelines... The PR placeholder is ignored. 😞

Please open a discussion first so we can discuss about the issue that motivated this PR.

PR #1648, opened 2022-05-24
Remove converter from path when generating `OpenAPI` schema

### LINKED ISSUE #746: Endpoint path parameters type specification compatibility with OAS
When specifying a convertor type for a path parameter like so:
```
...
Route('/users/{user_id:int}', user, methods=["GET", "POST"])
...
```

The OAS schema generated using `SchemaGenerator` interprets the whole portion within `{}`  eg. `'user_id:int'` as the required path parameter variable name, which then requires the parameter name in OAS to be `user_id:int` instead of just `user_id`.

user_idint

I think the convertor is great in that a GET request to `/users/foo` will result in `404: NOT FOUND`. Of course, another option is to have a `try-except` clause to handle non-integer values within the method... but I was wondering what the community thinks about this and if this even should be supported by the `SchemaGenerator`.

### LINKED ISSUE #1646: SchemaGenerator and path parameter converters
### Discussed in https://github.com/encode/starlette/discussions/1645

<sup>Originally posted by **jvail** May 23, 2022</sup>
Hi,

if I use type converters in a route definition ('/somepath/{parameter:int}') and generate an openAPI schema then I'd expect that the type, the converter is stripped from the URL template. But currently i get 

```json
"paths": {
        "//somepath/{parameter:int}": {}
}
```

instead of 

```json
"paths": {
        "//somepath/{parameter}": {}
}
```

Is that by purpose?
If so the parameter names inside the doc string don't match anymore if parsed e.g. inside postman. Postman would create a parameter named "parameter:int" instead of just "parameter".

Thank you
Jan
</div>

Kludex:
- Closes https://github.com/encode/starlette/issues/1646
- Closes #746

To be more correct, I'd have to add the information about the type of the path parameter...


starlette/schemas.py
```diff
@@ -49,10 +50,11 @@ def get_endpoints(
 
         for route in routes:
```
Kludex, line 51: `route` may not have path attribute, that's why the `path =...` is not here.

jvail:
Hi, thank you for fixing this so quickly. Just a quick question: Why not put it into the next bug fix release e.g. 0.20.4 instead 0.21? Unfortunately I am a bit stuck continuing my code until this is fixed.

Kludex:
> Hi, thank you for fixing this so quickly. Just a quick question: Why not put it into the next bug fix release e.g. 0.20.4 instead 0.21? Unfortunately I am a bit stuck continuing my code until this is fixed.

Yeah, it can be on 0.20.4.

starlette/schemas.py
```diff
@@ -64,23 +66,29 @@ def get_endpoints(
...
-                    endpoints_info.append(
-                        EndpointInfo(route.path, method.lower(), func)
-                    )
+                    endpoints_info.append(EndpointInfo(path, method.lower(), func))
 
         return endpoints_info
 
+    def _remove_converter(self, path: str) -> str:
+        """
+        Remove the converter from the path.
```
lovelydinosaur, line 88: I'd suggest we add an example to make the intent clear. Something like?...

```suggestion
        Remove the converter from the path.
        For example, a route like this:
            Route("/users/{id:int}", endpoint=get_user, methods=["GET"])
        Should be represented as `/users/{id}` in the OpenAPI schema.
```

lovelydinosaur (approved):
Looks good, yup. One suggestion, but not a blocker.

starlette/schemas.py
Kludex: Sure! Thanks Tom! 👍

PR #1649, opened 2022-05-24
Add Mount(..., middleware=[...])

### LINKED ISSUE #685: Include routing information for middleware

adriangb:
- Closes #1464 

By @Kludex:
- Closes #685
- Related to #1286


docs/middleware.md
```diff
@@ -41,6 +41,37 @@ application would look like this:
 * Routing
 * Endpoint
 
+Middleware can also be added to `Mount`, which allows you to apply middleware to a single route, a group of routes or any mounted ASGI application:
```
florimondmanca: I might be tempted to move this into a section, probably as "Applying middleware on mounts" between "BaseHTTPMiddleware" and "Using middleware from other frameworks"?

My reasoning is, this seems like an advanced functionality / not something people might _typically_ need to do. So putting this so prominently, just below something as fundamental as "middleware runs top-to-bottom", feels a bit odd to me.

docs/middleware.md
```diff
@@ -41,6 +41,37 @@ application would look like this:
...
+    )
+]
+
+app = Starlette(
+    routes=routes,
+    middleware=[
+        Middleware(HTTPSRedirectMiddleware),
+    ],
+)
+```
```
florimondmanca: Should we try to find a more motivating example here?

Here: https://github.com/tiangolo/fastapi/issues/1174#issuecomment-820584014 the use case would be applying CSP headers only on certain parts of a site.

Another example could be "apply CORS only on a `/api` sub-mount, representing a public API".

Here, I have a use case of patching headers for a specific sub-route: https://github.com/florimondmanca/www/blob/969ab5039f6dcfd86bcec63e45af1e1e9138b46d/server/routes.py#L31-L38

So, looks like a "selectively apply headers" use case could be motivating enough?

---

Side note - I see this issue was also mentioned in the previous PR: https://github.com/encode/starlette/issues/685 but I'm not sure how its "find route name" use case would be answered by per-mount middleware. would `request.scope` on sub-mounts then contain a `"route"` key?

docs/middleware.md
adriangb: Yeah I was thinking the same thing 😃

docs/middleware.md
adriangb: I agree the current example is not the most motivating, but I also don't want to the point to get lost in a bunch of parameters to multiple middleware.

> would request.scope on sub-mounts then contain a "route" key?

I think Starlette needs to in general think about a strategy for `request.scope`. I would make a `scope["extensions"]["starlette"]` `TypedDict` with, yes, a key to include the path the route traversed. But currently I think we'd have to dump it in `scope`, which we should avoid. So let's punt on this. Edit: this discussion https://github.com/encode/starlette/discussions/1511

docs/middleware.md
adriangb: Moved

tests/test_routing.py
```diff
@@ -746,3 +758,93 @@ def __call__(self, request):
...
 def test_route_name(endpoint: typing.Callable, expected_name: str):
     assert Route(path="/", endpoint=endpoint).name == expected_name
+
+
+def assert_middleware_header_route(request: Request):
+    assert getattr(request.state, "middleware_touched") == "Set by middleware"
+    return Response()
+
+
+class AddHeadersMiddleware(BaseHTTPMiddleware):
```
Kludex: Can we avoid the `BaseHTTPMiddleware` on the tests? 👀

adriangb: Yeah good call. I'll change it.

adriangb: Donzo!

tkrause:
This would be really useful. Is there a plan to merge and release this still at some point?

Kludex:
I'll review this before the next minor release.

progamesigner:
The implementation is simple and works well. However, there is a different behavior in both if an exception raised in handler.

This exception will be handled by `ExceptionMiddleware` and thus app-level middleware will still be able to change response such as appending extra tracking header. But this is not the case using router-level middleware (wherher #1649 or #1464).

I wonder if this behavior changing is expected and should be documented or we need build the middleware stack per `Mount` or `Router`? I currently remove the app-level middleware and wrap router-level middleware between `ServerErrorMiddleware` and `ExceptionMiddleware`.

adriangb:
Great point @progamesigner

One option would be to add `ExceptionMiddleware` automatically if the user supplies any other middleware. But that would get awkward because the exception handlers wouldn't propagate from the application level, so you'd be left juggling two sets of exception handlers. Sounds like a footgun to me. So maybe the best thing we can do is simply document this behavior?

I think this really only applies to things that plan to modify or look at the response, right? If you are trying to modify the response (e.g. GZipMiddleware) you probably _want_ the exceptions to just tear through you (no point in compressing a `404` response). In the cases where you want to inspect the response it may make sense to have the route level middleware just "mark" the request by adding a key into the scope like `scope["foobar.log-request"] = True` and then have an application level middleware be what is actually doing the heavy lifting. That way the route-level middleware contains the routing specific logic but doesn't have to handle errors and if another middleware modifies the response (e.g. by compressing it) it will happen before the response info gets logged (so the `content-length` won't be off and such).

adriangb:
I added a note to the docs explaining https://github.com/encode/starlette/pull/1649#issuecomment-1189821544

Kludex:
I'm going to check this tonight.

docs/middleware.md
```diff
@@ -683,6 +683,39 @@ to use the `middleware=<List of Middleware instances>` style, as it will:
 * Ensure that everything remains wrapped in a single outermost `ServerErrorMiddleware`.
 * Preserves the top-level `app` instance.
 
+## Applying middleware to `Mount`s
+
+Middleware can also be added to `Mount`, which allows you to apply middleware to a single route, a group of routes or any mounted ASGI application:
+
+```python
+from starlette.applications import Starlette
+from starlette.middleware.gzip import GzipMiddleware
```
```suggestion
from starlette.middleware import Middleware
from starlette.middleware.gzip import GZipMiddleware
from starlette.routing import Mount, Route
```

docs/middleware.md
```diff
@@ -683,6 +683,39 @@ to use the `middleware=<List of Middleware instances>` style, as it will:
...
+routes = [
+    Mount(
+        "/",
+        routes=[
+            Route(
+                "/example",
+                endpoint=...,
+            )
+        ],
+        middleware=[GzipMiddleware]
```
```suggestion
        middleware=[Middleware(GZipMiddleware)]
```

docs/middleware.md
```diff
@@ -683,6 +683,39 @@ to use the `middleware=<List of Middleware instances>` style, as it will:
...
+app = Starlette(routes=routes)
+```
+
+Note that middleware used in this way is *not* wrapped in exception handling middleware like the middleware applied to the `Starlette` application is.
+This is often not a problem because it only applies to middleware that inspect or modify the `Response`, and even then you probably don't want to apply this logic to error responses.
+If you do want to apply the middelware logic to error responses only on some routes you have a couple of options:
+
+* Add an `ExceptionMiddleware` onto the `Mount`
+* Add a `try/except` block to your middleware and return an error response from there
+* Split up marking and processing into two middlewares, one that gets put on `Mount` which simply marks the response as needing processing (for example by setting `scope["log-response"] = True`) and another applied to the `Starlette` application that does the heavy lifting.
```
Kludex: Hmm... This works... 🤔 

```python


async def home(request):
    return Response("Hi there!")


async def exception_handler(request, exc):
    return Response("I have a 400 for you!", status_code=400)


class PotatoException(Exception):
    ...


class CustomMiddleware:
    def __init__(self, app):
        self.app = app

        raise PotatoException()


routes = [
    Mount(
        "/",
        routes=[
            Route(
                "/example",
                endpoint=home,
            )
        ],
        middleware=[Middleware(CustomMiddleware)],
    ),
]
app = Starlette(routes=routes, exception_handlers={PotatoException: exception_handler})
```

Kludex (changes requested):

Kludex:
I think we can take this opportunity to add a more real case usage instead of `GZipMiddleware`, maybe a middleware related to the issue that motivates this PR. What do you think?

docs/middleware.md
adriangb: Try this example:

```python

class PotatoException(Exception):
    ...

async def bad_endpoint(request):
    raise PotatoException

async def exception_handler(request, exc):
    return Response("I have a 400 for you!", status_code=400)

class CustomMiddleware:
    def __init__(self, app, name: str):
        self.app = app
        self.name = name

        async def wrapped_send(msg):
            if msg["type"] == "http.response":
                print(f"{self.name} called for {scope['raw_path']}")
            await send(msg)
        await self.app(scope, receive, wrapped_send)

routes = [
    Mount(
        "/mount",
        routes=[
            Route(
                "/bad",
                endpoint=bad_endpoint,
            )
        ],
        middleware=[Middleware(CustomMiddleware, name="on_mount")],
    ),
    Route(
        "/good",
        endpoint=bad_endpoint,
    )
]

app = Starlette(
    routes=routes,
    exception_handlers={PotatoException: exception_handler},
    middleware=[Middleware(CustomMiddleware, name="on_app")],
)
```

The "on_mount" middleware will never be called because PotatoException tears through `Mount`'s middleware stack. On the other hand "on_app" always gets called because it's "protected" by ErrorMiddleware.

adriangb:
> I think we can take this opportunity to add a more real case usage instead of `GZipMiddleware`, maybe a middleware related to the issue that motivates this PR. What do you think?

I was going to say yes but: https://github.com/encode/starlette/discussions/1837. So `GZipMiddleware` is actually a motivating example, right?

docs/middleware.md
Kludex: Ah, got it. But that's not what I got from what's written here...

Do you think an image with middleware/app/mount as "blocks" would be helpful understanding this? 🤔

(I can help if you think it makes sense, but I lack design skills to do it 😎 👍)

Kludex:
> > I think we can take this opportunity to add a more real case usage instead of `GZipMiddleware`, maybe a middleware related to the issue that motivates this PR. What do you think?
> 
> I was going to say yes but: #1837. So `GZipMiddleware` is actually a motivating example, right?

- I think the motivation was https://github.com/encode/starlette/issues/685, but fair enough. 👍

docs/middleware.md
adriangb: That sounds like a good idea, please give it a shot! I won't have time until next week (and also don't have the design skills 😆). Maybe do it with ascii text instead of an image since it will be easier to embed?

Kludex (approved):
Thanks @adriangb 🎉 

I think it will be great if we can add a test that fails with the behavior you mentioned on the documentation, and I know you like those kinds of tests... 😎 👍 What do you think?

A bit of reference about this PR, because my memory is short, and I like to make things clear:

There were two previous attempts to solve the issue, which the first one was actually approved (#1286), but I pushed back. On the second one (#1464), @florimondmanca suggested adding a `middleware` parameter only on the `Mount` class, and other members agreed, so it was rejected as well. And now, I think we found the solution that makes more sense. Short solution, great improvement. 👍

docs/middleware.md
Kludex: This is not a blocker. 👍
I'll not have time to do this, if someone eventually wants, feel free to open a PR with it. 👍 

(Please do not resolve this conversation so ppl see it)

docs/middleware.md
```diff
@@ -683,6 +683,41 @@ to use the `middleware=<List of Middleware instances>` style, as it will:
...
+        middleware=[Middleware(GZipMiddleware)]
+    )
+]
+
+app = Starlette(routes=routes)
+```
+
+Note that middleware used in this way is *not* wrapped in exception handling middleware like the middleware applied to the `Starlette` application is.
+This is often not a problem because it only applies to middleware that inspect or modify the `Response`, and even then you probably don't want to apply this logic to error responses.
+If you do want to apply the middelware logic to error responses only on some routes you have a couple of options:
```
florimondmanca: (Typo)

```suggestion
If you do want to apply the middleware logic to error responses only on some routes you have a couple of options:
```

tests/test_routing.py
```diff
@@ -768,6 +780,101 @@ def test_route_name(endpoint: typing.Callable, expected_name: str):
...
+    ),
+    middleware=[Middleware(AddHeadersMiddleware)],
+)
+
+
+@pytest.mark.parametrize(
+    "route",
+    [
+        mounted_routes_with_middleware,
+        mounted_routes_with_middleware,
```
florimondmanca: Duplicated line:

```suggestion
```

tests/test_routing.py
```diff
@@ -768,6 +780,101 @@ def test_route_name(endpoint: typing.Callable, expected_name: str):
...
+    response = test_client.get("/http")
+    assert response.status_code == 200
+    assert response.headers["X-Test"] == "Set by middleware"
+
+
+@pytest.mark.parametrize(
+    "route",
+    [
+        mounted_routes_with_middleware,
+        mounted_routes_with_middleware,
```
florimondmanca: Duplicated as well, so I think we can get rid of the `@parametrize`.

tests/test_routing.py
```diff
@@ -768,6 +780,101 @@ def test_route_name(endpoint: typing.Callable, expected_name: str):
...
+        mounted_app_with_middleware,
+    ],
+)
+def test_mount_middleware(
+    test_client_factory: typing.Callable[..., TestClient],
+    route: BaseRoute,
+) -> None:
+    test_client = test_client_factory(Router([route]))
+    response = test_client.get("/http")
```
florimondmanca, line 849: Should we also `assert` that calling a route not under the `Mount` does _not_ apply the middleware? E.g. testing a `test_client.get("/")`.

docs/middleware.md
florimondmanca: I'm not sure I understand why requests to sub-mounted error routes bypass middleware associated with that mount? Or maybe that's not the problem here? I'm confused. Isn't `ErrorMiddleware` supposed to wrap the _entire_ `Starlette.router`, which includes sub-apps?

Kludex: I got this when I ran the snippet above: https://github.com/encode/starlette/pull/1649/files#r960241635

docs/middleware.md
```diff
@@ -683,6 +683,41 @@ to use the `middleware=<List of Middleware instances>` style, as it will:
...
+
+* Add an `ExceptionMiddleware` onto the `Mount`
+* Add a `try/except` block to your middleware and return an error response from there
+* Split up marking and processing into two middlewares, one that gets put on `Mount` which simply marks the response as needing processing (for example by setting `scope["log-response"] = True`) and another applied to the `Starlette` application that does the heavy lifting.
```
florimondmanca: Not that simple :-)

```suggestion
* Split up marking and processing into two middlewares, one that gets put on `Mount` which marks the response as needing processing (for example by setting `scope["log-response"] = True`) and another applied to the `Starlette` application that does the heavy lifting.
```

docs/middleware.md
florimondmanca: Lemme try to draw this out to make sure I understand...

Looking at the Starlette code, we have this layering:

```python
routes = [
    Route("/", ...),
    Mount("/sub", sub, middleware=[Middleware(MountMiddlewareA)]),
]

app = Starlette(
    routes=routes,
    middleware=[Middleware(MiddlewareB), Middleware(MiddlewareC)],
)
```

```python
ServerErrorMiddleware(
  MiddlewareC(
    MiddlewareB(
      ExceptionMiddleware(
        Router(
          routes=[
            Route("/", ...),
            Mount("/sub", MiddlewareA(sub))
          ]
        )
      )
    )
  )
)
```

So, if an exception occurs in `/`, it gets handled by `ExceptionMiddleware` and `Middleware{B,C}` get a well-defined response _[perhaps an error response]_ to process.

But if an exception occurs in `/sub*`, it gets sent to `MiddlewareA`, which has to handle it if it hopes to do some response processing (otherwise that processing isn't run).

Is that right?

Kludex: Yes, that's right.

Adrian's example was missing something (I had to add it myself before, and I forgot)...

```python

class PotatoException(Exception):
    ...

async def bad_endpoint(request):
    raise PotatoException

async def exception_handler(request, exc):
    return Response("I have a 400 for you!", status_code=400)

class CustomMiddleware:
    def __init__(self, app, name: str):
        self.app = app
        self.name = name

        async def wrapped_send(msg):
            if msg["type"] == "http.response.start":  # IT WAS MISSING `.start`
                print(f"{self.name} called for {scope['raw_path']}")
            await send(msg)
        await self.app(scope, receive, wrapped_send)

routes = [
    Mount(
        "/mount",
        routes=[
            Route(
                "/bad",
                endpoint=bad_endpoint,
            )
        ],
        middleware=[Middleware(CustomMiddleware, name="on_mount")],
    ),
    Route(
        "/good",
        endpoint=bad_endpoint,
    )
]

app = Starlette(
    routes=routes,
    exception_handlers={PotatoException: exception_handler},
    middleware=[Middleware(CustomMiddleware, name="on_app")],
)
```

Kludex: We can use @florimondmanca 's drawing here for the explanation 👍

adriangb: Sorry for the broken example 😫, but yes you got it right Florimond!

tests/test_routing.py
adriangb: Fixed in eee6a6f

tests/test_routing.py
adriangb: Added in eee6a6f

docs/middleware.md
adriangb: I added a test in aec580f

BilalAlpaslan (approved):

adriangb:
I'm going to go ahead and merge it so we can proceed with the next release, if anyone has any last minute concerns we can address them in a separate PR before the release.

docs/middleware.md
adriangb: @florimondmanca I'm realizing now that it was not clear if this was completely resolved, sorry if I missed it. Do you think we should tweak the docs more, maybe adding something along the lines of your explanation in https://github.com/encode/starlette/pull/1649#discussion_r962144793?

adriangb:
A thought about this PR, Request objects and error handling. I wonder if we maybe should’ve had ExceptionMiddleware wrap each Route instead of being a middleware at the application level? That would have both resolved the concerns in https://github.com/encode/starlette/pull/1649#discussion_r962144793 and also https://github.com/encode/starlette/pull/1692#issuecomment-1266401636 because we’d be able to (1) make sure there is nothing in between ExceptionMiddleware and the route and (2) use the same Request object for error handlers and the endpoint. I wonder if this could be made backwards compatible…

PR #1654, opened 2022-05-31
Shield background tasks

### LINKED ISSUE #1438: Background tasks are cancelled if the client closes connection

Kludex:
I'm not sure if we should be touching `BackgroundTasks`.

This PR allows `BackgroundTasks` to run until they finish. 

- Closes https://github.com/encode/starlette/issues/1438

### DIFF
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -1,5 +1,8 @@
 import sys
 import typing
+from asyncio import shield
+
+import anyio
 
 if sys.version_info >= (3, 10):  # pragma: no cover
     from typing import ParamSpec
@@ -39,5 +42,7 @@ class BackgroundTasks(BackgroundTask):
         self.tasks.append(task)
 
     async def __call__(self) -> None:
-        for task in self.tasks:
-            await task()
+        async with anyio.create_task_group():
+            with anyio.CancelScope(shield=True):
+                for task in self.tasks:
+                    await task()
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -1,7 +1,5 @@
 import sys
 import typing
-from asyncio import shield
-
 import anyio
 
 if sys.version_info >= (3, 10):  # pragma: no cover
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -1,5 +1,6 @@
 import sys
 import typing
+
 import anyio
 
 if sys.version_info >= (3, 10):  # pragma: no cover


starlette/background.py
```diff
@@ -1,5 +1,8 @@
 import sys
 import typing
+from asyncio import shield
+
```
Kludex:
```suggestion
```

starlette/background.py
```diff
@@ -1,5 +1,6 @@
 import sys
 import typing
+import anyio
```
Kludex, line 4:
```suggestion

import anyio
```

starlette/background.py
```diff
@@ -39,5 +41,7 @@ def add_task(
         self.tasks.append(task)
 
     async def __call__(self) -> None:
-        for task in self.tasks:
-            await task()
+        async with anyio.create_task_group():
+            with anyio.CancelScope(shield=True):
+                for task in self.tasks:
```
adriangb, line 46: Could we make this opt in on a per-task basis? For backwards compatibility and because I don't think in some cases it _does_ make sense to cancel the background task if the client closes the connection.

starlette/background.py
```diff
@@ -39,5 +41,7 @@ def add_task(
         self.tasks.append(task)
 
     async def __call__(self) -> None:
-        for task in self.tasks:
-            await task()
+        async with anyio.create_task_group():
+            with anyio.CancelScope(shield=True):
```
adriangb, line 45: Is this task group needed? I think just the cancel scope should be enough.

starlette/background.py
Kludex: Which cases?

adriangb: Business logic. Like if you're doing something that is expensive and you don't care if it succeeds or fails if the client closes the connection. No _specific_ use case in mind, just thinking in general.

Although now I realize this is even more complicated because the behavior changes depending on if there is a `BaseHTTPMiddleware` involved, right? What happens if there isn't a `BaseHTTPMiddleaware` involved? Does it depend on the ASGI server?

Kludex: > fails if the client closes the connection

bkg tasks runs after the response 🤔

adriangb: Sorry I was maybe getting a bit confused as to what actually causes the issue.

Kludex:
This is not what we want. 😞

Trying to fix issues of `BaseHTTPMiddleware` outside of it doesn't look good.

PR #1655, opened 2022-05-31
Improve documentation on `BaseHTTPMiddleware`

Kludex:
The statement "It's not possible to use multiple `BaseHTTPMiddleware` based middlewares." is not true. I've removed that.

I've also added a reference for the `BackgroundTasks` issue, and a hyperlink to `ContextVar`.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -229,9 +229,8 @@ around explicitly, rather than mutating the middleware instance.
 !!! bug
     Currently, the `BaseHTTPMiddleware` has some known issues:
 
-    - It's not possible to use multiple `BaseHTTPMiddleware` based middlewares.
-    - It's not possible to use `BackgroundTasks` with `BaseHTTPMiddleware`.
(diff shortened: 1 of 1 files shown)


adriangb (approved):

erewok:
I still think we need an additional comment that states that using `BaseHTTPMiddleware` will evaluate responses entirely in memory, so using it with StreamingResponses or FileResponses is dangerous.

tiangolo (approved):

Kludex:
> I still think we need an additional comment that states that using `BaseHTTPMiddleware` will evaluate responses entirely in memory, so using it with StreamingResponses or FileResponses is dangerous.

I've looked a bit into https://github.com/encode/starlette/issues/919, to try to find out where this statement was first raised... I couldn't find how to reproduce it.

If I run the following code (shouldn't matter if it's Starlette or FastAPI):
```python
import asyncio

from fastapi import FastAPI
from fastapi.responses import StreamingResponse


app = FastAPI()


@app.middleware("http")
async def middleware(request, call_next):


@app.get("/")
async def home():
    async def stream():
        while True:
            yield "haha"
            await asyncio.sleep(1)

    return StreamingResponse(stream())
```
And then run `uvicorn`:
```bash
curl --raw http://localhost:8000/
```
I'm able to see:
```bash
(tmp5) ➜  Development curl --raw http://localhost:8000/
4
haha
4
haha
...
```

Am I missing something?

Also, I'm going to merge this, but I'll happily add that note as well.

PR #1656, opened 2022-05-31
Document how to create ASGI middlewares

### LINKED ISSUE #1029: Improve Documentation on Writing Custom ASGI Middleware and BaseHTTPMiddleware

Kludex:
- Closes https://github.com/encode/starlette/issues/1029

~This PR is incomplete. A text with some explanations on how to do this is missing.~

~I'm not sure if I need to provide an explanation or just link to those articles... Think about it.~

image


florimondmanca:
When it comes to writing middleware, I think one thing that those two articles (appreciate linking one of mine btw! :blush:) don't mention or dive into is the pattern of **wrapping receive/send** to perform custom behavior on the request/response. E.g. this is how [CORSMiddleware](https://github.com/encode/starlette/blob/master/starlette/middleware/cors.py) is implemented: wraps `send()` to add some response headers.

Another thing is a insisting on _not storing per-request state on middleware_ -- middleware classes should be "**stateless**". Otherwise that state would conflict during concurrent requests. One pattern that solves this is a "Responder" helper ASGI app (not the only pattern -- could also do with a bunch of closures in `__call__`). This is how [GZipMiddleware](https://github.com/encode/starlette/blob/master/starlette/middleware/gzip.py) works: wraps `send()` in a `GZipResponder` that buffers the initial `http.response.start` until it knows if it can send gzip, as well as store some other gzip info. That's a pattern I've also used for [msgpack-asgi](https://github.com/florimondmanca/msgpack-asgi) (which combines that with extensive receive/send wrapping).

So, in terms of structure, we could do:

```markdown
## Pure ASGI middleware

<!-- Introduction that mentions the benefits of working at the ASGI level?
Eg. ensuring compatibility with any framework... -->

### Guiding principles

<!-- Maybe some general ideas when writing ASGI middleware? -->
- Accept a parent `app` as first argument
- Implement `__call__(scope, receive, send)` that should call into parent `app` to run the underlying app or middleware.
- Handle all three `scope` types: `lifespan`, `http`, `websocket`
- Middleware classes should be stateless -- see [Per-request state](#per-request-state) if you do need to store per-request state.
- You may use Starlette datastructures (`Headers`, `URL`, etc) to ease dealing with raw ASGI components.

### Wrapping the request

<!-- Explains modifying the scope (for request headers, URLs, etc)
or wrapping receive() (for request body) -->

### Wrapping the response

<!-- Explains wrapping send() (response headers or body) -->

### Per-request state

<!-- Explains the Responder pattern -->

### Storing context in `scope`

<!-- Something about storing stuff in `scope` for access in endpoints?
See AuthMiddleware, SessionMiddleware... -->

### Examples

<!-- Perhaps a list of 'best practice' implementations?
Or a full complete example that does all the above? -->
```

Kludex:
I still need to polish it, but reviews are already welcome.

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+!!! note
+    The function pattern is not commonly spread, but you can check a more advanced implementation of it on
+    [asgi-cors](https://github.com/simonw/asgi-cors/blob/10ef64bfcc6cd8d16f3014077f20a0fb8544ec39/asgi_cors.py).
+
+#### `Scope` types
+
+As we mention, the scope holds the information about the connection. There are three types of `scope`s:
+- `lifespan` is a special type of scope that is used for the lifespan of the ASGI application.
+- `http` is a type of scope that is used for HTTP requests.
+- `websocket` is a type of scope that is used for WebSocket connections.
```
florimondmanca: Do we want to link to the ASGI docs pages as a handy reference here?

```suggestion
- [`lifespan`](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope) is a special type of scope that is used for the lifespan of the ASGI application.
- [`http`](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) is a type of scope that is used for HTTP requests.
- [`websocket`](https://asgi.readthedocs.io/en/latest/specs/www.html#websocket-connection-scope) is a type of scope that is used for WebSocket connections.
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            # Do something here!
+        await self.app(scope, receive, send)
+```
+In the example above, if the `scope` type is `lifespan` or `websocket`, we'll directly call the `self.app`.
+
+The same applies for the other scopes.
+
+You can read more about those scopes on the [ASGI documentation]().
```
florimondmanca: Link missing, but would be made obsolete if we directly link to docs on each scope:

```suggestion
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+```python
+class ASGIMiddleware:
+    def __init__(self, app):
+        self.app = app
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            # Do something here!
+        await self.app(scope, receive, send)
```
florimondmanca: We might prefer this "exit early" style? (Inspired by [guard clauses](https://medium.com/lemon-code/guard-clauses-3bc0cd96a2d3))

Usually this allows for cleaner code when then defining receive/send wrappers that might be fairly long...

```suggestion
    async def __call__(self, scope, receive, send):
        if scope["type"] != "http":
            await self.app(scope, receive, send)
            return

        # Do something here!
        await self.app(scope, receive, send)
```

I noticed in Starlette we tend to mix both, but as soon as the main branch (case when the middleware actually does something) gets long, we switch to this style. Perhaps better to encourage it right away?

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+        status_code = 500
+        if scope["type"] == "http":
+                async def send_wrapper(message):
+                    if message["type"] == "http.response.start":
+                        status_code = message["status"]
+                    await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
+        print("This is a primitive access log")
+        print(f"status = {status_code}")
```
florimondmanca: If we reorganize the code using the "exit early" style above, we'd end up with this:

```suggestion
        if scope["type"] != "http":
            await self.app(scope, receive, send)
            return
    
        status_code = 500

        async def send_wrapper(message):
            if message["type"] == "http.response.start":
                status_code = message["status"]
            await send(message)

        await self.app(scope, receive, send_wrapper)

        print("This is a primitive access log")
        print(f"status = {status_code}")
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+!!! note
+    Middleware classes should be stateless -- see [Per-request state](#per-request-state) if you do need to store per-request state.
+
+#### Wrapping `send` and `receive`
+
+A common pattern, that you'll probably need to use is to wrap the `send` or `receive` callables.
+
+```python
+class ASGIMiddleware:
```
florimondmanca: Maybe we could call this a descriptive name? "Here's how you'd build a middleware that logs the response status code..."

```suggestion
class LogStatusCodeMiddleware:
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+                async def send_wrapper(message):
+                    if message["type"] == "http.response.start":
+                        status_code = message["status"]
+                    await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
+        print("This is a primitive access log")
+        print(f"status = {status_code}")
+```
+On the example above, we are wrapping the `send` function with the `send_wrapper` one.
```
florimondmanca: I often find it's clearer if we _introduce_ example code rather than describe if after the fact?

We might combine this with a short description of "here's how we'd build a middleware that does X":

> For example, here's how we could write a middleware that logs the response status code, which we'd obtain by wrapping `send`...

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+!!! info
+    You can check a more advanced implementation of the same rationale on [asgi-logger](https://github.com/Kludex/asgi-logger/blob/main/asgi_logger/middleware.py).
+
+#### Annotation
```
florimondmanca:
```suggestion
#### Type hinting
```

(Or "type annotations", but at first it was not clear to me this section would be about)

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+            async def send_wrapper(message: ASGISendCallable) -> None:
+                await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
+```
+
+The `ASGI3Application` is meant to represent an ASGI application that follows the third version of the standard.
+Starlette itself is an ASGI 3 application.
+
+### Wrapping the request
```
florimondmanca: It seems this section would be better described like this... ?

```suggestion
### Reusing Starlette components
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+On Starlette, we have data structures that make it more convenient to organize your ASGI middleware.
```
florimondmanca: I think this is a really good point to make -- Starlette ASGI components make it a joy to write ASGI cross-compatible middleware without reinventing the ASGI wheel each time. Suggesting a bit of editing:

```suggestion
If you need to work with request or response data, you may find it more convenient to reuse Starlette datastructures (`Request`, `Headers`, `QueryParams`, `URL`, etc) rather than work with raw ASGI data. All these components can be built from the ASGI `scope`, `receive` and `send`, allowing you to work on pure ASGI middleware at a higher level of abstraction.
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+        if scope["type"] == "http":
+            request = Request(scope, receive, send)
+            # Do something here!
+        await self.app(scope, receive, send)
+```
+
+Also, it's possible to make use of `MutableHeaders` to change the headers:
+
+```python
+class ASGIMiddleware:
```
florimondmanca: Would this be a descriptive name?

```suggestion
class ExtraResponseHeadersMiddleware:
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            async def send_wrapper(message):
+                if message["type"] == "http.response.start":
+                    headers = MutableHeaders(scope=message)
+                    for key, value for self.headers:
+                        headers.append(key, value)
+                await send(message)
+        await self.app(scope, receive, send_wrapper)
```
florimondmanca: I we accept some of the previous suggestions, then for consistency this might change to this:

```suggestion
        if scope["type"] != "http":
            await self.app(scope, receive, send)

        async def wrapped_send(message):
            if message["type"] == "http.response.start":
                headers = MutableHeaders(scope=message)
                for key, value for self.headers:
                    headers.append(key, value)
            await send(message)

        await self.app(scope, receive, wrapped_send)
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+```
+
+### Per-request state
+
+The ASGI middleware class is meant to be stateless, as we don't want to spread the state from one request to another.
```
florimondmanca: (Nit) Taking inspiration from "MUST / SHOULD /etc" wording in formal specs, as this is really a key requirement to prevent a common "gotcha"

```suggestion
ASGI middleware classes should be stateless, as we typically don't want to spread the state from one request to another.
```

(Using 'typically' as I'm sure there may be valid, more exotic cases where we e.g. accumulate some state on each request... E.g. a request ID, or summary statistics, etc?)

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+        await self.app(scope, receive, send)
+
+
+class Responder:
+    def __init__(self, app):
+        self.app = app
+        self.started = False
+
+    def __call__(self, scope, receive, send):
```
florimondmanca:
```suggestion
    async def __call__(self, scope, receive, send):
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+        self.app = app
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            request = Request(scope, receive, send)
+            # Do something here!
+        await self.app(scope, receive, send)
+```
+
+Also, it's possible to make use of `MutableHeaders` to change the headers:
```
florimondmanca: Reformulating to suggest that these are just examples and folks can actually be pretty creative about what parts they reuse...

```suggestion
Or we might use `MutableHeaders` to change the response headers:
```

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            responder = Responder(self.app)
+            return await responder(scope, receive, send)
+
+        await self.app(scope, receive, send)
+
+
+class Responder:
```
florimondmanca: To be fair, this section currently feels somewhat artifical because in the closure style (`def wrapped_send() ...`), request state is stored directly in the `__call__` scope (eg `status_code = 500` in one of the examples above), so there shouldn't be much risk of storing state on middleware attributes directly.

The thing is, for more advanced middleware the closure style might become a bit out of hand, and people may want to use proper methods instead. Like this...

```python
class ASGIMiddleware:
    def __init__(self, app, headers):
        self.app = app
        self.headers

    async def _send_with_headers(self, message, send):
            ...

        await send(message)

        if scope["type"] != "http":
            return

        send = functools.partial(self._send_with_headers, send=send)
```

But then _if_ you need to store state, you might be tempted to put that on `self`, which is when we start leaking state if we're not careful.

A very good example of this is `GZipMiddleware` which needs to keep a reference to the `initial_message` to decide if it should encode the response. It could very well be implemented as a single long `__call__()` function on the middleware, just turns out not to be (for readability? The middleware behavior is fairly complex).

So... Should we more explicitly write up about this "partials" style, and highlight this "middleware state potential gotcha" _in that particular context_?

This might turn up a bit long but maybe something like this...

```markdown
### Per-request state

ASGI middleware classes should be stateless, as we typically don't want to leak state across requests.

The risk is low when defining wrappers inside `__call__`, as state would typically be defined as inline variables.

But if the middleware grows larger and more complex, you might be tempted to refactor wrappers as methods. Still, state should not be stored in the middleware instance. Instead, if you need to manipulate per-request state, you may write a separate `Responder` class:

```python
from functools import partial

class TweakMiddleware:
    """
    Make a change to the response body if 'X-Tweak' is
    present in the reponse headers.
    """

    async def __call_(self, scope, receive, send):
        ...
        responder = MaybeTweakResponder(self.app)
        await responder(scope, receive, send)

class TweakResponder:
    def __init__(self, app):
        self.app = app
        self.should_tweak = False

        send = partial(self.maybe_send_with_tweaks, send=send)

    async def maybe_send_with_tweaks(self, message, send):
            headers = Headers(raw=message["headers"])
            self.should_tweak = headers.get("X-Tweak") == "1"
            await send(message)
            return

        if message["type"] == "http.response.body":
            if not self.should_tweak:
                await send(message)
                return

            # Actually tweak the response body...
```

See also [`GZipMiddleware`](https://github.com/encode/starlette/blob/9ef1b91c9c043197da6c3f38aa153fd874b95527/starlette/middleware/gzip.py) for a full example of this pattern.
```

File: docs/middleware.md [hunk 16]
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+The `Responder` object, in this case, will save the information about when the response body has "started" to be sent.
+
+!!! info
+    You can see a more advanced implementation of this pattern on our [GZipMiddleware](), and also on the [msgpack-asgi](https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py).
+
+### Storing context in `scope`
+
+As we know by now, the `scope` holds the information about the application. To be precise, the `scope` holds
+the stateless information of the application. IS THIS CORRECT?
```
florimondmanca: Rather, the `scope` stores information about the current request. It's got information such as "headers", "path", "query", etc.

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+
+As per the ASGI specifications, any application can store custom information on the `scope`.
+To be precise, it should be stored under the `extensions` key.
```
florimondmanca: > As per the ASGI specifications, any application can store custom information on the `scope`.
> To be precise, it should be stored under the `extensions` key.

Not really sure about this. As per https://github.com/encode/starlette/discussions/1511 it seems like storing non-ASGI state in `scope` is not really spec'd yet.

That's also different from [ASGI extensions](https://asgi.readthedocs.io/en/latest/extensions.html) which are meant to convey information from and to the ASGI server (eg. Uvicorn).

So the current status seems to be:

* Middleware may conveniently add items to `scope` to make data available to endpoints.
* Make sure to properly namespace to avoid conflicts.
* This is mostly a convention -- not standardized yet.

So, _maybe_ the example could be more like this:

```python
scope["my-special-feature"] = {"enabled": True}
```

But it's not even clear if we _should_ do that _or_ edit `extensions` (which we'd need to `.setdefault()` since it might already be set due to other middleware...?). I think it's a safer, "sure not to interfere with other middleware" option.

docs/middleware.md
```diff
@@ -232,6 +232,252 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope, receive, send):
+        await self.app(scope, receive, send)
+```
+
+The middleware above is the most basic ASGI middleware. It just receives an ASGI application as an argument for its constructor, and implements the `__call__` method, with the parameters that the ASGI protocol defines: `scope`, `receive`, and `send`.
+
+The `scope` parameter is a dictionary that contains the information about the connection.
+You can see more about it [here](https://asgi.readthedocs.io/en/latest/specs/index.html).
+
+Both `receive` and `send` parameters are async callables that work in opposite ways. The `send` callable allows the application to send event messages to the client, and the `receive` allows the application to receive event messages from the client.
```
florimondmanca: Maybe we should be a bit more concise on presenting scope/receive/send, and defer to the ASGI spec? I'm afraid any longer-form explanation might end up counter-productive as the ASGI spec has precise definitions.

```suggestion
The middleware above is the most basic ASGI middleware. It receives an ASGI application as an argument for its constructor, and implements the `async __call__` method. This method should accept `scope`, which contains information about the current connection, and `receive` and `send` which allow to exchange ASGI event messages with the ASGI server (learn more in the [ASGI specification](https://asgi.readthedocs.io/en/latest/specs/index.html)).
```

florimondmanca (commented):
This is _great_ :) I love that we'll have a guide on writing pure ASGI middleware, as this style enables more creative / detailed functionality as well as interoperability across ASGI frameworks.

One general thought -- how does this look in the docs? It seems to get pretty long. For good reason, since ASGI middleware is a fairly rich topic in terms of techniques to peek into ASGI data or wrap behavior... But since this topic is by definition not Starlette-specific (apart from the section mentioning reusing Starlette components), it does feel a bit odd having such a large section in a page that otherwise focuses on Starlette-specific middleware matters.

Not sure how we'd resolve this, though...

docs/middleware.md
Kludex: Sure! I thought it would be easier to explain with the other way, but I guess we should teach what we all do...
```suggestion
    async def __call__(self, scope, receive, send):
        if scope["type"] != "http":
            return await self.app(scope, receive, send)
            
        # Do something here!
        await self.app(scope, receive, send)
```

docs/middleware.md
Kludex:
```suggestion
        if scope["type"] != "http":
            return await self.app(scope, receive, send)
    
        status_code = 500

        async def send_wrapper(message):
            if message["type"] == "http.response.start":
                status_code = message["status"]
            await send(message)

        await self.app(scope, receive, send_wrapper)

        print("This is a primitive access log")
        print(f"status = {status_code}")
```

docs/middleware.md
Kludex: I'll apply this later. Writing it here so I don't forget.

docs/middleware.md
Kludex:
```suggestion
#### Type annotations
```

docs/middleware.md
Kludex: Wonderful!

docs/middleware.md
Kludex:
```suggestion
        if scope["type"] != "http":
            return await self.app(scope, receive, send)

        async def wrapped_send(message):
            if message["type"] == "http.response.start":
                headers = MutableHeaders(scope=message)
                for key, value for self.headers:
                    headers.append(key, value)
            await send(message)

        await self.app(scope, receive, wrapped_send)
```

docs/middleware.md
Kludex: I'll adapt. Wring this so I don't forget.

docs/middleware.md
Kludex: "IS THIS CORRECT" was more about the wording. I'll reformulate the sentence.

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope, receive, send):
+        await self.app(scope, receive, send)
+```
+
+The middleware above is the most basic ASGI middleware. It receives an ASGI application as an argument for its constructor, and implements the `async __call__` method. This method should accept `scope`, which contains information about the current connection, and `receive` and `send` which allow to exchange ASGI event messages with the ASGI server (learn more in the [ASGI specification](https://asgi.readthedocs.io/en/latest/specs/index.html)).
+
+As an alternative for the class approach, you can also use a function:
+
+```python
+def asgi_middleware():
```
tiangolo, line 262: Maybe just to be more explicit and make the example snippet complete.

```suggestion
import functools

def asgi_middleware():
```

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+    return asgi_decorator
+```
+
+!!! note
+    The function pattern is not commonly spread, but you can check a more advanced implementation of it on
+    [asgi-cors](https://github.com/simonw/asgi-cors/blob/10ef64bfcc6cd8d16f3014077f20a0fb8544ec39/asgi_cors.py).
+
+#### `Scope` types
+
+As we mention, the scope holds the information about the connection. There are three types of `scope`s:
```
tiangolo: Nitpick typo:

```suggestion
As we mentioned, the scope holds the information about the connection. There are three types of `scope`s:
```

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+
+        print("This is a primitive access log")
+        print(f"status = {status_code}")
+```
+On the example above, we are wrapping the `send` function with the `send_wrapper` one.
+
+!!! info
+    You can check a more advanced implementation of the same rationale on [asgi-logger](https://github.com/Kludex/asgi-logger/blob/main/asgi_logger/middleware.py).
+
+#### Type annotations
```
tiangolo, line 336: Amazing having docs to add proper type annotations! 🤩 🎉

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+```
+
+!!! note
+    The function pattern is not commonly spread, but you can check a more advanced implementation of it on
+    [asgi-cors](https://github.com/simonw/asgi-cors/blob/10ef64bfcc6cd8d16f3014077f20a0fb8544ec39/asgi_cors.py).
+
+#### `Scope` types
+
+As we mention, the scope holds the information about the connection. There are three types of `scope`s:
+- [`lifespan`](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope) is a special type of scope that is used for the lifespan of the ASGI application.
```
tiangolo, line 279: Just an extra new line to fix the list formatting:

```suggestion

- [`lifespan`](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope) is a special type of scope that is used for the lifespan of the ASGI application.
```

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+            if message["type"] == "http.response.body" and not self.started:
+                self.started = True
+            await send(message)
+        await self.app(scope, receive, send_wrapper)
+```
+
+The `Responder` object, in this case, will save the information about when the response body has "started" to be sent.
+
+!!! info
+    You can see a more advanced implementation of this pattern on our [GZipMiddleware](), and also on the [msgpack-asgi](https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py).
```
tiangolo: The current permalink to the `GzipMiddleware`:

```suggestion
    You can see a more advanced implementation of this pattern on our [GZipMiddleware](https://github.com/encode/starlette/blob/b588ebed36c2934b33ad3b0221d79bd8b7a4817f/starlette/middleware/gzip.py), and also on the [msgpack-asgi](https://github.com/florimondmanca/msgpack-asgi/blob/6261b1e22b3689551f68038ee00adda5fbc04670/src/msgpack_asgi/_middleware.py).
```

docs/middleware.md
tiangolo: I think it would be "information about the connection" instead of about the application.

Maybe it could help to add an example in the explanation, as Florimond says, in the case of HTTP it includes information about one request, like the headers, path, etc. In the case of WebSockets, it also includes headers, paths, etc, but it's not a single request, instead, it's a WebSocket connection that can have multiple messages received and sent.

tiangolo (commented):
This is looking great! Thanks for doing this. 🤩 

Just a couple of small suggestions.

docs/middleware.md
```diff
@@ -232,6 +232,255 @@ around explicitly, rather than mutating the middleware instance.
...
+            return await self.app(scope, receive, send)
+            
+        # Do something here!
+        await self.app(scope, receive, send)
+```
+In the example above, if the `scope` type is `lifespan` or `websocket`, we'll directly call the `self.app`.
+
+The same applies for the other scopes.
+
+
```
Kludex, line 300:
```suggestion
```

docs/middleware.md
Kludex: > Not really sure about this. As per https://github.com/encode/starlette/discussions/1511 it seems like storing non-ASGI state in scope is not really spec'd yet.

Right. 🤔 

Thanks for this comment. I want to have that clarified by the ASGI specs as well. I've created an issue there, and cc you.

Kludex (commented):
I've updated the PR.

- I've removed the "Examples" section, but I'd like your opinion about this. Should I link code sources around that implement different styles of middlewares? Or should I create examples based on the `scope` types?
- I still didn't finish the "Storing context in `scope`" section. I've created an issue on `asgiref` to have clarifications: https://github.com/django/asgiref/issues/332
- I've applied all the comments.

Another round of reviews is welcome. 🙏

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
     - It's not possible to use `BackgroundTasks` with `BaseHTTPMiddleware`. Check [#1438](https://github.com/encode/starlette/issues/1438) for more details.
     - Using `BaseHTTPMiddleware` will prevent changes to [`contextlib.ContextVar`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar)s from propagating upwards. That is, if you set a value for a `ContextVar` in your endpoint and try to read it from a middleware you will find that the value is not the same value you set in your endpoint (see [this test](https://github.com/encode/starlette/blob/621abc747a6604825190b93467918a0ec6456a24/tests/middleware/test_base.py#L192-L223) for an example of this behavior).
 
+## Pure ASGI Middleware
+
+Due to how ASGI was designed, we are able to build a chain of ASGI applications, on which each application calls the next one.
```
florimondmanca: Small reformulation:

```suggestion
Thanks to how ASGI was designed, it is possible to implement middleware as a chain of ASGI applications, where each application calls into the next one.
```

florimondmanca (commented):
Looks like this documentation is getting narrowed down! Nice work

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        # Do something here!
+        await self.app(scope, receive, send)
+```
+In the example above, if the `scope` type is `lifespan` or `websocket`, we'll directly call the `self.app`.
+
+The same applies for the other scopes.
```
florimondmanca:
```suggestion
The same applies to other scopes.
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+from asgiref.typing import ASGIReceiveEvent, ASGISendEvent
+
+
+class ASGIMiddleware:
+    def __init__(self, app: ASGI3Application) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable) -> None:
+        if scope["type"] == "http":
+            async def send_wrapper(message: ASGISendCallable) -> None:
```
florimondmanca: Shouldn't this be... ?

```suggestion
            async def send_wrapper(message: ASGISendEvent) -> None:
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+class ASGIMiddleware:
+    def __init__(self, app: Starlette) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] == "http":
+            async def send_wrapper(message: Message) -> None:
+                await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
```
florimondmanca: I'd argue it is somewhat odd that we present "here's how to do HTTP-only" with the guard-style, yet we use the 

I know we've discussed this before, but my point here is I think we should be consistent for the entire length of this documentation.

So:

```suggestion
        if scope["type"] != "http":
            await self.app(scope, receive, send)
            return

        async def send_wrapper(message: Message) -> None:
            # ... Do something
            await send(message)

        await self.app(scope, receive, send_wrapper)
```

florimondmanca (approved):
This is looking good overall; that's a nice documentation addition that I think folks will be able to link to from various places.

Leaving my approval for when we think this is ready.

docs/middleware.md
adriangb: agreed

docs/middleware.md
adriangb: +1. The `!=` style also results in less indented code

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
     - It's not possible to use `BackgroundTasks` with `BaseHTTPMiddleware`. Check [#1438](https://github.com/encode/starlette/issues/1438) for more details.
     - Using `BaseHTTPMiddleware` will prevent changes to [`contextlib.ContextVar`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar)s from propagating upwards. That is, if you set a value for a `ContextVar` in your endpoint and try to read it from a middleware you will find that the value is not the same value you set in your endpoint (see [this test](https://github.com/encode/starlette/blob/621abc747a6604825190b93467918a0ec6456a24/tests/middleware/test_base.py#L192-L223) for an example of this behavior).
 
+## Pure ASGI Middleware
+
+Thanks to how ASGI was designed, it is possible to implement middleware as a chain of ASGI applications, where each application calls into the next one.
+Each element of the chain is an [`ASGI`](https://asgi.readthedocs.io/en/latest/) application by itself, which per definition, is also a middleware.
+
+This is also an alternative approach in case the limitations of `BaseHTTPMiddleware` are a problem.
```
adriangb, line 240:
```suggestion
This is also an alternative approach in case the limitations of `BaseHTTPMiddleware` are a problem or if you want to modify the request or response body.
```

I think we should encourage users to use pure ASGI middleware if they want to modify the request or response body. I think for BaseHTTPMiddleware you can replace the response body entirely, but you can't wrap the request or response body.

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable) -> None:
+        if scope["type"] == "http":
+            async def send_wrapper(message: ASGISendCallable) -> None:
+                await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
+```
+
+The `ASGI3Application` is meant to represent an ASGI application that follows the third version of the standard.
+Starlette itself is an ASGI 3 application.
```
adriangb:
```suggestion
The `ASGI3Application` is meant to represent an ASGI application that follows the third version of the standard.
Starlette itself is an ASGI 3 application. You probably want to use `ASIG3Application`.
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        async def wrapped_send(message):
+            if message["type"] == "http.response.start":
+                headers = MutableHeaders(scope=message)
+                for key, value for self.headers:
+                    headers.append(key, value)
```
adriangb:
```suggestion
                headers = MutableHeaders(scope=message)
                for key, value for self.headers:
                    headers.append(key, value)
                message["headers"] = headers.raw  # ASGI headers is a List[Tuple[bytes, bytes]]
```

I think this more explicit assignment is clear than relying on `MutableHeaders` to modify `message` in-place.

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+        await self.app(scope, receive, wrapped_send)
+```
+
+### Per-request state
+
+ASGI middleware classes should be stateless, as we typically don't want to leak state across requests.
+
+The risk is low when defining wrappers inside `__call__`, as state would typically be defined as inline variables.
+
+But if the middleware grows larger and more complex, you might be tempted to refactor wrappers as methods. Still, state should not be stored in the middleware instance. Instead, if you need to manipulate per-request state, you may write a separate `Responder` class:
```
adriangb: Another option which I personally like is using a generator:

```python
async def tweaked_send() -> AsyncGenerator[None, Message]:
    message = yield
    headers = Headers(raw=message["headers"])
    should_tweak = headers.get("X-Tweak") == "1"
    await send(message)

    while True:
        message = yield
        if not should_tweak:
            await send(message)
        else:
            raise NotImplementedError  # TODO: tweak

async with aclosing(tweaked_send()) as wrapped_send:
    await wrapped_send.__anext__()  # start the generator
    await self.app(scope, receive, wrapped_send.asend)
```

Maybe it's worth mentioning?

docs/middleware.md
adriangb: Yep not sure what to do with this, we could remove it for now, wait until https://github.com/django/asgiref/issues/332 reaches some consensus or just make our own recommendation here?

adriangb (approved):
Overall this looks good. I only have minor nits which are not blockers and can be addressed later. Leaving an approval.

docs/middleware.md
florimondmanca: That’s wild!

Also, hmm… I’m less convinced about documenting the « responder class » now, although I’m aware I’m the one who brought it up initially. That’s because that’s a style I’ve been using. I think I was itches by the fact we’d be redefining a function every time with the closure style.

There’s a couple of styles that could be used, eg. responder class, or this generator style, etc. I think this section is complex and new enough for most readers already that we should stick to just one.

I’m curious what you would think about dropping this section. We’ve got a note somewhere on the fact that middleware classes should be stateless. The closure style actually doesn’t incite users to use class attributes, so it doesn’t pose a « per request state » problem, so we could drop this entire section.

docs/middleware.md
florimondmanca: Actually ASGI 2 is deprecated / obsolete now (eg Uvicorn doesn’t support it anymore), so users _definitely_ must use ASGI3Application.

I’m actually not sure we should provide details about this at all. Import the ASGI3Application thing from asgiref, is all?

Kludex: Uvicorn does supports it 🤔 
 https://github.com/encode/uvicorn/blob/d2b14eeed32ac793e1588c98aa8a283e8771a817/uvicorn/config.py#L42

Kludex: I'm not against deprecating it, tho... 👀

adriangb: Edit: I am _for_ removing ASGI2, not against. Late night typo.

I would be **for** removing ASGI2 support from Uvicorn and Starlette (I think it still exists in TestClient?) but that should be another topic/discussion.

As far as these docs, I am in agreement with Florimond: we should just not mention ASGI2/ASGI3 and instead only refer to ASGI3 implicitly as "ASGI".

docs/middleware.md
adriangb: > That’s wild!

It is a neat pattern huh! Generators are great little state machines that you can ratchet forwards and pass messages to. And of course you can go full functional:

```python
def functional_middleware(app: ASGIApp) -> ASGIApp:
    async def new_app(scope: Scope, receive: Receive, send: Send) -> None:
        async def tweaked_send() -> "AsyncGenerator[None, Message]":
            message = yield
            headers = Headers(raw=message["headers"])
            should_tweak = headers.get("X-Tweak") == "1"
            await send(message)

            while True:
                message = yield
                if not should_tweak:
                    await send(message)
                else:
                    await send(message)

        async with aclosing(tweaked_send()) as wrapped_send:
            await wrapped_send.__anext__()  # start the generator
            await app(scope, receive, wrapped_send.asend)

    return new_app
```

> I’m curious what you would think about dropping this section.

I do think this section is getting a bit in depth, maybe at this point we should let users fly and use their imagination and dev skills to figure out which solution they want to use? We'll still need to explain _why_ they should not use attributes on the middleware. And maybe the best way to do that is with a counterexample, and if I had to pick one style I'd pick the responder class because I think it'll be easier to understand.

docs/middleware.md
Kludex:
```suggestion
        if scope["type"] != "http":
            return await self.app(scope, receive, send)

        async def send_wrapper(message: Message) -> None:
            # ... Do something
            await send(message)

        await self.app(scope, receive, send_wrapper)
```

docs/middleware.md
euri10: I had to re read this section several times, I think it would be extremely clear to have a `don't do this example` and another one following, `do this instead`

docs/middleware.md
```diff
@@ -232,6 +232,274 @@ around explicitly, rather than mutating the middleware instance.
...
+Using Starlette, you can do as:
+
+```python
+from starlette.types import Message, Scope, Receive, Send
+from starlette.applications import Starlette
+
+
+class ASGIMiddleware:
+    def __init__(self, app: Starlette) -> None:
+        self.app = app
```
jhominal: Shouldn't we use `starlette.types.ASGIApp` here instead of `starlette.applications.Starlette`?
```suggestion
from starlette.types import ASGIApp, Message, Scope, Receive, Send


class ASGIMiddleware:
    def __init__(self, app: ASGIApp) -> None:
        self.app = app
```

adriangb: agreed

Kludex: Sure. Thanks 🙏

docs/middleware.md
Kludex: Can't we add this as a limitation or a note on the `BaseHTTPMiddleware` section?

adriangb: I think that list is to list foot guns, not features that were never implemented / in scope.

docs/middleware.md
```diff
@@ -232,6 +232,273 @@ around explicitly, rather than mutating the middleware instance.
...
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        # Do something here!
+        await self.app(scope, receive, send)
+```
+In the example above, if the `scope` type is `lifespan` or `websocket`, we'll directly call the `self.app`.
```
```suggestion
In the example above, if the `scope` type is **not** `http`, meaning that is either `lifespan` or `websocket`, we'll directly call the `self.app`.
```

docs/middleware.md
Kludex: Actually, I don't think it's a huge effort to maintain the support for ASGI2 on `uvicorn`.

I'm going to go with removing this section, and improve the note below.

docs/middleware.md
```diff
@@ -232,6 +232,273 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope: Scope, receive: ASGIReceiveCallable, send: ASGISendCallable) -> None:
+        if scope["type"] == "http":
+            async def send_wrapper(message: ASGISendEvent) -> None:
+                await send(message)
+            return await self.app(scope, receive, send_wrapper)
+        await self.app(scope, receive, send)
+```
+
+The `ASGI3Application` is meant to represent an ASGI application that follows the third version of the standard.
+Starlette itself is an ASGI 3 application.
```
```suggestion
```

docs/middleware.md
```diff
@@ -232,6 +232,273 @@ around explicitly, rather than mutating the middleware instance.
...
+
+!!! note
```
```suggestion
!!! info
```

docs/middleware.md
```diff
@@ -232,6 +232,273 @@ around explicitly, rather than mutating the middleware instance.
...
+    You can read more about ASGI versions on the [Legacy Applications section on the ASGI documentation](https://asgi.readthedocs.io/en/latest/specs/main.html#legacy-applications).
```
```suggestion
    If you're curious about the `ASGI3Application` type on the snippet above, you can read more about ASGI versions on the [Legacy Applications section on the ASGI 
    documentation](https://asgi.readthedocs.io/en/latest/specs/main.html#legacy-applications).
```

docs/middleware.md
adriangb: Why would anyone want to write an ASGI2 middleware in 2022? I don't think there's a single server that _only_ supports ASGI2, so if you're writing a new middleware you should just write it as ASGI3. So no need to even mention ASGI2, we can just say `ASGI` and it means `ASGI3`. Right?

docs/middleware.md
Kludex: I'm going to simplify this section and apply @euri10 's comment. I may remove the snippets here, and point to external implementations...

docs/middleware.md
Kludex: I'm going to change this according to the asgiref issue mentioned above.

Kludex (commented):
I'm going to include an example on how to manipulate the request body on a middleware.

docs/middleware.md
adriangb: Maybe we can even point to our own internal implementations? GZipMiddleware would be a good example.

docs/middleware.md
Kludex: The implementations I've seen use it in the way I'm using it. I do agree with you on which one is clear, but I'd rather prefer the user to think a bit on a small documentation like this instead of another more complicated code source.

That being said, if you are strong about it, we can add a note about it below.

adriangb: I wouldn't say I feel _strongly_ about it, but it is 1 LOC

docs/middleware.md
Kludex: The most correct type is `ASGI3Application`. As per Uvicorn's ASGI2 support, it's out of scope.

docs/middleware.md
florimondmanca: Also there’s a typo here: for key, value _in_ self.headers

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+        self.headers = headers
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] != "http":
+            return await self.app(scope, receive, send)
+
+        async def wrapped_send(message):
+            if message["type"] == "http.response.start":
+                headers = MutableHeaders(scope=message)
+                for key, value for self.headers:
```
adriangb:
```suggestion
                for key, value in self.headers:
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+        await self.app(scope, receive, wrapped_send)
+```
+
+### Per-request state
+
+ASGI middleware classes should be stateless, as we typically don't want to leak state across requests.
+
+The risk is low when defining wrappers inside `__call__`, as state would typically be defined as inline variables.
+
+But if the middleware grows larger and more complex, you might be tempted to refactor wrappers as methods. Still, state should not be stored in the middleware instance. Instead, if you need to manipulate per-request state, you may write a separate `Responder` class:
```
tiangolo: I would make it explicit that "state in the middleware instance" means using attributes, just in case, for less experienced developers. But maybe that's not necessary.

```suggestion
But if the middleware grows larger and more complex, you might be tempted to refactor wrappers as methods. Still, state should not be stored in the middleware instance. This means that the middleware should not have attributes that would change across requests or connections. If the middleware has an attribute, for example set in the `__init__()` method, nothing else should change it afterewards. Instead, if you need to manipulate per-request state, you may write a separate `Responder` class:
```

docs/middleware.md
tiangolo: I agree on *not* using the `extensions` word here. I would expect extensions to be used for generic/generalizable things that could become a standard in some way, and as Florimond says, that interact with the server (Uvicorn) directly in some way.

Also, I would expect this guide to be a bit more directed to users creating a middleware for their own app, not a generic/generalizable thing that could become a standard. I would just remove the `extensions` word and say something around trying to avoid collisions in names.

```suggestion
As per the ASGI specifications, any application can store custom information on the `scope`.
Have in mind that other components could also store data in the same `scope`, so it's important to use a key that has a low chance of being used by other things.

For example, if you are building an application called `supper-app` you could have that as a prefix for any keys you put in the `scope`, and then you could have a key called `super-app-transaction-id`.
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+To be precise, it should be stored under the `extensions` key.
+
+```python
+class ASGIMiddleware:
+    def __init__(self, app):
+        self.app = app
+
+    async def __call__(self, scope, receive, send):
+        scope["extensions"] = {"super.extension": True}
+        await self.app(scope, receive, send)
```
tiangolo: Continuing with my example and idea above, I would update this example like this:

```suggestion
from uuid import uuid4


class ASGIMiddleware:
    def __init__(self, app):
        self.app = app

    async def __call__(self, scope, receive, send):
        scope["super-app-transaction-id"] = uuid4()
        await self.app(scope, receive, send)
```

docs/middleware.md
```diff
@@ -232,6 +232,271 @@ around explicitly, rather than mutating the middleware instance.
...
+```
+On the example above, we stored an extension called "super.extension". That can be used by the application itself, as the scope is forwarded to it.
```
tiangolo: Continuing with my example and idea above, I would update this text like this:

```suggestion
On the example above, we stored a key called "super-app-transaction-id" in the scope. That can be used by the application itself, as the scope is forwarded to it.
```

I had to put this on another comment because GitHub doesn't like the nested triple backticks. :sweat_smile:

tiangolo (commented):
I think this is great! :rocket: 

I just have a couple of minor comments.

docs/middleware.md
tiangolo: Hehe, I always struggle with generators. :sweat_smile: What I don't like about `yield` for receiving stuff is that there's no way to have type inference, any values obtained at the left side would have to be manually annotated to have typing support (autocompletion, inline errors, mypy, etc). I'm also not sure what better approach is there. But anyway...

I would leave this PR without more examples, to try and limit the scope. Maybe a future PR could include suggesting more examples, but I think it would be good to have this close to what is already there, without adding too much.

I would just add another phrase clarifying the stateless stuff. I'm adding that as a separate suggestion below. :nerd_face:

docs/middleware.md
```diff
@@ -232,6 +232,276 @@ around explicitly, rather than mutating the middleware instance.
...
+See also [`GZipMiddleware`](https://github.com/encode/starlette/blob/9ef1b91c9c043197da6c3f38aa153fd874b95527/starlette/middleware/gzip.py) for a full example of this pattern.
+
+### Storing context in `scope`
+
+As we know by now, the `scope` holds the information about the connection.
+
+As per the ASGI specifications, any application can store custom information on the `scope`.
+Have in mind that other components could also store data in the same `scope`, so it's important to use a key that has a low chance of being used by other things.
+
+For example, if you are building an application called `supper-app` you could have that as a prefix for any keys you put in the `scope`, and then you could have a key called `super-app-transaction-id`.
```
```suggestion
For example, if you are building an application called `super-app` you could have that as a prefix for any keys you put in the `scope`, and then you could have a key called `super-app-transaction-id`.
```

docs/middleware.md
```diff
@@ -232,6 +232,276 @@ around explicitly, rather than mutating the middleware instance.
...
+            return await self.app(scope, receive, send)
+
+        # Do something here!
+        await self.app(scope, receive, send)
+```
+In the example above, if the `scope` type is **not** `http`, meaning that is either `lifespan` or `websocket`, we'll directly call the `self.app`.
+
+The same applies to other scopes.
+
+!!! note
```
```suggestion
!!! tip
```

docs/middleware.md
```diff
@@ -232,6 +232,276 @@ around explicitly, rather than mutating the middleware instance.
...
+def asgi_middleware():
+    def asgi_decorator(app):
+        @functools.wraps(app)
+        async def wrapped_app(scope, receive, send):
+            await app(scope, receive, send)
+        return wrapped_app
+    return asgi_decorator
+```
+
+!!! note
```
```suggestion
!!! info
```

docs/middleware.md
```diff
@@ -232,6 +232,276 @@ around explicitly, rather than mutating the middleware instance.
...
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http":
+            request = Request(scope, receive, send)
+            # Do something here!
+        await self.app(scope, receive, send)
+```
+
+Or we might use `MutableHeaders` to change the response headers:
+
+```python
```
Kludex, line 405:
```suggestion
```python
from starlette.datastructures import MutableHeaders
```

Kludex (commented):
I've done my last self review.

I think documentation is not appreciated as it should, so thanks for all the comments and rounds of reviews. 🎉

I'd like to add some points for some decisions or things that I didn't apply that I think we can improve in the future:
- There are no examples of wrapping the `receive` method (some example modifying the request body would be cool).
- The "Per-request state" section is too complex. I think we could add an example of "what not to do" and then following with the "right way", as mentioned by @euri10. 
- I think not adding further patterns (e.g. generator way) on the "Per-request section" was the best approach. I think that pattern is neat, yep, but it's also hard to understand, and not everybody has the same knowledge, so we need to try to give the information in a way that everybody can understand by only reading this documentation (or reference external content that makes reading this possible). That said, I do think that it's worth adding a `note` to showcase some advanced patterns.

In any case, once again, thanks everybody. 🙏

tiangolo:
Thanks for all the work and patience with this @Kludex ! Great job! 😎🚀

And awesome to see all the additional feedback from everybody 😍🍰

adriangb:
This is awesome @Kludex ! Thank you everyone for your work on this. Agreed about the not including the generator stuff, too complicated for intro docs 😅

docs/middleware.md
Kludex: I think we can include not as bug then, but as "limitations" there 🤔

PR #1659, opened 2022-05-31
Fix `authentication.requires` type

sobolevn:
I've noticed that `authentication.requires` does not preserve type information about the passed function.

So, I've fixed that!

```python
from starlette.authentication import requires


@requires(['authenticated', 'admin'])
async def dashboard(request: str) -> int:  # Don't mind these fake types here
    ...

reveal_type(dashboard)
# Before:
# note: Revealed type is "Any"

# After:
# note: Revealed type is "def (request: builtins.str) -> typing.Coroutine[Any, Any, builtins.int]"
```


Kludex:
I like the enthusiasm! Thanks for the PR @sobolevn ! 🙇 

But... This is not the right annotation, that's why the `# type: ignore` was needed. Besides, I think that's something we need to discuss first, as it opens precendece to a lot of annotations in code.

For those reasons, I'll be closing this PR.

sobolevn:
> But... This is not the right annotation

It looks right to me 🤔 
Can you please point out what's wrong?

This is the common way of expressing "this function returns a decorator that preserves the signature".
I might be missing something 🙂 

> that's why the # type: ignore was needed

`type: ignore` is needed, because we cannot really proove to `mypy` that `_CallableType` is the same thing as `decorator`. Mypy cannot inference that by default. There are two ways:
- `cast()` with runtime effect (small, but still)
- `# type: ignore` which does not have this effect

> Besides, I think that's something we need to discuss first, as it opens precendece to a lot of annotations in code.

Sure thing, I will open an issue about that.
Update: https://github.com/encode/starlette/discussions/1662

adriangb:
@Kludex I do think this is the right annotation, unless we want to be more explicit and encode `Callable[[Request], Response]` in there to avoid a typevar.

adriangb (approved):

starlette/authentication.py
```diff
@@ -95,7 +97,7 @@ def sync_wrapper(*args: typing.Any, **kwargs: typing.Any) -> Response:
 
             return sync_wrapper
 
-    return decorator
+    return decorator  # type: ignore
```
Kludex: Can we add at least the error on `ignore[<error>]`? :pray:

adriangb:
```suggestion
    return decorator
```

to get the error from mypy

adriangb:
```suggestion
    return decorator  # type: ignore[return-value]  # type checkers can't verify the type of the wrapper
```

Kludex: i'm already happy with just the return-value thingy, no need the comment stuff

Kludex:
> > But... This is not the right annotation
> 
> It looks right to me thinking Can you please point out what's wrong?
> 
> This is the common way of expressing "this function returns a decorator that preserves the signature". I might be missing something slightly_smiling_face
> 
> > that's why the # type: ignore was needed
> 
> `type: ignore` is needed, because we cannot really proove to `mypy` that `_CallableType` is the same thing as `decorator`. Mypy cannot inference that by default. There are two ways:
> 
> * `cast()` with runtime effect (small, but still)
> * `# type: ignore` which does not have this effect
> 
> > Besides, I think that's something we need to discuss first, as it opens precendece to a lot of annotations in code.
> 
> Sure thing, I will open an issue about that. Update: #1662

Thanks for the explanation! I guess I got excited closing PRs that day. :cold_sweat: 

I retract my words from before, and sorry for closing the PR. :)

Kludex (approved):

adriangb:
Thank you for the contribution @sobolevn !

PR #1670, opened 2022-06-02
Adding __reduce__ to HTTPException to enable pickle/unpickle

nexusdk:
[HTTPException pickle fails when constructed using named parameter status_code](https://github.com/encode/starlette/discussions/1669)

### DIFF
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -22,7 +22,7 @@ class HTTPException(Exception):
         class_name = self.__class__.__name__
         return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
 
-    def __reduce__(self) -> tuple[typing.Any, ...]:
+    def __reduce__(self) -> typing.Tuple[typing.Any, ...]:
         return type(self), (self.status_code, self.detail, self.headers)
 
 
--- a/starlette/exceptions.py
+++ b/starlette/exceptions.py
@@ -22,6 +22,9 @@ class HTTPException(Exception):
         class_name = self.__class__.__name__
         return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
 
+    def __reduce__(self) -> tuple[typing.Any, ...]:
+        return type(self), (self.status_code, self.detail, self.headers)
+
 
 __deprecated__ = "ExceptionMiddleware"
 
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -1,6 +1,7 @@
 import warnings
 
 import pytest
+import pickle
 
 from starlette.exceptions import HTTPException
 from starlette.middleware.exceptions import ExceptionMiddleware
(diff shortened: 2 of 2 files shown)


nexusdk:
Looks like the test suite is failing due to some issue that does not relate to the code.
`Can't use 'tar -xzf' extract archive file`

Kludex:
> Looks like the test suite is failing due to some issue that does not relate to the code. `Can't use 'tar -xzf' extract archive file`

Yeah, it was a non-related issue. I've rerun the pipeline. JFYK

starlette/exceptions.py
```diff
@@ -22,7 +22,7 @@ def __repr__(self) -> str:
         class_name = self.__class__.__name__
         return f"{class_name}(status_code={self.status_code!r}, detail={self.detail!r})"
 
-    def __reduce__(self) -> tuple[typing.Any, ...]:
+    def __reduce__(self) -> typing.Tuple[typing.Any, ...]:
```
Kludex, line 25: We can improve this type annotation.

```python
Self = TypeVar("Self", bound="HTTPException")

def __reduce__(self: Self) -> typing.Tuple[Type[Self], typing.Typle[int, typing.Optional, typing.dict]]:
```

Kludex (commented):
I'm unsure about this PR.

What's the use-case here? Is it something you could do by yourself without changing Starlette?

nexusdk:
> I'm unsure about this PR.
> 
> What's the use-case here? Is it something you could do by yourself without changing Starlette?

We use FastAPI with Gunicorn and multiprocessing.managers.BaseManager in order to share a large object between processes. When an HTTPException is thrown (instantiated using the named parameter) in one of the object's methods, it always results in a 500 response to the client, regardless of the actual exception thrown, due to the unpickling of the HTTPException failing.

The error in the logs is then always the same, making it difficult to debug what actually went wrong:
`TypeError: __init__() missing 1 required positional argument: 'status_code'`

What is the harm in making Starlette's HTTPException more picklable?

adriangb:
You could use [`copyreg.pickle`](https://docs.python.org/3/library/copyreg.html#copyreg.pickle)

Kludex:
As @adriangb pointed out. It's already possible to this without modifying `Starlette` itself.

```python
import copyreg
import pickle

def pickle_http_exception(exc):
    return HTTPException, (exc.status_code, exc.detail, exc.headers)

copyreg.pickle(HTTPException, pickle_http_exception)
print(pickle.loads(pickle.dumps(HTTPException(status_code=404))))
```

That being the case, I'll be closing this PR. 🙏

nexusdk:
The actual pickle happens within Python's built-in multiprocessing library. I can't change the way it does the pickling.
https://github.com/python/cpython/blob/68f6a5da199826fef1098f869d8d2260aaccff8e/Lib/multiprocessing/connection.py#L210

adriangb:
You just have to call copyreg before you fork/spawn the processes

CrAzE124:
@Kludex the workaround solutions suggested here, while reasonable, still lead to hacky (and imo unnecessary) monkey patching. Would you mind explaining your reservations for this PR? I don't understand the harm in adding something that would make multiprocessing a bit smoother?

Kludex:
> Would you mind explaining your reservations for this PR?

Sure!

To start, `HTTPException` was added in 4 Sep 2018. In almost 4 years, no one asked for this, this means that it's not a wanted change overall.

Also, we try to avoid unnecessary changes to Starlette. If there's an alternative, that has a few lines, then it's a good reason for pushing this back.

We can continue the discussion on https://github.com/encode/starlette/discussions/1669, and if there are good arguments, we can go back on this decision. :)

aniketmaurya:
+1 to have HTTPException picklable

PR #1681, opened 2022-06-10
Revert "Allow staticfiles to follow symlinks outside directory"

Kludex:
- Reverts encode/starlette#1377

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -3,7 +3,6 @@ import os
 import stat
 import typing
 from email.utils import parsedate
-from pathlib import Path
 
 import anyio
 
@@ -52,7 +51,7 @@ class StaticFiles:
         self.all_directories = self.get_directories(directory, packages)
         self.html = html
         self.config_checked = False
-        if check_dir and directory is not None and not Path(directory).is_dir():
+        if check_dir and directory is not None and not os.path.isdir(directory):
             raise RuntimeError(f"Directory '{directory}' does not exist")
 
     def get_directories(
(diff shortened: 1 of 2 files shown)


lovelydinosaur (approved):

Kludex:
Ok. The incident has been resolved. It's time to explain what happened.

## Description

Last night, we received a message on Gitter asking for a secure communication channel to report a security issue (much appreciated btw). I sent a message to that person, with @tomchristie in cc. Promptly, they replied with the following message:

image

There was also an application to reproduce the issue attached:

```python
import uvicorn
from starlette.routing import Mount

app = Starlette(
    routes=[
        Mount('/static', app=StaticFiles(directory='static')),
    ]
)

if __name__ == '__main__':
    uvicorn.run('main:app')
```

As mentioned, the problem was that we were allowing any symlink outside the specified directory from `StaticFiles` to be exposed. You can see the snippet that introduced it below. 

https://github.com/encode/starlette/blob/d3dccdc477652b6de5a7b6b14a2bf3fa2f94be2c/starlette/staticfiles.py#L166-L175

## How this happened? How to avoid it in the future?

There were three reviewers on the [PR that introduced the security issue](https://github.com/encode/starlette/pull/1377), and two approvals (mine included). Even with that, we failed to see the issue.

As a self evaluation, and retrospective, I think this incident could have been avoided if we either had paid more attention, or maybe don't being overconfident about our knowledge on the [`pathlib`](https://docs.python.org/3/library/pathlib.html#module-pathlib) module, or the same for the syscalls involved. In any case, we interpreted the logic flow wrong.

## What we did to solve the issue?

As soon as the report was received, we reverted the PR, created a new release, and "yanked" the previous one on PyPI.

The version 0.20.2 was live for less than three days. It's also good to mention that none of the FastAPI users were affected.

## Lessons learned

- We need a security report policy, to allow a secure channel to report issues.
- Being too cautious and push back PRs is not a problem.

Notes:
- I didn't expose the person who reported the issue, because I didn't ask, but I'd be more than happy to give merits if they allow me to do so! :)

EDIT: As @m1ckey said some words below, I guess I can now give the merits to him. Thanks so much @m1ckey! ❤️ 🏆

aminalaee:
@Kludex @tomchristie Thank you for taking care of this, do we still want to add this? I can update the previous code and add the fix.

tiangolo:
Thanks a lot for doing this, the quick response, and the full report!

m1ckey:
Thank you for the professional handling.
I am happy that I could help to improve Starlette. :star2:

PR #1683, opened 2022-06-10
Allow StaticFiles follow symlinks

### LINKED ISSUE #1083: StaticFiles middleware doesn't follow symlinks

aminalaee:
After the previous attempt #1377 , this PR will allow StaticFiles to follow symlinks outside the directory.

Fixes https://github.com/encode/starlette/issues/1083

Thanks to @Kludex and @m1ckey


tests/test_staticfiles.py
```diff
@@ -441,3 +441,26 @@ def mock_timeout(*args, **kwargs):
...
+    statics_file_path = os.path.join(statics_path, "index.html")
+    os.symlink(symlink_file_path, statics_file_path)
+
+    app = StaticFiles(directory=statics_path)
+    client = test_client_factory(app)
+
+    response = client.get("/index.html")
+    assert response.url == "http://testserver/index.html"
+    assert response.status_code == 200
+    assert response.text == "<h1>Hello</h1>"
```
aminalaee, line 466: Maybe a test for disallowing path traversal can be useful but the TestClient (requests and urllib) won't allow path traversal, at least I haven't seen sensible one yet.

starlette/staticfiles.py
```diff
@@ -161,7 +161,7 @@ def lookup_path(
         self, path: str
     ) -> typing.Tuple[str, typing.Optional[os.stat_result]]:
         for directory in self.all_directories:
-            full_path = os.path.realpath(os.path.join(directory, path))
+            full_path = os.path.abspath(os.path.join(directory, path))
             directory = os.path.realpath(directory)
             if os.path.commonprefix([full_path, directory]) != directory:
                 # Don't allow misbehaving clients to break out of the static files
```
aminalaee, line 173: As far as I understand `os.path.realpath` will follow symbolic links and here we will stop it from going outside of statics directory.
Instead we can build the path with `abspath` and not follow symlinks (yet) and prevent breaking outside of directory and if file is a symlink it will do `os.stat` and by default it will follow symlinks.

yezz123 (approved):

requirements.txt
```diff
@@ -18,7 +18,7 @@ trio==0.19.0
 
 # Documentation
 mkdocs==1.3.0
-mkdocs-material==8.2.8
+mkdocs-material==8.3.3
```
ischaojie: This change is not related to this PR :), In my opinion, don't put irrelevant changes together—just advice.

starlette/staticfiles.py
scientificRat: Why not change line 165 `realpath` to `abspath`?  If the directory is a symbolic link, line 166 will fail.

tests/test_staticfiles.py
```diff
@@ -448,3 +448,52 @@ def mock_timeout(*args, **kwargs):
...
+
+    symlink_path = os.path.join(tmpdir, "symlink")
+    os.mkdir(symlink_path)
+
+    symlink_file_path = os.path.join(symlink_path, "index.html")
+    with open(symlink_file_path, "w") as file:
+        file.write("<h1>Hello</h1>")
+
+    temp_path = os.path.join(tmpdir, "index.html")
+    os.symlink(symlink_file_path, temp_path)
```
Kludex: Isn't the name of those to be the other way around?

The `symlink_file_path` should be inside the `statics`, and the `temp_path` outside of it. :thinking:

starlette/staticfiles.py
Kludex: Can we add a test with the scenario @scientificRat mentioned?

Kludex (changes requested):

starlette/staticfiles.py
aminalaee: sure, I'll take a look.

tests/test_staticfiles.py
```diff
@@ -448,3 +449,70 @@ def mock_timeout(*args, **kwargs):
...
+    os.mkdir(source_path)
+
+    source_file_path = os.path.join(root_source_path, "index.html")
+    with open(source_file_path, "w") as file:
+        file.write("<h1>Hello</h1>")
+
+    os.symlink(source_path, statics_path)
+
+    app = StaticFiles(directory=statics_path)
+    # We can't test this with 'requests', so we test the app directly here.
```
Kludex: we don't use `requests` anymore :eyes:

starlette/staticfiles.py
aminalaee: Not sure if I got the point correctly, but I added the tests for this, but I think this change is not needed.

tests/test_staticfiles.py
aminalaee: Thanks, updated.

Kludex (changes requested):
I'm still not 100% sure about this. :disappointed:

I think I need to write a small summary about this, and what are the alternatives if we don't do it. :eyes:

tests/test_staticfiles.py
Kludex:
```suggestion
    # We can't test this with 'httpx', so we test the app directly here.
```

Kludex:
My previous concern on this PR was:
- If a developer uses the same directory as the `StaticFiles` to upload files, we can have a scenario on which a symlink is uploaded there, and the machine can be exploited.

Should we be concerned about it? Is this feature wanted that bad? :thinking:

Kludex:
Can we make this feature opt-in i.e. add `follow_symlink` on the `StaticFiles`?

Let me know what you think @aminalaee

aminalaee:
> Can we make this feature opt-in i.e. add `follow_symlink` on the `StaticFiles`?

Yeah that works, I will update the PR.

Kludex:
Thanks

Kludex:
I guess only docs are missing, and we are ready here 👀

aminalaee:
@Kludex I just updated the docs. Let me know what you think.

docs/staticfiles.md
```diff
@@ -9,6 +9,7 @@ Signature: `StaticFiles(directory=None, packages=None, check_dir=True)`
 * `packages` - A list of strings or list of tuples of strings of python packages.
 * `html` - Run in HTML mode. Automatically loads `index.html` for directories if such file exist.
 * `check_dir` - Ensure that the directory exists upon instantiation. Defaults to `True`.
+* `follow_symlink` - A boolean indicating if symbolic links for files and directories should be followed. Defaults to `False`.
```
Kludex, line 12: It's missing on the signature above as well, jfyk

aminalaee: thanks, added it.

Kludex (approved):

Kludex:
Thanks @aminalaee :)

Check your Twitter please.

PR #1684, opened 2022-06-10
Upgrade mkdocs-material to 8.3.3

aminalaee:
I faced this in https://github.com/encode/starlette/pull/1683 and checking [here](https://pypi.org/project/mkdocs-material/#history) seems like mkdocs-material 8.2.8 is not available anymore.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ trio==0.19.0
 
 # Documentation
 mkdocs==1.3.0
-mkdocs-material==8.2.8
+mkdocs-material==8.3.3
 mkautodoc==0.1.0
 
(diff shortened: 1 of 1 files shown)


Kludex:
What do you mean? 🤔 
image

aminalaee:
@Kludex Sorry the I meant 8.2.8, but you're right. it is there in pypi but strangely the pipeline failed: https://github.com/encode/starlette/runs/6831795662?check_suite_focus=true

maybe some transient issue with PyPI or Github.

Kludex:
https://status.python.org/

aminalaee:
Ok good, let's close this.

PR #1685, opened 2022-06-11
Add more TestClient examples

### LINKED ISSUE #1677: Document how to send files using `TestClient`
Currently, people need to rely on lost issues to understand how to send files using the `TestClient`. I guess we'd want to document how to do that. Even adding a reference to `requests` would be already helpful here.

aminalaee:
Fixes #1677


docs/testclient.md
```diff
@@ -26,6 +26,30 @@ function calls, not awaitables.
...
+response = client.get("/", headers={"Authorization": "..."})
+```
+
+And for example to send files with the TestClient:
+
+```python
+client = TestClient(app)
+
+files = {"file": open("example.txt", "rb")}
+response = client.post("/form", files=files)
```
Kludex: Can we also have how to send multiple files? 🙏

docs/testclient.md
```diff
@@ -26,6 +26,35 @@ function calls, not awaitables.
...
+```python
+client = TestClient(app)
+
+# Send a single file
+files = {"file": open("example.txt", "rb")}
+response = client.post("/form", files=files)
+
+# Send multiple files
+files = {"file1": open("example.txt", "rb"), "file2": open("example.png", "rb")}
+response = client.post("/form", files=files)
```
Kludex: With `open()` like this, pytest will warn about unclosed resources, if we set to raise errors on warnings. 🤔

Also, I wrote this comment years ago, which kind of mismatch the written here: https://github.com/tiangolo/fastapi/issues/1536#issuecomment-640781718

aminalaee: Yeah that's a good point, but we also have this case in Starlette `test_staticfiles.py` which pytest warns about this. But I agree it's better to use the contextt manager. Let me update..

aminalaee: I think this should be ok now, thank you Marcelo.

docs/testclient.md
```diff
@@ -26,6 +26,37 @@ function calls, not awaitables.
...
+    response = client.post("/form", files={"file": f})
+
+# Send multiple files
+with open("example.txt", "rb") as f1:
+    with open("example.png", "rb") as f2:
+        files = {"file1": f1, "file2": ("filename", f2, "image/png")}
+        response = client.post("/form", files=files)
+```
+
+For more information you can view `requests.Session` [docs](https://requests.readthedocs.io/en/master/user/advanced/) and [API](https://requests.readthedocs.io/en/master/api/#requests.Session).
```
Kludex: There's already a mention on the "requests API" above in this file.

What do you think about?
```suggestion
For more information you can check the `requests` [documentation](https://requests.readthedocs.io/en/master/user/advanced/).
```

Kludex (approved):

PR #1691, opened 2022-06-14
Experiment a high-level HTTPMiddleware

florimondmanca:
Refs #1678, cc @adriangb 

I experimented with what a helper middleware to do some simple request/response peeks and edits might look like, and yet be pure ASGI -- no Starlette dependencies at all, accept any ASGI app as a `parent`.

I was inspired by what we did for [HTTPX auth flows](https://www.python-httpx.org/advanced/#customizing-authentication). Turns out we can use the same generator-based API to provide a somewhat clean interface that goes like this:

```python
class MyMiddleware(HTTPMiddleware):
    async def dispatch(self, request):
        # Access or tweak the `request`...

        try:
            response = yield  # Call into the underlying app.
        except Exception:
            # Handle any exception that occurred, and optionally
            # respond on behalf of the failed app.
            yield Response(status_code=500)

        # Tweak the response
        response.headers["Custom"] = "Example"
```

The structure of this API does NOT allow:

* Accessing or modifying the request body. The request is built with `scope` only (request method, URL, headers, etc), so we haven't got any access to `receive()`.

I copy-pasted the `BaseHTTPMiddleware` tests, switching to `HTTPMiddleware`, and got everything passing.

I also looked up some example usage on https://grep.app and found [this fairly complex usage](https://github.com/Netflix/dispatch/blob/b2eade54f1b159019a45c9cd95b24626610faf22/src/dispatch/main.py#L160-L179) from Netflix's Dispatch project. Here's the diff showing how they might upgrade:

```diff
- class MetricsMiddleware(BaseHTTPMiddleware):
-     async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
+ class MetricsMiddleware(HTTPMiddleware):
+     async def dispatch(self, request: Request) -> AsyncGenerator[None, Response]:
          path_template = get_path_template(request)

          method = request.method
          tags = {"method": method, "endpoint": path_template}

          try:
              start = time.perf_counter()
-             response = await call_next(request)
+             response = yield
              elapsed_time = time.perf_counter() - start
          except Exception as e:
              metric_provider.counter("server.call.exception.counter", tags=tags)
              raise e from None
          else:
              tags.update({"status_code": response.status_code})
              metric_provider.timer("server.call.elapsed", value=elapsed_time, tags=tags)
              metric_provider.counter("server.call.counter", tags=tags)

-         return response
```

Some missing pieces might be:

- Refine the 'response patching' API -- I've reused the `Response` class, but we might want to use a specific 'stub' that clearly limits what's allowed within the "on-response-headers" part of the middleware. E.g. it should be clear that it's not possible to patch `response.body`.


adriangb:
Just curious, did you look at [my branch]( https://github.com/encode/starlette/compare/master...adriangb:safe-basehttpmiddleware?expand=1) (linked in https://github.com/encode/starlette/issues/1678#issuecomment-1155605380)? I ask because if you did not it seems we arrived at the exact same solution independently, which is a good sign.

florimondmanca:
@adriangb Woah, I did _not_ look at that branch — a _very_ good sign indeed!

adriangb:
> Woah, I did not look at that branch — a very good sign indeed!

Agreed. It seems like we even did the same thing with `early_response`.

> async support

I think we should just make it an async generator and _always_ an async generator.

> it should be clear that it's not possible to patch response.body

There is _one_ alternative: if `response.body` is set / changed we can return that instead. But maybe we want to allow _replacing_ the entire response?

```python
response = yield
yield Response()
```

We can support this by checking if the generator stops or is still available.

Another issue to think about: do we want to allow (safely) reading the request body?

- If the user calls `request.body()` we store the entire body in memory and replace the `receive` stream with that body
- If the user calls `request.stream()` we replace `receive` with a single message with an empty body

starlette/middleware/http.py
```diff
@@ -0,0 +1,94 @@
...
+            await self.app(scope, receive, send)
+            return
+
+        async with aclosing(self.dispatch(scope)) as flow:
+            # Kick the flow until the first `yield`.
+            # Might respond early before we call into the app.
+            early_response = await flow.__anext__()
+
+            if early_response is not None:
+                await early_response(scope, receive, send)
```
adriangb:
```suggestion
            early_response_option = await flow.__anext__()

            if early_response_option is not None:
                await early_response_option(scope, receive, send)
```

starlette/middleware/http.py
```diff
@@ -0,0 +1,94 @@
...
+                return
+
+            response_started = set[bool]()
+
+            wrapped_send = partial(
+                self._send,
+                flow=flow,
+                response_started=response_started,
+                send=send,
+            )
```
adriangb: Can we just make this a closure?

tests/middleware/test_http.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def test_exception_on_mounted_apps(test_client_factory):
+    sub_app = Starlette(routes=[Route("/", exc)])
+    app = Starlette(routes=[Mount("/sub", app=sub_app)])
+
+    client = test_client_factory(app)
+    with pytest.raises(Exception) as ctx:
+        client.get("/sub/")
+    assert str(ctx.value) == "Exc"
```
adriangb: IMO we should move this test, it's not relevant to BaseHTTPMiddleware. Right?

tests/middleware/test_http.py
```diff
@@ -0,0 +1,208 @@
...
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        ctxvar.set("set by middleware")
+        await self.app(scope, receive, send)
+        assert ctxvar.get() == "set by endpoint"
+
+
+class CustomMiddlewareUsingHTTPMiddleware(HTTPMiddleware):
+    async def dispatch(self, scope: Scope) -> HTTPDispatchFlow:
+        ctxvar.set("set by middleware")
+        yield None
```
adriangb: The `yield None` can be avoided by users if they write `-> AsyncGenerator[None, Response]` or `-> AsyncIterable[None]`

starlette/middleware/http.py
florimondmanca: No strong sentiments on that, I tend to prefer the partial method style as it keeps the main body of the middleware free from clutter, but the result here with everything inline is acceptable too.

starlette/middleware/http.py
florimondmanca: Not sure what prompts this? We know from type hints that this is an `Response | None`, I don't think we need to add its "optional" nature in the name.

adriangb: Yeah you don't have to accept it. I got a little bit confused reading it without type hints because I thought that maybe they use her was returning an ASGI app (well they are, it's a Response, I thought an arbitrary ASGI callable). I also tend to use `optional_...` or `maybe_...`, but that's just my preference

florimondmanca: > maybe they use her was returning an ASGI app

Indeed, I called the variable `response` but it can actually be any `ASGIApp`. Not sure if that should be made clearer? Edit: OH! I think the `YieldType` should be `ASGIApp` then.

I actually like and tend to use `maybe_*` as well, so lemme switch that over.

starlette/middleware/http.py
adriangb: I seem to remember someone in another issue complaining about us changing an internal method. If it's not intended for subclasses to use, can we name mangle it (`__` I think)?

florimondmanca: You might be looking at the outdated diff -- I switched to the closure style, so no more `_send()` method. :+1:

setup.py
```diff
@@ -39,6 +39,7 @@ def get_long_description():
     install_requires=[
         "anyio>=3.4.0,<5",
         "typing_extensions>=3.10.0; python_version < '3.10'",
+        "async_generator; python_version < '3.10'",
```
adriangb:
```suggestion
```

starlette/_compat.py
```diff
@@ -27,3 +32,9 @@ def md5_hexdigest(
 
     def md5_hexdigest(data: bytes, *, usedforsecurity: bool = True) -> str:
         return hashlib.md5(data).hexdigest()
+
+
+try:
+    from contextlib import aclosing  # type: ignore[attr-defined]
+except ImportError:  # Python < 3.10
+    from async_generator import aclosing  # type: ignore
```
adriangb:
```suggestion
# see contextlib.aclosing
class aclosing:
    def __init__(self, thing: AbstractAsyncContextManager):
        self.thing = thing
 
    async def __aenter__(self):
        return self.thing
   
    async def __aexit__(self, *exc_info: Any):
        await self.thing.aclose()
```

adriangb (commented):
`aclosing` is only a couple LOC, we can just vendor it

starlette/_compat.py
adriangb: (need the import for `AbstractAsyncContextManager`, or it could just be `Any` honestly)

starlette/middleware/http.py
```diff
@@ -0,0 +1,81 @@
+from typing import AsyncGenerator, Callable, Optional, Union
+
+from .._compat import aclosing
+from ..datastructures import MutableHeaders
+from ..responses import Response
+from ..types import ASGIApp, Message, Receive, Scope, Send
+
+_HTTPDispatchFlow = Union[
+    AsyncGenerator[None, Response], AsyncGenerator[ASGIApp, Response]
+]
```
adriangb:
```suggestion
_HTTPDispatchFlow = Union[
    AsyncGenerator[None, Response], AsyncGenerator[Response, Response]
]
```

Just because we _can_ accept an arbitrary ASGI app doesn't mean we should. I think we should go with the more restrictive version for now because:
1. A user that wants an arbitrary ASGI app can just add a `type: ignore` and it will work
2. It's easier to add an API than remove it. We can loosen the type annotation if/when someone asks for it. This way we can assume it's a `Response` if we want to use that to implement some other feature down the road.

florimondmanca: Sounds fair!

florimondmanca:
@adriangb As for bodies -- I'm tempted we explicitly don't allow a helper `HTTPMiddleware` to read or alter them. Buffering like that might get into issues with improper streaming support again. The rationale would be, "if you do need that, then go pure ASGI".

adriangb:
Yeah I think that's a fair position to have. I _think_ it can be done without issues (ref https://github.com/adriangb/starlette/commit/0bf1f3fa810740f6ab8fa43df2b41494b841efba). But we can always add that feature later, it wouldn't be breaking this API.

starlette/middleware/http.py
```diff
@@ -0,0 +1,86 @@
...
+    AsyncGenerator[Response, Response],
+    AsyncGenerator[Optional[Response], Response],
+]
+
+
+class HTTPMiddleware:
+    def __init__(
+        self,
+        app: ASGIApp,
+        dispatch: Optional[Callable[[Scope], _HTTPDispatchFlow]] = None,
```
adriangb: Can we make it accept an `HTTPConnection` it's basically a `Request` without access to the body.

florimondmanca:
I switched things so that `.dispatch()` accepts an unbound `request`. Saves users from manually building the components they need from `scope`, eg `Headers(raw=scope["headers"])`. Instead they can do `request.headers`, and `.body()` or `.stream()` will raise errors because `receive` wasn't passed to the `Request`.

starlette/middleware/http.py
adriangb: Ha you read my mind!

florimondmanca: Oh that's nice. Using this type hints don't reveal any body method. Better than passing a `Request(scope)` that would fail at runtime if calling `.body()` or something.

erewok:
This looks great! FWIW, I support not allowing accessing or modifying request or response body. It's a fair limitation, in my opinion.

starlette/_compat.py
```diff
@@ -1,4 +1,10 @@
 import hashlib
+from typing import Any
```
adriangb:
```suggestion
from typing import Any, AbstractAsyncContextManager
```

adriangb:
```python
import sys
```

starlette/_compat.py
```diff
@@ -1,4 +1,10 @@
 import hashlib
```
adriangb, line 1:
```suggestion
import hashlib
import sys
```

starlette/_compat.py
```diff
@@ -27,3 +33,18 @@ def md5_hexdigest(
 
     def md5_hexdigest(data: bytes, *, usedforsecurity: bool = True) -> str:
         return hashlib.md5(data).hexdigest()
+
+
+try:
+    from contextlib import aclosing  # type: ignore[attr-defined]
+except ImportError:  # Python < 3.10  # pragma: no cover
```
adriangb:
```suggestion
if sys.version_info >= (3, 10):
    from contextlib import aclosing
else:
```

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+    # if condition:
+    #     yield Response(...)
+    #     return
+    # try:
+    #     response = yield None
+    # except Exception:
+    #     yield Response(...)
+    # else:
+    #    ...
+    AsyncGenerator[Optional[Response], Response],
```
adriangb: Thanks for the comments, great stuff 👍

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+        dispatch: Optional[Callable[[HTTPConnection], _DispatchFlow]] = None,
+    ) -> None:
+        if dispatch is None:
+            dispatch = self.dispatch
+
+        self.app = app
+        self._dispatch_func = dispatch
+
+    def dispatch(self, conn: HTTPConnection) -> _DispatchFlow:
+        raise NotImplementedError  # pragma: no cover
```
adriangb: Do we maybe want to call this `_dispatch`? I played around with the idea. My reasoning is that the main advantage of subclassing is being able to "ship" a self contained middleware, e.g. in a library. And in that case, you'd want to reduce the public API as much as possible and this way we are forcing a public method onto the class.

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+        async with aclosing(self._dispatch_func(conn)) as flow:
+            # Kick the flow until the first `yield`.
+            # Might respond early before we call into the app.
+            maybe_early_response = await flow.__anext__()
+
+            if maybe_early_response is not None:
+                await maybe_early_response(scope, receive, send)
+                return
+
+            response_started: set = set()
```
adriangb: While I understand that this works, I think personally I would prefer a `response_started = False` even if it requires a `nonlocal response_started` below. Just preference.

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+
+            async def wrapped_send(message: Message) -> None:
+                if message["type"] == "http.response.start":
+                    response_started.add(True)
+
+                    response = Response(status_code=message["status"])
+                    response.raw_headers.clear()
```
adriangb: This was really smart 👍

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+
+                    try:
+                        await flow.asend(response)
+                    except StopAsyncIteration:
+                        pass
+                    else:
+                        raise RuntimeError("dispatch() should yield exactly once")
+
+                    headers = MutableHeaders(raw=message["headers"])
+                    headers.update(response.headers)
```
adriangb:
```suggestion
                    headers = MutableHeaders(raw=message["headers"])
                    headers.update(response.headers)
                    message["headers"] = headers.raw
```

I'm guessing `MutableHeaders` mutates the list it receives, but it's not super obvious and I'm not sure we should rely on that behavior.

starlette/middleware/http.py
```diff
@@ -0,0 +1,103 @@
...
+        app: ASGIApp,
+        dispatch: Optional[Callable[[HTTPConnection], _DispatchFlow]] = None,
+    ) -> None:
+        if dispatch is None:
+            dispatch = self.dispatch
+
+        self.app = app
+        self._dispatch_func = dispatch
+
+    def dispatch(self, conn: HTTPConnection) -> _DispatchFlow:
```
adriangb:
```suggestion
    def dispatch(self, __conn: HTTPConnection) -> _DispatchFlow:
```

This way it's a positional-only parameter in the base class and subclasses can name it as they please. Especially helpful if we end up swapping out `HTTPConnection` for some `Request` subclass in the future (`conn: Request` doesn't look very nice).

florimondmanca: Is that `__param` trick an official Python thing? I see my code editor does interpret this as "positional only", as it now shows `dispatch(self, __conn: HTTPConnection, /)` on hover despite not putting the `/` separator in code.

adriangb: Yep: https://peps.python.org/pep-0484/#positional-only-arguments

starlette/middleware/http.py
florimondmanca: This is a pattern we use in a number of places in the codebase already, eg. here: https://github.com/encode/starlette/blob/5b56e7dc429de946be8c782e0ae69bcc5d5b5585/starlette/middleware/cors.py#L157-L158

`MutableHeaders` takes care of deduplicating and normalization for us, which we wouldn't get with eg. `message["headers"].extend(response.raw_headers)`.

starlette/middleware/http.py
florimondmanca: I think I get your point. But I'm not in favor of making `_dispatch()` private -- as that's public API as far as `HTTPMiddleware` is concerned. It feels odd to be required to override a private method.

Library authors who do really want to hide a dispatch implementation from their users could always do this:

```python
class MyScopedMiddleware:
    def __init__(self, app):
        self._app = HTTPMiddleware(app, dispatch=self._dispatch)

    async def _dispatch(self, __conn):
        ...

        await self._app(scope, receive, send)
```

adriangb: Yep reasonable, resolving this thread

starlette/middleware/http.py
adriangb: I get that, the part I don't like is that we don't explicitly assign `message["headers"] = MutableHeaders.raw`. I don't like relying on `MutableHeaders` modifying the list it gets passed in place. This is counterintuitive: `a = [1]; list(a).pop()` doesn't modify `a`.  So even if in practice it is unnecessary I think it is worth explicitly assigning to be clear about what is happening.

florimondmanca: Oh, I see.

adriangb:
This is looking great. I think the last issue we have to carry over from #1694 is what to do if a user tries to set `Response.media_type` or `Response.body`? I think the safest path for now is to subclass `Response` and raise exceptions if they try to set either of those attributes.

florimondmanca:
@adriangb I made an attempt at controlling what attributes can be read/written in the context of the middleware. It works, but I'm far from satisfied by it. Overall that seems overly complex and defensive...

The core problem is, Starlette's `Response` wasn't designed to be built-and-swapped like this. On the other hand that's a central principle in `httpcore` for example.

Maybe, _for now_ (? how 'temporary' would that actually be?), we should do `status_code, headers = yield`? That would be structurally more appropriate: you only get a read-only access to `status_code`, and you can access or tweak the `headers`, is all. No access given to a full-fledged `Response` instance that wasn't designed for setter usage. The signature would become: `AsyncGenerator[None, Tuple[int, (Mutable)Headers]]`.

adriangb:
I think your attempt is quite reasonable. I share your same concerns, but as a counter point this API (a `Response` object that can't really be modified in some ways) has existed in BaseHTTPMiddleware for a long time and I do not recall anyone complaining about that part of BaseHTTPMiddleware, so maybe what you've got is fine.

I also don't particularly like the tuple response, especially as something temporary (changing it would be a breaking change).

Kludex:
What should we do with this PR?

adriangb:
The conclusion from discussion in the meeting was that it's probably best to just push users towards writing pure ASGI middleware or solutions provided by higher-level packages like FastAPI. The logic behind this decision is that users generally are not writing one-off middleware and any sort of reusable middleware component can take the time to be implemented as a pure ASGI middleware.

So I think we should close this PR? Florimond, what do you think? Maybe this can be a 3rd party package?

Kludex:
Closing it just to organize a bit the PRs.

Feel free to reopen if you think it's still relevant. :pray:

PR #1692, opened 2022-06-15
Reuse Request's body buffer for call_next in BaseHTTPMiddleware

### LINKED ISSUE #493: Question about Request.Body()

### LINKED ISSUE #495: Consume request body in middleware is problematic

### LINKED ISSUE #847: Middleware Request parse Hangs forever

adriangb:
Thought of this while working on related work to #1691. Sorry if this has already been proposed, I know we've gone around a couple times with similar stuff.

Background:
- Closes #848 
- Closes #1519 
- Closes #944 
- Fixes #847
- Fixes #495

Basically if the user calls `Request.body()` from their dispatch function we cache the entire request body in memory and pass that to downstream middlewares but if they call `Request.stream()` then all we do is send an empty body so that downstream things don't hang forever.

I think this behavior is pretty sensible and doesn't use any unexpected memory (e.g. caching all of the body if `Request.stream()` was called). It also doesn't break the ASGI flow: if a downstream middleware replaces `receive()` or modified messages the downstream ASGI app (including the final endpoint) will see the new `receive()`.

If this approach works well we could upstream this into the base Request so that arbitrary uses of Request can inter operate with ASGI apps/call chains.

Note that **this does not fix https://github.com/encode/starlette/issues/493#issuecomment-487566994**  or any other case where the body is consumed in the endpoint and then `Request.body()` is called somewhere else (e.g. in BaseHTTPMiddleware after `call_next()` or in an exception handler).


adriangb:
> It also doesn't break the ASGI flow: if a downstream middleware replaces receive() or modified messages the downstream ASGI app (including the final endpoint) will see the new receive().

TODO: add a test to verify this.

Other misc TODO:
- handle client disconnects (the test for this is fugly)

starlette/middleware/base.py
```diff
@@ -4,14 +4,33 @@
...
+                "more_body": False,
+            }
+        elif self._stream_consumed:
+            # stream() was called to completion
+            return {
+                "type": "http.request",
+                "body": b"",
+                "more_body": False,
+            }
+        return await self.receive()
```
adriangb: I think handling here should be more stateful. In particular:
-  if the stream gets consumed by the downstream app we should mark the stream as consumed on this app instance
- if we've already retuned all of the body we should move onto polling receive() again (so that we can wait for the client to disconnect, etc)

starlette/requests.py
```diff
@@ -187,15 +188,19 @@ async def empty_send(message: Message) -> typing.NoReturn:
...
     def __init__(
         self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send
     ):
         super().__init__(scope)
         assert scope["type"] == "http"
         self._receive = receive
         self._send = send
-        self._stream_consumed = False
-        self._is_disconnected = False
+        self._stream_state = self._StreamState.connected
```
adriangb: This makes the state of "stream not consumed but disconnected" unrepresentable.

starlette/requests.py
```diff
@@ -211,21 +216,24 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
             yield b""
             return
 
-        if self._stream_consumed:
+        if self._stream_state is self._StreamState.disconnected:
+            raise RuntimeError("Client disconnected")
```
adriangb: This wasn't checked or tested but I think it makes sense to check for it

starlette/middleware/base.py
```diff
@@ -2,16 +2,62 @@
...
+    async def wrapped_receive(self) -> Message:
+        if self._wrapped_rcv_state is self._StreamState.connected:
+            if getattr(self, "_body", None) is not None:
+                # body() was called
+                self._wrapped_rcv_state = self._StreamState.consumed
+                return {
+                    "type": "http.request",
+                    "body": self._body,
+                    "more_body": False,
+                }
```
adriangb: Note that there is no _additional_ buffering going on here: if the user calls `request.body()` on this `Request` instance within `dispatch` we would already be keeping around all of the bytes for the duration of the `dispatch()` call. We are just re-using that buffering for the downstream ASGI app. If the user never called `request.body()` then we consume the body here, just like if the user had called `request.stream()`.

starlette/requests.py
adriangb: To be clear: this PR _can_ be implemented without this change, but I think it cleans things up nicely and makes the new code portion of this PR simpler.

starlette/requests.py
```diff
@@ -187,15 +188,19 @@ async def empty_send(message: Message) -> typing.NoReturn:
 
 
 class Request(HTTPConnection):
+    class _StreamState(enum.Enum):
+        connected = enum.auto()
+        consumed = enum.auto()
+        disconnected = enum.auto()
```
Kludex: Conceptually speaking, why would a `stream` have `connected` and `disconnected` state?

starlette/requests.py
Kludex: Why?

Kludex (commented):
What does this means exactly:
> If this approach works well we could upstream this into the base Request so that arbitrary uses of Request can inter operate with ASGI apps/call chains.

You mean that if we create a `Request` object inside a middleware/ASGI app, we can't get the body again if we use the same middleware twice, for example?

adriangb:
> What does this means exactly:
> 
> > If this approach works well we could upstream this into the base Request so that arbitrary uses of Request can inter operate with ASGI apps/call chains.
> 
> You mean that if we create a `Request` object inside a middleware/ASGI app, we can't get the body again if we use the same middleware twice, for example?

What I mean here is that if we moved this implementation into the base `Requests` class it would allow folks not using `BaseHTTPMiddleware` but using `Request` (https://github.com/encode/starlette/blob/master/docs/middleware.md#reusing-starlette-components) to read bodies without breaking the downstream ASGI app as long as they do `receive = request.wrapped_receive`.

starlette/requests.py
adriangb: The way I see it there are three states. We start in a "connected" state. If we call `receive()` and get `"more_body": False` then we transition to the "consumed" state. If at any point we receive an `"http.disconnect` message we transition to the "disconnected" state. I think this is the same thing that was previously being represented by two boolean instance variables but by using an enum we disallow representation of the "disconnected but stream is not consumed state"

starlette/requests.py
adriangb: It's for this test: https://github.com/encode/starlette/blob/1dad9de11cd6c65a5f937664930051e0d5b20c6a/tests/test_requests.py#L270-L296

Re-reading the linked asgiref issue, I now realize that with Uvicorn at least it's not a problem. Which makes this much more tangential and unnecessary. I'm going to remove this to minimize the changes.

starlette/requests.py
adriangb: But again I could not make this change, it just adds a bit more boilerplate / repeated logic into the `wrapped_receive()` implementation. If you'd prefer me to not make this change just let me know and I'll work around that.

adriangb:
Note that **this does not fix https://github.com/encode/starlette/issues/493#issuecomment-487566994**  or any other case where the body is consumed in the endpoint and then `Request.body()` is called somewhere else (e.g. in BaseHTTPMiddleware after `call_next()` or in an exception handler). These seem like finicky use cases, I don't think there's any general way to maintain full ASGI compatibility and supporting these use cases since they essentially require information to be transmitted "upstream". We _could_ mitigate the "hang forever" behavior by detecting stream double consumption and erroring out, but that's about it.

adriangb:
I tried to figure out a way to get use cases like https://github.com/encode/starlette/issues/493#issuecomment-487566994 working but failed. Here's a comparison of before and after for this change:

| Request.body called first in | Endpoint | Middleware | Error handler |
| - | - | - | - |
| Endpoint | ✅ |  ❌ | N/A | 
| Middleware | ❌ | ✅ | N/A | 
| Error handler | ❌| ❌ | ✅ | 

| Request.body called first in | Endpoint | Middleware | Error handler |
| - | - | - | - |
| Endpoint | ✅ |  ✅ | N/A | 
| Middleware | ❌ | ✅ | N/A | 
| Error handler | ❌| ❌ | ✅ | 

So this only really fixes the case that could already be fixed by moving to pure ASGI middleware, as discussed in https://github.com/encode/starlette/pull/1519#issuecomment-1060633787. The use cases this doesn't fix (reading the request body for the second time in an error handler or a middleware) are also not use cases you can get around using pure ASGI middleware.

So the question is: do we move forward with this to fix one specific use case, or do we leave it as -is?

ZhymabekRoman:
Thank you all for developing and contributing to this library. Today I want to make a logging for the back-end app, and I'm stuck on it. Is there any change in the status of this PR? I've seen this PR change and I think this implementation is too confusing, we have to look at this issue differently. Sorry if I don't express myself that way, get it right, I have no doubt about your professionalism, but the Middleware codebase requires a refactor as to me

adriangb:
Thanks for the feedback @ZhymabekRoman.

> I've seen this PR change and I think this implementation is too confusing, we have to look at this issue differently.

Is it confusing to you conceptually or the code itself?

> but the Middleware codebase requires a refactor as to me

I don't get what you mean by this. Are you referring to `BaseHTTPMiddleware`? To all middleware?

adriangb:
We could also detect stream double consumption here and prohibit calling stream() and body() after call_next() is called if the stream is consumed. That way there’s no error/hanging. With that I think we will have pretty much solved all of BaseHTTPMiddleware’s problems?

Kludex:
> We could also detect stream double consumption here and prohibit calling stream() and body() after call_next() is called if the stream is consumed. That way there’s no error/hanging. With that I think we will have pretty much solved all of BaseHTTPMiddleware’s problems?

Besides the ContextVar?

adriangb:
Yep

starlette/requests.py
adriangb: I gave up and I'm just using the same 2 booleans we were using before now.

starlette/middleware/base.py
```diff
@@ -14,6 +14,61 @@
 T = typing.TypeVar("T")
 
 
+class _CachedRequest(Request):
```
lovelydinosaur, line 17: Your description in the pull request helped me understand the functionality here.

> Basically if the user calls Request.body() from their dispatch function we cache the entire request body in memory and pass that to downstream middlewares but if they call Request.stream() then all we do is send an empty body so that downstream things don't hang forever.

Perhaps it could be adapted into something we can use for the docstring?

adriangb: Added!

adriangb:
@tomchristie are there any other changes you'd like, or anything I can do to help get this reviewed?

starlette/requests.py
```diff
@@ -220,10 +220,12 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
         while True:
             message = await self._receive()
             if message["type"] == "http.request":
+                if not message.get("more_body", False):
+                    self._stream_consumed = True
```
adriangb, line 224: This gets used by `_CachedRequest`, adding it here avoids having to overwrite all of async def stream.

starlette/middleware/base.py
```diff
@@ -14,6 +14,68 @@
...
+    """
+    If the user calls Request.body() from their dispatch function
+    we cache the entire request body in memory and pass that to downstream middlewares,
+    but if they call Request.stream() then all we do is send an
+    empty body so that downstream things don't hang forever.
+    """
+
+    def __init__(self, *args: typing.Any, **kwargs: typing.Any) -> None:
```
Kludex: Why not naming the parameters?

starlette/middleware/base.py
```diff
@@ -26,6 +88,8 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await self.app(scope, receive, send)
             return
 
+        request = _CachedRequest(scope, receive)
```
Kludex, line 104: This doesn't need to be here, does it? Can it be in the same place it was instantiated before, or am I missing something?

starlette/requests.py
```diff
@@ -223,7 +223,7 @@ async def stream(self) -> typing.AsyncGenerator[bytes, None]:
                 body = message.get("body", b"")
                 if body:
                     yield body
-                if not message.get("more_body", False):
+                if self._stream_consumed:
```
Kludex: Does this change is needed? The `_CachedRequest` doesn't change the value of `self._stream_consumed`. 🤔

Kludex: The test `test_read_request_stream_in_dispatch_after_app_calls_body` fails without this logic.

Hmm... Why the `more_body` doesn't matter? Like, not considering the `BaseHTTPMiddleware`, why the `more_body` doesn't matter to exit?

Hmmm... If we receive 2 chunks of body, how this works? It doesn't look like we have a test that covers standalone `Request` with multiple chunks. 🤔

Kludex: Or am I missing something?

starlette/middleware/base.py
```diff
@@ -14,6 +14,68 @@
...
+            if getattr(self, "_body", None) is not None:
+                # body() was called, we return it even if the client disconnected
+                self._wrapped_rcv_consumed = True
+                return {
+                    "type": "http.request",
+                    "body": self._body,
+                    "more_body": False,
+                }
+            elif self._stream_consumed:
+                # stream() was called to completion or client disconnected
```
Kludex: This "or disconnected" is not right, right? The disconnected logic is not handled here. Or...?

starlette/middleware/base.py
```diff
@@ -14,6 +14,68 @@
...
+                    self._wrapped_rcv_consumed = self._stream_consumed
+                    return {
+                        "type": "http.request",
+                        "body": chunk,
+                        "more_body": self._stream_consumed,
+                    }
+                except ClientDisconnect:
+                    self._wrapped_rcv_disconnected = True
+                    return {"type": "http.disconnect"}
+        # wrapped_rcv is either disconnected or consumed
```
Kludex: The "or consumed" also shouldn't be here, right?

Kludex:
Sorry the time it took to review this.

starlette/middleware/base.py
adriangb: Because I don't want to repeat a bunch of boilerplate. I'm just adding two attributes, not interacting with those other parameters at all. This is a very reasonable use of args kwargs.

starlette/middleware/base.py
adriangb: The next line references this. It, maybe, can be moved to where it was before but at the very least it will need a new variable name  like`outer_request` to differentiate it from the `request: Request` on line 95. It makes more sense to just move it up here, there is no harm in that.

starlette/middleware/base.py
Kludex: Isn't it only scope, receive and send? 🤔

adriangb: okay I'll name them

adriangb: done

starlette/middleware/base.py
adriangb: I'll re-organize this a bit to clarify

starlette/middleware/base.py
adriangb: re-organized for clarity

starlette/requests.py
adriangb: We really should have tests for `Request` as a standalone thing since it is a standalone thing in the public API and... I've been encouraging people to use it e.g. in ASGI middleware.

Kludex: We do, we just don't cover what I mention

adriangb: I'll add it. If you already prototyped it out in your head or on paper please comment it here and save me a few min haha.

Kludex: I don't recall how to do it from the TestClient's POV, but I thought about sending a stream with 2 chunks. Maybe you can use httpx directly if you can't do it with the TestClient. 

I guess that would be enough to break this logic here, since the value of stream_consumed will not change

adriangb: Ok yes, you were right, I did have a bug, good catch. I still need to modify `Request` a bit, I added a couple of tests to explain why. TLDR is we were marking the stream as consumed as soon as you call `stream()` but in reality you can call stream, get one message and then _call steam again_ before it is consumed. Let me know if it's clear now.

adriangb:
@Kludex  thank you for the review. No worries with the timeline, the only issue becomes that I start forgetting why I did things like I did so sorry if answers to some of your questions weren't super clear.

I fixed the bug you found and reworked things to try to make the answers to your questions clearer by just looking at the code. I also added more tests, now there are 400+ lines of tests for 70 lines of implementation 😅 .

Kludex (approved):
Sorry for the long wait here.

Let's go 👍

starlette/middleware/base.py
larsagny: Just to be clear @adriangb : if I call `request.body()` _before_ `call_next` I'll be fine, right? Any endpoint that calls `request.body` will get the cached data? 

```
        my_body = await request.body()
```

adriangb: yes that's the idea. And if you call it after things won't hang, you'll get an empty request body (if I remember correctly)

larsagny: Thanks for getting back to me to quickly.  The request doesn't hang anymore indeed, but you'll get a `RuntimeError`

```
    if await request.body():
  File "/home/basicuser/.local/lib/python3.10/site-packages/starlette/requests.py", line 236, in body
    async for chunk in self.stream():
  File "/home/basicuser/.local/lib/python3.10/site-packages/starlette/requests.py", line 218, in stream
    raise RuntimeError("Stream consumed")
RuntimeError: Stream consumed
```

adriangb: That makes sense, you consumed the stream in the request body so it's not available in the middleware.

PR #1694, opened 2022-06-15
copy over some tests from alternative implementation

adriangb:
Copying over tests from my implementation.

This drops the `contextvar` tests which in my opinion are unnecessary with this implementation.
It also adds types to the test file.

There's two main test cases being added:
1. yielding after an early response (I added the fix for this, it's pretty straightforward)
2. Modifying `Response.media_type`. In my implementation I was detecting this and overwriting the header, but I guess we could subclass `Response` and raise an exception if someone attempts to assign it?

### DIFF
--- a/starlette/middleware/http.py
+++ b/starlette/middleware/http.py
@@ -36,7 +36,7 @@ class HTTPMiddleware:
         self.app = app
         self._dispatch_func = dispatch
 
-    def dispatch(self, conn: HTTPConnection) -> _DispatchFlow:
+    def dispatch(self, __conn: HTTPConnection) -> _DispatchFlow:
         raise NotImplementedError  # pragma: no cover
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
@@ -52,6 +52,13 @@ class HTTPMiddleware:
             maybe_early_response = await flow.__anext__()
 
             if maybe_early_response is not None:
+                try:
+                    await flow.__anext__()
+                except StopAsyncIteration:
+                    pass
+                else:
+                    raise RuntimeError("dispatch() should yield exactly once")
+
                 await maybe_early_response(scope, receive, send)
                 return
 
@@ -98,6 +105,3 @@ class HTTPMiddleware:
 
                 await response(scope, receive, send)
                 return
-
-            if not response_started:
-                raise RuntimeError("No response returned.")
--- a/tests/middleware/test_http.py
+++ b/tests/middleware/test_http.py
@@ -1,32 +1,27 @@
-import contextvars
-from typing import AsyncGenerator, Optional
+from typing import Any, AsyncGenerator, Callable, Optional
 
 import pytest
 
 from starlette.applications import Starlette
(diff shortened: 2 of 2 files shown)


starlette/middleware/http.py
```diff
@@ -98,6 +105,3 @@ async def wrapped_send(message: Message) -> None:
 
                 await response(scope, receive, send)
                 return
-
-            if not response_started:
-                raise RuntimeError("No response returned.")
```
adriangb, line 103: I don't think we need this here, we can just let the server handle it which is what would happen if we didn't install this middleware. I think we only do it in `BaseHTTPMiddleware` out of necessity / implementation detail.

florimondmanca:
@adriangb I think there are some changes here that you also mentioned in comments in #1691, so I'll first review and make some changes there later, then come back to reviewing this, if that's OK.

starlette/middleware/http.py
florimondmanca: True, eg. Uvicorn handles that here: https://github.com/encode/uvicorn/blob/f3c33fe7bca90326e38f66181f4623d8558571fb/uvicorn/protocols/http/h11_impl.py#L386-L389

tests/middleware/test_http.py
```diff
@@ -247,43 +259,24 @@ async def dispatch(
...
-
-class PureASGICustomMiddleware:
-    def __init__(self, app: ASGIApp) -> None:
-        self.app = app
+def test_modify_content_type(
+    test_client_factory: Callable[[ASGIApp], TestClient]
+) -> None:
+    async def dispatch(request: HTTPConnection) -> AsyncGenerator[None, Response]:
+        resp = yield
+        resp.media_type = "text/csv"
```
florimondmanca, line 267: This would be strange usage without modifying the response body, wouldn't it? Is there a situation where we'd actually do this w/o having to modify the response body and `Content-Type` header, which would require pure ASGI middleware (or the body modification API you thought about as a possible future improvement)?

adriangb: Yeah I'm not sure what the use case would be, but you are able to set the attribute, so maybe we should error in this situation instead by overriding `Response.__setattr__`?

adriangb:
I think you copied most of this over, I'll close this, thanks @florimondmanca

tests/middleware/test_http.py
florimondmanca: I'm not sure we need to be that defensive, to be honest. If we're clear enough in the documentation (which ought to be written before we consider merging this anyway, right?) that this is a "simplified response interface, allowing to inspect the status code and headers, or tweak headers", then maybe that's enough.

Though, now that I think of it — what if users _do_ want to inspect `response.body` in the "on response" part of the dispatch flow? Right now, it's empty, and actually makes no sense because this is called before we start processing any chunk of the body. Hmm.

adriangb: Yeah I don't think there's any behavior that's right or wrong here, it's just about what will be least confusing.

PR #1695, opened 2022-06-15
allow inspecting response body

adriangb:
Pretty complex, but maybe it leads to some better ideas? TLDR is move passing the response to the dispatch function until we've received the response body. This allows inspecting the response body.


adriangb:
Closing this, I don't think it's really worth it

PR #1700, opened 2022-06-22
Move BackgroundTask execution outside of middleware stack

### LINKED ISSUE #919: Background tasks don't work with middleware that subclasses `BaseHTTPMiddleware`

### LINKED ISSUE #1438: Background tasks are cancelled if the client closes connection

adriangb:
Fixes #1438, fixes #919, closes #1654, closes #1699

This will currently break backwards compatibility if someone tries to use the "new" Response implementation with an "old" Starlette app (or an app that is not Starlette, like FastAPI which copies a lot of Starlette code and will have to manually update 1 LOC). We could easily work around this with an `if "starlette.background" in scope` but it will require a couple more tests. Otherwise this should be backwards compatible.


kigawas:
It's better to add some background task tests

https://github.com/encode/starlette/pull/1699/files#diff-f4ca7144ea672966ffaaabe2fc8e7aa2230b6b029a253e3b2dc4e1d9136d5b5fR89

adriangb:
There is already a test which does not pass on `master` but passes on this branch. The test you linked also passes on `master`.

kigawas:
```
tests/middleware/test_base.py::test_background_tasks[asyncio] background task 1 started
PASSED
tests/middleware/test_base.py::test_background_tasks[trio] background task 1 started
PASSED
```

It didn't actually "pass" the test on master - the background tasks are cancelled during the `sleep(2)`.

The expected output is:
```
tests/middleware/test_base.py::test_background_tasks[asyncio] background task 1 started
background task 1 completed
background task 2 started
background task 2 completed
background task sync started
background task sync completed
PASSED
tests/middleware/test_base.py::test_background_tasks[trio] background task 1 started
background task 1 completed
background task 2 started
background task 2 completed
background task sync started
background task sync completed
PASSED
```

adriangb:
Apologies, I did not look at the print statements. In this codebase print statements are not a valid way to make a test fail.

That said, you do make a point: I had added a test for #1438 but not #919. I added a test for #919 now.

jhominal:
When I saw this PR about a week ago, I thought that this was a pretty good solution to #1438.

However, since then I have been able to think about that issue more deeply (after being prompted by another of your PRs, #1697), and I believe that #1715 also solves #1438, but keeps background tasks as being run in the request/response cycle.

I do not mean to say that "putting background tasks in a separate, app-initialized task group" is not a worthwhile change - I only mean to say that it may not be necessary to fix #1438.

tests/test_background.py
```diff
@@ -1,13 +1,76 @@
...
+
+def streaming_response_app_factory(task: BackgroundTask) -> ASGIApp:
+    async def app(scope: Scope, receive: Receive, send: Send):
+        async def stream() -> AsyncIterable[bytes]:
+            yield b"task initiated"
+
+        response = StreamingResponse(stream(), media_type="text/plain", background=task)
+        await response(scope, receive, send)
+
+    return app
```
adriangb, line 59: This is all just to get 100% code coverage with both code paths. If we removed the old code path (i.e. backgroundTask does not work without the middleware) there would be (almost) no changes to these tests and it would look a lot like https://github.com/xpresso-devs/asgi-background

adriangb:
Glad one of my messy PRs actually inspired something instead of just being noise 😅

> I do not mean to say that "putting background tasks in a separate, app-initialized task group" is not a worthwhile change

There is not TaksGroup in this PR. That's the name of the branch and that being the original idea, but I realized it was too ambitious so I toned it down (I still think a `TaskGroup` launched by a lifespan event would be the right thing to do, but not today).

> I only mean to say that it may not be necessary to fix https://github.com/encode/starlette/issues/1438.

👍 gotcha I do think there's a lot to be said for not changing other parts of the codebase in an effort to fix bugs caused by BaseHTTPMiddleware. That said, I think this is a pretty _natural_ change. I _think_ that if I looked at this in 5 years without any comments it'd still make sense. Doesn't make it a better option per-se, but if this wasn't the case I think it'd be a bad option for sure.

tests/middleware/test_base.py
```diff
@@ -206,3 +213,81 @@ async def homepage(request):
...
+    async def recv_gen() -> AsyncGenerator[Message, None]:
+        yield {"type": "http.request"}
+        disconnected.set()
+        yield {"type": "http.disconnect"}
+
+    async def send_gen() -> AsyncGenerator[None, Message]:
+        msg = yield
+        assert msg["type"] == "http.response.start"
+        await disconnected.wait()
+        raise AssertionError("Should not be called")  # pragma: no cover
```
jhominal: I would say that this sequence of events does not look natural to me, in terms of expressing what happens in #1438.

I believe the main event that should be modeled here is "the client disconnects once they have received the response body" - that means, `disconnected` should be set in `send_gen` once the last `http.response.body` has been sent. (Whereas `receive` blocks when it has no available message to send)

```suggestion
    async def recv_gen() -> AsyncGenerator[Message, None]:
        yield {"type": "http.request"}
        await disconnected.await()
        while True:
            yield {"type": "http.disconnect"}

    async def send_gen() -> AsyncGenerator[None, Message]:
        msg = yield
        if msg["type"] == "http.response.body" and not msg.get("more_body", False):
            disconnected.set()
```

tests/middleware/test_base.py
```diff
@@ -206,3 +213,81 @@ async def homepage(request):
...
+@pytest.mark.anyio
+async def test_background_tasks_client_disconnect() -> None:
+    # test for https://github.com/encode/starlette/issues/1438
+    container: List[str] = []
+
+    disconnected = anyio.Event()
+
+    async def slow_background() -> None:
+        await disconnected.wait()
+        container.append("called")
```
jhominal: I do not think that the background task itself should wait on the `disconnected` event - that places an explicit link between the ASGI events of the test and the background task, whereas the link that we seek to test (around cancellation and request processing in the presence of a `BaseHTTPMiddleware`) is implicit.

I think a `sleep(0.1)` to allow cancellation would be more appropriate.

```suggestion
    async def slow_background() -> None:
        # small delay to give BaseHTTPMiddleware a chance to cancel us
        # this is required to make the test fail prior to fixing the issue
        # so do not be surprised if you remove it and the test still passes
        await anyio.sleep(0.1)
        container.append("called")
```

jhominal (commented):
I have realized that I did not actually read your test case for #1438 in detail, so I made a few comments.

I do realize that these comments would make the tests for #1438 and #919 very close, but as I believe the underlying issue is the same, that is not wholly unexpected.

adriangb:
I ended up just removing the test for #919. I think if we end up merging one of these solutions we should probably comment on that issue saying "hey we think this is fixed and also this issue has become an agglomeration of possibly multiple bugs that have changed over time so we are closing it for now. If you are still suffering from this issue please open a new issue with a self contained example"

gnat:
For reference I do believe this is the pattern employed by sanic: Task group outside of the middleware stack. https://sanic.dev/en/guide/basics/tasks.html

Seems sound.

adriangb:
Also used by Quart:
https://github.com/pgjones/quart-trio/blob/d0b775b2dda60cba916df6989c2e44fdf14f30d1/src/quart_trio/asgi.py#L114
https://github.com/pgjones/quart-trio/blob/d0b775b2dda60cba916df6989c2e44fdf14f30d1/src/quart_trio/app.py#L253

Quart uses a TaskGroup/Nuersery created in the lifespan event. That latter implementation was the original goal of this PR, I've thought that was the right move since before knowing Quart does that, which I think indicates it is indeed the right move long term. But for today this will have to do.

PR #1701, opened 2022-06-23
Add `AsyncTemplateResponse` to Jinja2Templates

amisadmin:
Help Jinja2 make it easier and faster to use async template rendering.


yezz123 (approved):

tests/test_templates.py
```diff
@@ -32,3 +32,34 @@ def test_template_response_requires_request(tmpdir):
...
     with pytest.raises(ValueError):
         templates.TemplateResponse("", {})
+
+
+def test_async_templates(tmpdir, test_client_factory):
+    path = os.path.join(tmpdir, "index.html")
+    with open(path, "w") as file:
+        file.write("<html>Hello,{{ get_name() }}</html>")
+
+    async def get_name():
```
aminalaee, line 42: Is this needed?

starlette/templating.py
```diff
@@ -27,14 +27,14 @@ def __init__(
         self,
         template: typing.Any,
         context: dict,
+        content: typing.Optional[typing.Any] = None,
```
aminalaee:
```suggestion
        content: typing.Optional[str] = None,
```

starlette/templating.py
```diff
@@ -56,6 +56,12 @@ class Jinja2Templates:
     templates = Jinja2Templates("templates")
 
     return templates.TemplateResponse("index.html", {"request": request})
+
+    ..enable_async:
```
aminalaee, line 60: Starlette is not using Sphinx, you can write a simple docstring and maybe add it to the `docs` directory with `Asynchronous template rendering`.

aminalaee (commented):
I think this looks good, I'd wait for @tomchristie 's input on this too.

euri10:
is this not valid anymore https://github.com/encode/starlette/blob/master/docs/templates.md#asynchronous-template-rendering ?

aminalaee:
> https://github.com/encode/starlette/blob/master/docs/templates.md#asynchronous-template-rendering

Yeah I mentioned if this is going to be merged the docs should be updated.

starlette/templating.py
amisadmin: The `content` attribute in the parent `Response` is of type `Any`. I'm not sure it's necessarily of type `str` here.

tests/test_templates.py
amisadmin: Yes.

starlette/templating.py
amisadmin: Ok, thanks for your advice. The document has been updated.

docs/templates.md
```diff
@@ -88,4 +88,33 @@ Instead we'd recommend that you ensure that your endpoints perform all I/O,
...
+
+In asynchronous mode you can use 'AsyncTemplateResponse' to render asynchronous functions
+or coroutine objects in jinja2 templates.
+
+```python
+from starlette.templating import Jinja2Templates
+
+templates = Jinja2Templates(directory='templates', enable_async=True)
+
+async def get_name():
```
lovelydinosaur, line 103: I'd suggest some kind of comment noting that this'd actually only make sense in the case of an actual async operation. (Eg. an async database lookup.)

lovelydinosaur:
> is this not valid anymore https://github.com/encode/starlette/blob/master/docs/templates.md#asynchronous-template-rendering ?

Subjective. Personally I think it's probably decent advice. But. 🤷‍♂️

starlette/templating.py
aminalaee: Well in this case it is always 'str' as output of 'render' sync and async in Jinja2.
I might be missing something but we could narrow the type down here as 'str' and that would still satisfy the parent with 'Any'.
What do you think @Kludex?

docs/templates.md
amisadmin: Ok, thanks for your advice.

starlette/templating.py
```diff
@@ -56,6 +56,12 @@ class Jinja2Templates:
...
+
+    templates = Jinja2Templates("templates",enable_async=True)
```
```suggestion
    templates = Jinja2Templates("templates", enable_async=True)
```

starlette/templating.py
```diff
@@ -95,9 +101,35 @@ def TemplateResponse(
...
+        name: str,
+        context: dict,
+        status_code: int = 200,
+        headers: typing.Optional[typing.Mapping[str, str]] = None,
+        media_type: typing.Optional[str] = None,
+        background: typing.Optional[BackgroundTask] = None,
+    ) -> _TemplateResponse:
+        if "request" not in context:
+            raise ValueError('context must include a "request" key')
+        assert self.env.is_async, "please enable async rendering mode."
```
```suggestion
        if not self.env.is_async:
            raise RuntimeError("please enable async rendering mode.")
```

tests/test_templates.py
```diff
@@ -32,3 +32,34 @@ def test_template_response_requires_request(tmpdir):
...
+
+    async def homepage(request):
+        return await templates.AsyncTemplateResponse("index.html", {"request": request})
+
+    app = Starlette(
+        debug=True,
+        routes=[Route("/", endpoint=homepage)],
+    )
+    templates = Jinja2Templates(directory=str(tmpdir), enable_async=True)
+    templates.env.globals["get_name"] = get_name
```
Kludex, line 53: Why is this not added on line 46?

Kludex (commented):
We also need to tweak a bit the documentation, as the introduced section doesn't connect well with the previous section. 🤔

starlette/templating.py
Kludex: I'm not even sure if the `Optional` is needed. :sweat_smile: 

But yeah, it should be `str`. :+1: 

The return type of the render methods is `str`, and `_TemplateResponse` is a private class.

Kludex:
@amisadmin Are you still interested on working on this PR?

amisadmin:
> @amisadmin Are you still interested on working on this PR?

Yes, I am. But I'm a terrible document writer.

Kludex:
I'm going to pass on this for the time being. We can come back to this after 1.0 is released.

Thanks for the PR, and sorry for the waiting time.

PR #1706, opened 2022-06-25
Shield send "http.response.start" from cancellation

acjh:
Fixes #1634 
- Discussion #1527 
- Caused by #1157 

`RuntimeError: No response returned.` is raised in `BaseHTTPMiddleware` if request is disconnected, due to `task_group.cancel_scope.cancel()` in `StreamingResponse.__call__.<locals>.wrap` and cancellation check in `await checkpoint()` of `MemoryObjectSendStream.send`.

Let's fix this behaviour change caused by anyio integration in 0.15.0.

---

I managed to make this error reproducible in 0.14.2 by partially emulating 0.15.0 logic: https://github.com/acjh/starlette/commit/37dd8ace69993bf437712bc54a2fdb93116ab918

starlette/concurrency.py:

```diff
  async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
+     async def run(func: typing.Callable[[], typing.Coroutine]) -> None:
+         await func()
+         # (starlette 0.15.0) starlette.concurrency.run_until_first_complete `task_group.cancel_scope.cancel()`
+         for task in tasks:
+             if not task.done() and task != asyncio.current_task():
+                 task.cancel()
+
-     tasks = [create_task(handler(**kwargs)) for handler, kwargs in args]
+     tasks = [create_task(run(functools.partial(handler, **kwargs))) for handler, kwargs in args]
      (done, pending) = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
      [task.cancel() for task in pending]
-     [task.result() for task in done]
+     for task in done:
+         try:
+             task.result()
+         except asyncio.CancelledError:
+             pass
```

starlette/middleware/base.py:

```diff
  class BaseHTTPMiddleware:
      ...

      async def call_next(self, request: Request) -> Response:
          ...
-         send = queue.put
+
+         async def send(item: typing.Any) -> None:
+             await asyncio.sleep(0)  # anyio.streams.memory.MemoryObjectSendStream.send `await checkpoint()`
+             await queue.put(item)

          ...
```

PR #1710, opened 2022-06-27
Shield send "http.response.start" from cancellation (BaseHTTPMiddleware)

### LINKED ISSUE #1634: Seemingly random error `RuntimeError: No Reponse returned.`
- [x] Related discussion: https://github.com/encode/starlette/discussions/1527 

Same error though I don't know if cause is the same, we have not been able to reproduce the error in our environments.

### The issue
Our frontend is getting HTTP 500 errors from our FastAPI backend. The error is shown below. The error is seemingly random, we have not been able to reproduce it. The traceback does not lead anywhere to our own source code. It has happened on multiple different endpoint requests. We have wrapped our endpoint code in `try/except` blocks to catch any "hidden" error in our code to trigger this, but it did not catch anything, leading me to believe that this is an underlying error in the framework.

We discovered the error on FastAPI v0.71.0 (Starlette v0.17.1). 
We tried upgrading FastAPI to v0.76.0 (Starlette v0.18.0), the error became more frequent.
We have now downgraded FastAPI to v0.68.2 hoping that the error dissipates.

python v3.8
uvicorn v0.15.0

```
INFO:   xxx - "POST /api/xxx HTTP/1.1" 500
ERROR:  Exception in ASGI application
  File "/xxx/venv/lib64/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 375, in run_asgi
  File "/xxx/venv/lib64/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
  File "/xxx/venv/lib64/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
  File "/xxx/venv/lib64/python3.8/site-packages/starlette/applications.py", line 112, in __call__
  File "/xxx/venv/lib64/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/xxx/venv/lib64/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
  File "/xxx/venv/lib64/python3.8/site-packages/starlette/middleware/base.py", line 63, in __call__
    response = await self.dispatch_func(request, call_next)
  File "/xxx/venv/lib64/python3.8/site-packages/prometheus_fastapi_instrumentator/instrumentation.py", line 172, in dispatch_middleware
    raise e from None
  File "/xxx/venv/lib64/python3.8/site-packages/prometheus_fastapi_instrumentator/instrumentation.py", line 169, in dispatch_middleware
  File "/xxx/venv/lib64/python3.8/site-packages/starlette/middleware/base.py", line 45, in call_next
RuntimeError: No Reponse returned.
```

acjh:
Fixes #1634 
- Discussion #1527 
- Caused by #1157 

`await recv_stream.receive()` will raise `anyio.EndOfStream` if request is disconnected, due to:
- `task_group.cancel_scope.cancel()` in `StreamingResponse.__call__.<locals>.wrap` and
- cancellation check in `await checkpoint()` of `MemoryObjectSendStream.send`,

and then `RuntimeError: No response returned.` will be raised in `BaseHTTPMiddleware`.

Let's shield send "http.response.start" from cancellation, since the message is ready to be sent to the receiver.

This is an alternative implementation of #1706 in `BaseHTTPMiddleware` instead of `StreamingResponse`.
We should not force the shielding in `StreamingResponse`, since the cancellation check is an intended feature of `MemoryObjectSendStream`. `BaseHTTPMiddleware`, which uses both, should be responsible for the compatibility.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -30,9 +30,11 @@ class BaseHTTPMiddleware:
 
             async def send(msg: Message) -> None:
                 # Shield send "http.response.start" from cancellation.
-                # Otherwise, `await recv_stream.receive()` will raise `anyio.EndOfStream` if request is disconnected,
-                # due to `task_group.cancel_scope.cancel()` in `StreamingResponse.__call__.<locals>.wrap`
-                # and cancellation check in `await checkpoint()` of `MemoryObjectSendStream.send`,
+                # Otherwise, `await recv_stream.receive()` will raise
+                # `anyio.EndOfStream` if request is disconnected,
+                # due to `task_group.cancel_scope.cancel()` in
+                # `StreamingResponse.__call__.<locals>.wrap` and cancellation check
+                # during `await checkpoint()` in `MemoryObjectSendStream.send`,
                 # and then `RuntimeError: No response returned.` will be raised below.
                 shield = msg["type"] == "http.response.start"
                 with anyio.CancelScope(shield=shield):
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -31,11 +31,13 @@ class BaseHTTPMiddleware:
             async def send(msg: Message) -> None:
                 # Shield send "http.response.start" from cancellation.
                 # Otherwise, `await recv_stream.receive()` will raise
-                # `anyio.EndOfStream` if request is disconnected,
+                # `anyio.EndOfStream` if the connection is disconnected,
                 # due to `task_group.cancel_scope.cancel()` in
-                # `StreamingResponse.__call__.<locals>.wrap` and cancellation check
-                # during `await checkpoint()` in `MemoryObjectSendStream.send`,
-                # and then `RuntimeError: No response returned.` will be raised below.
+                # `StreamingResponse.__call__.<locals>.wrap`
+                # and cancellation check during `await checkpoint()` in
+                # `MemoryObjectSendStream.send`.
+                # This would trigger the check we have in this middleware resulting in
+                # `RuntimeError: No response returned.` being raised below.
                 shield = msg["type"] == "http.response.start"
                 with anyio.CancelScope(shield=shield):
                     await send_stream.send(msg)
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -4,7 +4,7 @@ import anyio
 
 from starlette.requests import Request
 from starlette.responses import Response, StreamingResponse
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 RequestResponseEndpoint = typing.Callable[[Request], typing.Awaitable[Response]]
 DispatchFunction = typing.Callable[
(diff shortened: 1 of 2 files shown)


starlette/middleware/base.py
```diff
@@ -28,12 +28,22 @@ async def call_next(request: Request) -> Response:
             app_exc: typing.Optional[Exception] = None
             send_stream, recv_stream = anyio.create_memory_object_stream()
 
+            async def send(msg: Message) -> None:
+                # Shield send "http.response.start" from cancellation.
+                # Otherwise, `await recv_stream.receive()` will raise `anyio.EndOfStream` if request is disconnected,
+                # due to `task_group.cancel_scope.cancel()` in `StreamingResponse.__call__.<locals>.wrap`
+                # and cancellation check in `await checkpoint()` of `MemoryObjectSendStream.send`,
+                # and then `RuntimeError: No response returned.` will be raised below.
```
adriangb: thank you for the comment explaining what's going on! I shortened the lines to fix listing and did some minor rewording

```suggestion
                # Shield send "http.response.start" from cancellation.
                # Otherwise, `await recv_stream.receive()` will raise `anyio.EndOfStream`
                # if the connection is disconnected,
                # due to `task_group.cancel_scope.cancel()` in
                # `StreamingResponse.__call__.<locals>.wrap`
                # and cancellation check in `await checkpoint()` of
                # `MemoryObjectSendStream.send`
                # This triggers the check we have in this middleware resulting in
                # `RuntimeError: No response returned.` will be raised below.
```

adriangb:
I tried my patch to `base.py` from https://github.com/encode/starlette/pull/1706#issuecomment-1166342235 on this branch and the current test (in this branch, not the patch) also passes (I can confirm the test fails on `master` so it's not just that it passes on any branch). I do think your solution is nicer in some sense, but does it fix any edge cases that the raise an exception solution doesn't?

acjh:
I don't think this solution fixes any additional edge cases.

Both solutions pretty much target this specific edge case, where `StreamingResponse` does `listen_for_disconnect` and handles the disconnection in a way that is not compatible with how `MemorySendStream` is used in `BaseHTTPMiddleware`.

adriangb:
If there are no pros to doing cancellation, I would prefer to just raise an exception, but that's just my opinion and I'm biased, so I'd like others to weigh in.

Otherwise, this PR looks good and I'd be happy to move forward with it, thank you for the contribution and your persistence @acjh !

acjh:
> If there are no pros to doing cancellation

I believe you mean *not* doing cancellation. More accurately, it's temporarily shielding from cancellation.

> I would prefer to just raise an exception

I'm totally fine with closing this PR in favour of that, btw. 😄 

This PR simply handles the compatibility between `StreamingResponse` and usage of anyio memory streams.
Raising an exception takes control of the flow by forcing exception handling, ignoring the streams' behaviour.

adriangb:
> > If there are no pros to doing cancellation
> 
> I believe you mean _not_ doing cancellation. More accurately, it's temporarily shielding from cancellation.

Yep sorry for being unclear

> > I would prefer to just raise an exception
> 
> I'm totally fine with closing this PR in favour of that, btw. 😄
> 
> This PR simply handles the compatibility between `StreamingResponse` and usage of anyio memory streams. Raising an exception takes control of the flow by forcing exception handling, ignoring the streams' behaviour.

Let's just wait to see what others think, I'm also fine with either 😄

acjh:
Sure. Thank you for your patience and prompt responses!

scottherlihy:
@acjh @adriangb from my perspective I do not think an exception should be thrown because I don't view this as an error condition and I'm curious what your thoughts are. We are encountering this while utilizing a callback url pattern with long running tasks. The client is posting a request, it includes a callback url where to send the response, as such the client disconnects immediately. We have all the information we need to send the response and don't need an active client connection yet this error indeterminately causes dropping of responses via the error documented above. We can certainly catch and handle an error thrown, but I don't think there should be an expectation that the client needs to be connected through the lifetime of the process. Does this make sense or is there something I'm missing?

acjh:
Are you using background tasks? This PR does not fix #1438.

scottherlihy:
@acjh we are, I will read more about that other ticket, thank you

PR #1711, opened 2022-06-28
fix type annotations in responses.py

adriangb:
Adding/fixing a couple minor typing nits in `responses.py`

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -112,7 +112,7 @@ class Response:
         httponly: bool = False,
         samesite: typing.Optional[Literal["lax", "strict", "none"]] = "lax",
     ) -> None:
-        cookie: http.cookies.BaseCookie = http.cookies.SimpleCookie()
+        cookie: "http.cookies.BaseCookie[str]" = http.cookies.SimpleCookie()
         cookie[key] = value
         if max_age is not None:
             cookie[key]["max-age"] = max_age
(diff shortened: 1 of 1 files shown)


starlette/responses.py
```diff
@@ -112,7 +112,7 @@ def set_cookie(
         httponly: bool = False,
         samesite: typing.Optional[Literal["lax", "strict", "none"]] = "lax",
     ) -> None:
-        cookie: http.cookies.BaseCookie = http.cookies.SimpleCookie()
+        cookie: "http.cookies.BaseCookie[str]" = http.cookies.SimpleCookie()
```
Kludex, line 115: Why is this correct? I can see it fixes the issue with pyright, but I don't understand how a single argument works, as `BaseCookie` is a `dict` 🤔

adriangb: Are you asking why `str` is the correct type? Both `key` and `value` are `str` (from the function signature) and `str` made type checker happy so pretty sure `str` is correct.

Kludex: As pointed on our conversation outside GitHub... It's because of the definition on typeshed: https://github.com/python/typeshed/blob/544c7c963333a4a415b951f13a6a6bbec2421161/stdlib/http/cookies.pyi#L54

Thanks @adriangb 😗

Kludex (approved):

PR #1712, opened 2022-06-28
fix type annotations in requests.py

adriangb:
Minor fixes / addition to annotations in requests.py


starlette/requests.py
```diff
@@ -60,7 +59,7 @@ class ClientDisconnect(Exception):
     pass
 
 
-class HTTPConnection(Mapping):
+class HTTPConnection(typing.Mapping[str, typing.Any]):
```
adriangb, line 62: I remember trying this before and it not working. I think it broke on Python 3.6. But now that we've dropped 3.6 we can remove 1 import and get better typing as well 🥳

starlette/requests.py
```diff
@@ -249,7 +248,8 @@ async def form(self) -> FormData:
                 parse_options_header is not None
             ), "The `python-multipart` library must be installed to use form parsing."
             content_type_header = self.headers.get("Content-Type")
-            content_type, options = parse_options_header(content_type_header)
+            content_type: bytes
+            content_type, _ = parse_options_header(content_type_header)
```
Kludex, line 252: What's the problem with `parse_options_header`?

adriangb: It's untyped. It also returns a tuple and the second variable was an unused variable.

Kludex (approved):
For reference, `python-multipart` PR to drop Python 3: https://github.com/andrew-d/python-multipart/pull/42.

The above would have avoided the `content_type: bytes` if annotation was added on `python-multipart`.

PR #1727, opened 2022-07-02
Test StreamingResponse behavior on receiving http.disconnect

jhominal:
- Initially raised as discussion #1726 

This issue simply adds a test to check that `StreamingResponse` will interrupt its streaming when the client disconnects.

### DIFF
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -391,3 +391,34 @@ def test_streaming_response_known_size(test_client_factory):
     client: TestClient = test_client_factory(app)
     response = client.get("/")
     assert response.headers["content-length"] == "10"
+
+
+@pytest.mark.anyio
+async def test_streaming_response_stops_if_receiving_http_disconnect():
+    streamed = 0
+
+    disconnected = anyio.Event()
+
+    async def receive_disconnect():
+        await disconnected.wait()
+        return {"type": "http.disconnect"}
+
+    async def send(message):
+        nonlocal streamed
+        if message["type"] == "http.response.body":
+            streamed += len(message.get("body", b""))
+            # Simulate disconnection after download has started
+            if streamed >= 16:
+                disconnected.set()
+
+    async def stream_indefinitely():
+        while True:
+            # Need a sleep for the event loop to switch to another task
+            await anyio.sleep(0)
+            yield b"chunk "
+
+    response = StreamingResponse(content=stream_indefinitely())
+
+    with anyio.move_on_after(1) as cancel_scope:
+        await response({}, receive_disconnect, send)
+    assert not cancel_scope.cancel_called, "Content streaming should stop itself."


tests/test_responses.py
```diff
@@ -391,3 +391,34 @@ def test_streaming_response_known_size(test_client_factory):
...
+
+    async def send(message):
+        nonlocal streamed
+        if message["type"] == "http.response.body":
+            streamed += len(message.get("body", b""))
+            # Simulate disconnection after download has started
+            if streamed >= 16:
+                disconnected.set()
+
+    async def _stream_indefinitely():
```
adriangb:
```suggestion
    async def stream_indefinitely():
```

no reason for this to be "private"

tests/test_responses.py
```diff
@@ -391,3 +391,34 @@ def test_streaming_response_known_size(test_client_factory):
...
+        while True:
+            # Need a sleep for the event loop to switch to another task
+            await anyio.sleep(0)
+            yield b"chunk "
+
+    response = StreamingResponse(content=_stream_indefinitely())
```
adriangb:
```suggestion
    response = StreamingResponse(content=stream_indefinitely())
```

tests/test_responses.py
```diff
@@ -391,3 +391,34 @@ def test_streaming_response_known_size(test_client_factory):
...
+
+    with anyio.move_on_after(1) as cancel_scope:
+        await response({}, receive_disconnect, send)
+    assert not cancel_scope.cancel_called, "Content streaming should stop itself."
```
adriangb, line 424: this may just be my personal opinion, but I'd just let the test hang, feel free to ignore (I think you'll need a `pragma: no cover` if you do)

```suggestion
    await response({}, receive_disconnect, send)
```

adriangb (changes requested):

tests/test_responses.py
Kludex: I'm failing to see why you'd prefer that 🤔

adriangb: I try to optimize tests for the passing case, i.e. not put in code that handles the failing case, but like I said just my preference. Now realizing too that no `pragma` is needed here, so really what's here is totally fine 😄

tests/test_responses.py
jhominal: Yes, these underscores were really a remnant of the previous form of the test, where the function was defined outside of the test 😅 

Thank you for catching it.

adriangb (approved):
This looks good to me, I love me a test documenting existing behavior!

tests/test_responses.py
jhominal: Deadlocks in tests are, in my opinion, a pain in the neck:
 * they push the test execution on CI up to the timeout (which can often be 10 to 30 minutes), even for quick unit tests;
 * they consume CI resources during that time (which is not irrelevant given that free CI options generally limit the amount of computing resources that can be used);
 * they prevent the generation of a test report - which complicates the "fix the test" workflow;

I believe adding a simple, small line is well worth ensuring that this test will never be the cause of such issues. (It really makes the "test the regression" workflow easier, too!)

adriangb: Yeah all fair arguments!

Kludex (approved):
Thanks! 🙏

PR #1732, opened 2022-07-03
type config with `None` default as `str | None` instead of `Any`

chdsbd:
A config with a default of `None` now returns `str | None` instead of `Any`.

```python
config = Config()
PRIVATE_KEY = config("PRIVATE_KEY", default=None)

# before
reveal_type(PRIVATE_KEY) # Revealed type is "Any"

# after
reveal_type(PRIVATE_KEY) # Revealed type is "Union[builtins.str, None]"
```

Initially raised as discussion https://github.com/encode/starlette/discussions/1730

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ databases[sqlite]==0.5.5
 flake8==3.9.2
 isort==5.10.1
 mypy==0.961
+typing_extensions==4.2.0
 types-requests==2.26.3
 types-contextvars==2.4.7
 types-PyYAML==6.0.4
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "anyio>=3.4.0,<5",
-        "typing_extensions>=3.10.0; python_version < '3.10'",
+        "typing_extensions>=4.2.0; python_version < '3.10'",
     ],
     extras_require={
         "full": [
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "anyio>=3.4.0,<5",
-        "typing_extensions>=4.2.0; python_version < '3.10'",
+        "typing_extensions>=3.10.0; python_version < '3.10'",
     ],
     extras_require={
         "full": [
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1,10 +1,66 @@
+from __future__ import annotations
 import os
 from pathlib import Path
+from typing import Any
 
 import pytest
 
 from starlette.config import Config, Environ, EnvironError
 from starlette.datastructures import URL, Secret
+
+
+def is_str(x: list[str]) -> None:
+    ...
+
+
+def is_bool(x: list[bool]) -> None:
+    ...
+
+
+def is_int(x: list[int]) -> None:
+    ...
+
+
+def is_bool_or_none(x: list[bool | None]) -> None:
+    ...
+
+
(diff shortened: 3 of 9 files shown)


tests/test_config.py
```diff
@@ -1,12 +1,80 @@
...
 from pathlib import Path
+from typing import Any
 
 import pytest
 
 from starlette.config import Config, Environ, EnvironError
 from starlette.datastructures import URL, Secret
 
 
+def test_config_types() -> None:
```
chdsbd, line 12: I added some tests for the type annotations. Open to feedback on how to make these nicer

tests/test_config.py
```diff
@@ -1,12 +1,80 @@
...
+    is_str(STR_CAST_STR)
+
+    BOOL = [config("BOOL", cast=bool)]
+    is_bool(BOOL)
+    BOOL_DEFAULT = [config("BOOL_DEFAULT", cast=bool, default=False)]
+    is_bool(BOOL_DEFAULT)
+    BOOL_NONE = [config("BOOL_NONE", cast=bool, default=None)]
+    is_bool_or_none(BOOL_NONE)
+
+    # defaults with non-bool types are allowed by our type annotations.
```
chdsbd: See https://github.com/encode/starlette/discussions/1731

In a future PR, I'd like to require that the default be either `None` or the same type as the `cast`.

tests/test_config.py
Kludex: Maybe parametrize those:
```python
STR = [config("STR")]
is_str(STR)
```
And move the function definitions outside this test function (above) 🤔

Kludex (commented):
I'm not sure if we should take an approach in which we do assertions instead of depending on the type checker.

I've asked around how people do this, not sure if helpful, but there are some ideas: https://twitter.com/marcelotryle/status/1544374655481532422

setup.py
```diff
@@ -38,7 +38,7 @@ def get_long_description():
     include_package_data=True,
     install_requires=[
         "anyio>=3.4.0,<5",
-        "typing_extensions>=3.10.0; python_version < '3.10'",
+        "typing_extensions>=4.2.0; python_version < '3.10'",
```
Kludex: Please undo this 😗 
```suggestion
        "typing_extensions>=3.10.0; python_version < '3.10'",
```

tests/test_config.py
```diff
@@ -1,12 +1,43 @@
...
+from typing import Any, Optional
 
 import pytest
+from typing_extensions import assert_type
 
 from starlette.config import Config, Environ, EnvironError
 from starlette.datastructures import URL, Secret
 
 
+def test_config_types() -> None:
```
Kludex, line 12: Maybe adding a docstring explaining a bit that we are relying on mypy? 🙏

Kludex (commented):
I added some comments. I'm overall fine with this PR. 🙏

requirements.txt
```diff
@@ -9,6 +9,7 @@ databases[sqlite]==0.5.5
 flake8==3.9.2
 isort==5.10.1
 mypy==0.961
+typing_extensions==4.2.0
```
chdsbd, line 12: I added typing_extensions here because we need a newer version that contains assert_type

Kludex: Thanks 🙏

Kludex (approved):
Thanks @chdsbd 🙌

I'll let merge this in a few days if there are no further comments from other members. 😁

tests/test_config.py
```diff
@@ -1,12 +1,46 @@
+from __future__ import annotations
+
```
Kludex:
```suggestion
```

PR #1746, opened 2022-07-09
Remove default http scheme in url_for

aebrahim:
A more sensible default is a link starting with `//` which will use the same scheme as the current page.

The starting point for contributions should usually be [a discussion](https://github.com/encode/httpx/discussions)


Kludex:
So... How can I help? 😗

Kludex:
Closing it as author seems away, and there was no discussion about the topic.

PR #1765, opened 2022-07-14
✨ Remove body from requests with a status code that don't allow it

### LINKED ISSUE #1764: RuntimeError: Response content longer than Content-Length in Uvicorn, for status codes without body (e.g. 204, 304), in responses with Body
## Related discussion

- [x] Initially raised as discussion https://github.com/encode/starlette/discussions/1635

I think this is worth as an issue/bug in Starlette, and I'm gonna work on it.

## Description

When a Starlette application sends a response with a status code that would have no content, e.g. 204, 304, but the response actually has content, it is accepted by Starlette as valid, the code looks valid, and there's no exception raised by Starlette, it sends the response ASGI message to the server like that.

But Uvicorn takes that as an erroneous message and raises an error of:

```
RuntimeError: Response content longer than Content-Length
```

So, an application and code that would look fine in Starlette, and Starlette would accept as valid, results in an error in Uvicorn.

This means that Starlette is accepting some code from the user as valid, but sending an invalid message to Uvicorn.

## Minimal Example

Here's a minimal example to reproduce the issue with Starlette:

```Python


def endpoint(request: Request):
    return Response("null", status_code=204)


app = Starlette(
    routes=[Route("/", endpoint=endpoint)],
)
```

Then call the endpoint, e.g.:

```console
$ curl localhost:8000
```

The problem is even more notorious when using FastAPI, as the same response from above is generated automatically because a function implicitly always returns `None` if nothing else is returned, and FastAPI will use a `JSONResponse` by default, so it will try to send the JSON string `"null"`:

```Python
from fastapi import FastAPI

app = FastAPI()


@app.get("/", status_code=204)
def endpoint():
    print("Oh, hi mark")
```

## On Uvicorn

Uvicorn checks if there's a `content-length` header and uses it to set the `expected_content_length`, if there's none, it is not modified and keeps its default of `0`.

Then that default of `0` for `expected_content_length` is checked here, and that's where the error is raised: https://github.com/encode/uvicorn/blob/fd4386fefb8fe8a4568831a7d8b2930d5fb61455/uvicorn/protocols/http/httptools_impl.py#L532-L533

The only case where that is not affected is when `chunked_encoding` is used. And that is used by default when the `content-length` is not set **only if** the status code is not one of the status codes without body: https://github.com/encode/uvicorn/blob/fd4386fefb8fe8a4568831a7d8b2930d5fb61455/uvicorn/protocols/http/httptools_impl.py#L498-L505

So, from Uvicorn's point of view, Starlette is sending a bogus message.

## Next steps

I think this has to be solved on Starlette, not on Uvicorn.

I think the sensible thing would be to remove the body of a request with a status code that doesn't contain a body, Uvicorn won't send that body either way, it will just error out.

Another option is to raise an exception to tell the user that they can't provide a body for a status code that doesn't contain a body, but it should be done in Starlette, not in Uvicorn. And if that was the approach chosen, then it would make sense to add a lot of typing `@overload`s to the `__init__` methods of the request classes with the `Literal` for each status code that would raise an error, and making the `content` as exclusively `None` in those overloads, to make those errors show up when editing and checking the code, not only at runtime (in production :scream:). This second option would be a lot more code and it wouldn't solve the problem for FastAPI, I would still have to do the first option above in FastAPI.

tiangolo:
✨ Remove body from requests with a status code that don't allow it

This will solve/prevent errors in Uvicorn caused by Starlette sending invalid responses.

Closes https://github.com/encode/starlette/issues/1764, also probably solves/related to https://github.com/encode/starlette/discussions/1635

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -35,6 +35,10 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
+def _is_body_allowed_for_status_code(status_code: int) -> bool:
+    return not (status_code < 200 or status_code in {204, 304})
+
+
 class Response:
     media_type = None
     charset = "utf-8"
@@ -51,7 +55,11 @@ class Response:
         if media_type is not None:
             self.media_type = media_type
         self.background = background
-        self.body = self.render(content)
+        self.body = (
+            self.render(content)
+            if _is_body_allowed_for_status_code(self.status_code)
+            else b""
+        )
         self.init_headers(headers)
 
     def render(self, content: typing.Any) -> bytes:
@@ -81,7 +89,7 @@ class Response:
         if (
             body is not None
             and populate_content_length
-            and not (self.status_code < 200 or self.status_code in (204, 304))
+            and _is_body_allowed_for_status_code(self.status_code)
         ):
             content_length = str(len(body))
             raw_headers.append((b"content-length", content_length.encode("latin-1")))
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -358,6 +358,22 @@ def test_empty_204_response(test_client_factory):
     assert "content-length" not in response.headers
 
 
+def test_empty_204_response_trims_content(test_client_factory):
+    app = Response(content="hi", status_code=204)
+    client: TestClient = test_client_factory(app)
+    response = client.get("/")
+    assert "content-length" not in response.headers
+    assert response.content == b""
+
+
+def test_empty_json_response_204_response_trims_content(test_client_factory):
+    app = JSONResponse(content="hi", status_code=204)
+    client: TestClient = test_client_factory(app)
+    response = client.get("/")
+    assert "content-length" not in response.headers
+    assert response.content == b""
+
+
 def test_non_empty_response(test_client_factory):
     app = Response(content="hi")
     client: TestClient = test_client_factory(app)


jhominal:
I do not think that I like filtering the request body in that way.

What I mean, the `Response` object has contradictory instructions, and I do not think Starlette should decide which of the status code or the content specified by the user should be sacrificed. Personally, I would be more comfortable with Starlette raising an exception in that case - that keeps Starlette at the conceptual HTTP level, where I feel it was decided we would let it stay.

For the related idea of adding a lot of `typing` overloads using `typing.Literal` on the status code to signal that content cannot be set in that case, I do not know that it would help beyond very simple code patterns.

## About the underlying issue in FastAPI

I know that you want to solve a FastAPI issue where an exception is raised if someone writes a minimal endpoint that returns a 204 status code - but in my view, that issue is the manifestation of an impedance mismatch between:
 * python functions - where `return` and `return None` are indistinguishable;
 * and HTTP responses - where "HTTP response has JSON content `null`" is distinct from "no HTTP response content";

As it is FastAPI that owns the process of transforming regular Python return values into JSON responses, I think that it is FastAPI that should own the handling of that case.

My suggestion would be for FastAPI to replace the default for that transformation from `JSONResponse` into a FastAPI-specific subclass of `starlette.responses.Response` (e.g. call it `ImplicitResponse` or `ImplicitJSONResponse`), that would:
 * behave like `JSONResponse` if the status code allows a response body;
 * set an empty response body if the status code does not allow a response body *and* the value returned by the function was `None`;
 * raise an error, or filter the content (probably with at least a warning), if the status code does not allow a response body and the value returned by the function was not `None`;

That approach also allows more future flexibility for FastAPI, such as adding a flag on the endpoint on whether `None` should be translated as "`null` JSON response" or as "HTTP response without content", even when the status code allows for a response body. (E.g. I personally had the case where I wanted to return an empty response body with a 202 status code.)

tiangolo:
Thanks for the feedback @jhominal.

On the FastAPI side, I already solved it and merged the PR there.

I'm gonna wait to hear more feedback here, hopefully @tomchristie will have the chance to chime in.

Kludex:
I'll check this later, but if I recall correctly, it was a conscious decision.

Maybe the reason is in one of those links:
- https://github.com/encode/starlette/pull/1270
- https://github.com/encode/starlette/issues/1178
- https://github.com/encode/starlette/pull/1397
- https://github.com/encode/starlette/issues/1282

I'll be back on this later.

tiangolo:
@Kludex found the previous conversation where it seems this was decided: https://gitter.im/encode/community?at=61d82b1a82a4667b2563bfc3

I'm gonna wait and see if @tomchristie has any other opinion. Or if it should be an exception for the user to know that they created an invalid response object (that Uvicorn will reject) or something like that.

florimondmanca:
I read that discussion. I'm kind of on the fence, because the [HTTP/1.1 spec](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5) says this about `204 No Content`:

> A 204 response is terminated by the first empty line after the header fields because it cannot contain a message body.

On the other hand, I agree with @jhominal that this seems something that's much more possible in FastAPI than Starlette.

Starlette does the right thing by default, because it uses `b""` as the default body (whereas FastAPI passes that anything the view returns through `JSONResponse` by default). So if you write `return PlainTextResponse(status_code=204)` then there's no body returned.

To put oneself in the situation akin to the one which FastAPI shoot themselves with quite frequently (myself included), you'd have to explicitly write `return PlainTextResponse(b"Some content", status_code=204)`, which, if knowing about the HTTP spec, should look suspicious.

As an illustration, I've never found myself in a "204 with body" when using solely Starlette, whereas this happened sometimes in FastAPI due to the issue which got fixed by https://github.com/tiangolo/fastapi/pull/5145. (I also think it might have been more structured to tweak the view -> JSONResponse transformation code into a FastAPI-specific response implementation that takes care of the impendance mismatch which @jhominal described. But I don't know the FastAPI codebase enough to really tell.)

Isn't the FastAPI fix enough? Does FastAPI _need_ this change in Starlette to work as expected?

lovelydinosaur:
I'd mostly likely agree with @florimondmanca's assessment here.

Starlette forces the user to be explicit about the content and the status code, and *yes* there are some broken things that a user could choose to do with that, but they'd have to actively and explicitly do so.

tiangolo:
Thanks for the feedback everyone! Yep, I get it and it makes sense.

I already solved the problem on the FastAPI side, so having it here in Starlette or not won't affect that. 🚀  But I just felt the "duty" to bring it up here and handle it here (if there was anything to handle).

Knowing that the conclusion and decision is to leave it as it is and that it's intentional, I'll close this PR (and the issue) now.

Thanks everyone for the feedback and discussion! 🍰 ☕

Issue #1336, opened 2021-11-15
Handling of `scope["root_path"]` and `scope["path"]` differs from ASGI spec


silane:
### Checklist

- [x] The bug is reproducible against the latest release and/or `master`.
- [x] There are no similar issues or pull requests to fix it yet.

### Describe the bug

Accoring to https://github.com/django/asgiref/issues/229#issuecomment-765583185, `scope["path"]` includes the `scope["root_path"]`.
So if root_path is `/foo` and request for `/foo/bar` has come, the upstream application servers (such as uvicorn) are passing `scope["path"] == "/foo/bar" and scope["root_path"] == "/foo"`, which is correct behavior. 

But starlette does not conform this. It assumes `scope["path"]` is a remainder of stripping `scope["root_path"]` from original request path. In this case, `scope["path"] == "/bar" and scope["root_path"] == "/foo"`. This ASGI incompatible assumption can be seen [here](https://github.com/encode/starlette/blob/6af5c515e0a896cbf3f86ee043b88f6c24200bcf/starlette/routing.py#L368) or [here](https://github.com/encode/starlette/blob/6af5c515e0a896cbf3f86ee043b88f6c24200bcf/starlette/datastructures.py#L23) and many other places.

The following "To reproduce" section shows just one aspect of this wrong assumption.

### To reproduce

1. Save `main.py` as follows.
```python
from starlette.applications import Starlette
from starlette.responses import PlainTextResponse
from starlette.routing import Route, Mount


class Bar:
    async def __call__(self, scope, receive, send):
        await PlainTextResponse(f"bar {scope['root_path']=} {scope['path']=}")(scope, receive, send)

class FooBar:
    async def __call__(self, scope, receive, send):
        await PlainTextResponse(f"foobar {scope['root_path']=} {scope['path']=}")(scope, receive, send)

routes =[
    Route('/bar', endpoint=Bar()),
    Mount('/foo', routes=[
        Route('/bar', endpoint=FooBar())
    ])
]
app = Starlette(routes=routes)
```

2. Run `uvicorn main:app --port 8000 --root-path /foo`
3. Access by `curl http://localhost:8000/foo/bar`

### Expected behavior

Receives
```
bar scope['root_path']='/foo' scope['path']='/foo/bar'
```

### Actual behavior

Receives
```
foobar scope['root_path']='/foo/foo' scope['path']='/bar'
```

Some points here are,
[...]

PR #1774, opened 2022-07-22
[WIP] Handle scope["root_path"] and scope["path"] as per ASGI spec 

frankie567:
- [X] Initially raised as issue #1336 

## Rationale

This is an exploratory work as how to solve issue #1336. The issue details the problem quite well but in a nutshell:

* The issue arises when using a `Mount` router alongside the `root_path` option.
* We consider that `path` is the remainder of the whole path striped from `root_path`. As pointed out [here](https://github.com/django/asgiref/issues/229#issuecomment-765583185), this is not how the ASGI spec was imagined: `path` *should* contain `root_path`.
* As a consequence, every route matching is done assuming the `root_path* is *not present* in the path.
* Besides, we use this behavior to  recursively handle `Mount` routers, by adding the router prefix on `root_path` and stripping `path` until we match the single route:

https://github.com/encode/starlette/blob/e3de71f1433568a3f7ac3dc69ef86cf4f72f5651/starlette/routing.py#L378-L390

Notice that we define an `app_root_path` to keep track of the original `root_path`.

The way I understand it, we should keep both `path` and `root_path` *untouched*. It's a shame of course because we won't be able to handle `Mount` elegantly as we do now.

## Current work status

First of all, I added a unit test to reproduce the issue.

Then, I've tried a first approach to solve it which consists of two things:

* Instead of using `path` and `root_path` directly, use internal variables `current_path` and `current_root_path` (admittedly, this is a very uninspired naming) to allow our recursive approach
* Take care of stripping the `root_path` from the `path` so we can match properly, both in `Route` and `Mount`

This solves the specific problem but breaks other parts of the code. Obviously, every parts of the routing which makes use of the `path` will need to account for this `root_path`, e.g. in `WebSocketRoute` or in `Router`, in [the part where we make slash redirects](https://github.com/encode/starlette/blob/e3de71f1433568a3f7ac3dc69ef86cf4f72f5651/starlette/routing.py#L694-L707).

An idea I had to tackle this would be to have a common `prepare_scope` which could be called by the Route classes to prepare the path, add the "internal" variables, while keeping the original `path` and `root_path` untouched.
 
Before going further, I would be interested in your views and opinions about this so we can take the right path.

Cheers!

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -234,7 +234,11 @@ class Route(BaseRoute):
 
     def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
         if scope["type"] == "http":
-            match = self.path_regex.match(scope["path"])
+            root_path = scope.get("current_root_path", scope.get("root_path", ""))
+            path = scope.get(
+                "current_path", re.sub(r"^" + root_path, "", scope["path"])
+            )
+            match = self.path_regex.match(path)
             if match:
                 matched_params = match.groupdict()
                 for key, value in matched_params.items():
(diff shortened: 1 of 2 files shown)


Kludex:
@frankie567 Do you mind if I keep this open?

frankie567:
Oh yes, sure, @Kludex! I closed it in order to not keep too much stale things, but definitely willing to work on this when needed 🙂

Kludex:
Sorry for the long waiting time.

I find this too complex, and I need time to sit, and get more knowledge around it.

frankie567:
No worries 🙂 Feel free to ping me when needed!

Kludex:
@frankie567 I know that I said that I'd like to keep this around, but I'm going to close it for now (hope it's fine).

In any case, I'm sorry for the long time to do something here.

frankie567:
No problem, @Kludex 😄 Keep up the great work!

PR #1789, opened 2022-08-04
Added support for middleware as: `func(request, next)`

protibimbok:
[Add support for per route middleware](https://github.com/encode/starlette/discussions/1781)

## Route Middlewares
The route middlewares behaves more like an endpoint than a middleware.
These middlewares receive two parameters `request: Request` and `next: async function`

```python
async def check_admin(request: Request, next):
    if request.headers['Authorization'] == 'I am admin':
        request.isAdmin = True
        request.perms = ['a', 'b', 'c']
        return await next()
    else:
        return JSONResponse({'error':'Unauthorized access'})


async def check_perm(request: Request, next):
    if request.isAdmin and 'c' in request.perms:
        response = await next() # you can now do stuff with the response
        if type(response) == str:
            response = Response(response, media_type="text/plain")
        elif type(response) == dict:
            response = JSONResponse(response)

        return response
    else:
        return JSONResponse({'error':'Unauthorized access'})



routes = [
    Route("/admin", endpoint = admin_page, middlewares = check_admin),
    Route(
        "/admin/secret", endpoint = admin_secret,
        middlewares = [check_admin, check_perm] # Middlewares will be ran in the same order
    )
]

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ venv*/
 .python-version
 build/
 dist/
+.vscode
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -146,6 +146,45 @@ routes = [
 app = Starlette(routes=routes)
 ```
 
+## Route Middlewares
+The route middlewares behaves more like an endpoint than a middleware.
+These middlewares receive two parameters `request: Request` and `next: async function`
+
+```python
+async def check_admin(request: Request, next):
+    if request.headers['Authorization'] == 'I am admin':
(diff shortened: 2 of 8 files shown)


Kludex:
Thanks for the PR!

- This was discussed on #1286, then the idea was reworked on #1464, and finally the idea was _rereworked_ on #1649.

Also, this PR is more complex, and breaks how we propose to structure Starlette: blocks of ASGI applications.

For that reason, I'll be closing this PR.

PR #1799, opened 2022-08-09
Update Github Actions for Python 3.11.0-rc.1

aminalaee:
If `3.11-dev` includes the RC release we can close this.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
 
     steps:
       - uses: "actions/checkout@v3"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-suite.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11.0-rc.1"]
```
aminalaee, line 17: Not really sure if `3.11-dev` will include `rc` releases or not. And the `3.11.0-rc.1` is not yet available in GH actions.

Kludex: `-dev` includes all releases, not only alpha and beta.

Kludex: ref.: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python

aminalaee: It's kind of ambiguous, it does mention alpha and beta but not RC. But I guess you're right.

PR #1820, opened 2022-08-24
put template in scope to avoid assertion errors in other middleware

### LINKED ISSUE #472: AssertionError with middleware and TemplateResponse
```diff
diff --git i/tests/test_templates.py w/tests/test_templates.py
index b48373e..57dee91 100644
--- i/tests/test_templates.py
+++ w/tests/test_templates.py
@@ -20,6 +20,11 @@ def test_templates(tmpdir):
     async def homepage(request):
         return templates.TemplateResponse("index.html", {"request": request})

+    @app.middleware("http")
+    async def noop(request, call_next):
+        response = await call_next(request)
+        return response
+
     client = TestClient(app)
     response = client.get("/")
     assert response.text == "<html>Hello, <a href='http://testserver/'>world</a></html>"
```

causes:
```
tests/test_templates.py:29: in test_templates
    response = client.get("/")
.venv/lib/python3.7/site-packages/requests/sessions.py:546: in get
starlette/testclient.py:421: in request
    json=json,
.venv/lib/python3.7/site-packages/requests/sessions.py:533: in request
.venv/lib/python3.7/site-packages/requests/sessions.py:646: in send
starlette/testclient.py:238: in send
    raise exc from None
starlette/testclient.py:235: in send
/usr/lib/python3.7/asyncio/base_events.py:584: in run_until_complete
    return future.result()
starlette/applications.py:134: in __call__
    await self.error_middleware(scope, receive, send)
starlette/middleware/errors.py:122: in __call__
    raise exc from None
starlette/middleware/errors.py:100: in __call__
starlette/middleware/base.py:25: in __call__
    response = await self.dispatch_func(request, self.call_next)
tests/test_templates.py:21: in noop
starlette/middleware/base.py:47: in call_next
E   AssertionError
```
`message` is `{'type': 'http.response.template', 'template': <Template 'index.html'>, 'context': {'request': <starlette.requests.Request object at 0x7ff4a2d21e48>}}` there.

This appears to be an issue with the test client only though.
Do I understand it correctly that the "http.response.template" extension is only used with tests?

adriangb:
I think fixes #472 

The [original concern](https://github.com/encode/starlette/issues/472#issuecomment-484144812) with this change was:

> We don't really want to pack it into the state because it's possible that some middleware might return a different response instead, in which case we'd erronously be reporting that we got a template response.

This resolves the concern by putting this info into `scope` _only if TemplateResponse is the one sending the response`.

### DIFF
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -3,7 +3,7 @@ from os import PathLike
 
 from starlette.background import BackgroundTask
 from starlette.responses import Response
-from starlette.types import Receive, Scope, Send
+from starlette.types import Message, Receive, Scope, Send
 
 try:
     import jinja2
@@ -41,13 +41,20 @@ class _TemplateResponse(Response):
         request = self.context.get("request", {})
         extensions = request.get("extensions", {})
         if "http.response.template" in extensions:
-            await send(
-                {
-                    "type": "http.response.template",
+            original_send = send
+
+            async def wrapped_send(message: Message) -> None:
+                scope["extensions"] = extensions = scope.get("extensions", None) or {}
+                extensions["http.test_info"] = test_info = (
+                    extensions.get("http.test_info", None) or {}
+                )
+                test_info["template"] = {
                     "template": self.template,
                     "context": self.context,
                 }
-            )
+                await original_send(message)
+
+            send = wrapped_send
         await super().__call__(scope, receive, send)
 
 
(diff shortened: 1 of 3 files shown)


adriangb:
Hmm so this doesn't really do what I was hoping to achieve. `_TemplateResponse` is in the middle of the onion, replacing `send` has no meaning. The original issue is that an external middleware might replace `send` altogether.

PR #1824, opened 2022-08-26
Use explicit origin when with credentials

### LINKED ISSUE #1832: CORSMiddleware does not provide explicit origin although Authorization header is present

> [!IMPORTANT]

<picture>
  Fund with Polar
</picture>
</a>

gyusang:
- [ ] Initially raised as discussion #1823 

Edit by @Kludex:
- Closes #1832

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -158,10 +158,11 @@ class CORSMiddleware:
         headers.update(self.simple_headers)
         origin = request_headers["Origin"]
         has_cookie = "cookie" in request_headers
+        has_authorization_header = "authorization" in request_headers
 
         # If request includes any cookie headers, then we must respond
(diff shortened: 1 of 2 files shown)


Kludex:
Can you add a test, please?

gyusang:
> Can you add a test, please?

Yes, I have added a test for the added functionality.

Kludex:
I'm not sure about this idea. I don't see any CORS implementation using a custom logic for cookies or authorization headers.

But I do see a different logic when "allow_credentials" is true, that we don't have.

Kludex:
@gyusang I think we still need a bit of discussion around this. I took a bit of time to search around, and I didn't find any implementation that does what we do, nor expand the logic as this PR suggests.

Let's continue this on #1832, please. 🙏

PR #1825, opened 2022-08-26
Update conditions

iQiexie:
I have recently stumbled upon a bug and created a little fix for this.

Traceback if anyone is interested https://gist.github.com/Neykuratick/175100a393155e5b38a8db2d48c78e14


Kludex:
I think you need to bump FastAPI. This is not a bug.

PR #1836, opened 2022-08-31
Add fake setup.py

Kludex:
- Analogous to https://github.com/encode/httpx/pull/2351

This is just to make GitHub to include Starlette on the package dependencies graph.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,29 @@
+import sys
+
+from setuptools import setup
+
+sys.stderr.write(
+    """
+===============================
+Unsupported installation method
+===============================
+httpx no longer supports installation with `python setup.py install`.
+Please use `python -m pip install .` instead.
+"""
+)
+sys.exit(1)
+
+
+# The below code will never execute, however GitHub is particularly
+# picky about where it finds Python packaging metadata.
+# See: https://github.com/github/feedback/discussions/6456
+#
+# To be removed once GitHub catches up.
+
+setup(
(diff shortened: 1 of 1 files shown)


michaeloliverx (approved):

setup.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+===============================
+Unsupported installation method
+===============================
+httpx no longer supports installation with `python setup.py install`.
+Please use `python -m pip install .` instead.
+"""
+)
+sys.exit(1)
+
```
aminalaee:
```suggestion
from setuptools import setup

```

setup.py
```diff
@@ -0,0 +1,29 @@
+import sys
+
+from setuptools import setup
+
+sys.stderr.write(
+    """
+===============================
+Unsupported installation method
+===============================
+httpx no longer supports installation with `python setup.py install`.
```
Kludex:
```suggestion
uvicorn no longer supports installation with `python setup.py install`.
```

setup.py
```diff
@@ -0,0 +1,29 @@
+import sys
+
+from setuptools import setup
+
+sys.stderr.write(
+    """
+===============================
+Unsupported installation method
+===============================
+uvicorn no longer supports installation with `python setup.py install`.
```
Kludex:
```suggestion
Starlette no longer supports installation with `python setup.py install`.
```

PR #1848, opened 2022-09-05
[FIX] Installation script and requirements.txt for dev

joaopfg:
Related to the discussion [here](https://github.com/encode/starlette/discussions/1847).


requirements.txt
```diff
@@ -1,20 +1,19 @@
...
--e .[full]
-
 # Testing
 autoflake==1.5.3
 black==22.8.0
 coverage==6.4.2
 databases[sqlite]==0.6.1
 flake8==3.9.2
 isort==5.10.1
+itsdangerous==2.1.2
```
```suggestion
```

requirements.txt
```diff
@@ -1,20 +1,19 @@
...
 mypy==0.971
 typing_extensions==4.3.0
 types-requests==2.26.3
 types-contextvars==2.4.7
 types-PyYAML==6.0.11
 types-dataclasses==0.6.6
 pytest==7.1.2
+python-multipart==0.0.5
```
```suggestion
```

scripts/install
```diff
@@ -16,4 +16,4 @@ else
 fi
 
 "$PIP" install -r "$REQUIREMENTS"
-"$PIP" install -e .
+"$PIP" install .
```
Kludex: Isn't this the only change needed? 🤔

joaopfg: @Kludex For me, the `scripts/test` only fully passed after also adding the two lines in the `requirements.txt`

scripts/install
```diff
@@ -16,4 +16,4 @@ else
 fi
```
Kludex, line 17: If you undo all the changes you've made, and just add this line:
```suggestion
"$PIP" install pip setuptools --upgrade
```
Does it work?

joaopfg: @Kludex It installs without problems but fails the test suite with:

```
______________________________________________________________________________________ ERROR collecting tests/middleware/test_session.py ______________________________________________________________________________________
ImportError while importing test module '/home/eyepick/starlette/tests/middleware/test_session.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../.pyenv/versions/3.9.5/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/middleware/test_session.py:5: in <module>
starlette/middleware/sessions.py:6: in <module>
    import itsdangerous
E   ModuleNotFoundError: No module named 'itsdangerous'
```

Adding the two lines to the requirements.txt pass all the tests though

Kludex: How are you running that?

Kludex: You're running `./scripts/test`? Why is your pyenv venv being used then? 🤔

joaopfg: Indeed!  I had a bad setup. Sorry, it works with your changes.

Kludex: Would you like to update this PR? 👀

joaopfg: @Kludex PR updated. The github CI tests only pass with the following requirements:

```
httpx==0.23.0
itsdangerous==2.1.2
python-multipart==0.0.5
```

It seems it's not an issue with my setup.

Kludex: Can you undo all the changes and only let the pip and setuptools installation?

joaopfg: @Kludex It was like that in my previous commit. It failed some github CI tests. The tests can only pass after adding the requirements I mentioned.

scripts/install
```diff
@@ -15,5 +15,6 @@ else
     PIP="pip"
 fi
 
+"$PIP" install pip setuptools --upgrade
 "$PIP" install -r "$REQUIREMENTS"
-"$PIP" install -e .
+"$PIP" install -e .
```
Kludex:
```suggestion
"$PIP" install -e .

```

scripts/install
Kludex: Pipeline is passing...

Kludex: Does it work on your local setup?

joaopfg: @Kludex Yes, now it's working fine

Kludex:
This PR solves a problem that I've mentioned on https://github.com/encode/starlette/discussions/1847#discussioncomment-3559038 which I'm not able to reproduce anymore... If someone is able to reproduce it, we can reopen this PR.

Thanks for the efforts @joaopfg :pray:

PR #1872, opened 2022-09-27
Disable TestClient follow_redirects by default

aminalaee:
As discussed in https://github.com/encode/starlette/discussions/1871

In the migration of `TestClient` from `requests` to `httpx` in https://github.com/encode/starlette/pull/1376 we enabled `follow_redirects` in httpx by default. I think this was a side effect of that PR and was not on purpose.

There are two points to disable it by default:
- This was not enabled by default before, even though it's not a major breaking change, it can still break some code and it was not planned for.
- In httpx this is also disabled by default so if you have both async and sync tests, you have to also enable it for AsyncClient to get the same behaviour of TestClient.

There's a sample code in https://github.com/encode/starlette/discussions/1871 if needed.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -386,7 +386,6 @@ class TestClient(httpx.Client):
             base_url=base_url,
             headers={"user-agent": "testclient"},
             transport=transport,
-            follow_redirects=True,
             cookies=cookies,
         )
 
--- a/tests/test_testclient.py
+++ b/tests/test_testclient.py
@@ -9,7 +9,7 @@ import trio.lowlevel
 
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
-from starlette.responses import JSONResponse
+from starlette.responses import JSONResponse, RedirectResponse
 from starlette.routing import Route
 from starlette.websockets import WebSocket, WebSocketDisconnect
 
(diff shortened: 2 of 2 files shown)


aminalaee:
@Kludex Not entirely sure but seems like this change only was for POST requests and now we are in a better position compared to Starlette 0.20.1:

```py
def test_sync():
    client = TestClient(app)

    response = client.get("/bar")
    assert response.url == "http://testserver/foo"

    # Starlette 0.21.0
    response = client.post("/bar")
    assert response.url == "http://testserver/foo"

    # Starlette 0.20.1
    response = client.post("/bar")
    assert response.url == "http://testserver/bar"
    assert response.status_code == 302. # Not following redirects for POST but it does for GET
```

Even though this is different from httpx AsyncClient,  I think it's at least behaving the same way for both POST and GET.
I'll spend some time on it and close it if I have nothing to add.

euri10:
unless I'm mistaken that's one of the difference between requests and httpx.

requests follows redirects by default, httpx doesnt

so unless the previous TestClient was overriding requests default behaviour (or requests in fact does not follow redirects by default, which means this part in httpx docs should then be corrected: https://www.python-httpx.org/compatibility/) it's fair to say the new TestClient should force the flag to True.

aminalaee:
@euri10 yeah looks like it’s ok.
I was only seeing an edge case that POST request with httpx TestClient follows redirects but previously with requestst it didn’t follow.
But both now and in the past they followed redurrcts for GET requests. I have tried to do a minimal example.

if I understand it correctly it’s actually even better now so I will close it.

Kludex:
Thanks @aminalaee  🙏

PR #1874, opened 2022-09-27
Remove note about `BackgroundTasks` being a limitation on the `BaseHTTPMiddleware`

Kludex:


docs/middleware.md
```diff
@@ -238,14 +238,11 @@ Middleware classes should not modify their state outside of the `__init__` metho
 Instead you should keep any state local to the `dispatch` method, or pass it
 around explicitly, rather than mutating the middleware instance.
 
-### Limitations
+### Limitation
```
adriangb: I think we can keep this plural and keep it a list even if there is only 1 currently known limitation

Kludex: Do you intend to extend this list? :thinking:

adriangb: Not necessarily, I still think it's better style, but that's just my opinion, not a blocker so I'm approving the PR regardless

adriangb (approved):

PR #1876, opened 2022-09-28
Add support for HTTP trailers

Kludex:
No intention to get this merged or to even discuss for the time being. Just pushing for future reference. :pray:

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -261,6 +261,7 @@ class StreamingResponse(Response):
                 "type": "http.response.start",
                 "status": self.status_code,
                 "headers": self.raw_headers,
+                "trailers": bool(self.trailers),
             }
         )
         async for chunk in self.body_iterator:
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -12,6 +12,7 @@ from urllib.parse import quote
 import anyio
 
 from starlette._compat import md5_hexdigest
+from starlette._utils import is_async_callable
 from starlette.background import BackgroundTask
 from starlette.concurrency import iterate_in_threadpool
 from starlette.datastructures import URL, MutableHeaders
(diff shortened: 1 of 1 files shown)


starlette/responses.py
```diff
@@ -262,14 +270,36 @@ async def stream_response(self, send: Send) -> None:
...
+                    "headers": [
+                        (key.encode(self.charset), value.encode(self.charset))
+                        for key, value in trailers
+                    ],
+                    "more_trailers": False,
+                }
+            )
+
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        support_trailers = "http.response.trailers" in scope.get("extensions", {})
```
Kludex, line 294: Do we need this? The server should be able to ignore it... :thinking:

starlette/responses.py
```diff
@@ -247,12 +255,13 @@ async def listen_for_disconnect(self, receive: Receive) -> None:
             if message["type"] == "http.disconnect":
                 break
 
-    async def stream_response(self, send: Send) -> None:
+    async def stream_response(self, send: Send, support_trailers: bool) -> None:
         await send(
             {
                 "type": "http.response.start",
                 "status": self.status_code,
                 "headers": self.raw_headers,
```
Kludex, line 263: Should we append the `Trailer` header? The key of the `trailers` attribute can be used to populate it... Or it's not Starlette's job?

Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer

starlette/responses.py
```diff
@@ -262,14 +271,36 @@ async def stream_response(self, send: Send) -> None:
...
 
+        if support_trailers and self.trailers:
+            trailers: typing.List[typing.Tuple[str, str]] = []
+            for key, call in self.trailers.items():
+                if not is_async_callable(call):
+                    trailers.append(key, await anyio.to_thread.run_sync(call))
+                else:
+                    trailers.append(key, await call())
+
+            await send(
```
Kludex, line 282: Multiple `http.response.trailers` can be sent, but... Is there a reason for it? Maybe with another kind of API it would make sense... :thinking:

starlette/responses.py
```diff
@@ -220,6 +221,10 @@ def __init__(
 AsyncContentStream = typing.AsyncIterable[Content]
 ContentStream = typing.Union[AsyncContentStream, SyncContentStream]
 
+Trailers = typing.Mapping[
+    str, typing.Callable[[], typing.Union[str, typing.Awaitable[str]]]
```
Kludex, line 225: We need to let the user populate those values at the end of the response.

Kludex:
That's it. :wave:

Kludex:
This is implemented in https://github.com/Kludex/uvicorn-extensions/blob/e01b7048aa55752c024fc45ab0e946e4972ece87/src/python/asgi-trailers/asgi_trailers/response.py.

PR #1888, opened 2022-10-03
Version 1.0.0

### LINKED ISSUE #51: Version 1.0

Kludex:
- Closes #51

## Description

This PR bumps Starlette to 1.0.0. :tada:

## Changes

- Bump Starlette to 1.0.0.
- Add deprecation policy (`docs/deprecation.md`).
- Remove all code that was marked as deprecated.
- ~Deprecate `BaseHTTPMiddleware`.~
- ~Deprecate decorator functions in the `Router` method, which were discouraged for some time.~

> **Note**
> UPDATE [05/10/22]: PRs created according to comments on this PR:
> - https://github.com/encode/starlette/pull/1898 
> - https://github.com/encode/starlette/pull/1897

## Blockers

A blocker is PR or idea that can potentially make us jump to 2.0.0 too fast. If you see one around, please comment.

- [x] Decision on `UploadFile` PRs:
	- [x] https://github.com/encode/starlette/pull/1405
	- [x] https://github.com/encode/starlette/pull/1413
- [x] Decision on `Request.url_for` or related PRs:
	- [x] https://github.com/encode/starlette/pull/1385
	- [x] https://github.com/encode/starlette/pull/611
	- [x] https://github.com/encode/starlette/pull/1738 - No need to be included.
	- [x] https://github.com/encode/starlette/pull/1416 - No need to be included.
- [x] **Important**: check if #1858 is really an issue.  - No need to be included.

> **Note**
> UPDATE [22/10/22]: We need to fix the `python-multpart` issues.
> - [x] #1059

## Missing

- [X] Deprecate decorators on `Starlette` - I've only deprecated on `Router` until now.

> **Note**
> UPDATE [05/10/22]: Based on the comments on this PR, there's an agreement that Starlette 1.0 will not have deprecation warnings. We shall remove all deprecated code before that.

## Questions

1. Should we remove the code around the `lifespan` warnings? Is that part of the code supposed to be removed?
2. Should we remove the `allow_redirects` deprecation? It was recently added.
    - Yes. It's been months already.
4. Should we think about release cycle?
    - We are going to support only the latest version.

## How to review this PR?

1. Check `warnings.warn` in code.
2. Read the `Deprecation Policy` section using `./scripts/docs serve`.
3. Try hyperlinks used in the code, and release notes.


alex-oleshkevich:
This is good news! Here are my thoughts about deprecation.

As Starlette goes 1.0 then it is a good time to drop all deprecated stuff and apply breaking changes. If this is too risky at this moment, then I would advise rolling out a few more 0.xx releases with deprecation warnings and then release 1.0 by the end of the year.

alex-oleshkevich:
We may want to get back to https://github.com/encode/starlette/pull/1803

docs/deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+## Process to Remove a Feature
+
+We'll consider two kinds of processes: **drop Python version support** and **feature removal**.
+
+### Python Version
+
+Starlette will aim to support a Python version until the [EOL date of that version](https://endoflife.date/python).
+When a Python version reaches EOL, Starlette will drop support for that version in the next **minor** release.
+
+The drop of Python version support will be documented in the release notes, but the user will **not** be warned it in code.
```
tiangolo, line 25:
```suggestion
The drop of Python version support will be documented in the release notes, but the user will **not** be warned about it in code.
```

docs/deprecation.md
```diff
@@ -0,0 +1,61 @@
...
+Starlette follows [Semantic Versioning](https://semver.org/).
+
+## Process to Remove a Feature
+
+We'll consider two kinds of processes: **drop Python version support** and **feature removal**.
+
+### Python Version
+
+Starlette will aim to support a Python version until the [EOL date of that version](https://endoflife.date/python).
+When a Python version reaches EOL, Starlette will drop support for that version in the next **minor** release.
```
tiangolo, line 23: I agree that dropping support for old Python versions makes sense in minors. Because there's no way pip or other tools would install a version that doesn't support the environment's old Python. This means that the *breaking* change would not really break anything, because the version would not get installed in the environment.

I just tested with:

```console
$ docker run -it python:3.6 pip install starlette

Successfully installed anyio-3.6.1 contextlib2-21.6.0 contextvars-2.4 dataclasses-0.8 idna-3.4 immutables-0.19 sniffio-1.2.0 starlette-0.19.1 typing-extensions-4.1.1
```

It uses the last Starlette version that supports Python 3.6: `starlette-0.19.1`

docs/release-notes.md
```diff
@@ -1,3 +1,31 @@
...
+* Deprecated `Router.on_event` decorator. Please refer to [Registering events](https://www.starlette.io/events/#registering-events)
+  for recommended approach.
+* Deprecated `Router.websocket_route` decorator. Please refer to [WebSocket Routing](https://www.starlette.io/routing/#websocket-routing)
+  for recommended approach.
+* Deprecated `Router.route` decorator. Please refer to [HTTP Routing](https://www.starlette.io/routing/#http-routing) for recommended approach.
+* Deprecated `Router.host` method. Please refer to [Host-based routing](https://www.starlette.io/routing/#host-based-routing)
+  for recommended approach.
+* Deprecated `Router.mount` method. Please refer to [Submounting routes](https://www.starlette.io/routing/#submounting-routes)
+  for recommended approach.
+
```
tiangolo: For FastAPI, at some point, I plan on deprecating some things, and then removing them, all that before the first 1.0.0 release.

I would like the first 1.0.0 release in FastAPI to not have sort of dangling or pending things to do. It's also easier to deprecate, remove, revert, add, and change things in 0.x.x land. Currently removing deprecated things requires a minor bump (the same as in Starlette right now). But removing things can break others and require updates or reverts. I want to have those things a bit more stable and not expecting changes before releasing the first FastAPI 1.0.0.

But then again, that's just how I plan on doing it on FastAPI, it doesn't have to be that way, there's no obvious single right way to do things. But knowing that I plan on doing that in FastAPI might relax the requirements to release a first 1.0.0 here before deprecating and removing those things.

Also, I have Starlette strictly pinned in FastAPI, so if these changes were done in 0.x.x versions they wouldn't affect FastAPI users, as the internal parts would be updated before the version pin is raised.

I don't have strong opinions about how it should be done here, just my random thoughts.

starlette/status.py
```diff
@@ -5,89 +5,6 @@
...
-    "WS_1007_INVALID_FRAME_PAYLOAD_DATA",
-    "WS_1008_POLICY_VIOLATION",
-    "WS_1009_MESSAGE_TOO_BIG",
-    "WS_1010_MANDATORY_EXT",
-    "WS_1011_INTERNAL_ERROR",
-    "WS_1012_SERVICE_RESTART",
-    "WS_1013_TRY_AGAIN_LATER",
-    "WS_1014_BAD_GATEWAY",
-    "WS_1015_TLS_HANDSHAKE",
-)
```
tiangolo, line 90: I like removing these `__all__`! 🚀 

I wonder if there was some rationale for having them before and if there's one for removing them now.

---

I understand that could be needed for example in `__init__.py` files to define the things that are exported, but in those cases I also prefer the alternative of `import x as x`, just to let mypy and others that it is explicitly exported.

Anyway, I like and agree with removing this, it's just curiosity.

tiangolo (commented):
Awesome! Exciting! I just have a couple of minor comments.

starlette/status.py
Kludex: They were actually added because they were needed: https://github.com/encode/starlette/pull/1580

Otherwise, I couldn't make use of [PEP 562](https://peps.python.org/pep-0562/) properly.

tiangolo:
Another thing to have in mind that might be worth deprecating and can create some strange behaviors/issues now: `app.add_middleware` would create multiple instances of any previous middlewares as described here: https://github.com/encode/starlette/discussions/1161

There are not many cases where this is a problem, but it is in some, for example when dealing with Prometheus, that expects that an object is created and registered only once.

---

I was chatting with @Kludex and he pointed me to that discussion, so, mentioning it here as well.

starlette/status.py
tiangolo: Ah! Amazing, thanks for the clarification. ✅

docs/deprecation.md
Kludex: Gracias 🙏

Kludex:
I also agree on what @alex-oleshkevich (if I got what you said right) and @tiangolo said about the deprecation.

The plan changes a bit then, we should include the deprecation warnings on the next release (`0.22.0`).

Then, on this PR, we remove all the deprecation warnings - Let's not have any pendency on our stable release. Conceptually, it makes more sense to do this.

tiangolo:
Last night I remembered a couple of extra things I had pending and probably worth discussing here.

## Closing `UploadFile`

When enabling tests for Python 3.11 in FastAPI, I'm getting errors about `UploadFile` not being closed. It seems Python 3.10 and below don't complain about that. But it would make sense that it would expect those files to be closed.

It seems those errors don't show up in Starlette's tests for Python 3.11, but maybe it's just because that code path is currently not tested, not sure.

The other thing is, it might be the decision in Starlette that users/developers are responsible for closing those `UploadFile` objects or something, in that case, I should just handle that in FastAPI, but I wanted to mention it here in case it's something that should be fixed.

I know this should be a proper issue/example. But I thought it was worth at least mentioning it here before the changes and releases are fixed.

## Routers supporting empty trailing slash

What I refer with this is that there could be a router that handles all the routes under some prefix, for example: `https://github.com/encode`, and it would also handle `https://github.com/encode/starlette` and `https://github.com/encode/uvicorn`. But the same router handles `https://github.com/encode` without requiring it to end in a trailing slash. BTW, these GitHub URLs work like that.

Currently, Starlette doesn't internally support that, at least not in the same router.

FastAPI has supported that in a strange way, because in FastAPI, when using `app.include_router()`, currently it just copies all the paths, and in the end, it's just a single global router with everything, in FastAPI routers are currently not really "mounted" as in Starlette. So, if a router has a route with a path of `""`, (instead of `"/"`), that would handle that top level route without the trailing slash.

Now, I've wanted to refactor that in FastAPI, so that routes are not copied but routers are actually mounted (or something similar). This will allow some things on the FastAPI side. But one of the blockers I found was that I couldn't use the standard Starlette Routers and Mounts because that would not support prefixes without a trailing slash.

I need to support that and a couple of other things in FastAPI.

I'm totally fine if the decision is to not support that in Starlette, but I thought it was worth mentioning that here too, in case it seems it's something that would make sense to support.

adriangb:
> Routers supporting empty trailing slash

Thank you for bringing that up Sebastián. I've wanted to fix that for a while (I wrote the implementation at one point) but it required some minor breaking changes (I think it had to do with what errors are raised). If we want to make the change I'm happy to make the PR.

docs/release-notes.md
Kludex: https://github.com/encode/starlette/pull/1888#issuecomment-1267325021 :+1:

adriangb:
With respect to UploadFile: can you add the .close() to the AsyncExitStack for dependencies? It works in Xpresso: https://github.com/adriangb/xpresso/blob/ae1586513b8ee48bfe9a8fdb2e7cc7038c37053d/xpresso/binders/_binders/file_body.py#L76

Kludex:
> Another thing to have in mind that might be worth deprecating and can create some strange behaviors/issues now: `app.add_middleware` would create multiple instances of any previous middlewares as described here: #1161
> 
> There are not many cases where this is a problem, but it is in some, for example when dealing with Prometheus, that expects that an object is created and registered only once.
> 
> I was chatting with @Kludex and he pointed me to that discussion, so, mentioning it here as well.

`app.add_middleware` doesn't have any issues since 0.24.0 thanks to @adriangb. :pray:

adriangb:
Let’s add #2068

Kludex:
I'm going to review both (which I also invite others to review):
- https://github.com/encode/starlette/pull/1692
- https://github.com/encode/starlette/pull/2026

After a decision is made there, let's release the last zero version release, then wait some weeks, and then prepare for 1.0 with:
- https://github.com/encode/starlette/pull/2068
- https://github.com/encode/starlette/pull/1888

Comments, feedback, or anything that you want to include before 1.0, please feel free to talk cc @encode/maintainers

BrettMoan:
I'm just here to write a thank you to all the people here supporting `starlette` and `fastapi` and other related repos. And congratulate you on a massive undertaking (an official release you are feeling prod and confident of) 

Often open source is thankless and while I may be a random person on the internet you don't know, know I'm a person out there like you, typing away as part of a living supporting my family via a career in programming. Using `starlette` and `fastapi` has had helped me tremendously at my job as a backend developer. It has been a joy to work with. 

(this includes `pydantic` as well where theres also a decent amount of contributor overlap)

Thank you. Really. My job wouldn't be nearly as sane without all the of the great work y'all have already done, both in implementation and documentation.

Kludex:
@BrettMoan Thank you for saying it. Much appreciated! ❤️

Kludex:
Let's go to https://github.com/encode/starlette/pull/2384.

I've cleaned the branch. There were too many conflicts, I thought it would be easier to create a new one.

PR #1890, opened 2022-10-04
touch up TestClient type annotations

adriangb:
Before in Visual Studio Code:

Screen Shot 2022-10-04 at 3 06 41 PM

After:

Screen Shot 2022-10-04 at 3 07 04 PM

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -576,7 +576,7 @@ class TestClient(httpx.Client):
         return super().post(
             url,
             content=content,
-            data=data,
+            data=data,  # type: ignore
             files=files,
             json=json,
             params=params,
@@ -613,7 +613,7 @@ class TestClient(httpx.Client):
         return super().put(
             url,
             content=content,
-            data=data,
+            data=data,  # type: ignore
             files=files,
             json=json,
             params=params,
@@ -650,7 +650,7 @@ class TestClient(httpx.Client):
         return super().patch(
             url,
             content=content,
-            data=data,
+            data=data,  # type: ignore
             files=files,
             json=json,
             params=params,
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -452,8 +452,7 @@ class TestClient(httpx.Client):
             method,
             url,
             content=content,
-            data=data,  # type: ignore
-            files=files,
+            data=data,  # type: ignore[arg-ty[e            files=files,
             json=json,
             params=params,
             headers=headers,
@@ -576,7 +575,7 @@ class TestClient(httpx.Client):
         return super().post(
             url,
             content=content,
-            data=data,  # type: ignore
+            data=data,  # type: ignore[arg-type]
             files=files,
             json=json,
             params=params,
@@ -613,7 +612,7 @@ class TestClient(httpx.Client):
         return super().put(
             url,
             content=content,
-            data=data,  # type: ignore
+            data=data,  # type: ignore[arg-type]
             files=files,
             json=json,
             params=params,
@@ -650,7 +649,7 @@ class TestClient(httpx.Client):
         return super().patch(
             url,
             content=content,
-            data=data,  # type: ignore
+            data=data,  # type: ignore[arg-type]
             files=files,
             json=json,
             params=params,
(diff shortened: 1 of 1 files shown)


starlette/testclient.py
```diff
@@ -426,30 +429,30 @@ def request(  # type: ignore[override]
...
+        extensions: typing.Optional[typing.Dict[str, typing.Any]] = None,
     ) -> httpx.Response:
         url = self.base_url.join(url)
         redirect = self._choose_redirect_arg(follow_redirects, allow_redirects)
         return super().request(
             method,
             url,
             content=content,
-            data=data,
+            data=data,  # type: ignore
```
Kludex, line 455: Please add the code error :pray:

Kludex: Please add the code error :pray:

Kludex (changes requested):

starlette/testclient.py
```diff
@@ -33,6 +33,9 @@
 ASGI3App = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
 
 
+_RequestData = typing.Mapping[str, typing.Union[str, typing.Iterable[str]]]
```
euri10, line 36: probably dumb question but  cant we use httpx types directly into starlette ?

adriangb: If we're willing to bump the min version to the next unreleased version, then yes: https://github.com/encode/httpx/commit/8152c4facd0f71e0f376287e41a0810a60fec9c6

I figured I'd just fix it here since I'm doing related stuff and it's a one liner, that way we don't have to require everyone to bump up their httpx version just for a single type annotation

starlette/testclient.py
adriangb: Added. This is the same reason as https://github.com/encode/starlette/pull/1890#discussion_r987574795 (https://github.com/encode/starlette/actions/runs/3185871438/jobs/5195913711#step:5:19)

starlette/testclient.py
euri10: I'd rather bump the req and have the types manages in a single place, but ymmv :)

starlette/testclient.py
Kludex: I didn't get it. The reason is that httpx defines its own type?

adriangb: The reason is that httpx's current type is `dict` and we'll be giving it a `Mapping`. The correct type is `Mapping` and it will be fixed in the next httpx release.

Kludex: Ah! Got it. :+1:

Kludex: Ok, another question then... Why are you being more strict here than `httpx`? Or the intention is to also be more strict there?

adriangb: I didn't even pay attention to that to be honest. I would say yes, we should be more strict there. I can make a PR for that. Happy to change this to `Mapping[str, Any]` in the meantime, but I think it's better to be strict and then relax than the other way around (to avoid breaking changes)

Kludex (approved):

starlette/testclient.py
```diff
@@ -426,30 +429,30 @@ def request(  # type: ignore[override]
...
+        params: typing.Optional[httpx._types.QueryParamTypes] = None,
+        headers: typing.Optional[httpx._types.HeaderTypes] = None,
+        cookies: typing.Optional[httpx._types.CookieTypes] = None,
         auth: typing.Union[
             httpx._types.AuthTypes, httpx._client.UseClientDefault
         ] = httpx._client.USE_CLIENT_DEFAULT,
-        follow_redirects: bool = None,
-        allow_redirects: bool = None,
+        follow_redirects: typing.Optional[bool] = None,
+        allow_redirects: typing.Optional[bool] = None,
```
Kludex, line 443: Why are those `Optional[bool]`? Is it the same in `httpx`?

adriangb: I think you may be right. These should have been USE_CLIENT_DEFAULT.

Kludex: No, it's fine. I found out why.

Kludex: It's because I had to maintain `allow_redirects` alive. :+1:

adriangb: Ok so no need for a followup PR?

Kludex: nop, everything is cool. :pray:

PR #1894, opened 2022-10-05
Install latest pip on `install` script

Kludex:
Same as https://github.com/encode/httpx/pull/2396.


michaeloliverx (approved):

ahopkins:
This seems like a very strange and overly opinionated decision. Why introduce this side effect?

Kludex:
Hmmm.. The context of this PR is on the httpx analogous one. It's because old versions of pip can't install the project anymore (because of hatch).

PR #1898, opened 2022-10-05
Deprecate `BaseHTTPMiddleware`

### LINKED ISSUE #1678: Deprecating BaseHTTPMiddleware

Kludex:
- Closes https://github.com/encode/starlette/issues/1678

## Description

To be continued... Here you'll see an explanation about this decision.

## Documentation

image

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -182,6 +182,11 @@ The following arguments are supported:
 
 ## BaseHTTPMiddleware
 
+!!! warning
+    The `BaseHTTPMiddleware` is deprecated since version `0.22.0`.
+
+    Please refer to [Pure ASGI Middleware](#pure-asgi-middleware) for the recommended way to implement middleware.
+
 An abstract class that allows you to write ASGI middleware against a request/response
 interface.
 
@@ -569,7 +574,7 @@ import time
 class MonitoringMiddleware:
     def __init__(self, app):
         self.app = app
-    
+
     async def __call__(self, scope, receive, send):
         start = time.time()
         try:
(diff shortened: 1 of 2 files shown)


Kludex:
Does it make sense for me to create a package that contains only the `BaseHTTPMiddleware`?

Kludex:
I'm still not confident about this. :disappointed:

Kludex:
Let's continue on #1678.

PR #1901, opened 2022-10-05
Bypass GZipMiddleware when response includes `Content-Encoding`

kklingenberg:
This is a suggestion for `GZipMiddleware` to not encode responses that already include the `Content-Encoding` header. It is a minor issue for me when using the middleware in an application that includes regular routes and _proxies_ based on this:

<https://gist.github.com/tomchristie/5765e10a90a41c7e57470e2dc700f9db>

If the proxied response already has an encoding set, starlette's middleware double-encodes the response.

I can get around it by writing the middleware on my own, such as shown in this PR, but I thought you might want to consider it also. I'm basically imitating [Django's GZipMiddleware](https://docs.djangoproject.com/en/4.1/ref/middleware/#module-django.middleware.gzip) which claims to have this behaviour.

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -33,6 +33,7 @@ class GZipResponder:
         self.send: Send = unattached_send
         self.initial_message: Message = {}
         self.started = False
+        self.content_encoding_set = False
         self.gzip_buffer = io.BytesIO()
         self.gzip_file = gzip.GzipFile(
             mode="wb", fileobj=self.gzip_buffer, compresslevel=compresslevel
@@ -48,12 +49,19 @@ class GZipResponder:
             # Don't send the initial message until we've determined how to
             # modify the outgoing headers correctly.
             self.initial_message = message
+            self.content_encoding_set = any(
+                header == b"content-encoding"
+                for header, _ in message.get("headers", [])
+            )
         elif message_type == "http.response.body" and not self.started:
             self.started = True
             body = message.get("body", b"")
             more_body = message.get("more_body", False)
-            if len(body) < self.minimum_size and not more_body:
-                # Don't apply GZip to small outgoing responses.
+            if (
+                len(body) < self.minimum_size and not more_body
+            ) or self.content_encoding_set:
+                # Don't apply GZip to small outgoing responses, or
+                # those already encoded.
                 await self.send(self.initial_message)
                 await self.send(message)
             elif not more_body:
@@ -86,19 +94,22 @@ class GZipResponder:
                 await self.send(message)
 
         elif message_type == "http.response.body":
-            # Remaining body in streaming GZip response.
-            body = message.get("body", b"")
-            more_body = message.get("more_body", False)
+            if self.content_encoding_set:
+                await self.send(message)
+            else:
+                # Remaining body in streaming GZip response.
+                body = message.get("body", b"")
+                more_body = message.get("more_body", False)
 
-            self.gzip_file.write(body)
-            if not more_body:
-                self.gzip_file.close()
+                self.gzip_file.write(body)
+                if not more_body:
+                    self.gzip_file.close()
 
-            message["body"] = self.gzip_buffer.getvalue()
-            self.gzip_buffer.seek(0)
-            self.gzip_buffer.truncate()
+                message["body"] = self.gzip_buffer.getvalue()
+                self.gzip_buffer.seek(0)
+                self.gzip_buffer.truncate()
 
-            await self.send(message)
+                await self.send(message)
 
 
 async def unattached_send(message: Message) -> typing.NoReturn:
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -49,19 +49,19 @@ class GZipResponder:
             # Don't send the initial message until we've determined how to
             # modify the outgoing headers correctly.
             self.initial_message = message
-            self.content_encoding_set = any(
-                header == b"content-encoding"
-                for header, _ in message.get("headers", [])
-            )
+            headers = Headers(raw=self.initial_message["headers"])
+            self.content_encoding_set = "content-encoding" in headers
+        elif message_type == "http.response.body" and self.content_encoding_set:
+            if not self.started:
+                self.started = True
+                await self.send(self.initial_message)
+            await self.send(message)
         elif message_type == "http.response.body" and not self.started:
             self.started = True
             body = message.get("body", b"")
             more_body = message.get("more_body", False)
-            if (
-                len(body) < self.minimum_size and not more_body
-            ) or self.content_encoding_set:
-                # Don't apply GZip to small outgoing responses, or
-                # those already encoded.
+            if len(body) < self.minimum_size and not more_body:
+                # Don't apply GZip to small outgoing responses.
                 await self.send(self.initial_message)
                 await self.send(message)
             elif not more_body:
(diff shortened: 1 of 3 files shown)


starlette/middleware/gzip.py
```diff
@@ -48,12 +49,19 @@ async def send_with_gzip(self, message: Message) -> None:
             # Don't send the initial message until we've determined how to
             # modify the outgoing headers correctly.
             self.initial_message = message
+            self.content_encoding_set = any(
+                header == b"content-encoding"
+                for header, _ in message.get("headers", [])
+            )
         elif message_type == "http.response.body" and not self.started:
```
Kludex, line 59: Can we only change it here instead?
```suggestion
        elif message_type == "http.response.body" and self.content_encoding_set:
            if not self.started:
                await send(self.initial_message)
            await send(message)
        elif message_type == "http.response.body" and not self.started:
```

starlette/middleware/gzip.py
```diff
@@ -48,12 +49,19 @@ async def send_with_gzip(self, message: Message) -> None:
...
             self.started = True
             body = message.get("body", b"")
             more_body = message.get("more_body", False)
-            if len(body) < self.minimum_size and not more_body:
-                # Don't apply GZip to small outgoing responses.
+            if (
+                len(body) < self.minimum_size and not more_body
+            ) or self.content_encoding_set:
+                # Don't apply GZip to small outgoing responses, or
+                # those already encoded.
```
```suggestion
            if len(body) < self.minimum_size and not more_body:
                # Don't apply GZip to small outgoing responses.
```

starlette/middleware/gzip.py
```diff
@@ -86,19 +94,22 @@ async def send_with_gzip(self, message: Message) -> None:
...
         elif message_type == "http.response.body":
-            # Remaining body in streaming GZip response.
-            body = message.get("body", b"")
-            more_body = message.get("more_body", False)
+            if self.content_encoding_set:
+                await self.send(message)
+            else:
+                # Remaining body in streaming GZip response.
+                body = message.get("body", b"")
+                more_body = message.get("more_body", False)
```
```suggestion
            # Remaining body in streaming GZip response.
            body = message.get("body", b"")
            more_body = message.get("more_body", False)
```

starlette/middleware/gzip.py
```diff
@@ -86,19 +94,22 @@ async def send_with_gzip(self, message: Message) -> None:
...
 
-            self.gzip_file.write(body)
-            if not more_body:
-                self.gzip_file.close()
+                self.gzip_file.write(body)
+                if not more_body:
+                    self.gzip_file.close()
```
```suggestion
            self.gzip_file.write(body)
            if not more_body:
                self.gzip_file.close()
```

starlette/middleware/gzip.py
```diff
@@ -86,19 +94,22 @@ async def send_with_gzip(self, message: Message) -> None:
...
 
-            message["body"] = self.gzip_buffer.getvalue()
-            self.gzip_buffer.seek(0)
-            self.gzip_buffer.truncate()
+                message["body"] = self.gzip_buffer.getvalue()
+                self.gzip_buffer.seek(0)
+                self.gzip_buffer.truncate()
```
```suggestion
            message["body"] = self.gzip_buffer.getvalue()
            self.gzip_buffer.seek(0)
            self.gzip_buffer.truncate()
```

starlette/middleware/gzip.py
```diff
@@ -86,19 +94,22 @@ async def send_with_gzip(self, message: Message) -> None:
...
 
-            await self.send(message)
+                await self.send(message)
```
```suggestion
            await self.send(message)
```

Kludex (changes requested):
I think this is valid. Can you check my suggestion? 🙏

starlette/middleware/gzip.py
```diff
@@ -48,6 +49,14 @@ async def send_with_gzip(self, message: Message) -> None:
...
             # modify the outgoing headers correctly.
             self.initial_message = message
+            self.content_encoding_set = any(
+                header == b"content-encoding"
+                for header, _ in message.get("headers", [])
+            )
+        elif message_type == "http.response.body" and self.content_encoding_set:
+            if not self.started:
+                await send(self.initial_message)
+            await send(message)
```
Kludex: Ups 👀 
```suggestion
                await self.send(self.initial_message)
            await self.send(message)
```

tests/middleware/test_gzip.py
```diff
@@ -76,3 +76,27 @@ async def generator(bytes, count):
...
+            streaming, status_code=200, headers={"Content-Encoding": "brotli"}
+        )
+
+    app = Starlette(
+        routes=[Route("/", endpoint=homepage)],
+        middleware=[Middleware(GZipMiddleware)],
+    )
+
+    client = test_client_factory(app)
+    response = client.get("/", headers={"accept-encoding": "gzip"})
```
Kludex: Not sure if it's relevant, but the client needs to show that it can understand `brotli`... 👀 
```suggestion
    response = client.get("/", headers={"accept-encoding": "gzip, br"})
```

starlette/middleware/gzip.py
```diff
@@ -48,6 +49,14 @@ async def send_with_gzip(self, message: Message) -> None:
             # Don't send the initial message until we've determined how to
             # modify the outgoing headers correctly.
             self.initial_message = message
+            self.content_encoding_set = any(
+                header == b"content-encoding"
+                for header, _ in message.get("headers", [])
+            )
```
Kludex: We have a preference to use the wrappers `Headers`, and `MutableHeaders`. 👀 
```suggestion
            headers = Headers(raw=self.initial_message["headers"])
            self.content_encoding_set = b"content-encoding" in headers
```

starlette/middleware/gzip.py
kklingenberg: I think I caught this one and it didn't auto-resolve for some reason :thinking:

kklingenberg:
Great suggestions, they reduce the diff size and make it more readable. Thanks :)

Kludex (approved):
Thanks @kklingenberg 🙏

papb:
I was about to start working on a PR for this, thankfully I noticed that you already did so I just have to update my Starlette version. Thanks @kklingenberg!!

Edit: actually I'm using FastAPI and the latest version (0.87.0) still uses Starlette 0.21.0, so I will have to wait... Thankfully there is already [a PR](https://github.com/tiangolo/fastapi/pull/5659) to update it to 0.22.0 so it will probably happen soon... Sorry for the noise.

tests/middleware/test_gzip.py
```diff
@@ -76,3 +76,27 @@ async def generator(bytes, count):
...
+        routes=[Route("/", endpoint=homepage)],
+        middleware=[Middleware(GZipMiddleware)],
+    )
+
+    client = test_client_factory(app)
+    response = client.get("/", headers={"accept-encoding": "gzip, br"})
+    assert response.status_code == 200
+    assert response.text == "x" * 4000
+    assert response.headers["Content-Encoding"] == "br"
+    assert "Content-Length" not in response.headers
```
papb, line 102: Wait, this test creates a `StreamingResponse` that claims to be encoded as Brotli but actually isn't, since its body is just `b"x" * 4000`, right? How is it possible that this test passes? Doesn't the client try to process brotli automatically on the response and [fails](https://github.com/encode/httpx/blob/c0e9dba32091ad0d58ed9d46d5bafaf892964b15/httpx/_decoders.py#L118-L121)?

```
>>> import brotli
>>> decompressor = brotli.Decompressor()
>>> decompressor.process(b"x" * 4000)
  File "<stdin>", line 1, in <module>
brotli.error: BrotliDecoderDecompressStream failed while processing the stream
```

papb: I've created what I think is a better test [here](https://github.com/fullonic/brotli-asgi/pull/24/files#diff-f455f302936271d755f6892d443786094d2607b505ff180ca817cb9eeabe1e5eR166-R187). @Kludex If you're interested, I can submit a PR updating the test here as well.

Kludex:
Gratitude is rare, and I don't consider it as noise. Thanks for the message :)

tests/middleware/test_gzip.py
Kludex: Yeah... It was my bad. This test was modified, you can see how it is on master.

papb: Oh indeed, [I found it](https://github.com/encode/starlette/pull/1962)

PR #1903, opened 2022-10-06
allow using Request.form() as a context manager

adriangb:
Coming from https://github.com/encode/starlette/pull/1888#issuecomment-1268165441

Evidently forgetting / not knowing to call `UploadFile.close()` is a common issue.
I think FastAPI should be able to automatically call that internally, but Starlette should also do more to nudge users in the right direction. I think the right tool for that in this situation is a context manager.

This PR allows users to use `Request.form()` both directly (as before) and as a context manager.

I would suggest that we then deprecate the `await request.form()` usage so that there is only 1 correct way to do things.


starlette/_utils.py
```diff
@@ -10,3 +17,50 @@ def is_async_callable(obj: typing.Any) -> bool:
...
+    async def __aenter__(self) -> T_co:
+        ...  # pragma: no cover
+
+    async def __aexit__(
+        self,
+        __exc_type: typing.Optional[typing.Type[BaseException]],
+        __exc_value: typing.Optional[BaseException],
+        __traceback: typing.Optional[TracebackType],
+    ) -> typing.Union[bool, None]:
+        ...  # pragma: no cover
```
adriangb, line 46: Because of https://github.com/python/typing_extensions/issues/34. This is unnecessary on 3.8+

starlette/requests.py
```diff
@@ -265,8 +267,11 @@ async def form(self) -> FormData:
                 self._form = FormData()
         return self._form
 
+    def form(self) -> AwaitableOrContextManager[FormData]:
```
tiangolo, line 270: Should this be async?

```suggestion
    async def form(self) -> AwaitableOrContextManager[FormData]:
```

I understand it would work either way just because the awaitable thing is returned, but maybe it could help to make it async, to make it explicit in the function that it returns something to be awaited. What do you think?

tiangolo (approved):
Amazing! And beautiful types! 🤩 👏

starlette/requests.py
adriangb: Unfortunately I don't think that will work: we want to pass an await able object to `AwaitableOrContextManagerWrapper` so that it can be used like `async with request.form()` and not `async with await request.form()`.

starlette/_utils.py
Kludex: What part exactly? I didn't get it.

Kludex:
> I would suggest that we then deprecate the await request.form() usage so that there is only 1 correct way to do things.

I can live with it. :man_shrugging:

Also, since 3.11 warns, it shouldn't be much of an issue, I guess?

Kludex (changes requested):
@adriangb Would you mind adding some lines to the docs about this? :pray:

Kludex:
> > I would suggest that we then deprecate the await request.form() usage so that there is only 1 correct way to do things.
> 
> I can live with it. man_shrugging
> 
> Also, since 3.11 warns, it shouldn't be much of an issue, I guess?

I'd say for us to focus on getting 1.0 out. It's not a big deal to further deprecate later on.

adriangb:
I opted to just edit the usage in the docs examples. I do think we should recommend this usage over the `await request.form()` usage so I felt like adding a note along the lines of "by the way you can do it both ways" would create confusion with no benefit to users. If they're used to doing it the other way they can continue to do so (unless we decide to deprecate it at some point, which I am in favor of but I agree doesn't need to happen right now).

docs/requests.md
```diff
@@ -137,9 +137,9 @@ As all these methods are `async` methods, you need to "await" them.
 For example, you can get the file name and the contents with:
 
 ```python
-form = await request.form()
-filename = form["upload_file"].filename
-contents = await form["upload_file"].read()
+async await request.form() as form:
```
Kludex:
```suggestion
async with request.form() as form:
```

starlette/_utils.py
Kludex: Ah, got it. Should we add a note or a conditional so we can remove when 3.8 EOL happens?

adriangb: Nah let’s just remove the boilerplate when 3.8 becomes EOL; I don’t think it will be a breaking change at runtime

Kludex: I mean, I think we need a way for us to remember to remove by then :thinking: 

There's nothing in the code that points that we need to replace it when we don't support 3.8 anymore.

florimondmanca (approved):

starlette/_utils.py
adriangb: I don't think there's an easy way to remind ourselves of this.

adriangb, line 46:
```suggestion
# TODO: once 3.8 is the minimum supported version (27 Jun 2023)
# this can just become
# class AwaitableOrContextManager(typing.Awaitable[T_co], typing.AsyncContextManager[T_co]):    pass
class AwaitableOrContextManager(Protocol[T_co]):
    def __await__(self) -> typing.Generator[typing.Any, None, T_co]:
        ...  # pragma: no cover

    async def __aenter__(self) -> T_co:
        ...  # pragma: no cover

    async def __aexit__(
        self,
        __exc_type: typing.Optional[typing.Type[BaseException]],
        __exc_value: typing.Optional[BaseException],
        __traceback: typing.Optional[TracebackType],
    ) -> typing.Union[bool, None]:
        ...  # pragma: no cover
```

adriangb: I added a comment. I think that's the best we can do.

Kludex: Yeah, that's what I meant. :+1:

Kludex (approved):

starlette/_utils.py
```diff
@@ -10,3 +17,53 @@ def is_async_callable(obj: typing.Any) -> bool:
...
         callable(obj) and asyncio.iscoroutinefunction(obj.__call__)
     )
+
+
+T_co = typing.TypeVar("T_co", covariant=True)
+
+
+# TODO: once 3.8 is the minimum supported version (27 Jun 2023)
+# this can just become
+# class AwaitableOrContextManager(typing.Awaitable[T_co], typing.AsyncContextManager[T_co]):    pass
```
Kludex:
```suggestion
# class AwaitableOrContextManager(Awaitable[T_co], AsyncContextManager[T_co]):    pass
```

adriangb: But then this is not a protocol, it's a class, and we're not returning an instance of it.

Kludex: I'm just hiding the `typing` module to make the comment shorter, so it passes the linter :thinking:

adriangb: oh sorry didn't see the suggestion was on the comment. I already reformatted it.

PR #1908, opened 2022-10-08
Support `str` and `datetime` on `expires` parameter on the `set_cookie` method

### LINKED ISSUE #1878: Wrong typing for expires in set_cookie
I think the type for 
https://github.com/encode/starlette/blob/55b0cbd045269a28a3f1b18f32ae3cb47226ab65/starlette/responses.py#L108

This should be a date according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie 

So I think a better type is `Optional[str]` , or Optional[str | datetime] and provide the right formatting.

oskipa:
- Closes https://github.com/encode/starlette/issues/1878

# What?

The PR implements the type changes suggested in the discussion. We are adding `typing.Optional[typing.Union[datetime, str]` to set cookie.

# Why?
Expiration was previously set as an integer. Reviewing https://www.rfc-editor.org/rfc/rfc6265 and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie doesn't show that an integer is the right type for a cookie date.

# How?
Following the instructions given in the discussion.

# Notes
This is my first PR to this project, as part of Hacktober fest :)

Edit by @Kludex :

## Missing

- [ ] Update documentation.
- [ ] Add test with naive timezone.


starlette/responses.py
```diff
@@ -148,14 +150,20 @@ def delete_cookie(
         self.set_cookie(
             key,
             max_age=0,
-            expires=0,
+            expires="0",
```
Kludex: Is this needed tho? What happens if this field is not set?

starlette/responses.py
```diff
@@ -148,14 +150,20 @@ def delete_cookie(
...
             httponly=httponly,
             samesite=samesite,
         )
 
+    def _format_expires(self, date: typing.Union[datetime, str]) -> str:
+        if isinstance(date, str):
+            return date
+        else:
+            return date.strftime("%a, %d %b %Y %H:%M:%S GMT")
+
```
```suggestion
```

starlette/responses.py
```diff
@@ -117,7 +118,8 @@ def set_cookie(
         if max_age is not None:
             cookie[key]["max-age"] = max_age
         if expires is not None:
-            cookie[key]["expires"] = expires
+            expires = typing.cast(typing.Union[datetime, str], expires)
+            cookie[key]["expires"] = self._format_expires(expires)
```
Kludex: Can we simplify this a bit?
```suggestion
            if isinstance(expires, datetime):
                cookie[key]["expires"] = expires.strftime(COOKIE_DATETIME_FORMAT)
            else:
                cookie[key]["expires"] = expires
```
Also, when creating the `COOKIE_DATETIME_FORMAT` constant, would you mind adding a comment above it with the link on which specifies the format? 🙏

starlette/responses.py
```diff
@@ -105,7 +106,7 @@ def set_cookie(
         key: str,
         value: str = "",
         max_age: typing.Optional[int] = None,
-        expires: typing.Optional[int] = None,
+        expires: typing.Optional[typing.Union[datetime, str]] = None,
```
Kludex: Let's add `int` back. It's working with it, even if it's not in any standard that I've found.

The question now becomes if we should accept `str`. Because objectively speaking, we can just accept a `datetime`, and it's already enough to solve the issue.

starlette/responses.py
foolishhugo: If we don't set it, `expires` is not set. I can remove it so that the semantic meaning is kept

starlette/responses.py
foolishhugo: Sure thing. I will update the typing

starlette/responses.py
oskipa: Although the point is to delete it. We can use the current type to achieve the same

starlette/responses.py
oskipa: Ok

starlette/responses.py
oskipa: Done

starlette/responses.py
```diff
@@ -35,6 +36,12 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
+# Format based on MDN HTTP documentation on Set-Cookie and Date format
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
+# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date
```
Kludex: Those are not the RFC... :eyes:

starlette/responses.py
```diff
@@ -148,7 +158,7 @@ def delete_cookie(
         self.set_cookie(
             key,
             max_age=0,
-            expires=0,
+            expires=datetime.now(timezone.utc),
```
Kludex: Do we need the field expires at all to delete a cookie?

tests/test_responses.py
```diff
@@ -309,6 +312,61 @@ async def app(scope, receive, send):
...
+            httponly=True,
+            samesite="none",
+        )
+        await response(scope, receive, send)
+
+    client = test_client_factory(app)
+    response = client.get("/")
+    cookie: SimpleCookie = SimpleCookie(response.headers.get("set-cookie"))
+
+    assert cookie["mycookie"]
```
Kludex: What are you making sure here?

Kludex (commented):
One test is enough. Make sure to pass a `datetime`, and assert the cookie has the format specified.

starlette/responses.py
oskipa: It is a cleaner interpretation than the RFC, but I will link the RFC too :)

starlette/responses.py
oskipa: I think so. But let me confirm this one

tests/test_responses.py
oskipa: That the cookie exists

starlette/responses.py
oskipa: In the next update, I replaced the links to mdn to RFCs

starlette/responses.py
oskipa: So we don't need an `expires` if we are setting `max_age` to `0`.  
`max_age` takes precedent over `expires`
https://www.rfc-editor.org/rfc/rfc6265#section-4.1.2.2

tests/test_responses.py
oskipa: I personally believe that is best to keep a test to test a single thing. One test to confirm that the cookie was created. One test to check the the `expire` format.  Having a single test where we only test the format is testing both scenarios at the same time.

That said, I am probably following these rules too strictly. Nothing terrible wrong will happen, and having a single test will speed up the tests. And more than anything else, it is your project, not mine. So one test it is!

oskipa:
The build is failing on some database test. The tests pass locally. I don't think it is related to the changes I made, but if it is, please let me know, and I can try to figure out what is going on to fix it.

I saw another PR that has the same database error.

oskipa:
The tests are passing now :)

Kludex:
Cool. Thank you.

I just have one doubt here... Why does being an `int` works on Chrome? Does it work on other browsers as well?

oskipa:
> Cool. Thank you.
> 
> I just have one doubt here... Why does being an `int` works on Chrome? Does it work on other browsers as well?

Let me investigate that this morning

oskipa:
Yes, `int` works in Firefox too. This behavior doesn't match the behavior described in https://www.rfc-editor.org/rfc/rfc6265#section-5.1.1 It does match the `max-age` behavior, which does take an integer that represents the seconds which the cookie is allowed to live.

My guess is that at some point browser programmers thought that `max-age` and `expires` were semantically the same, and allowed both integers and dates to set one or the other. In Firefox, dev tools shows `expires` as `Expires/Max-Age`. Then, although not part of the RFC, it seems to be a de-facto standard, either because the other browsers implement it that way, or because it was implemented this way years ago and no one touches the cookie code since it works.

oskipa:
Is there anything else to do?

Kludex:
Actually, I've found out why we accept `int`. It's because of how SimpleCookie is implemented in CPython:

https://github.com/python/cpython/blob/921f2353675ded86899c6dfdfc1c5c57bb998638/Lib/http/cookies.py#L408

`werkzeug` implements this logic like this: https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/http.py#L1228-L1230C29

Just some references.

starlette/responses.py
```diff
@@ -35,6 +36,12 @@ def guess_type(
     return mimetypes_guess_type(url, strict)
 
 
+# Cookie expire format
+# As described by https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1
+# And defined by https://www.rfc-editor.org/rfc/rfc2616#section-3.3.1
+COOKIE_DATETIME_FORMAT = "%a, %d %b %Y %H:%M:%S GMT"
```
Kludex, line 42: Let's use https://docs.python.org/3/library/email.utils.html#email.utils.formatdate instead. :pray:

Kludex: Let's use https://docs.python.org/3/library/email.utils.html#email.utils.formatdate instead, please. :pray:

tests/test_responses.py
```diff
@@ -309,6 +312,37 @@ async def app(scope, receive, send):
...
+        response.set_cookie(
+            "mycookie",
+            "myvalue",
+            max_age=10,
+            expires=expires,
+            path="/",
+            domain="localhost",
+            secure=True,
+            httponly=True,
+            samesite="none",
```
Kludex: Can we set only the fields we are interested i.e. `expires`, `key` and `value` here?

Kludex (changes requested):

tests/test_responses.py
```diff
@@ -307,6 +313,35 @@ async def app(scope, receive, send):
...
+
+    async def app(scope, receive, send):
+        response = Response("Hello, world!", media_type="text/plain")
+        response.set_cookie("mycookie", "myvalue", expires=expires_factory())
+        await response(scope, receive, send)
+
+    client = test_client_factory(app)
+    response = client.get("/")
+    cookie: SimpleCookie = SimpleCookie(response.headers.get("set-cookie"))
+    assert cookie["mycookie"]["expires"] == "Fri, 22 Jan 2100 12:00:10 GMT"
```
Kludex: If someone can do better here, please feel free to suggest. @encode/maintainers

Kludex:
If no one has anything against, I'm going to merge this when the release is ready.

Kludex (approved):

iudeen:
Looks fine to me.

tests/test_responses.py
```diff
@@ -307,6 +312,35 @@ async def app(scope, receive, send):
     client = test_client_factory(app)
     response = client.get("/")
     assert response.text == "Hello, world!"
+    assert (
+        response.headers["set-cookie"]
+        == "mycookie=myvalue; Domain=localhost; expires=Fri, 22 Jan 2100 12:00:10 GMT; "
```
florimondmanca, line 319: I don't understand this. Why do we get `expires=<date>`? We pass `expires=10` in `set_cookie()`, so we should get that in the Set-Cookie header as well, right?

tests/test_responses.py
florimondmanca: I don't think we need the `timemachine` trick. Can we pass a well-defined datetime as input, rather than `datetime.now()`?

tests/test_responses.py
Kludex: Sorry, I didn't get the question. 🤔

florimondmanca: In the `app` above, we call `.set_cookie(..., expire=10)`. Then we get a `Set-Cookie` header where `expires` is set to an actual Internet date. I noticed just now the date seems to be `10s in the future from now`. Out of curiosity, where does this logic live?

Kludex:
@florimondmanca I tried on this commit https://github.com/encode/starlette/commit/6db2870a8a54a45bd7e58fb3474c3793a81dabf3

But it was flaky... Do you mean something different?

tests/test_responses.py
Kludex: It's how the http.SimpleCookie works. It's here: https://github.com/encode/starlette/blob/master/starlette/responses.py#L115

tests/test_responses.py
florimondmanca: Here I meant:

```python
@pytest.mark.parametrize(
    "expires, expires_output",
    [
        pytest.param(datetime.datetime(2023, 1, 1, 0, 0, 10, tzinfo=timezone.utc), "Fri, 22 Jan 2100 12:00:10 GMT", id="datetime"),
        pytest.param("Fri, 22 Jan 2100 12:00:10 GMT", "Fri, 22 Jan 2100 12:00:10 GMT", id="str"),
    ],
)

...

response.set_cookie("mycookie", "myvalue", expires=expires)

...

assert cookie["mycookie"]["expires"] == expires_output
```

If we drop the `expires=10` case (already tested above), then we don't need a `time_machine` call in this test anymore.

Kludex: The goal of the assertion on the previous test is to make sure the header is the correct, and here is to check the expires field itself.

If I don't check with an integer, then I don't need the time-machine here, but I think I still need it for the above... 🤔

florimondmanca:
@Kludex I played with it locally. I feel like pytest's `monkeypatch` is just enough for our use case here, not really a need for a full-fledged extra library?

Diff:

```diff
diff --git a/requirements.txt b/requirements.txt
index 500a9b5..15aa541 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,6 @@ flake8==3.9.2
 importlib-metadata==4.13.0
 isort==5.10.1
 mypy==0.991
-time-machine==2.9.0
 typing_extensions==4.4.0
 types-contextvars==2.4.7
 types-PyYAML==6.0.12.3
diff --git a/tests/test_responses.py b/tests/test_responses.py
index bf54772..aa2f5c4 100644
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -1,6 +1,6 @@
 import os
-from datetime import datetime, timedelta, timezone
-from email.utils import format_datetime
+import datetime as dt
+import time
 from http.cookies import SimpleCookie
 
 import anyio
@@ -291,8 +291,10 @@ def test_file_response_with_inline_disposition(tmpdir, test_client_factory):
     assert response.headers["content-disposition"] == expected_disposition
 
 
-def test_set_cookie(test_client_factory, time_machine):
-    time_machine.move_to(datetime(2100, 1, 22, 12, 0, 0))
+def test_set_cookie(test_client_factory, monkeypatch):
+    # Mock time used as a reference for `Expires` by stdlib `SimpleCookie`.
+    mocked_now = dt.datetime(2100, 1, 22, 12, 0, 0, tzinfo=dt.timezone.utc)
+    monkeypatch.setattr(time, "time", lambda: mocked_now.timestamp())
 
     async def app(scope, receive, send):
         response = Response("Hello, world!", media_type="text/plain")
@@ -320,21 +322,23 @@ def test_set_cookie(test_client_factory, time_machine):
 
 
 @pytest.mark.parametrize(
-    "expires_factory",
+    "expires",
     [
-        lambda: datetime.now(timezone.utc) + timedelta(seconds=10),
-        lambda: format_datetime(
-            datetime.now(timezone.utc) + timedelta(seconds=10), usegmt=True
+        pytest.param(
+            dt.datetime(2100, 1, 22, 12, 0, 10, tzinfo=dt.timezone.utc), id="datetime"
         ),
-        lambda: 10,
+        pytest.param("Fri, 22 Jan 2100 12:00:10 GMT", id="str"),
+        pytest.param(10, id="int"),
     ],
 )
-def test_expires_on_set_cookie(test_client_factory, expires_factory, time_machine):
-    time_machine.move_to(datetime(2100, 1, 22, 12, 0, 0))
+def test_expires_on_set_cookie(test_client_factory, monkeypatch, expires):
+    # Mock time used as a reference for `Expires` by stdlib `SimpleCookie`.
+    mocked_now = dt.datetime(2100, 1, 22, 12, 0, 0, tzinfo=dt.timezone.utc)
+    monkeypatch.setattr(time, "time", lambda: mocked_now.timestamp())
 
     async def app(scope, receive, send):
         response = Response("Hello, world!", media_type="text/plain")
-        response.set_cookie("mycookie", "myvalue", expires=expires_factory())
+        response.set_cookie("mycookie", "myvalue", expires=expires)
         await response(scope, receive, send)
 
     client = test_client_factory(app)
```

Kludex:
> @Kludex I played with it locally. I feel like pytest's `monkeypatch` is just enough for our use case here, not really a need for a full-fledged extra library?
> 
> Diff:
> 
> ```diff
> diff --git a/requirements.txt b/requirements.txt
> index 500a9b5..15aa541 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -9,7 +9,6 @@ flake8==3.9.2
>  importlib-metadata==4.13.0
>  isort==5.10.1
>  mypy==0.991
> -time-machine==2.9.0
>  typing_extensions==4.4.0
>  types-contextvars==2.4.7
>  types-PyYAML==6.0.12.3
> diff --git a/tests/test_responses.py b/tests/test_responses.py
> index bf54772..aa2f5c4 100644
> --- a/tests/test_responses.py
> +++ b/tests/test_responses.py
> @@ -1,6 +1,6 @@
>  import os
> -from datetime import datetime, timedelta, timezone
> -from email.utils import format_datetime
> +import datetime as dt
> +import time
>  from http.cookies import SimpleCookie
>  
>  import anyio
> @@ -291,8 +291,10 @@ def test_file_response_with_inline_disposition(tmpdir, test_client_factory):
>      assert response.headers["content-disposition"] == expected_disposition
>  
>  
> -def test_set_cookie(test_client_factory, time_machine):
> -    time_machine.move_to(datetime(2100, 1, 22, 12, 0, 0))
> +def test_set_cookie(test_client_factory, monkeypatch):
> +    # Mock time used as a reference for `Expires` by stdlib `SimpleCookie`.
> +    mocked_now = dt.datetime(2100, 1, 22, 12, 0, 0, tzinfo=dt.timezone.utc)
> +    monkeypatch.setattr(time, "time", lambda: mocked_now.timestamp())
>  
>      async def app(scope, receive, send):
>          response = Response("Hello, world!", media_type="text/plain")
> @@ -320,21 +322,23 @@ def test_set_cookie(test_client_factory, time_machine):
>  
>  
>  @pytest.mark.parametrize(
> -    "expires_factory",
> +    "expires",
>      [
> -        lambda: datetime.now(timezone.utc) + timedelta(seconds=10),
> -        lambda: format_datetime(
> -            datetime.now(timezone.utc) + timedelta(seconds=10), usegmt=True
> +        pytest.param(
> +            dt.datetime(2100, 1, 22, 12, 0, 10, tzinfo=dt.timezone.utc), id="datetime"
>          ),
> -        lambda: 10,
> +        pytest.param("Fri, 22 Jan 2100 12:00:10 GMT", id="str"),
> +        pytest.param(10, id="int"),
>      ],
>  )
> -def test_expires_on_set_cookie(test_client_factory, expires_factory, time_machine):
> -    time_machine.move_to(datetime(2100, 1, 22, 12, 0, 0))
> +def test_expires_on_set_cookie(test_client_factory, monkeypatch, expires):
> +    # Mock time used as a reference for `Expires` by stdlib `SimpleCookie`.
> +    mocked_now = dt.datetime(2100, 1, 22, 12, 0, 0, tzinfo=dt.timezone.utc)
> +    monkeypatch.setattr(time, "time", lambda: mocked_now.timestamp())
>  
>      async def app(scope, receive, send):
>          response = Response("Hello, world!", media_type="text/plain")
> -        response.set_cookie("mycookie", "myvalue", expires=expires_factory())
> +        response.set_cookie("mycookie", "myvalue", expires=expires)
>          await response(scope, receive, send)
>  
>      client = test_client_factory(app)
> ```

I've applied this. Thanks!

PR #1917, opened 2022-10-19
Make sure MutableHeaders._list is actually a list

adriangb:
Fixed #1909


starlette/datastructures.py
```diff
@@ -585,6 +585,16 @@ def __repr__(self) -> str:
...
 class MutableHeaders(Headers):
+
+    def __init__(
+        self,
+        headers: typing.Optional[typing.Mapping[str, str]] = None,
+        raw: typing.Optional[typing.List[typing.Tuple[bytes, bytes]]] = None,
+        scope: typing.Optional[typing.Mapping[str, typing.Any]] = None,
+    ) -> None:
+        super().__init__(headers, raw, scope)
+        self._list = list(self._list)
```
alex-oleshkevich: I think doing this (line 595) in the parent class is good enough.

alex-oleshkevich: Also a test case covering the issue is wanted.

adriangb: Feel free to update the branch, I’m on my phone and AFK the rest of the day

alex-oleshkevich: Same thing. Let's postpone it till morning.

adriangb:
Looks like the test revealed a very real issue: we assume that `MutableHeaders` updates the headers but if we cast it to a `list` were making a copy. So I had to assign this new list back to `scope`. Guess that makes sense. What do you think @alex-oleshkevich? Thanks for pushing for a test btw, it was the right call here.

alex-oleshkevich (approved):

jose-rojas0:
> Looks like the test revealed a very real issue: we assume that `MutableHeaders` updates the headers but if we cast it to a `list` were making a copy. So I had to assign this new list back to `scope`. Guess that makes sense. What do you think @alex-oleshkevich? Thanks for pushing for a test btw, it was the right call here.

thank you! this was the problem I was alluding to in https://github.com/encode/starlette/discussions/1909 - this definitely fixes it. 💯

PR #1919, opened 2022-10-20
Remove `databases` from tests

Kludex:
It looks like `databases` was a dependency when `DatabaseMiddleware` was still a thing.

I'm trying to understand the need for this test file. Is it needed? :thinking:


lovelydinosaur (approved):
> Is it needed? 🤔

Nope. Take this appendix out, doctor. 👨‍⚕️

PR #1922, opened 2022-10-25
Use Base64Url encoding and decoding in Session Cookie

### LINKED ISSUE #1259: SessionMiddleware may create malformed cookie
### Describe the bug

The cookie created may include "=" and be wrongly formatted, ignored by the browser.

### To reproduce

starlette/starlette/middleware/sessions.py:55
data = b64encode(json.dumps(scope["session"]).encode("utf-8"))

E.g. when scope["session"] = {"a":"bc"}
data becomes "eyJhIjogImJjIn0="
which generates a wrong cookie, ignored by the browser.

### Expected behavior

Equals sign should not be included in the cookie.

### Actual behavior

Equals sign might be included in the cookie.

### Environment

Python 3.9.6
MacOS
Starlette Version: 0.14.2

### Additional context

Actual cookie generated was something like this:

session=gICAgIH1cbiAgICB9XG59In0=.YQ5bnw.jlYF8N3XGf8tYCGDzrnZzW4BHPA; path=/; Max-Age=1209600; httponly; samesite=lax

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/1259">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/1259/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

allezxandre:
- [x] Initially raised as discussion #1259 

This Pull Request makes the Session JWS use Base64Url encoding _without padding_, as described by [RFC 7515, Appendix C](https://tools.ietf.org/html/rfc7515#appendix-C).

Should close #1259.

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -1,7 +1,7 @@
 import json
 import sys
 import typing
-from base64 import b64decode, b64encode
+from base64 import urlsafe_b64decode, urlsafe_b64encode
 
 import itsdangerous
 from itsdangerous.exc import BadSignature
@@ -48,7 +48,7 @@ class SessionMiddleware:
             data = connection.cookies[self.session_cookie].encode("utf-8")
             try:
                 data = self.signer.unsign(data, max_age=self.max_age)
-                scope["session"] = json.loads(b64decode(data))
+                scope["session"] = json.loads(base64url_decode(data))
                 initial_session_was_empty = False
             except BadSignature:
                 scope["session"] = {}
@@ -59,7 +59,9 @@ class SessionMiddleware:
             if message["type"] == "http.response.start":
                 if scope["session"]:
                     # We have session data to persist.
-                    data = b64encode(json.dumps(scope["session"]).encode("utf-8"))
+                    data = base64url_encode(
+                        json.dumps(scope["session"]).encode("utf-8")
+                    )
                     data = self.signer.sign(data)
                     headers = MutableHeaders(scope=message)
                     header_value = "{session_cookie}={data}; path={path}; {max_age}{security_flags}".format(  # noqa E501
@@ -84,3 +86,21 @@ class SessionMiddleware:
             await send(message)
 
         await self.app(scope, receive, send_wrapper)
+
+
+def base64url_encode(data: bytes) -> bytes:
+    """
+    Encode a byte-string to base64url.
+
+    See https://tools.ietf.org/html/rfc7515#appendix-C
+    """
+    return urlsafe_b64encode(data).rstrip(b"=")
+
+
+def base64url_decode(data: bytes) -> bytes:
+    """
+    Decode a base64url byte-string.
+
+    See https://tools.ietf.org/html/rfc7515#appendix-C
+    """
+    return urlsafe_b64decode(data + b"=" * (-len(data) % 4))


Kludex:
Thanks for the PR.

@allezxandre Would you mind rebasing it? I'm not allowed to make changes to this branch.

Kludex:
If you can provide an MRE comparing how it works on Flask and DJango, it will speed up the process here.

What I want is to see that the behavior between the web frameworks are matching.

allezxandre:
My experience with Flask is quite dated, but looking through [Flask's code](https://github.com/pallets/flask/blob/88069bd417f5fd0e271a83768391a280267d2994/src/flask/sessions.py#L409), it looks like they are encoding Cookies using a custom URL-safe serialiser from their `itsdangerous` package : https://itsdangerous.palletsprojects.com/en/2.1.x/url_safe/

I have never used Django so I wouldn't really know how to read their code.

Kludex:
But we also use `itsdangerous` on the `SessionMiddleware`... :thinking:

Kludex:
You mean to use `URLSafeTimedSerializer` instead of `json`?

allezxandre:
We could use their custom encoder.

The current code change simply proposes to replace Base64 encoding with Cookie/url-safe Base64 encoding as suggested by the JWS RFC.

I think using `itsdangerous.url_safe.URLSafeSerializer` encoding in place of url-safe Base64 would also be fine, but such a change would increase coupling with `itsdangerous`, and it would be a bigger code change as it looks like `URLSafeSerializer` makes some code (i.e. the signing) redundant.

Kludex:
> If you can provide an MRE comparing how it works on Flask and DJango, it will speed up the process here.
> 
> What I want is to see that the behavior between the web frameworks are matching.

I need this to get this PR reviewed. To speed up, if you can provide it, it will be cool. Otherwise, I'll create them over the weekend.

allezxandre:
Sorry, but what’s an MRE? Wikipedia tells me it might be a Minimum Reproducible Example?

Kludex:
Yep. A snippet with those other web frameworks that shows how they behave considering this PR.

Kludex:
Why should we use this approach, and not the one that flask uses (with double quotes)?

What does Django do?

allezxandre:
I don't think the Flask code only uses double quotes? From what I understand they first serialize the cookie using their custom URL-safe serializer.

The point of this PR was more about making the current serializing URL-safe, as it currently isn't and might create malformed cookies (see #1259), so I didn't really put much thought into what serializing method would be more appropriate, I just kept JSON Web Signature Base64 encoding and made it URL Safe.

Currently, it might sometimes serialize a Cookie that might look like this: `abcdef1234=` (with an `=` sign used for Base64 encoding), which is unsafe as it can be parsed as a key-value separator:

```http
Set-Cookie: session=abcdef1234=; max-age=31536000; domain=example.com; path=/; secure; SameSite=Lax
```

For this reason, some browsers (Safari) will escape `=` before saving, but will not un-escape it reliably when sending it back, so the Session middleware interprets it as malformed and ignores it.

hasansezertasan:
I have copied the commited `starlette/middleware/sessions.py`s content and tried the MRE below, mounted the Flask Application to a FastAPI application. Tried to get the same session data from FastAPI side that has set by Flask side and vice versa, didn't work for me.

Flask Application:

```py
# app.py
from flask import Flask, jsonify, session
import datetime

app = Flask(__name__)

config = {
    "SECRET_KEY": "super-secret",
    "PERMANENT_SESSION_LIFETIME": datetime.timedelta(days=7),
}
app.config.update(config)


@app.get("/")
def index():
    return jsonify({"message": "Hello World"})

@app.get("/set-session")
def set_session():
    session["application"] = "flask"
    session.modified = True
    return jsonify({"message": "Session set"})

@app.get("/get-session")
def get_session():
    return jsonify({"message": session.get("application")})

@app.get("/delete-session")
def delete_session():
    session.pop("application")
    session.modified = True
    return jsonify({"message": "Session deleted"})

@app.get("/get-session-data")
def get_session_data():
    print(session.keys())
    print(session.items())
    print(session.values())
    return {"message": "session"}

```

FastAPI Application:

```py
# main.py
from fastapi import FastAPI, Request, Response
from starlette.middleware.wsgi import WSGIMiddleware

from app import app as flask_app

app = FastAPI(
    title="FastAPI",
    version="0.0.1",
)
app.mount("/flask-application", WSGIMiddleware(flask_app))
app.add_middleware(
    SessionMiddleware,
    secret_key="super-secret",
)


@app.middleware("http")
async def add_process_time_header(request: Request, call_next):
    response.headers["X-Process-Time"] = "100"
    print(response.headers)
    return response


@app.get("/")
async def root(req: Request, res: Response):
    return {"message": "Hello World"}


@app.get("/set-session")
async def set_session(req: Request, res: Response):
    req.session.update({"application": "fastapi"})
    return {"message": "Session set"}


@app.get("/get-session")
async def get_session(req: Request, res: Response):
    return {"message": req.session.get("application")}


@app.get("/delete-session")
async def delete_session(req: Request, res: Response):
    req.session.pop("application")
    return {"message": "Session deleted"}


@app.get("/get-session-data")
async def get_session_data(req: Request, res: Response):
    return {"message": req.session.items()}

```

I implemented a solution in this discussion [tiangolo/fastapi#9318](https://github.com/tiangolo/fastapi/discussions/9318), just yesterday for this, it works quite well. I can send a PR if you want.

Kludex:
Closing this.

Reason on https://github.com/encode/starlette/issues/1259#issuecomment-1868470263.

PR #1936, opened 2022-11-04
♻️ Add compatibility with the next version of AnyIO

tiangolo:
♻️ Add compatibility with the next version of AnyIO

The next version of AnyIO (https://github.com/agronholm/anyio/commit/9f97cd7bb5f59f79f1bcda64230ab2127e79935a), not released yet, doesn't export `start_blocking_portal` from the top level of the package, it has to be imported from directly where it is declared, in `from_thread`.

This keeps compatibility with the current version but adds compatibility with the next (current main branch in AnyIO's repo).

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -12,6 +12,7 @@ from types import GeneratorType
 from urllib.parse import unquote, urljoin
 
 import anyio
+import anyio.from_thread
 import httpx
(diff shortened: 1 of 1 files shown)


adriangb (approved):

tiangolo:
That was fast, thanks for the review @adriangb! 🚀

PR #1609, opened 2022-04-29
Do not send empty data to `StreamingResponse` on `BaseHTTPMiddleware`

Kludex:
- This seems more correct than #1608.
- Alternative to #1579
- ~Closes #803~ not really related.

PR #1579, opened 2022-04-07
GZIP content length mismatch fix

hiranp:
This addresses concerns raised by the following discussions:
 - [Gzip Middleware content-length is incorrect #803](https://github.com/encode/starlette/issues/803)
 - [GZip Middleware Throws "Response content longer than Content-Length" when given a 304 empty-body response #4050](https://github.com/tiangolo/fastapi/issues/4050)
 - [Gzip middleware randomly raise RuntimeError but does not affect use , what are the possible causes?](https://github.com/tiangolo/fastapi/issues/2818)

We encountered this problem when developing our internal application, when we implemented both Brotli and GZIP compression. We tested the following fix using several use cases.

The solution tries to address the problem using two(2) approaches:
- Setting the respond body if NULL bytes are detected
- Not writing the message body when its empty or contains NULL byte char

- [x] Initially raised as discussion #803

PR #1940, opened 2022-11-08
Only stop receiving stream on `body_stream` if body is empty

### LINKED ISSUE #1977: middleware causes exceptions to not be raised/handled silently
### Discussed in https://github.com/encode/starlette/discussions/1976

<sup>Originally posted by **fraser-langton** December  6, 2022</sup>
From [issue in FastAPI](https://github.com/tiangolo/fastapi/issues/5173) - issue was said to be from starlette

Was initially noticed in FastAPI 0.74.0, was fixed after 0.79.0 but has since regressed (FastAPI 0.88.0 it isn't working)

When exceptions are raised on a subapp, the exceptions are not propagated all the way to see in console
In the example code
/info raises an exception and the full stacktrace is seen in console
/private/info does not raise the exception and only `INFO: 127.0.0.1:56308 - "GET /info HTTP/1.1" 500 Internal Server Error` is shown in console

```python
import uvicorn
from fastapi import FastAPI


app = FastAPI()


@app.get("/info")
def info():
    # raises Exception as expected, the traceback is seen in console
    raise Exception


private_api = FastAPI()


@private_api.get("/info")
def info():
    # exception is handled silently, no traceback is seen in console
    raise Exception


app.mount("/private", private_api)


class Middleware(BaseHTTPMiddleware):

app.add_middleware(Middleware)  # when this is removed, the exceptions are raised for all routes


if __name__ == '__main__':
    uvicorn.run(app, port=8000)
```
</div>

Kludex:
- Closes #1977

This solves the issue reported by @MrSalman333 on https://github.com/encode/starlette/pull/1609#issuecomment-1306742910.

## Test 1 (https://github.com/encode/starlette/pull/1609#issuecomment-1306786049)

```python
import logging

import anyio
from fastapi import Depends, FastAPI, Response


app = FastAPI(middleware=[Middleware(CustomMiddleware)])


def dependency_a(response: Response):
    try:
        yield "A"
    finally:
        logging.warning("A")


def dependency_b(response: Response):
    try:
        yield "B"
    finally:
        logging.warning("B")


def dependency_c(response: Response):
    try:
        yield "C"
    finally:
        logging.warning("C")


@app.get(
    "/issue",
    dependencies=[
        Depends(dependency_a),
    ],
)
def show_dependencies_issue(
    b=Depends(dependency_b),
    c=Depends(dependency_c),
):
    print(b, c)
    return {"status": "ok"}
```

## Test 2 (https://github.com/encode/starlette/pull/1579#issuecomment-1097119520)

```python
from typing import Awaitable, Callable

class NoopMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request: Request, call_next: Callable[[Request], Awaitable]):

def endpoint(request: Request):
    return Response(status_code=304)


app2 = Starlette(
    routes=[Route("/", endpoint=endpoint)],
    middleware=[Middleware(GZipMiddleware), Middleware(NoopMiddleware)]
)
```


Kludex:
Now I need to understand what's happening :rofl:

starlette/middleware/base.py
```diff
@@ -89,7 +89,7 @@ async def body_stream() -> typing.AsyncGenerator[bytes, None]:
                         body = message.get("body", b"")
                         if body:
                             yield body
-                        if not message.get("more_body", False):
+                        elif not message.get("more_body", False):
                             break
```
Kludex: Removing those two lines also solves the issue.

Kludex:
```suggestion
```

Kludex: I've removed them instead.

Kludex:
I don't know why we have the issue... It would be cool if we could find out the reason, but this change solves the issue and doesn't bring any regression.

adriangb:
I don't see how this change can hurt, but I'd like to get some more input and maybe do a little post mortem on how we got here since quickly merging a "this can't hurt" PR last time is what hurt us in the end. BaseHTTMiddleware is a complex beast!

adriangb:
@jhominal I would really appreciate your thoughts on this if you can take a look

MrSalman333:
This is what I found originally ,, 
I didn't open a PR though since the PR that added this line was done to solve a bug since it was alternative to https://github.com/encode/starlette/pull/1579

Kludex:
Some findings: this only happens with more than one _sync_ dependency.

That's it, no more findings.

Kludex:
Ok, if we add a `await asyncio.sleep` after the `yield` on the async dependency, I'm able to see the cancellation.

Kludex:
I know why we are able to see the logs after some more HTTP calls: it's because the garbage collector is acting there.

Kludex:
Got it. The `contextmanager_in_threadpool` fails completely, but it's the `contextmanager` itself that is able to conclude when the garbage collector acts, see:
```python
async def solve_generator(
    *, call: Callable[..., Any], stack: AsyncExitStack, sub_values: Dict[str, Any]
) -> Any:
    if is_gen_callable(call):
        cm = contextmanager_in_threadpool(contextmanager(call)(**sub_values))
    elif is_async_gen_callable(call):
        cm = asynccontextmanager(call)(**sub_values)
    return await stack.enter_async_context(cm)
```

Kludex:
I'm tired. I'm going to write a report tomorrow morning, but the last messages show what is happening (maybe not in a clear way).

Kludex:
This happens in FastAPI because they change the middleware order on the `build_stack_middleware`, adding the `AsyncExitMiddleware`. That's not actually a problem, just pointing out why FastAPI is able to experience this issue. :sweat_smile:

We are wrong here in breaking the _receive stream_.

Kludex:
To test this, we can add a middleware wrapped by the `BaseHTTPMiddleware` and verify that it runs until the end.

I'll add this test tomorrow.

(I think this makes sense... :eyes:)

adriangb:
Thank you for digging @Kludex !

MrSalman333:
> To test this, we can add a middleware wrapped by the `BaseHTTPMiddleware` and verify that it runs until the end.
> 
> I'll add this test tomorrow.
> 
> (I think this makes sense... 👀)

thanks @Kludex !!

keep in mind. this behavior is not exposed in the TestClient.

Kludex:
One more finding... Actually, the issue doesn't exist in the latest Starlette, so FastAPI just needs to bump Starlette to 0.21.0.

@jhominal Solved this on https://github.com/encode/starlette/pull/1715 as well... Specifically mentioned on this sentence:
image

Kludex:
On my tests, I was going back in time to the PR that introduced, but once I used the latest Starlette I was not able to reproduce it... :eyes:

Kludex:
We even have a test for this behavior:

https://github.com/encode/starlette/blob/bf5d7357b544b48b96844d3274ea558df079571d/tests/middleware/test_base.py#L266-L325

Kludex:
In any case, I'll write a report because there was a mistake.

Some months ago, there was a PR (#1579) that tried to fix an issue reported in FastAPI (https://github.com/tiangolo/fastapi/issues/4050). I was able to [reproduce that issue](https://github.com/encode/starlette/pull/1579#issuecomment-1097119520), but I concluded that the issue was not in the `GZipMiddleware` (as the PR proposed), but actually in the `BaseHTTPMiddleware`. For that reason, I've created a PR that fixed the issue on the `BaseHTTPMiddleware` (#1609).

The PR introduced two changes:
1. It didn't send empty data to the `StreamingResponse`;
2. and it closed the `receive` stream.

I overstepped there, as the `2` was not needed, and actually caused an issue: inner middlewares wouldn't be able to run as expected, as there was a `CancellationError`. On the other hand, the `1` is fine.

On the process of merging this PR, what we lacked on the author (me) side was to create a test for the behavior change to avoid regressions. Specifically, he needed to come up with a test case like what [was created](https://github.com/encode/starlette/blob/bf5d7357b544b48b96844d3274ea558df079571d/tests/middleware/test_base.py#L266-L325) on #1715. As for the reviewers side, they thought the changes were fine based on the amount of changes, and considering that there was no drop of coverage, I think it made sense to have that consideration. What the reviewers could have done differently was to insist on a test case, **maybe**.

We took a lot to notice this, and it's not even relevant right now because FastAPI pins each version of Starlette. Currently, latest FastAPI runs with Starlette 0.20.4, which contains the issue. The thing is, this behavior was fixed on #1715, which was released on Starlette 0.21.0. On the description of that PR, we can already see the intention of solving the case of the `receive` stream being cancelled.

To conclude:
- This issue doesn't exist on the latest Starlette.
- This PR is valid, but not urgent.
- There's already a [test for it](https://github.com/encode/starlette/blob/bf5d7357b544b48b96844d3274ea558df079571d/tests/middleware/test_base.py#L266-L325).
- FastAPI needs to bump Starlette to 0.21.0.

Kludex:
This is the solution for https://github.com/encode/starlette/issues/1977. It looks like my report above is not quite precise.

And it looks like we could benefit from another test case here... Ideas?

lovelydinosaur (approved):
Given that we're not regressing any test cases, I think code removal *without additional tests* is okay.

Kludex:
I still want a test to cover this, but I'll merge this as is.

Issue #1931, opened 2022-11-02
IPv6 handling bug in starlette.datastructures.URL


dimaqq:
`URL` uses stdlib `urlsplit` to parse the string URL,

however, it uses this code to recombine the host and port.

```py
            netloc = hostname                                                           
            if port is not None:                                                        
                netloc += f":{port}"  
```

Consider an IPv6 URL, like `http://[::1]:8080/`
* urlsplit will strip the quare brackets and produce `::1` 
* manual concatenation will yield `::1:8080`
* which is not a valid netloc

dimaqq:
MRE

```
In [5]: import starlette.datastructures

In [6]: starlette.datastructures.URL("https://[fe::2]:12345").replace(port=42)
Out[6]: URL('https://fe::2:42')
```

dimaqq:
@Kludex I think `good-first-issue` tag would apply here 🙏🏻

Kludex:
Ok :eyes:

So... If it's too complicated, it's on @dimaqq :eyes:

kousikmitra:
Hi @Kludex I have raised a PR for this https://github.com/encode/starlette/pull/1965, This is my first PR in this repo so do let me know if I missed anything.

brycefisher:
@kousikmitra thanks for your fix PR!! Looks like you might need to pull in recent changes from the main branch.

@Kludex any other feedback you have for @kousikmitra? I'd love to get this fix merged and brought into FastAPI for my own project. Thanks everyone for all your contributions!

Kludex:
This was solved on #1965, and it will be available on Starlette 0.24.0.

PR #1965, opened 2022-11-25
Fix url parsing of ipv6 urls on `URL.replace`

kousikmitra:


starlette/datastructures.py
```diff
@@ -114,11 +114,18 @@ def replace(self, **kwargs: typing.Any) -> "URL":
...
             port = kwargs.pop("port", self.port)
             username = kwargs.pop("username", self.username)
             password = kwargs.pop("password", self.password)
 
+            if not hostname:
+                netloc = self.netloc
+                _, _, hostname = netloc.rpartition("@")
+
+                if hostname[-1] != "]":
+                    hostname = hostname.rsplit(":", 1)[0]
```
dimaqq: I guess the new approach breaks the existing undocumented, untested functionality...

May I propose a test like this:

`assert URL("http://u:p@host/").replace(hostname="bar") == URL("http://u:p@bar/")`

dimaqq: Or maybe like this:

`assert URL("http://u:p@host:80").replace(port=88) == URL("http://u:p@host:88")`

dimaqq (commented):
Would it be better to use `urlunparse` or `urlunsplit` from python stdlib instead of manual string concatention?

starlette/datastructures.py
kousikmitra: added, these are passing

kousikmitra:
> Would it be better to use `urlunparse` or `urlunsplit` from python stdlib instead of manual string concatention?

@dimaqq not sure how you wanted to use `urlunsplit` here! `urlunsplit` won't join netloc fields. can you explain more?

dimaqq:
> @kousikmitra
[Merge branch 'master' into fix/ipv6-url-bug](https://github.com/encode/starlette/pull/1965/commits/85464a965305326af6f7d0141d424896fd1e0aeb)

Instead of the above, I would generally recommend to rebase your branch on top of master and then force-push your branch.
I don't know what git flow this particular project uses though.

adriangb:
We squash merge into master, so it's not really important how the branch is kept up to date with master or how the commits on the branch are organized. Of course a nice linear history with well described atomic commits is a huge plus as a reviewer, but we certainly don't require it.

kousikmitra:
> > @kousikmitra
> > [Merge branch 'master' into fix/ipv6-url-bug](https://github.com/encode/starlette/pull/1965/commits/85464a965305326af6f7d0141d424896fd1e0aeb)
> 
> Instead of the above, I would generally recommend to rebase your branch on top of master and then force-push your branch. I don't know what git flow this particular project uses though.

I prefer the same, my bad I didn't notice github used merge method to update branch.

dimaqq:
my 2c: it's best to disable GitHub's magic "update the branch" in repo settings.
anyway I'm not the maintainer, I don't get to make these decisions.

dimaqq (approved):
Looks good and solves the problem that I had.

All of these work fine:

```py
In [8]: URL("http://[1::fe]/123").replace(port=88, username="uu", password="pp").replace(port=None, username=None, password=None)
Out[8]: URL('http://[1::fe]/123')
```

There's still one gotcha, but I suspect this also didn't really work before either, the IPv6 hostname needs to be bracketed, which is a bit unexpected:

```py

In [14]: URL("http://foo.com/123").replace(hostname="fe::1")
Out[14]: URL('http://fe::1/123')

In [15]: URL("http://foo.com/123").replace(hostname="[fe::1]")
Out[15]: URL('http://[fe::1]/123')
```

Which also leads, in one case to an exception when using IPv6 shorthands:
```py
In [18]: URL("http://foo.com/123").replace(hostname="1::23").replace(port=88)
ValueError: Port could not be cast to integer value as ':23'
```

This path triggered same error before.

So, strictly speaking, this PR fixes #1931 and doesn't introduce a regression, I'd say let's merge it!

dimaqq:
P.S. by comparison, other libraries take care of bracketing ipv6 addresses themselves, they don't ask the caller to do it:
```py
In [7]: yarl.URL("http://foo.com/1").with_host("123::fe")
Out[7]: URL('http://[123::fe]/1')
```

kousikmitra:
> There's still one gotcha, but I suspect this also didn't really work before either, the IPv6 hostname needs to be bracketed, which is a bit unexpected:

> Which also leads, in one case to an exception when using IPv6 shorthands:

To tackle this we should not be relying on `urlsplit` hostname instead we can infer hostname, port from netloc.

dimaqq:
🤔 I was under the impression that `urlsplit` does the right thing; and this library doesn't.

my 2c: `.update(hostname=...)` value needs to be examined, if it contains a colon, then wrap it in square brackets.
Edit: and maybe the inner logic of splitting the netloc....

Anyway, that can be a separate PR.
Edit: perhaps maintainers would like a discussion on hacking URLs in this data structure (extra code and maintenance) vs. adopting a 3p URL library (an extra dep).

starlette/datastructures.py
```diff
@@ -114,11 +114,18 @@ def replace(self, **kwargs: typing.Any) -> "URL":
             or "hostname" in kwargs
             or "port" in kwargs
         ):
-            hostname = kwargs.pop("hostname", self.hostname)
+            hostname = kwargs.pop("hostname", None)
             port = kwargs.pop("port", self.port)
             username = kwargs.pop("username", self.username)
             password = kwargs.pop("password", self.password)
 
+            if not hostname:
```
florimondmanca: Should we check against `None` instead ("not given in `kwargs`")?

```suggestion
            if hostname is None:
```

florimondmanca (approved):
LGTM

PR #1966, opened 2022-11-28
Test client init to support headers argument

roy-pstr:

- [x] Initially raised as discussion #1961

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -389,7 +389,8 @@ class TestClient(httpx.Client):
         if headers is None:
             headers = {}
         if "user-agent" not in headers:
-            headers["user-agent"] = "testclient" # this is for backwards compatibility in case no user-agent is set in the given headers
+            # for backwards compatibility in case no user-agent is set
+            headers["user-agent"] = "testclient"
         super().__init__(
             app=self.app,
             base_url=base_url,
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -388,13 +388,12 @@ class TestClient(httpx.Client):
         )
         if headers is None:
             headers = {}
+        if "user-agent" not in headers:
+            headers["user-agent"] = "testclient" # this is for backwards compatibility in case no user-agent is set in the given headers
         super().__init__(
             app=self.app,
             base_url=base_url,
-            headers=headers
-            | {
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -385,10 +386,15 @@ def __init__(
             raise_server_exceptions=raise_server_exceptions,
             root_path=root_path,
         )
+        if headers is None:
+            headers = {}
+        if "user-agent" not in headers:
```
danielballan: This might be a good place to use `headers.setdefault("user-agent", "testclient")`.

starlette/testclient.py
```diff
@@ -385,10 +386,15 @@ def __init__(
             raise_server_exceptions=raise_server_exceptions,
             root_path=root_path,
         )
+        if headers is None:
+            headers = {}
+        headers.setdefault(
+            "user-agent", "testclient"
+        )  # for backwards compatibility in case no user-agent is set
```
```suggestion
        headers.setdefault("user-agent", "testclient")
```

tests/test_testclient.py
```diff
@@ -67,6 +68,27 @@ def homepage(request):
     assert response.json() == {"mock": "example"}
 
 
+def test_testclient_headers_behavior():
+    """
+    We should be able to use the test client with user defined headers.
+
+    This could useful if we need to set custom headers for authentication
```
Kludex:
```suggestion
    This is useful if we need to set custom headers for authentication
```

tests/test_testclient.py
```diff
@@ -67,6 +68,27 @@ def homepage(request):
...
+    during tests or in development.
+
+    make sure default value of "user-agent" is "testclient" unless overridden.
```
Kludex:
```suggestion
```

Kludex (changes requested):

Kludex (approved):

Kludex:
Thanks @roy-pstr ! :)

Kludex:
Just for record keeping...

Even though it's possible to add the headers to the `TestClient` after the client is instantiated, it's more comfortable to add the headers' parameter. It's not only about practicality, but I believe people will make a good use of this feature.

Also, do not take this PR as an argument to add the other parameters supported by `httpx` client.

PR #1980, opened 2022-12-09
middleware/wsgi: remove unnecessary threads

uSpike:
With the addition of `send_nowait()` to memory object streams in anyio 3.0, there are some extraneous threads that can be removed as the underlying stream is configured to buffer to infinity and this will never block.

I didn't start a discussion, just noticed that this can be simplified a bit.  I bet it'd improve performance a bit.

---


Kludex:
On the other hand... The `WSGIMiddleware` is deprecated. :thinking:

uSpike:
Ah I didn't notice that, OK I will close then thanks!

PR #1984, opened 2022-12-12
add invalid token status code to status.py

likecodingloveproblems:
INVALID_TOKEN status code is 498, but it's not available in status.
so I add it.


Kludex:
I don't see 498 mentioned on https://www.rfc-editor.org/rfc/rfc9110...

likecodingloveproblems:
Isn't [this](https://http.dev/498) valid? @Kludex

Kludex:
If you click on your own link, it says: "...is an unofficial client error..." 👀

PR #1987, opened 2022-12-20
Use `MiddlewareType` as a type hint for middlewares

Viicos:
This might brake things in other repositories (fastapi for instance), but this improves type safety in my opinion.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -134,7 +134,9 @@ class Starlette:
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(self, middleware_class: MiddlewareType, **options: typing.Any) -> None:
+    def add_middleware(
+        self, middleware_class: MiddlewareType, **options: typing.Any
+    ) -> None:
         if self.middleware_stack is not None:  # pragma: no cover
             raise RuntimeError("Cannot add middleware after an application has started")
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -134,9 +134,9 @@ class Starlette:
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(
-        self, middleware_class: MiddlewareType, **options: typing.Any
-    ) -> None:  # pragma: no cover
+    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+        if self.middleware_stack is not None:  # pragma: no cover
+            raise RuntimeError("Cannot add middleware after an application has started")
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
 
     def add_exception_handler(
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -9,7 +9,7 @@ from starlette.middleware.exceptions import ExceptionMiddleware
 from starlette.requests import Request
 from starlette.responses import Response
 from starlette.routing import BaseRoute, Router
-from starlette.types import ASGIApp, Lifespan, Receive, Scope, Send
+from starlette.types import ASGIApp, Lifespan, MiddlewareType, Receive, Scope, Send
 
 AppType = typing.TypeVar("AppType", bound="Starlette")
 
@@ -135,7 +135,7 @@ class Starlette:
         self.router.host(host, app=app, name=name)
 
     def add_middleware(
-        self, middleware_class: typing.Type[ASGIApp], **options: typing.Any
+        self, middleware_class: MiddlewareType, **options: typing.Any
     ) -> None:  # pragma: no cover
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
 
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -134,9 +134,9 @@ class Starlette:
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
-        if self.middleware_stack is not None:  # pragma: no cover
-            raise RuntimeError("Cannot add middleware after an application has started")
+    def add_middleware(
+        self, middleware_class: typing.Type[ASGIApp], **options: typing.Any
+    ) -> None:  # pragma: no cover
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
 
     def add_exception_handler(
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -134,7 +134,7 @@ class Starlette:
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(self, middleware_class: MiddlewareType, **options: typing.Any) -> None:
         if self.middleware_stack is not None:  # pragma: no cover
             raise RuntimeError("Cannot add middleware after an application has started")
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
--- a/starlette/middleware/__init__.py
+++ b/starlette/middleware/__init__.py
@@ -1,8 +1,10 @@
 import typing
 
+from starlette.types import ASGIApp
+
 
 class Middleware:
-    def __init__(self, cls: type, **options: typing.Any) -> None:
+    def __init__(self, cls: typing.Type[ASGIApp], **options: typing.Any) -> None:
         self.cls = cls
         self.options = options
 
--- a/starlette/middleware/__init__.py
+++ b/starlette/middleware/__init__.py
@@ -1,10 +1,10 @@
 import typing
 
-from starlette.types import ASGIApp
+from starlette.types import MiddlewareType
 
 
 class Middleware:
-    def __init__(self, cls: typing.Type[ASGIApp], **options: typing.Any) -> None:
+    def __init__(self, cls: MiddlewareType, **options: typing.Any) -> None:
         self.cls = cls
         self.options = options
 
--- a/starlette/types.py
+++ b/starlette/types.py
@@ -1,4 +1,10 @@
+import sys
 import typing
+
+if sys.version_info < (3, 8):  # pragma: no cover
+    from typing_extensions import Protocol
+else:  # pragma: no cover
+    from typing import Protocol
 
 AppType = typing.TypeVar("AppType")
 
@@ -15,3 +21,15 @@ StatefulLifespan = typing.Callable[
     [AppType], typing.AsyncContextManager[typing.Mapping[str, typing.Any]]
 ]
 Lifespan = typing.Union[StatelessLifespan[AppType], StatefulLifespan[AppType]]
+
+
+# This callable protocol can both be used to represent a function returning
+# an ASGIApp, or a class with an __init__ method matching this __call__ signature
+# and a __call__ method matching the ASGIApp signature.
+class MiddlewareType(Protocol):
+    __name__: str
+
+    def __call__(
+        self, *args: typing.Any, **kwargs: typing.Any
+    ) -> ASGIApp:  # pragma: no cover
+        ...
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -1,4 +1,5 @@
 import contextvars
+import typing
 from contextlib import AsyncExitStack
 
 import anyio
@@ -193,7 +194,7 @@ class CustomMiddlewareUsingBaseHTTPMiddleware(BaseHTTPMiddleware):
         ),
     ],
 )
-def test_contextvars(test_client_factory, middleware_cls: type):
+def test_contextvars(test_client_factory, middleware_cls: typing.Type[ASGIApp]):
     # this has to be an async endpoint because Starlette calls run_in_threadpool
     # on sync endpoints which has it's own set of peculiarities w.r.t propagating
     # contextvars (it propagates them forwards but not backwards)
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -1,5 +1,4 @@
 import contextvars
-import typing
 from contextlib import AsyncExitStack
 
 import anyio
@@ -11,7 +10,7 @@ from starlette.middleware import Middleware
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.responses import PlainTextResponse, StreamingResponse
 from starlette.routing import Route, WebSocketRoute
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp, MiddlewareType, Receive, Scope, Send
 
 
 class CustomMiddleware(BaseHTTPMiddleware):
@@ -194,7 +193,7 @@ class CustomMiddlewareUsingBaseHTTPMiddleware(BaseHTTPMiddleware):
         ),
     ],
 )
-def test_contextvars(test_client_factory, middleware_cls: typing.Type[ASGIApp]):
+def test_contextvars(test_client_factory, middleware_cls: MiddlewareType):
     # this has to be an async endpoint because Starlette calls run_in_threadpool
     # on sync endpoints which has it's own set of peculiarities w.r.t propagating
     # contextvars (it propagates them forwards but not backwards)
--- a/tests/middleware/test_middleware.py
+++ b/tests/middleware/test_middleware.py
@@ -1,8 +1,10 @@
 from starlette.middleware import Middleware
+from starlette.types import Receive, Scope, Send
 
 
 class CustomMiddleware:
-    pass
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        return None  # pragma: no cover
 
 
 def test_middleware_repr():


Kludex:
I think this was already discussed before. :thinking:

Viicos:
> I think this was already discussed before. 🤔

I've tried looking in the already existing issues/PRs/discussions but couldn't find anything 😕

Kludex:
- https://github.com/encode/starlette/pull/1176/

EDIT: I think my comment says that we should be doing what you are trying to improve, but that PR was doing too many things... :eyes: 

EDIT 2: This doesn't mean that I agree with those changes, I didn't even check them yet... jfyk

Viicos:
Note to self: import `Protocol` for `typing_extensions` to support 3.7

Viicos:
Waiting for directives regarding the coverage fail.

Kludex:
> This might brake things in other repositories (fastapi for instance), but this improves type safety in my opinion.

What does this break on FastAPI?

alex-oleshkevich:
Middlewares are just ASGI applications. `ASGIApp` type hint works for them as well.

Viicos:
> > This might brake things in other repositories (fastapi for instance), but this improves type safety in my opinion.
> 
> What does this break on FastAPI?

Sorry I wasn't clear on that point, it doesn't break anything for projects that depend on starlette, but it might break type checking if one is passing an unknown type to `add_middleware`

> Middlewares are just ASGI applications. `ASGIApp` type hint works for them as well.

True indeed, I went for the `Protocol` definition but it looks like using `Callable` is equivalent (and doesn't require `typing_extensions`). Having an explicit type for middlewares improve readability though, so maybe we could use something like:

```python
Middleware: TypeAlias = ASGIApp
```

This is up to you

starlette/types.py
```diff
@@ -1,9 +1,18 @@
...
+Scope: TypeAlias = typing.MutableMapping[str, typing.Any]
+Message: TypeAlias = typing.MutableMapping[str, typing.Any]
 
-ASGIApp = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
+Receive: TypeAlias = typing.Callable[[], typing.Awaitable[Message]]
+Send: TypeAlias = typing.Callable[[Message], typing.Awaitable[None]]
+
+ASGIApp: TypeAlias = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
+
+MiddewareType: TypeAlias = ASGIApp
```
Kludex: Is `TypeAlias` needed?

starlette/types.py
```diff
@@ -1,9 +1,18 @@
...
+"""Type alias used for middleware types."""
```
```suggestion
```

starlette/types.py
```diff
@@ -1,9 +1,18 @@
...
-Receive = typing.Callable[[], typing.Awaitable[Message]]
-Send = typing.Callable[[Message], typing.Awaitable[None]]
+Scope: TypeAlias = typing.MutableMapping[str, typing.Any]
+Message: TypeAlias = typing.MutableMapping[str, typing.Any]
 
-ASGIApp = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
+Receive: TypeAlias = typing.Callable[[], typing.Awaitable[Message]]
+Send: TypeAlias = typing.Callable[[Message], typing.Awaitable[None]]
+
+ASGIApp: TypeAlias = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
```
Kludex: Can we change the order back to what it was?

Kludex (dismissed):

starlette/types.py
Viicos: See https://peps.python.org/pep-0613/. Here it does not add value, as it would be implicitly determined as a `TypeAlias` any way, but as `typing_extensions` is used in this project I figured why not use it, that way we know explicitly it is a `TypeAlias`.

starlette/types.py
Viicos: Looking at https://github.com/encode/starlette/blob/master/starlette/types.py the order is still the same

Kludex: Oh, sorry. hahaha

Kludex:
Should we just use ASGIApp instead of creating this new type? :thinking:

Viicos:
> Should we just use ASGIApp instead of creating this new type? 🤔

This is purely to avoid confusion, and to avoid having people thinking they need to pass in a `ASGIApp` in the `Middleware`'s `__init__` constructor, even though as stated by @alex-oleshkevich they are the same 🙂

alex-oleshkevich:
@Kludex after thinking about this, I found that it makes sense to have dedicated types for such components. It makes things very specific. It should not affect only middlewares, but any other component we may find important.
`MiddlewareType`, `ContextProcessorType`, another other callback type, etc. In the same way, we have types for `Receive` and `Send`.

I would not drop this PR but rather initiate a deeper discussion on this topic.

odiseo0:
Is `MiddewareType` the correct name? It has a typo in it, or is intended?

starlette/types.py
```diff
@@ -1,9 +1,17 @@
...
+
+MiddewareType: TypeAlias = ASGIApp
```
abersheeran: https://github.com/encode/starlette/blob/337ae243b10d3c20db53b77bc3c7718bb4f1a164/starlette/applications.py#L103-L104

Considering this code, should we use
```python
class MiddlewareType(Protocol):
    def __call__(
        self,
        app: ASGIApp,
        **options: Any
    ) -> ASGIApp:
        ...
```

Viicos: If you look at the [`__iter__`](https://github.com/encode/starlette/blob/337ae243b10d3c20db53b77bc3c7718bb4f1a164/starlette/middleware/__init__.py#L9-L11) method of the `Middleware` class, you will see that it returns a tuple whose first element is the middleware class, which corresponds to the `cls` variable here

BilalAlpaslan (approved):

starlette/types.py
abersheeran: `cls` is `MiddlewareType`.

Viicos: Yes indeed sorry, there's some confusion here between the `__init__` and the `__call__` method. When doing `app = cls(app=app, **options)`, we are actually instantiating the middleware, not using the `__call__` method. (That is because `cls` is `Type[MiddlewareType]`, not `MiddlewareType`).

Defining the `MiddlewareType` as a `Callable` or as a `Protocol` implementing `__call__` is equivalent, as `Callable` is an ABC implementing `__call__`

starlette/middleware/__init__.py
```diff
@@ -1,8 +1,10 @@
 import typing
 
+from starlette.types import MiddlewareType
+
 
 class Middleware:
-    def __init__(self, cls: type, **options: typing.Any) -> None:
+    def __init__(self, cls: typing.Type[MiddlewareType], **options: typing.Any) -> None:
```
Viicos: Before this gets merged, maybe using another name as `MiddlewareType` is more suitable? (Perhaps `MiddlewareT`). Having `typing.Type[MiddlewareType]` could lead to some confusion (type of a type? 🤨). Originally, I've named it `MiddlewareType` to indicate that this is a type hint of a middleware hence my suggestion of using `MiddlewareT` instead, the `T` being pretty common to indicate type hints

starlette/types.py
abersheeran: This is a ASGI middleware that can be used in starlette.

```python
class ASGIMiddleware:
    def __init__(app: ASGIApp, p0: str, p1: int) -> None:
        ...
    async def __call__(self, scope, receive, send) -> None:
        ...
```

This also can be used in starlette.

```python
def asgi_middleware(app, p0: str, p1: int) -> ASGIApp:
    async def _asgi(scope, receive, send) -> None:
        ...
    return _asgi
```

But this can't be used in starlette. Even though `P` passes the type check you're giving now.

```python
class P:
    def __init__(self) -> None: ...
    def __call__(self, scope, receive, send) -> None: ...
```

abersheeran: If we are going to write Type Annotation, then the parameter type needs to be as precise as possible, otherwise it will not bring any benefit.

abersheeran: > https://github.com/encode/starlette/blob/337ae243b10d3c20db53b77bc3c7718bb4f1a164/starlette/applications.py#L103-L104
> 
> Considering this code, should we use
> 
> ```python
> class MiddlewareType(Protocol):
>     def __call__(
>         self,
>         app: ASGIApp,
>         **options: Any
>     ) -> ASGIApp:
>         ...
> ```

In other places where `typing.Type[MiddlewareType]` is now used, just change to `MiddlewareType`.

michaeloliverx: > If we are going to write Type Annotation, then the parameter type needs to be as precise as possible, otherwise it will not bring any benefit.

**Without** having read any of the other discussion here, I would tend to agree, incorrect type annotations are worse than none.

Viicos: Too bad protocols aren't supporting the `__init__` method (for reference: https://bugs.python.org/issue44807, where a use case similar to this one is discussed). If I understand correctly, by using your suggestion the `__call__` method is actually used as the `__init__` method? So we get the `__init__` method correctly typed but we loose information on the real async `__call__` :/

abersheeran: > Too bad protocols aren't supporting the `__init__` method (for reference: https://bugs.python.org/issue44807, where a use case similar to this one is discussed). If I understand correctly, by using your suggestion the `__call__` method is actually used as the `__init__` method? So we get the `__init__` method correctly typed but we loose information on the real async `__call__` :/

No, this is just a way of specifying the name of the `Callable` parameter. In fact it is not limited to classes.

Viicos: I'm not sure I'm following you on this one. What I understand is that middlewares should have the following implementation:

```python
class MyMiddleware:
    def __init__(self, app: ASGIApp, **options: Any) -> None:
        ...

        ...
```

So considering the following code: https://github.com/encode/starlette/blob/337ae243b10d3c20db53b77bc3c7718bb4f1a164/starlette/applications.py#L103-L104

`cls` is actually the class of the middleware, not an instance. Hence calling `cls(app=app, **options)` is actually calling the `__init__` method of the middleware and instantiating it. The problem I see we are facing is that there's no way to specify the `__init__` method signature in a `Protocol` definition

abersheeran:
```python
class MyMiddleware:
    def __init__(self, app: ASGIApp, **options: Any) -> None:
        ...

        ...
```

This class passes the following type checks.

```python
class MiddlewareType(Protocol):
    def __call__(
        self,
        app: ASGIApp,
        **options: Any
    ) -> ASGIApp:
        ...
```

Viicos: It might, but imo this is not the way to go:
- You are using the `__call__` method as a way to define the signature of the `__init__` method of a middleware, which is confusing
- the `MiddlewareType` you created does not have any signature for the real async `__call__` method that should be defined on a middleware class

starlette/middleware/__init__.py
odiseo0: I prefer `MiddlewareType`

starlette/types.py
abersheeran: If you don't know how to use typing.Protocol, please consider using mypy to give it a try. 😊

starlette/applications.py
```diff
@@ -138,7 +138,7 @@ def host(
         self.router.host(host, app=app, name=name)
 
     def add_middleware(
-        self, middleware_class: type, **options: typing.Any
+        self, middleware_class: typing.Type[MiddlewareType], **options: typing.Any
```
florimondmanca: Rather than introducing a new type alias, is there a reason not to reuse `ASGIApp` here? As in, `middleware_class: Type[ASGIApp]`

A middleware instance _is_ an ASGI application, so that would make sense to me.

It would also avoid adding a new item to the public API when we can already do with existing types.

starlette/types.py
```diff
@@ -1,9 +1,17 @@
...
-Scope = typing.MutableMapping[str, typing.Any]
-Message = typing.MutableMapping[str, typing.Any]
+if sys.version_info >= (3, 10):  # pragma: no cover
+    from typing import TypeAlias
+else:  # pragma: no cover
+    from typing_extensions import TypeAlias
 
-Receive = typing.Callable[[], typing.Awaitable[Message]]
-Send = typing.Callable[[Message], typing.Awaitable[None]]
+Scope: TypeAlias = typing.MutableMapping[str, typing.Any]
```
florimondmanca: Is there a motivation for explicit type aliases here? Does the lack of TypeAlias cause issues? These changes seem unrelated to the goal of this PR (refining the `middleware_class: type` declaration), maybe we should consider a separate PR and discussion for this.

starlette/applications.py
Viicos: It was discussed in the comments above (2 weeks ago), and I gave an explanation on why I've created an alias here. If you want to avoid having a new type, I can use `ASGIApp` instead.

starlette/types.py
Viicos: Was discussed in a review by @Kludex above as well.

> See https://peps.python.org/pep-0613/. Here it does not add value, as it would be implicitly determined as a `TypeAlias` any way, but as `typing_extensions` is used in this project I figured why not use it, that way we know explicitly it is a `TypeAlias`.

I can revert this one as well

starlette/types.py
Viicos: I know how Protocols work, but you haven't explained why you are using these parameters for the `__call__` method (see my answers above). As I understand it currently, you are using the `__call__` method of your protocol as a way to define the `__init__` method of a middleware class, right?

starlette/types.py
Kludex: Yep, please do. :pray:

starlette/applications.py
Kludex: Please do. :pray:

Kludex:
I don't know why the pipeline is not running, but maybe because you need to rebase (?).

Viicos:
woops rebase went wrong

starlette/types.py
Viicos: Will continue discussion in main thread for clarity. Sorry for the confusion, it seems your suggestion is working, I couldn't get it to work since I was still on `mypy==0.991`.

Viicos:
To clarify things, I wasn't careful enough when defining the `cls` type hint. To sum things up:

`ASGIApp` is currently defined as `Callable[[Scope, Receive, Send], typing.Awaitable[None]]`

so there're two ways of defining it, either with a class and a `__call__` method, or simply with a function respecting the signature above.

Now regarding the constructor of `Middleware`:

https://github.com/encode/starlette/blob/e594de77488e346b354f31d8900499ac20ebe7d3/starlette/middleware/__init__.py#L4-L5

`cls` is an object that can be called or initialized and should return a `ASGIApp` (so both [definitions](https://github.com/encode/starlette/pull/1987#discussion_r1106876601) by @abersheeran are correct).

If `ASGIApp` is a class, it makes sense to use the `Type[]` annotation, but as `ASGIApp` is defined as a callable, we can't use `Type[Callable[...]]` in this context (`mypy` throws an error when trying to instanciate such an object).

So the workaround provided by @abersheeran seems to be working:

```python
class MiddlewareType(Protocol):
    def __call__(self, app: ASGIApp, **options: Any) -> ASGIApp:
        ...
```

We can easily see that the following definition ~~passes the type check~~ (doesn't, see https://github.com/encode/starlette/pull/1987#issuecomment-1517418039):

```python
def asgi_middleware(app, p0: str, p1: int) -> ASGIApp:
    async def _asgi(scope, receive, send) -> None:
        ...
    return _asgi
```

Now what if our middleware is defined as a class?

```python
class MyMiddleware:
    def __init__(self, app: ASGIApp, **options: Any) -> None:
        ...

        ...
```

Well it seems that since `mypy>=1.0.0`, the `__init__` method will be treated as the `__call__` method. This is still a bit hacky, and I don't know if ~~this is intentional~~ (it is, see linked `mypy` issue below).

abersheeran:
Seems like you need to resolve conflict issues to run CI

Viicos:
I believe mypy is confused about the `__init__` and `__call__` methods. I'll open an issue @`mypy`, as it is not stated explicitly in the docs that this is supported.

Viicos:
@abersheeran sorry for the delay, this is how it looks now:

```python
class MiddlewareType(Protocol):
    __name__: str

    def __call__(self, *args: typing.Any, **kwargs: typing.Any) -> ASGIApp:  # pragma: no cover
        ...
```

Unfortunately, I can't enforce the `app` argument to be there, see the linked mypy issue for more detail. If you have a better solution, do not hesitate

abersheeran:
```python
class MiddlewareType(Protocol):
    def __call__(
        self,
        app: ASGIApp,
        **options: Any
    ) -> ASGIApp:
        ...
```

Be more precise.

Viicos:
> ```python
> class MiddlewareType(Protocol):
>     def __call__(
>         self,
>         app: ASGIApp,
>         **options: Any
>     ) -> ASGIApp:
>         ...
> ```
> 
> Be more precise.

This doesn't work, as I said, more details in the linked `mypy` issue. For instance, it won't match against this one:

https://github.com/encode/starlette/blob/01aa49a379520d3bbe6bc1aecc48a48169e6b004/starlette/middleware/errors.py#L125-L142

as it would expect `ServerErrorMiddleware` to have `**kwargs` in the `__init__` method

abersheeran:
https://github.com/encode/starlette/blob/01aa49a379520d3bbe6bc1aecc48a48169e6b004/starlette/applications.py#L95-L107

We can modify its definition a bit. Because it is like this when it is used.

Viicos:
> We can modify its definition a bit. Because it is like this when it is used.

Apart from moving  `handler: typing.Optional[typing.Callable] = None, debug: bool = False` to `**kwargs`, it won't match the `MiddlewareType` with `**options: Any`, and I don't think it's convenient to have it this way 🤔

My opinion on this is while the `__init__` method is incomplete, and middlewares without an `app: ASGIApp` argument will pass type checking, we at least make sure the `__call__` method is correct (or the returned callable in the case of a nested callable). Plus it makes a step towards supporting https://github.com/encode/starlette/pull/1176, which had a similar solution, apart from the fact that it was missing the `__name__` attribute.

abersheeran:
My meaning is to modify the code to

```python
    def __init__( 
         self, 
         app: ASGIApp, 
         *,
         handler: typing.Optional[typing.Callable] = None, 
         debug: bool = False, 
     ) -> None: 
```

Kludex:
I'm not following the discussion, but I guess is @Viicos turn. :eyes:

Viicos:
Yes I'll probably take a look this afternoon 👌

Viicos:
Unfortunately this doesn't seem to work either:
https://mypy-play.net/?mypy=master&python=3.11&gist=0f955a8ee146e68a1410d7bd9167726d

```sh
main.py:20: note: Revealed type is "def (builtins.str)"
main.py:22: note: Revealed type is "None"
main.py:24: error: Argument 1 to "test" has incompatible type "type[ExampleMiddleware]"; expected "MiddlewareType"  [arg-type]
main.py:24: note: Following member(s) of "ExampleMiddleware" have conflicts:
main.py:24: note:     Expected:
main.py:24: note:         def __call__(app: int, **options: Any) -> Callable[[str], None]
main.py:24: note:     Got:
main.py:24: note:         def __init__(app: int, *, other_arg: int | None = ...) -> ExampleMiddleware
```

But even if it did, I'm not sure having people defining their own middlewares to explicitly use the `*` argument is good idea 🤔

abersheeran:
> Unfortunately this doesn't seem to work either: https://mypy-play.net/?mypy=master&python=3.11&gist=0f955a8ee146e68a1410d7bd9167726d
> 
> ```shell
> main.py:20: note: Revealed type is "def (builtins.str)"
> main.py:22: note: Revealed type is "None"
> main.py:24: error: Argument 1 to "test" has incompatible type "type[ExampleMiddleware]"; expected "MiddlewareType"  [arg-type]
> main.py:24: note: Following member(s) of "ExampleMiddleware" have conflicts:
> main.py:24: note:     Expected:
> main.py:24: note:         def __call__(app: int, **options: Any) -> Callable[[str], None]
> main.py:24: note:     Got:
> main.py:24: note:         def __init__(app: int, *, other_arg: int | None = ...) -> ExampleMiddleware
> ```
> 
> But even if it did, I'm not sure having people defining their own middlewares to explicitly use the `*` argument is good idea 🤔

😯 The results of mypy are a bit beyond my imagination.

abersheeran (approved):
Limited by the inference ability of mypy, I temporarily agree with the solution given by this PR. 🙏 Thank you all for your patient communication.

Viicos:
Thanks for the reviews @abersheeran and sorry for taking some time to respond. Typing callables can sometimes be a pain, but hopefully with `ParamSpec` and `Concatenate` this will be easier in the future

Kludex:
Pyright doesn't seem to like the proposed changes from this PR. 🤔

abersheeran:
> Pyright doesn't seem to like the proposed changes from this PR. 🤔

Are there plans to add pyright to CI?

Viicos:
> Pyright doesn't seem to like the proposed changes from this PR. 🤔

Pyright doesn't seem to support `*args: Any, **kwargs: Any` treated as any callable (mypy has it with https://github.com/python/mypy/issues/5876). We could switch to `MiddlewareType = Callable[..., ASGIApp]`, but then loose the `__name__` attribute that is required by starlette 😕

Kludex:
> > Pyright doesn't seem to like the proposed changes from this PR. 🤔
> 
> Are there plans to add pyright to CI?

No, but since `httpx` is now using mypy strict, we can follow it and do the same here.

Viicos:
> Pyright doesn't seem to support `*args: Any, **kwargs: Any` treated as any callable (mypy has it with [python/mypy#5876](https://github.com/python/mypy/issues/5876)). We could switch to `MiddlewareType = Callable[..., ASGIApp]`, but then loose the `__name__` attribute that is required by starlette 😕

The other alternative would be to switch back to what I did at first:

```python
class MiddlewareType(Protocol):
        ...
```

This is equivalent to `ASGIApp`, but needs to be defined as a `Protocol` and not a `Callable`, otherwise `Type[]` couldn't be used.

We can also keep things as is if we don't want to support `pyright` yet.

Kludex:
> We can also keep things as is if we don't want to support pyright yet.

Let's first do what httpx did, and use mypy on strict mode. Then we can see what errors we have with pyright.

For the time being, I'll close this PR.

Thanks for the patience @Viicos

PR #1988, opened 2022-12-23
Run lifespans of mounted applications

daanvdk:
Runs lifespans of mounted apps.

Fixes issue #649.

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -634,23 +634,23 @@ async def _app_lifespan(scope: Scope, app: ASGIApp) -> typing.AsyncIterator[None
         else:
             raise RuntimeError("unexpected send")
 
-    # This wrapper is needed because TaskGroup.start_soon does not like that
-    # App returns Awaitable instead of Coroutine
     async def coro_app(scope: Scope, receive: Receive, send: Send) -> None:
         await app(scope, receive, send)
+        if exception is None and not shutdown_done.is_set():
+            raise RuntimeError("lifespan returned unexpectedly")
 
     try:
         async with create_task_group() as tg:
             tg.start_soon(coro_app, {**scope, "app": app}, receive, send)
             await startup_done.wait()
-            if exception:
+            if exception is not None:
                 raise LifespanException(exception)
             try:
                 yield
             finally:
                 shutdown_init.set()
                 await shutdown_done.wait()
-                if exception:
+                if exception is not None:
                     raise LifespanException(exception)
     except Exception:
         if lifespan_supported:
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -6,8 +6,10 @@ import traceback
 import types
 import typing
 import warnings
-from contextlib import asynccontextmanager
+from contextlib import AsyncExitStack, asynccontextmanager
 from enum import Enum
+
+from anyio import Event, create_task_group
 
 from starlette._utils import is_async_callable
 from starlette.concurrency import run_in_threadpool
@@ -17,7 +19,7 @@ from starlette.exceptions import HTTPException
 from starlette.middleware import Middleware
 from starlette.requests import Request
 from starlette.responses import PlainTextResponse, RedirectResponse
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
 from starlette.websockets import WebSocket, WebSocketClose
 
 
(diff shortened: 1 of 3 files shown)


tests/test_routing.py
```diff
@@ -1022,6 +1032,221 @@ def test_host_named_repr() -> None:
...
+            shutdown()
+        except Exception:
+            await send(
+                {
+                    "type": "lifespan.shutdown.failed",
+                    "message": format_exc(),
+                }
+            )
+            return
+        await send({"type": "lifespan.shutdown.complete"})
```
adriangb, line 1070: I'm curious what the rational was for implementing this here as opposed to just mounting a Starlette app within a Starlette app (`app = Starlette(routes=[Mount(app=Starlette(lifespan=...))], lifespan=...)`)

starlette/routing.py
```diff
@@ -572,6 +574,91 @@ def __call__(self: _T, app: object) -> _T:
...
+                exception = message.get("message", "")
+            else:
+                raise ValueError(f"unexpected type: {message['type']}")
+            shutdown_done.set()
+
+        else:
+            raise RuntimeError("unexpected send")
+
+    # This wrapper is needed because TaskGroup.start_soon does not like that
+    # App returns Awaitable instead of Coroutine
```
adriangb: Yup that is the case. If we use this somewhere else in the library (I don't recall) it might be worth doing something more generic since the issue relates to awaitables/coroutines and not specifically ASGI apps.

starlette/routing.py
```diff
@@ -572,6 +574,91 @@ def __call__(self: _T, app: object) -> _T:
...
+    async def coro_app(scope: Scope, receive: Receive, send: Send) -> None:
+        await app(scope, receive, send)
+
+    try:
+        async with create_task_group() as tg:
+            tg.start_soon(coro_app, {**scope, "app": app}, receive, send)
```
adriangb, line 644: After all of the issues caused by using tasks in `BaseHTTPMiddelware` I'd like to avoid using them if possible. And I think that in this case it _is_ possible: https://github.com/adriangb/asgi-routing/blob/main/asgi_routing/_lifespan_dispatcher.py#L11-L101. It's not really more LOC either. But maybe there's a bug in there that I didn't catch.

adriangb (commented):
Great work! I gave this an initial look, left some comments. Haven't reviewed the tests in detail but I appreciate how thorough the test cases are. If we were to move forward with this I plan to do an in-depth review of the test cases and would like several maintainers to do the same. I think this sort of code with lots of events and tasks is prone to all sorts of bugs from race conditions, cancellation, etc.

High level I think the main question (which I'd like input from other maintainers on) is if we want to adopt this complexity (in other words, is the feature worth the complexity). I know this is something that has been requested in the past, but I don't think it's been a _huge_ issue.

I also wonder if it might be better to just take control of the event loop and run Uvicorn programmatically instead of getting into such complex lifespans and mounted apps (that is, instead of lifespans you just do whatever setup you need and then instantiate your apps).

starlette/routing.py
daanvdk: Yeah makes sense, for now this wrapper turned out to be a good thing because I added some code at the end to check if the app did not return in a situation where that would block the context manager.

starlette/routing.py
daanvdk: Yeah I think the main thing is that you need a task to be able to wrap a lifespan in an async context manager, which does feel like a nice abstraction to have. But I definitely understand your reservations.

tests/test_routing.py
daanvdk: Yeah I mainly, wanted to have a lifespan where I could easily overwrite the result of what happens during startup/shutdown. Might indeed be possible to simplify with a starlette app indeed.

Kludex:
@adriangb if it's fine by you, I'd prefer for us to avoid this before 1.0.

adriangb:
I agree on that

racinmat:
Hi @Kludex by 1.0 you mean 1.0 of starlette?
Is there any timeline for this?
There are new use-cases that I think are relevant: in AI Agents, both a2a and adk frameworks have FastAPI integration and use lifespans:
- https://github.com/a2aproject/a2a-python/blob/v0.2.14/src/a2a/server/apps/jsonrpc/fastapi_app.py#L131
- https://github.com/google/adk-python/blob/v1.7.0/src/google/adk/cli/fast_api.py#L281

Being able to mount them to existing application would be great help.

Kludex:
> Hi @Kludex by 1.0 you mean 1.0 of starlette?

Yes.

> Is there any timeline for this?

No.

> There are new use-cases that I think are relevant: in AI Agents, both a2a and adk frameworks have FastAPI integration and use lifespans:

By the way, I maintain [FastA2A](https://github.com/pydantic/fasta2a).

---

This project runs is volunteer based, with not many sponsors. It's hard for me to give estimations. If Google is interested, I'd be happy to accept sponsorship from them. :)

PR #1990, opened 2022-12-29
Add Config.env_prefix option

alex-oleshkevich:
This PR adds `Config.env_prefix` argument.

```python
import os

os.environ['APP_DEBUG'] = 'yes'
os.environ['ENVIRONMENT'] = 'dev'

config = Config(env_prefix='APP_')

DEBUG = config('DEBUG') # lookups APP_DEBUG, returns "yes"
ENVIRONMENT = config('ENVIRONMENT') # lookups APP_ENVIRONMENT, raises KeyError as variable is not defined
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -40,9 +40,9 @@ ALLOWED_HOSTS=127.0.0.1, localhost
 
 The order in which configuration values are read is:
 
--   From an environment variable.
--   From the ".env" file.
--   The default value given in `config`.
+* From an environment variable.
+* From the ".env" file.
+* The default value given in `config`.
 
 If none of those match, then `config(...)` will raise an error.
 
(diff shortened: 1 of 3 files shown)


Kludex (commented):
@alex-oleshkevich If you don't mind, would you be able to give a bit more about the motivation here?

This is a pretty straightforward feature, but no one ever asked before.

alex-oleshkevich:
This is a common practice to avoid naming conflicts.
E.g. there is `ENV` variable defined by the OS, and the app itself may require `ENV` like "dev" or "prod". So you can prefix your envvars to separate app's ones and third-party.

This situation pretty common in non-docker environments. Or, it may happen in docker-compose when 2 services require the same variable but with different value.

These are basically two examples I recall so far.

docs/config.md
```diff
@@ -113,6 +113,25 @@ from starlette.config import environ
 environ['TESTING'] = 'TRUE'
 ```
 
+## Reading prefixed environment variables
+
+You can namespace the environment variables by setting `env_prefix` argument.
+
+**myproject/settings.py**:
+
+```python
```
Kludex:
```suggestion
```python title="myproject/settings.py"
```
Can't we do this?

Kludex (approved):
Ok, let's go. :shrug:

Kludex:
Please check if the documentation changes that I suggested work, it may need a plugin. 👀

alex-oleshkevich:
> Please check if the documentation changes that I suggested work, it may need a plugin. eyes

it looks good 
image

PR #2005, opened 2023-01-13
hotfix: checking if have any required scope

danbailo:
This PR hotfix the propose of decorator `@requires`. In this case, supposing that we have 3 scopes ...


Kludex:
The way to read that `@requires` is: "The user needs those scopes to access the endpoint." 🤔

danbailo:
> The way to read that `@requires` is: "The user needs those scopes to access the endpoint." thinking

@Kludex I opened a discussion here :arrow_right:  https://github.com/encode/starlette/discussions/2006 . There is more explained

PR #2017, opened 2023-01-23
Lazily build middleware stack

### LINKED ISSUE #2002: Middleware stacking and recursive initialization
### Discussed in https://github.com/encode/starlette/discussions/1161

<sup>Originally posted by **euri10** April  1, 2021</sup>
long story short I was writing a logging middleware whose handlers are initialized in the `__init__` of the middleware and ended up with multiple logs.

it seems to me the `__init__` of the middlewarea are recursively called  because of the `build_middleware_stack` method and I'm not sure if this is intended or a bug.

here's a test case that shows what I think is an issue:

```
class SimpleInitializableMiddleware:
    def __init__(self, app, counter: int):
        self.app = app
        self.counter = counter
        print(self.counter)


def test_middleware_stack_init():
    app = Starlette()
    app.add_middleware(SimpleInitializableMiddleware, counter=1)
    app.add_middleware(SimpleInitializableMiddleware, counter=2)
    app.add_middleware(SimpleInitializableMiddleware, counter=3)
    app.add_middleware(SimpleInitializableMiddleware, counter=4)
```

I would expect to see `1 2 3 4` as the output.
however it prints `1 1 2 1 2 3 1 2 3 4` because for each add_middlware call it goes down the stack and reinitialize it.

what you think @tomchristie ?</div>

adriangb:
- Closes https://github.com/encode/starlette/issues/2002

Edit by @Kludex:

This is an alternative solution for deprecation + removal of the `add_middlware`. The solution here is to build the application with its middlewares only when we receive the first ASGI event - lifespan, websocket or http, doesn't matter.

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -74,7 +74,7 @@ class Starlette:
             {} if exception_handlers is None else dict(exception_handlers)
         )
         self.user_middleware = [] if middleware is None else list(middleware)
-        self.middleware_stack = self.build_middleware_stack()
+        self.middleware_stack = None
 
     def build_middleware_stack(self) -> ASGIApp:
         debug = self.debug
@@ -115,13 +115,14 @@ class Starlette:
     @debug.setter
     def debug(self, value: bool) -> None:
         self._debug = value
-        self.middleware_stack = self.build_middleware_stack()
 
     def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
         return self.router.url_path_for(name, **path_params)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
+        if self.middleware_stack is None:
+            self.middleware_stack = self.build_middleware_stack()
         await self.middleware_stack(scope, receive, send)
 
     def on_event(self, event_type: str) -> typing.Callable:  # pragma: nocover
@@ -140,8 +141,11 @@ class Starlette:
     def add_middleware(
         self, middleware_class: type, **options: typing.Any
     ) -> None:  # pragma: no cover
+        if self.middleware_stack is not None:
+            raise RuntimeError(
+                "Cannot add middlewares after an application has started"
+            )
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
-        self.middleware_stack = self.build_middleware_stack()
 
     def add_exception_handler(
         self,
@@ -149,7 +153,6 @@ class Starlette:
         handler: typing.Callable,
     ) -> None:  # pragma: no cover
         self.exception_handlers[exc_class_or_status_code] = handler
-        self.middleware_stack = self.build_middleware_stack()
 
     def add_event_handler(
         self, event_type: str, func: typing.Callable
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -65,7 +65,7 @@ class Starlette:
             on_startup is None and on_shutdown is None
         ), "Use either 'lifespan' or 'on_startup'/'on_shutdown', not both."
 
-        self._debug = debug
+        self.debug = debug
         self.state = State()
         self.router = Router(
             routes, on_startup=on_startup, on_shutdown=on_shutdown, lifespan=lifespan
@@ -108,14 +108,6 @@ class Starlette:
     def routes(self) -> typing.List[BaseRoute]:
         return self.router.routes
 
-    @property
-    def debug(self) -> bool:
-        return self._debug
-
-    @debug.setter
-    def debug(self, value: bool) -> None:
-        self._debug = value
-
     def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
         return self.router.url_path_for(name, **path_params)
 
@@ -140,9 +132,7 @@ class Starlette:
 
     def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
         if self.middleware_stack is not None:  # pragma: no cover
-            raise RuntimeError(
-                "Cannot add middleware after an application has started"
-            )
+            raise RuntimeError("Cannot add middleware after an application has started")
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
 
     def add_exception_handler(
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -140,7 +140,7 @@ class Starlette:
 
     def add_middleware(
         self, middleware_class: type, **options: typing.Any
-    ) -> None:  # pragma: no cover
+    ) -> None:
         if self.middleware_stack is not None:
             raise RuntimeError(
                 "Cannot add middlewares after an application has started"
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -139,9 +139,9 @@ class Starlette:
         self.router.host(host, app=app, name=name)
 
     def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
-        if self.middleware_stack is not None:
+        if self.middleware_stack is not None:  # pragma: no cover
             raise RuntimeError(
-                "Cannot add middlewares after an application has started"
+                "Cannot add middleware after an application has started"
             )
         self.user_middleware.insert(0, Middleware(middleware_class, **options))
 
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -138,9 +138,7 @@ class Starlette:
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(
-        self, middleware_class: type, **options: typing.Any
-    ) -> None:
+    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
         if self.middleware_stack is not None:
             raise RuntimeError(
                 "Cannot add middlewares after an application has started"
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -74,7 +74,7 @@ class Starlette:
             {} if exception_handlers is None else dict(exception_handlers)
         )
         self.user_middleware = [] if middleware is None else list(middleware)
-        self.middleware_stack = None
+        self.middleware_stack: typing.Optional[ASGIApp] = None
 
     def build_middleware_stack(self) -> ASGIApp:
         debug = self.debug
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -515,11 +515,19 @@ def test_middleware_stack_init(test_client_factory: Callable[[ASGIApp], httpx.Cl
         app.add_middleware(NoOpMiddleware)
         return app
 
-    with test_client_factory(get_app()):
+    app = get_app()
+
+    with test_client_factory(app):
         pass
 
     assert SimpleInitializableMiddleware.counter == 1
 
-    test_client_factory(get_app()).get("/foo")
+    test_client_factory(app).get("/foo")
+
+    assert SimpleInitializableMiddleware.counter == 1
+
+    app = get_app()
+
+    test_client_factory(app).get("/foo")
 
     assert SimpleInitializableMiddleware.counter == 2
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -1,7 +1,9 @@
 import os
 from contextlib import asynccontextmanager
+from typing import Any, Callable
 
 import anyio
+import httpx
 import pytest
 
 from starlette import status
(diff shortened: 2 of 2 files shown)


starlette/applications.py
```diff
@@ -115,13 +115,14 @@ def debug(self) -> bool:
...
         self._debug = value
-        self.middleware_stack = self.build_middleware_stack()
 
     def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
         return self.router.url_path_for(name, **path_params)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
+        if self.middleware_stack is None:
+            self.middleware_stack = self.build_middleware_stack()
```
adriangb, line 117: Note that this will run when the lifespan is triggered most of the time but if the lifespan is disable it will run on the first request. So not checking `scope["type"]` is by design.

starlette/applications.py
```diff
@@ -140,16 +141,18 @@ def host(
     def add_middleware(
         self, middleware_class: type, **options: typing.Any
     ) -> None:  # pragma: no cover
+        if self.middleware_stack is not None:
```
alex-oleshkevich: This method is deprecated, isn't it @Kludex ?

starlette/applications.py
```diff
@@ -74,7 +74,7 @@ def __init__(
             {} if exception_handlers is None else dict(exception_handlers)
         )
         self.user_middleware = [] if middleware is None else list(middleware)
-        self.middleware_stack = self.build_middleware_stack()
+        self.middleware_stack: typing.Optional[ASGIApp] = None
```
alex-oleshkevich, line 77: IMO, `build_middleware_stack` is a very cheap function to call. It makes sense to call it here and call once again when you need it again, and avoid nullable attributes (and null-checks).

alex-oleshkevich: Moreover, it is not expected to be called often at runtime and affects only startup (configuration) time. 
So calling it multiple times seems ok to me.

adriangb: I’m not sure if you advocating for the old existing approach with the one being proposed here. Performance is not the goal, it is to avoid instantiating users middleware repeatedly as reported in #2002

starlette/applications.py
adriangb: Yes it is. #2002 still came up last meeting as a problem.

starlette/applications.py
```diff
@@ -140,16 +141,18 @@ def host(
     def add_middleware(
         self, middleware_class: type, **options: typing.Any
     ) -> None:  # pragma: no cover
```
```suggestion
    ) -> None:
```

starlette/applications.py
Kludex: It's not...

We didn't deprecate it. We only deprecated the decorators. :thinking: 

And "what came up on the meeting" was that I mentioned the `add_middleware` as an issue, due to the linked GitHub issue above, and the idea was to deprecate, and further remove, but Adrian came up with this alternative solution.

adriangb: Apologies, thank you for clarifying 😄

starlette/applications.py
Kludex: I've mentioned this in the description as well. :+1:

starlette/applications.py
```diff
@@ -115,13 +115,14 @@ def debug(self) -> bool:
     @debug.setter
     def debug(self, value: bool) -> None:
         self._debug = value
```
Kludex: We can now replace this setter by just using `self.debug`, can't we?

Kludex:
I think Flask 2.0 introduced a similar idea... I can't recall how was it. Someone knows?

Kludex:
I think there's no need to add anything to the documentation regarding this change.

This change shouldn't impact anyone - only if they are relying on the multiple instantiations' behavior, which I doubt.

Maybe we can test this against FastAPI to see if there's something that may not be obvious that breaks over there - I can do it before the next release. :+1:

starlette/applications.py
adriangb: Yep I guess so!

adriangb: I think this PR is now _less_ lines of code and less complexity 🥳

starlette/applications.py
```diff
@@ -137,19 +130,17 @@ def host(
     ) -> None:  # pragma: no cover
         self.router.host(host, app=app, name=name)
 
-    def add_middleware(
-        self, middleware_class: type, **options: typing.Any
-    ) -> None:  # pragma: no cover
+    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+        if self.middleware_stack is not None:  # pragma: no cover
```
Kludex, line 134: Can't we add a test for this case? :thinking:

Kludex (approved):

starlette/applications.py
adriangb: Seems like testing for the sake of testing but if you want I can add it.

tests/test_applications.py
```diff
@@ -486,3 +489,45 @@ async def startup():
 
         app.on_event("startup")(startup)
         assert len(record) == 1
+
+
+def test_middleware_stack_init(test_client_factory: Callable[[ASGIApp], httpx.Client]):
```
Kludex, line 494: This would be the first time using `httpx.Client` on this annotation on the test suite :eyes: 
```suggestion
def test_middleware_stack_init(test_client_factory: Callable[[ASGIApp], TestClient]):
```

adriangb: Hmm yeah I should have used `TestClient`. Luckily it's just a type annotation.

tests/test_applications.py
```diff
@@ -486,3 +489,45 @@ async def startup():
...
+
+    test_client_factory(app).get("/foo")
+
+    assert SimpleInitializableMiddleware.counter == 1
+
+    app = get_app()
+
+    test_client_factory(app).get("/foo")
+
+    assert SimpleInitializableMiddleware.counter == 2
```
Kludex, line 533: How relevant is this part for the test? :thinking:

Mark90:
@Kludex 

> This change shouldn't impact anyone - only if they are relying on the multiple instantiations' behavior, which I doubt.

I don't know if that's what we're doing. But I did just notice this in one of our FastAPI projects and reduced it to the following example:

```python
# main.py
from fastapi import FastAPI
from fastapi.requests import *
from fastapi.responses import *

class MyHandledException(Exception):
    """My handled exception."""

async def my_exception_handler(
    request: Request, exc: MyHandledException
) -> PlainTextResponse:
    # fastapi==0.90.1 starlette==0.23.1   -> my_exception_handler executes
    # fastapi==0.91.0 starlette==0.24.0   -> my_exception_handler does not execute
    print(">>> Executing my_exception_handler <<<")
    return PlainTextResponse(
        f"Caught the following exception: {str(exc)}", status_code=400
    )

app = FastAPI()

@app.on_event("startup")
async def on_startup() -> None:
    print(">>> Executing on_startup <<<")
    app.add_exception_handler(MyHandledException, my_exception_handler)

@app.get("/{thing}")
async def get_thing(thing):
    if thing == "boom":
        raise MyHandledException("Boom")
    return f"Here you go: {thing}"
```

In a Python 3.11 venv with `uvicorn==0.21.1 fastapi==0.90.1 starlette==0.23.1`, started the app with `uvicorn main:app`, called the endpoint with `"/boom"` -> MyHandledException is picked up by the exception handler just fine.

After upgrading to `fastapi==0.91.0 starlette=0.24.0` -> the on_startup is executed, but MyHandledException is not picked up by the handler. It doesn't seem to be in the middleware stack, I'm not completely sure why. Perhaps the stack is already initialized at this point?

When I register the handler like recommended in the FastAPI docs as below, then it does work
```python
@app.execution_handler(MyHandledException)
async def my_exception_handler(
...
```

In our real application the exception handler comes from another package, so adding the `@app.exception_handler` decorator is not an option. 

But I've also read somewhere that `on_startup` event is going to be replaced by "Lifespan", I'll have to read up on that (to see if it can be used for dynamically registering exception handlers)

Kludex:
> Perhaps the stack is already initialized at this point?

Yes. It's the middleware stack.

Note: I don't think we predicted users doing this @adriangb... :thinking:

adriangb:
Oh jeez, we certainly did not predict this. Technically with the previous implementation you could add middleware at any time, even within a running request/response cycle, which I don't think should be allowed, period. Adding it within the lifespan seams a bit more reasonable: you could have a middleware that depends on something you initialize in the lifespan.
The good news is that we can probably move building the stack until after the ASGI lifespan has been run, which would solve the example above.

starlette/applications.py
```diff
@@ -108,20 +108,13 @@ def build_middleware_stack(self) -> ASGIApp:
...
-        self._debug = value
-        self.middleware_stack = self.build_middleware_stack()
-
     def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
         return self.router.url_path_for(name, **path_params)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
+        if self.middleware_stack is None:
+            self.middleware_stack = self.build_middleware_stack()
```
Kludex, line 117:
```suggestion
        if scope["type"] != "lifespan" and self.middleware_stack is None:
            self.middleware_stack = self.build_middleware_stack()
```

adriangb: What if the middleware runs something on lifespan? Maybe we need to build once before running the lifespan and then, if it's changed, we re-build it after running the lifespan?

Kludex: Actually we want both scenarios...

adriangb: What do you mean?

Kludex: Yeah :P

Kludex: > What do you mean?

The same thing as your first message.

But actually... We don't want to rebuild, otherwise we introduce the issue that motivated this PR again.

adriangb: I think I have a solution that allows dynamically adding middleware at any point in time and never re-building:

```python
class ASGIAppProxy:
        self.app = app

```

The reason we need to re-build in the first place is that the last user middleware needs to point to `self.router/ExceptionMiddleware`. So say you have a single `A` user middelware, the "stack" looks like `ServerErrorMiddleware -> UserMiddlewareA -> ExceptionMiddleware -> Router`. Then you add a `UserMiddlewareB` and it should be `ServerErrorMiddleware -> UserMiddlewareA ->  UserMiddlewareB -> ExceptionMiddleware -> Router`. So you need to change the ASGIApp that `UserMiddlewareA` is pointing to from `ExceptionMiddleware` to `UserMiddlewareB`. We can't just assign a `.app` attribute on the users middleware because it could be anything. By introducing the above wrapper we can do something like:

```python
exc_middleware = ExceptionMiddleware(...)
tail = ASGIAppProxy(exc_middleware)
user_middleware_tail = UserMiddlewareA(tail)
# add UserMiddlewareB
new_tail = ASGIAppProxy(exc_middleware)
new_user_middleware_tail = UserMiddlewareB(new_tail)
tail.app = new_user_middleware_tail
tail = new_tail
```

Kludex:
> ...you could have a middleware that depends on something you initialize in the lifespan.

For example? You can (and I guess you should) use the `scope["state"]` in that middleware.

@Mark90 can you share what's the exact use-case that you have?

Kludex:
We can also add an error on `add_middleware` in case it's used after the middleware stack is built.

adriangb:
> We can also add an error on `add_middleware` in case it's used after the middleware stack is built.

We already have one: https://github.com/encode/starlette/blob/master/starlette/applications.py#L139

adriangb:
> For example?

An authentication middleware that needs a database connection or HTTP client.

> You can (and I guess you should) use the scope["state"] in that middleware.

I had not thought of that use case for scope[“state”], but it’s a good idea. Even without that feature of the middleware could just store whatever it needs on its instance.

Mark90:
@Kludex 

> @Mark90 can you share what's the exact use-case that you have?

We have a library containing endpoints / exceptions (and handlers) that we reuse among a few projects. Using the shared exception handler in those projects requires `.add_exception_handler()` as far as I know.

adriangb:
Why can’t you pass those to the constructor or call add_exception_handler before the lifespan?

Mark90:
Ah, I only checked FastAPI's docs for exception handlers, not Starlette's. Didn't know about that constructor parameter, it works perfect. Thank you!

PR #2026, opened 2023-02-06
Move exception handling logic to Route

### LINKED ISSUE #493: Question about Request.Body()

adriangb:
Fixes #493, partially resolves https://github.com/encode/starlette/pull/1692#issuecomment-1266401636, resolves https://github.com/encode/starlette/pull/1649#issuecomment-1404473365

By moving handling of exceptions to the `Route` level we can use the same `Request`/`WebSocket` object for the endpoint and exception handlers.

See #2020 for some more background


Kludex:
I'm positive towards this PR. Is any other member that wants to check it? @encode/maintainers

adriangb:
@tomchristie it would be great to get your eyes on this since no one has really changed this part of the codebase since you originally wrote it

starlette/middleware/exceptions/_wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+from starlette._utils import is_async_callable
+from starlette.concurrency import run_in_threadpool
+from starlette.exceptions import HTTPException
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+from starlette.websockets import WebSocket
+
+Handler = typing.Callable[..., typing.Any]
+ExcHandlers = typing.Dict[typing.Any, Handler]
```
alex-oleshkevich: Let's use a full version here: `ExceptionHandlers`

Kludex (commented):
Are you strong about this code structure?

Do you mind if we maintain the `exceptions.py` and create a `_<some_good_name_at_root_level>`? I feel like importing `wrap_app_handling_exceptions` from `middleware/exceptions` is not very logic. :thinking: 

@encode/maintainers If someone is against this idea, or has any concerns regarding this, or want more context, anything... Please let us know. Otherwise, I'll get this merged end of next week.

alex-oleshkevich:
This is useful as it unlocks future per-route error handlers, greatly simplifying view code.
```python
Route('/, view, error_handlers={})
```
However, this does not solve the problem when someone wants to access `request.body` in other middleware. This issue is still there, isn't it? This solution (https://github.com/encode/starlette/pull/1519) is more straightforward as fixes the issue everywhere.

adriangb:
> this does not solve the problem when someone wants to access `request.body` in other middleware. This issue is still there, isn't it? This solution (#1519) is more straightforward as fixes the issue everywhere.

https://github.com/encode/starlette/pull/1692 also resolves it in a way that compliments this PR, and without breaking other things.

adriangb:
Folks, it would be great to get some more input/review here.

starlette/middleware/exceptions/_wrapper.py
adriangb: Thanks, done in 3cf3881

Kludex:
The last bits here:

What's our plan regarding the `ExceptionMiddleware`? Are we going to remove it at some point?

Also, should we update the documentation: https://www.starlette.io/exceptions/#errors-and-handled-exceptions ?

Kludex (commented):
Do you mind if we change the structure of the files around (I can change it)?

I'd prefer to not remove the file `exceptions.py`, and just create a top level `_exception_handler.py`.

adriangb:
Sure go for it

Kludex:
@adriangb I didn't change the documentation, but I think we should do something there for 1.0.

Also, I applied this commit: [`3fb1118` (#2026)](https://github.com/encode/starlette/pull/2026/commits/3fb111835755c65a4144603c552d9d06c54d80e2). If you didn't like it, let me know, so I can revert it. 👍

Kludex (approved):

PR #2034, opened 2023-02-11
add sse-starlette to third-party page

sysid:
sse-starlette brings SSE to starlette.


Kludex:
Thanks, but it's already mentioned in the docs: https://www.starlette.io/responses/#eventsourceresponse

PR #2036, opened 2023-02-14
🐛 Close all the multipart files on error

tiangolo:
🐛 Close all the multipart files on error.

This is a small fix on top of the PR fixing the vulnerability here: https://github.com/encode/starlette/commit/8c74c2c8dba7030154f8af18e016136bea1938fa (it doesn't show as a PR as it was done in a security advisory private fork).

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -270,7 +270,7 @@ class MultiPartParser:
             # Close all the files if there was an error.
             for file in self._files_to_close_on_error:
                 file.close()
-            raise e
+            raise exc
 
         parser.finalize()
         return FormData(self.items)
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -266,7 +266,7 @@ class MultiPartParser:
                     await part.file.seek(0)
                 self._file_parts_to_write.clear()
                 self._file_parts_to_finish.clear()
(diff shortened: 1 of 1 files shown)


starlette/formparsers.py
```diff
@@ -247,21 +249,28 @@ async def parse(self) -> FormData:
...
+                # the main thread.
+                for part, data in self._file_parts_to_write:
+                    assert part.file  # for type checkers
+                    await part.file.write(data)
+                for part in self._file_parts_to_finish:
+                    assert part.file  # for type checkers
+                    await part.file.seek(0)
+                self._file_parts_to_write.clear()
+                self._file_parts_to_finish.clear()
+        except Exception as e:
```
Kludex: Isn't it only with MultiPartException?
```suggestion
        except MultiPartException as exc:
```

starlette/formparsers.py
```diff
@@ -247,21 +249,28 @@ async def parse(self) -> FormData:
...
+            # Close all the files if there was an error.
+            for file in self._files_to_close_on_error:
+                file.close()
+            raise e
```
Kludex: Just for consistency
```suggestion
            raise exc
```

Kludex (approved):

PR #1535, opened 2022-03-03
Initial support for Websocket Denial Response

paulo-raca:
When a websocket connection is rejected (e.g., wrong URL, missing permissions, etc),
starlette always returns a generic 403 response (triggered by "websocket.close" message)
or a generic HTTP500 response (Produced by ASGI gateway if there is an unhandled error)

Instead it is better to return useful HTTP error responses using the [Websocket Denial Response ASGI extension](https://asgi.readthedocs.io/en/latest/extensions.html#websocket-denial-response), if it is available

This PR makes the following changes:
- Added WebsocketDenialResponse class to return a HTTP Response to a websocket request, mapping the ASGI message as appropriate.
  if the extension is not supported, it just closes the websocket resulting and HTTP 403 like before

- Updated testclient to support denial responses.
  The extension must be explicitly enabled on a connection using `websocket_connect(..., denial_response=True)`,
  and if a denial response is returned a WebSocketDenied exception is thrown

- Refactored testclient code so that the code to parse HTTP Responses can be reused on websocket denial responses

This PR superseeds #1478 with better documentation and testing support. Unlike the previous PR,
it does not update the various middlewares.

Instead, I'll create separate PRs for those later, in order to keep this PR smaller

PR #2041, opened 2023-02-18
Support the WebSocket Denial Response ASGI extension

kristjanvalur:
This PR adds support for the "websocket.http.response" ASGI extension in the `WebSocket` class as well as in the
`TestClient` (see https://asgi.readthedocs.io/en/latest/extensions.html)

- A new `send_response()` method is added to `WebSocket`, taking a `Response` object. This must only be called before `accept()`.
  - If the extension is supported, the response is sent as per extension specification
  - Otherwise, a `websocket.close(1008)` is performed which an ASGI server should turn into a 403 response.
- A new `starlette.testlient.WebSocketReject` exception is added which represents a disconnect before a `websocket.accept()` call.  In a real HTTP client, it would correspond to a failure response to the Upgrade request.
- Added documentation about rejecting a WebSocket connection, either with `send_response()` or a `close()` before `accept()`

This PR was spurred by a perceived need to standardize better how to do authentication and connection rejection in ASGI frameworks.  In particular, FastAPI seems a bit wobbly there. 

the API presented is a suggestion.  We could use `close_response()` for example,.

We could also provide a `code` and `reason` for the fallback case (when the extension isn't supported by the server).  How exactly a server should construct 403 out of the `code` and `reason` isn't specified by ASGI  Different servers might or might not add that information to the response Body so providing that would be a pure convenience.


Kludex:
- Related to https://github.com/encode/starlette/pull/1535

starlette/testclient.py
```diff
@@ -66,6 +66,19 @@ def __init__(self, session: "WebSocketTestSession") -> None:
         self.session = session
 
 
+class WebSocketReject(WebSocketDisconnect):
+    def __init__(
+        self,
+        status_code: int,
+        body: bytes = b"",
+        code: int = 1000,
+        reason: typing.Optional[str] = None,
```
Kludex: I don't think we should have the "fallback case" that you mentioned.

```suggestion
```

tests/test_staticfiles.py
```diff
@@ -68,13 +68,13 @@ def test_staticfiles_with_package(test_client_factory):
     client = test_client_factory(app)
     response = client.get("/example.txt")
     assert response.status_code == 200
-    assert response.text == "123\n"
+    assert response.text.replace("\r\n", "\n") == "123\n"
```
Kludex: What's this?

starlette/websockets.py
```diff
@@ -181,6 +202,12 @@ async def close(
             {"type": "websocket.close", "code": code, "reason": reason or ""}
         )
 
+    async def send_response(self, response: Response) -> None:
+        if self._have_response_extension():
+            await response(self.scope, self.receive, self.send)
+        else:
+            await self.close(code=1008, reason=f"HTTP Response {response.status_code}")
```
Kludex: Maybe we can raise a runtime exception with "The server doesn't support the Websocket Denial Response extension." instead?

Why do you think we should have a fallback?

starlette/websockets.py
```diff
@@ -64,12 +69,18 @@ async def send(self, message: Message) -> None:
...
-                    'Expected ASGI message "websocket.connect", '
-                    f"but got {message_type!r}"
-                )
+                if not (
+                    self._have_response_extension()
+                    and message_type == "websocket.http.response.start"
+                ):
+                    raise RuntimeError(
+                        'Expected ASGI message "websocket.connect", '
+                        f"but got {message_type!r}"
```
Kludex: Now that I read this error message here, isn't it wrong? :thinking:

Kludex (commented):
Would you like to implement the support for Websocket Denial Response on Uvicorn?

tests/test_staticfiles.py
kristjanvalur: Helps test run on windows, where example.txt has this encoding.  Not-essential to this PR.

Kludex: Starlette's test suite fails on Windows?

starlette/websockets.py
kristjanvalur: Yes, I noticed it too but decided not to do anything.  I'll fix that too.

kristjanvalur:
> Would you like to implement the support for Websocket Denial Response on Uvicorn?

Sure, at least, I can have a look.

starlette/websockets.py
```diff
@@ -64,12 +69,18 @@ async def send(self, message: Message) -> None:
...
-                raise RuntimeError(
-                    'Expected ASGI message "websocket.connect", '
-                    f"but got {message_type!r}"
-                )
+                if not (
+                    self._have_response_extension()
+                    and message_type == "websocket.http.response.start"
+                ):
+                    raise RuntimeError(
+                        'Expected ASGI message "websocket.accept", '
```
```suggestion
                        'Expected ASGI message "websocket.accept" or "websocket.close", '
```

tests/test_staticfiles.py
kristjanvalur: I guess it depends on how it is checked out.  I cloned it in windows, then was testing it in WSL.  The file then has \r\n endings.  I guess it would be best to list this file in .gitattributes as having  \n line endings.   Anyway, I've removed this change from the PR

starlette/websockets.py
kristjanvalur: My thinking was that the application shouldn't have to have code dealing with server peculiarities. The application could simply always create an error response, and leave it to Starlette to decide what to do with it.  If the server supports it, then fine.  If not, Starlette automatically does the next best thing, simply close the socket, which is specified to create a 403 response.   
The fallback code and message are arbitrary, because ASGI does not specify what should happen to these codes in case of a 403.  _maybe_ the server adds that to the 403 payload, maybe not.  I haven't looked at what uvicorn does.

Kludex: If this is the path we want to go, then let's remove the code and reason. Both doesn't make sense, since the handshake didn't happen, and the server actually should ignore both code and reason.

Kludex: I'm still not sure if we want this... If you are using `send_response`, I'd expect that working out, and if it doesn't, as a developer I'd like to see an exception :thinking:

docs/websockets.md
```diff
@@ -109,3 +109,17 @@ correctly updated.
...
+e.g. because of failing authentication.  The standard way to do this
+is to call `websocket.close()` which will cause the ASGI web server
+to respond to the initator with a HTTP 403 error.
+It is also possible to construct a `Response` object and send that:
+
+* `await websocket.send_response(response)`
+
+If the ASGI server supports the `websocket.http.response` extension, this
+will cause the specified response to be sent to the initiator, before closing
+the connection.  Otherwise, it behaves like `websocket.close()`.
```
Kludex:
```suggestion
the connection.
```

starlette/testclient.py
```diff
@@ -66,6 +66,17 @@ def __init__(self, session: "WebSocketTestSession") -> None:
         self.session = session
 
 
+class WebSocketReject(WebSocketDisconnect):
+    def __init__(
+        self,
+        status_code: int,
+        body: bytes = b"",
+    ) -> None:
+        super().__init__(code, reason)
```
Kludex: There's no code or reason.

starlette/websockets.py
```suggestion
        raise RuntimeError(""The server doesn't support the Websocket Denial Response extension.")
```

Kludex (changes requested):

kristjanvalur:
I'm re-adding `close_code` and `close_reason` for `WebSocketReject`, because this is a TestClient class, designed to capture the state when a connection is _rejected_, either by a `close()` or a `send_response()` _before_ the connection is _accepted_.

The exception then contains the information as provided by the application.

Previously a `close()` in this situation would raise a `WebSocketClose`, now it raises a `WebSocketReject`, a special case with the same information, but allowing the test client application to distinguish this from a regular close.

docs/websockets.md
kristjanvalur: See my other comment.  I think it is useful for the application to provide a custom response, if the server provides it, otherwise it provides the only response, possible, the close().    I don't really see what benefit there is to moving that logic to the application, since there is only one course of action.

kristjanvalur:
I've also made sure that the Testclient now delivers the `websocket.disconnect` to the app, when the app initiates a close, either by doing a `websocket.close()` while open, or while rejecting a connection.

kristjanvalur:
I guess the only thing here that needs deciding is if `websocket.send_response()` should default to `websocket.close()` if there is no extension.  IMHO it should, because that's the only other possible thing that can be done (it is the standard, non-extended way of rejecting) and that relieves removes boilerplate logic from the app into Starlette. 

But I've stated my case and will do this however you like :)

starlette/websockets.py
kristjanvalur: But that exception appearing or not depends on which server you choose to run your app.  The response here is informative only.  Starlette is ASGI toolkit aimed at simplifying development.
As I see it, here are the options:
1) just close the connection().  This is always supported and is the original way of rejecting a connection.  Server provides a 403 response.
2) Close the response, but proide additional information for those interested.
  2a) Server supports the extension, and a response body is delivered along with the message.
  2b) Server does not support the response and body is ignored.
  
It is th case 2b) which worries you.  IMHO, the response body is informative only.  Depending on web server a response body can or can not be delivered.   If the information is somehow protocol critical (not recommended), then the application can always check for the presence of the extension.

Ok, how about this:  Instead of an exception, how about a warning?  This will get flagged to the developer but it won't crash your application if an extension isn't supported.   And we won't have to push the burden on the developer to always check the state for the presence of the extension before using a close call with an extended api?

kristjanvalur: See above.  I don't think we should bounce an exception on the developer when a purely informational message is not deliverable because of a choice of web server.  But I suggest that we add a warning instead?

kristjanvalur:
> > Would you like to implement the support for Websocket Denial Response on Uvicorn?
> 
> Sure, at least, I can have a look.

https://github.com/encode/uvicorn/pull/1916

starlette/websockets.py
```diff
@@ -181,6 +218,12 @@ async def close(
             {"type": "websocket.close", "code": code, "reason": reason or ""}
         )
 
+    async def send_response(self, response: Response) -> None:
+        if self._have_response_extension():
+            await response(self.scope, self.receive, self.send)
+        else:
+            await self.close(code=1008, reason=f"HTTP Response {response.status_code}")
```
Kludex: Instead of this API, does it make sense to allow `return Response()`?

starlette/websockets.py
Kludex: I think the `RuntimeError` makes more sense.

docs/websockets.md
```diff
@@ -109,3 +109,17 @@ correctly updated.
 
 * `await websocket.send(message)`
 * `await websocket.receive()`
+
+### Rejecting the connection
+
+Before the calling `websocket.accept()` it is possible to reject the connection,
```
Kludex:
```suggestion
Before calling `websocket.accept()` it is possible to reject the connection,
```

docs/websockets.md
Kludex: I don't agree.

tests/test_websockets.py
```diff
@@ -494,3 +593,19 @@ async def app(scope, receive, send):
...
+        nonlocal close_msg
+        websocket = WebSocket(scope, receive=receive, send=send)
+        await websocket.accept()
+        close_msg = await websocket.receive()
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/") as websocket:
+        websocket.send({"type": "websocket.disconnect"})
+
+    assert close_msg == {"type": "websocket.disconnect"}
```
Kludex: This test needs to be fixed, I guess?

Kludex (commented):
@kristjanvalur Can you review my changes here, please?

@paulo-raca if you want to review as well...

There's still a test, and documentation to be fixed.

starlette/testclient.py
```diff
@@ -165,30 +156,17 @@ def _handle_response(self, message: Message) -> None:
...
+        raise DenialResponse(
+            status_code=status_code,
+            headers=headers,
+            body=b"".join(body),
         )
 
-    def _raise_on_close(
-        self, message: Message, reject: typing.Optional[bool] = None
-    ) -> None:
+    def _raise_on_close(self, message: Message) -> None:
```
kristjanvalur, line 169: Not sure why you want to remove the special "reject" type of websocket close.  this is a test client and it is useful for the client to verify why the connection was closed ,whether it was a regular .close() during operation, or if it was a close() before an accept().  I see you have added a specific DenialResponse() exception.  There are two types of denials, just doing 'close' before accept and the other doing 'http.response.start' and a test client needs to distinguish between both and also just a regular (non-denial) close.

starlette/websockets.py
```diff
@@ -62,8 +53,6 @@ async def receive(self) -> Message:
                 )
             if message_type == "websocket.disconnect":
                 self.client_state = WebSocketState.DISCONNECTED
-                if "code" not in message:
```
kristjanvalur: yes, further reading of standard indicates that this code must be application level only.

starlette/websockets.py
```diff
@@ -28,20 +28,11 @@ def __init__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
-        self.app_disconnect_msg: typing.Optional[Message] = None
-
-    def _have_response_extension(self) -> bool:
-        return "websocket.http.response" in self.scope.get("extensions", {})
 
     async def receive(self) -> Message:
         """
         Receive ASGI websocket messages, ensuring valid state transitions.
         """
-        if self.app_disconnect_msg is not None:
```
kristjanvalur: Not sure why you are removing this.  If I read asgiref correctly, a receive on a disconnected connection should return the close message which caused the disconnection, even if it was a local message.

starlette/websockets.py
```diff
@@ -76,22 +65,19 @@ async def send(self, message: Message) -> None:
...
+                "websocket.http.response.start",
+            }:
+                raise RuntimeError(
+                    'Expected ASGI message "websocket.accept" or "websocket.close" or '
+                    f'"websocket.http.response.start", but got {message_type!r}'
+                )
             if message_type == "websocket.close":
                 self.application_state = WebSocketState.DISCONNECTED
-                # no close frame is sent, then the default is 1006
-                self.app_disconnect_msg = {"type": "websocket.disconnect", "code": 1006}
```
kristjanvalur: see above.  we are creating a close message to be read by the application, not to send over the wire.  1006 is appropriate code in this case

starlette/websockets.py
```diff
@@ -76,22 +65,19 @@ async def send(self, message: Message) -> None:
...
-                ):
-                    raise RuntimeError(
-                        'Expected ASGI message "websocket.accept"'
-                        ' or "websocket.close", '
-                        f"but got {message_type!r}"
-                    )
+            if message_type not in {
+                "websocket.accept",
+                "websocket.close",
+                "websocket.http.response.start",
```
kristjanvalur, line 73: I think we should be verifying that the messages that are sent correspond to protocol and enabled extensions.  You don't have to `await response()` to invoke the `send()` method.

starlette/websockets.py
```diff
@@ -219,10 +200,9 @@ async def close(
         )
 
     async def send_response(self, response: Response) -> None:
-        if self._have_response_extension():
-            await response(self.scope, self.receive, self.send)
-        else:
-            await self.close(code=1008, reason=f"HTTP Response {response.status_code}")
+        if "websocket.http.response" in self.scope.get("extensions", {}):
+            return await response(self.scope, self.receive, self.send)
+        raise RuntimeError("The server doesn't support the WebSocket Denial extension.")
```
kristjanvalur: For the record, I disagree here.  You are forcing the creation of boilerplate in the application.  When the server does not support the extension, a close() is the only available option.   This _requires_ the application to either always check for the presence of the extension, _or_ have a try_catch mechanism built in.  IMHO, sensible middleware should simply do the right thing, possibly issuing a warning.  Middleware like Starlette is supposed to  make the interaction with ASGI simpler, and remove such considerations from the application developer.

kristjanvalur (commented):
Well, this is a substantial rewrite of my PR so I don't think you need my approval for any of it. 
I didn't review the tests changes for that reason.

The changes, as I see them, are

- in the TestClient where you removed the ability to test for specific types of close events
- Removal of protocol verification in the `send()` command
- Removal of the return of the local disconnect event from `receive()`.  
- Requiring application to add boilerplate code if it wants to terminate with a response.

Fine by me, but like I said, it is almost not my PR anymore :)

starlette/testclient.py
```diff
@@ -68,6 +68,18 @@ def __init__(self, session: "WebSocketTestSession") -> None:
         self.session = session
 
 
+class DenialResponse(Exception):
+    def __init__(
+        self,
+        status_code: int,
+        headers: typing.List[typing.Tuple[bytes, bytes]] = [],
```
paulo-raca: Seems clumsy -- Why not `Dict[str, str]` (or `starlette.datastructures.Headers`)?

In fact, this should probably contain a `httpx.Response` -- this would allow reusing code more easily

starlette/websockets.py
paulo-raca: I agree, this should fallback to `close()` if the extension isn't supported

starlette/websockets.py
```diff
@@ -181,6 +199,11 @@ async def close(
             {"type": "websocket.close", "code": code, "reason": reason or ""}
         )
 
+    async def send_response(self, response: Response) -> None:
```
paulo-raca: nit: I would call it `send_denial_response`, as it is a better match to the specs and makes it clear that it is a failure scenario.

starlette/testclient.py
```diff
@@ -68,6 +68,18 @@ def __init__(self, session: "WebSocketTestSession") -> None:
         self.session = session
 
 
+class DenialResponse(Exception):
```
paulo-raca: nit: maybe make it clear it is websocket-only exception?

```suggestion
class WebsocketDenialResponse(Exception):
```

paulo-raca (commented):
Thank you for picking this up :heart:

Kludex:
I've reverted my changes. I'll take your comments in consideration.

starlette/websockets.py
Kludex: This is unrelated to the PR.

kristjanvalur: Strictly yes, although the PR was an overhaul of the disconnect mechanism and it seemed prudent to fill in the cracks, and it fitted well with the testing added. But I can remove the `app_disconnect_msg` system from this pr, and so an app won't `receive()` any local `close()`.

Kludex: If you can do that, it would be cool. Sorry for the back and forth here...

starlette/websockets.py
Kludex: I understand what you are saying, but I don't think the premise is right.

If you start using `send_denial_response`, and you don't have an error, you'll believe it's working as expected, when it actually is not.

starlette/websockets.py
```diff
@@ -66,12 +73,19 @@ async def send(self, message: Message) -> None:
...
-                )
+                if not (
+                    self._have_response_extension()
+                    and message_type == "websocket.http.response.start"
+                ):
+                    raise RuntimeError(
+                        'Expected ASGI message "websocket.accept"'
+                        ' or "websocket.close", '
+                        f"but got {message_type!r}"
+                    )
```
Kludex: Do we have a case where we don't have the extension, and we send a `"websocket.http.response.start"`?

The only place we send a `websocket.http.response.start` is on `send_denial_response`, isn't it?

starlette/websockets.py
```diff
@@ -53,6 +58,8 @@ async def receive(self) -> Message:
                 )
             if message_type == "websocket.disconnect":
                 self.client_state = WebSocketState.DISCONNECTED
+                if "code" not in message:
+                    message["code"] = 1005  # websocket spec
```
Kludex: This can also be removed from this PR, I guess?

starlette/testclient.py
```diff
@@ -152,11 +177,40 @@ async def _asgi_receive(self) -> Message:
...
-            )
+            if not reject:
+                raise WebSocketDisconnect(
+                    message.get("code", 1000), message.get("reason", "")
+                )
+            else:
+                # A webserver which gets a "close" before "accept"
+                # will return a 403 response.  It may or may not do anything
+                # with the close code and reason.  We store it in this exception
+                # for test introspection.
```
Kludex: If the `code` and `reason` are not being used, why the `WebSocketReject` would store it?

I think this is more confusing.

starlette/testclient.py
```diff
@@ -77,6 +77,27 @@ def __init__(self, session: WebSocketTestSession) -> None:
         self.session = session
 
 
+class WebSocketReject(WebSocketDisconnect):
```
Kludex: Can we rename this?

```suggestion
class WebSocketDenialResponse(WebSocketDisconnect):
```

starlette/testclient.py
```diff
@@ -152,11 +177,40 @@ async def _asgi_receive(self) -> Message:
...
+            else:
+                # A webserver which gets a "close" before "accept"
+                # will return a 403 response.  It may or may not do anything
+                # with the close code and reason.  We store it in this exception
+                # for test introspection.
+                raise WebSocketReject(
+                    response_status=403,
+                    close_code=message.get("code", 1000),
+                    close_reason=message.get("reason"),
+                )
```
Kludex: This is actually a change that doesn't need to be in this PR. This is actually "fixing" the current state, since the client would actually receive a 403, instead of the code, and reason on the WebSocket.

Please remove this from the PR. Let's discuss this after.

starlette/testclient.py
```diff
@@ -101,10 +122,14 @@ def __enter__(self) -> WebSocketTestSession:
...
             message = self.receive()
-            self._raise_on_close(message)
+            if message["type"] == "websocket.http.response.start":
+                self._handle_response(message)
+            else:
+                self._raise_on_close(message, reject=True)
         except Exception:
             self.exit_stack.close()
             raise
+        assert message["type"] == "websocket.accept"
```
Kludex: Is this here for type checking reasons?

starlette/testclient.py
```diff
@@ -77,6 +77,27 @@ def __init__(self, session: WebSocketTestSession) -> None:
...
+    or a websocket.close()
+    """
+
+    def __init__(
+        self,
+        response_status: int,
+        response_headers: typing.List[typing.Tuple[bytes, bytes]] = [],
+        response_body: bytes = b"",
+        close_code: int = 1000,
+        close_reason: typing.Optional[str] = None,
```
Kludex: I think this class should only have non-WebSocket related data i.e. `status_code`, `headers`, and `body`.

Kludex (changes requested):

Kludex:
@kristjanvalur I think we are almost there. 🙏

starlette/websockets.py
kristjanvalur: Yes, sure.

starlette/websockets.py
kristjanvalur: Yes, that is true, but `send()` is already doing a lot of state checking.  I guess we can remove the check from here.

starlette/websockets.py
kristjanvalur: My thinking is that the denial response is an informative event.  It augments the plain close event, possibly with
some nicely formatted data, information about what is missing, etc.  But it is not essential to the operation of the _service_.

I understand it as "I'm not accepting this connection.  Here is some additional information to the client, _if_ I can send it and _if_ he wants to receive it, but otherwise, I'm done with the connection."

This is why I suggested that maybe a warning were appropriate.

But I don't feel strongly about this anymore and am happy to do it either way, so having made my point, I can make it an exception.

starlette/testclient.py
kristjanvalur: sure.

starlette/testclient.py
kristjanvalur: So that tests, using the TestClient, can verify that the _application_ sent the code and reason.
This is a TestClient, designed to test if the application is behaving in a certain way.  If I write my websocket app, and return a close and reason, in my unit tests I want to make sure that that close and reason are correctly generated.

starlette/testclient.py
kristjanvalur: Okay

starlette/testclient.py
kristjanvalur: I don't think so, I think this is a genuine assert I put there, since this is a test client.
I can remove it if you like.

starlette/testclient.py
kristjanvalur: I'll remove it from the PR.  The intention is for the TestClient to be able to verify the behaviour of the application which is being tested, the arguments that it passes to the close() command.

kristjanvalur: I'll remove the parameters.  But it is still a subclass of WebSocketDisconnect.

kristjanvalur: Right, now I get it.  I had the parameters because I used this class for both kinds of "rejection" response.  The one
done with close(), and the wone with http.response.
Now I raise a WebSocketDisconnect for the close() type reject, but that means that a TestClient won't know the WebSocketDisconnect
is because of a reject.  it looks just like a regular close().  I can live with that.

starlette/testclient.py
kristjanvalur: WebSocketReject() (now WebSocketDenialResponse) was previously sent both for close() type rejection and also for http.response type rejection.  The idea was that this class would capture as much as possible about the rejection event, for the TestClient to be able to see how the application performed its connection rejection.

Latest change no longer sends a WebSocketDenialResponse() when a close() type rejection is performed.

kristjanvalur:
Latest commit separates the two possible denial ecetpions into WebSocketDenialClose and WebSocketDenialResponse.
Both are instances of WebSocketDenial, thus allowing tests to identify a denial.
This is just a suggestion, I can revert this if you like.

starlette/testclient.py
Kludex: Yes, please.
```suggestion
```

starlette/websockets.py
Kludex: You mean that you want to check if the extension is enabled here?

starlette/testclient.py
```diff
@@ -77,6 +77,33 @@ def __init__(self, session: WebSocketTestSession) -> None:
...
+
+    def __init__(
+        self,
+        response_status: int,
+        response_headers: typing.List[typing.Tuple[bytes, bytes]] = [],
+        response_body: bytes = b"",
+    ) -> None:
+        self.response_status = response_status
+        self.response_headers = response_headers
+        self.response_body = response_body
```
```suggestion
        status_code: int,
        headers: typing.List[typing.Tuple[bytes, bytes]] = [],
        body: bytes = b"",
    ) -> None:
        self.status_code = status_code
        self.headers = headers
        self.body = body
```

starlette/testclient.py
```diff
@@ -77,6 +77,33 @@ def __init__(self, session: WebSocketTestSession) -> None:
...
+    """
+    A special case of WebSocketDisconnect, raised in the TestClient if the
+    with a send_denial_response()
+    or a websocket.close()
+    """
+
+    def __init__(
+        self,
+        response_status: int,
+        response_headers: typing.List[typing.Tuple[bytes, bytes]] = [],
```
Kludex: The `response.headers` from the `TestClient` are a dictionary, right? Can we make this as well? (please confirm)

docs/websockets.md
```diff
@@ -109,3 +109,18 @@ correctly updated.
...
+
+If you want to send a different error response, you can use the
+`websocket.send_denial_response()` method. This will send the response
+and then close the connection.
+
+* `await websocket.send_denial_response(response)`
+
+This requires the ASGI server to support the `websocket.http.response`
+extension. If it is not supported a `RuntimeError` will be raised and
+the application can then elect to perform a `websocket.close()` instead.
```
Kludex: I don't think informing about `RuntimeError` makes sense. The error raised on the code should contain the information needed for the user to take action.
```suggestion

```

Kludex (changes requested):

Kludex:
> Latest commit separates the two possible denial ecetpions into WebSocketDenialClose and WebSocketDenialResponse. Both are instances of WebSocketDenial, thus allowing tests to identify a denial. This is just a suggestion, I can revert this if you like.

I'd prefer to revert.

docs/websockets.md
kristjanvalur: It helps the developer write the necessary boilerplate (a check, or an exception handler) without having to first experience the failure if his application is ever deployed with a server that does not offer the extension.  Since you don't want to support a fallback mechanism, at the very least we should document the failure mode.

starlette/websockets.py
kristjanvalur: I made the change already,
This code is doing sanity checks on the messages it is receiving based on the state machine.
Previous version would not accept "websocket.http.response.start" as a valid message at all if the extension was not supported.
After the change, we just send it to the server and leave the sanity checking to the server.
I'm not sure, though.  maybe we should validate messages in this place.

starlette/testclient.py
kristjanvalur: the `httpx.Response` returned by TestClient contains headers of this type:
```
HeaderTypes = Union[
    "Headers",
    Mapping[str, str],
    Mapping[bytes, bytes],
    Sequence[Tuple[str, str]],
    Sequence[Tuple[bytes, bytes]],
]
```
The WebsocketDenialResponse contains the raw http response, thus a sequence of bytes tuples.
We could of course add a helper function to return a str:str mapping, with utf8 decoding, if you think that is worthwhile.

Hm, I guess we could even subclass httpx.Response for this case, rather than write our own.  I"ll try that.

kristjanvalur: Done, we now subclass httpx.Response

starlette/websockets.py
```diff
@@ -65,13 +70,20 @@ async def send(self, message: Message) -> None:
...
+            if message_type not in {
+                "websocket.accept",
+                "websocket.close",
+                "websocket.http.response.start",
+            }:
                 raise RuntimeError(
-                    'Expected ASGI message "websocket.accept" or '
-                    f'"websocket.close", but got {message_type!r}'
+                    'Expected ASGI message "websocket.accept"'
+                    ' or "websocket.close", '
```
Kludex: Missing the new message.

starlette/websockets.py
```diff
@@ -185,6 +207,14 @@ async def close(self, code: int = 1000, reason: str | None = None) -> None:
             {"type": "websocket.close", "code": code, "reason": reason or ""}
         )
 
+    async def send_denial_response(self, response: Response) -> None:
+        if self._have_response_extension():
```
Kludex: No need for the method.

tests/test_websockets.py
```diff
@@ -290,19 +291,124 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+        msg = await websocket.receive()
+        assert msg == {"type": "websocket.connect"}
         await websocket.close(status.WS_1001_GOING_AWAY)
 
     client = test_client_factory(app)
     with pytest.raises(WebSocketDisconnect) as exc:
         with client.websocket_connect("/"):
             pass  # pragma: no cover
-    assert exc.value.code == status.WS_1001_GOING_AWAY
+    assert exc.value.code == 1001
```
Kludex:
```suggestion
    assert exc.value.code == status.WS_1001_GOING_AWAY
```

tests/test_websockets.py
```diff
@@ -290,19 +291,124 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
 
-def test_rejected_connection(test_client_factory: Callable[..., TestClient]):
+def test_rejected_connection(test_client_factory):
+    close_msg: Message = {}
+
     async def app(scope: Scope, receive: Receive, send: Send) -> None:
+        nonlocal close_msg
         websocket = WebSocket(scope, receive=receive, send=send)
+        msg = await websocket.receive()
+        assert msg == {"type": "websocket.connect"}
```
Kludex:
```suggestion
def test_rejected_connection(test_client_factory: Callable[..., TestClient]):
    async def app(scope: Scope, receive: Receive, send: Send) -> None:
        websocket = WebSocket(scope, receive=receive, send=send)
        msg = await websocket.receive()
        assert msg == {"type": "websocket.connect"}
```

tests/test_websockets.py
```diff
@@ -290,19 +291,124 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+                "status": response.status_code,
+                "headers": response.raw_headers,
+            }
+        )
+
+    client = test_client_factory(app)
+    with pytest.raises(RuntimeError) as exc:
+        with client.websocket_connect("/"):
+            pass  # pragma: no cover
+    assert exc.match("Expected ASGI message")
```
Kludex: Let's use the full match.

tests/test_websockets.py
```diff
@@ -290,19 +291,124 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+
+
+def test_subprotocol(test_client_factory):
```
Kludex:
```suggestion
def test_subprotocol(test_client_factory: Callable[..., TestClient]):
```

tests/test_websockets.py
```diff
@@ -522,3 +628,21 @@ async def app(scope: Scope, receive: Receive, send: Send):
...
+        await websocket.accept()
+        close_msg = await websocket.receive()
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/") as websocket:
+        websocket.send({"type": "websocket.disconnect"})
+
+    assert close_msg == {
+        "type": "websocket.disconnect",
+    }
```
Kludex:
```suggestion
    assert close_msg == {"type": "websocket.disconnect"}
```

tests/test_websockets.py
```diff
@@ -522,3 +628,21 @@ async def app(scope: Scope, receive: Receive, send: Send):
     with pytest.raises(RuntimeError):
         with client.websocket_connect("/") as websocket:
             websocket.send({"type": "websocket.connect"})
+
+
+def test_send_disconnect_no_code(test_client_factory):
```
Kludex: We need an explanation about this test, or maybe turn it into a `xfail`, with the code, and write the reason on the `xfail` decorator.

Kludex (changes requested):

tests/test_websockets.py
kristjanvalur: I expect you mean the "test_send_no_disconnect" test.  I'm adding a doc string.
The original test was written to verify that the app would supply a 1005 code.
I think that is the correct code to be put in place, it is an indication that no close code
was present on the wire.  But that can be a future PR.  As it is, the test just makes sure a close is
received, even without a result code and verifies whatever the current behaviour is.

tests/test_websockets.py
```diff
@@ -522,3 +628,22 @@ async def app(scope: Scope, receive: Receive, send: Send):
...
+        nonlocal close_msg
+        websocket = WebSocket(scope, receive=receive, send=send)
+        await websocket.accept()
+        close_msg = await websocket.receive()
+
+    client = test_client_factory(app)
+    with client.websocket_connect("/") as websocket:
+        websocket.send({"type": "websocket.disconnect"})
+
+    assert close_msg == {"type": "websocket.disconnect"}
```
Kludex: @kristjanvalur This test doesn't make sense to me. Also, this is more a test for the `TestClient` than the `WebSocket`, so it should go in `test_testclient.py`.

I'll remove it, but if you want, feel free to open a new PR adding the code on the `TestClient`.

starlette/testclient.py
```diff
@@ -152,11 +166,35 @@ async def _asgi_receive(self) -> Message:
...
-            )
+            if not reject:
+                raise WebSocketDisconnect(
+                    message.get("code", 1000), message.get("reason", "")
+                )
+            else:
+                raise WebSocketDisconnect(
+                    code=message.get("code", 1000),
+                    reason=message.get("reason"),
+                )
```
Kludex: Isn't this the same exception? The default empty string proves something?

starlette/testclient.py
```diff
@@ -77,6 +77,17 @@ def __init__(self, session: WebSocketTestSession) -> None:
...
 
 
+class WebSocketDenialResponse(  # type: ignore[misc]
+    httpx.Response,
+    WebSocketDisconnect,
+):
+    """
+    A special case of WebSocketDisconnect, raised in the TestClient if the
+    socket is closed before being accepted, either with a send_denial_response()
+    or a websocket.close()
```
```suggestion
    `WebSocket` is closed before being accepted with a `send_denial_response()`.
```

starlette/testclient.py
```diff
@@ -77,6 +77,16 @@ def __init__(self, session: WebSocketTestSession) -> None:
         self.session = session
 
 
+class WebSocketDenialResponse(  # type: ignore[misc]
+    httpx.Response,
+    WebSocketDisconnect,
+):
+    """
+    A special case of WebSocketDisconnect, raised in the TestClient if the
```
```suggestion
    A special case of `WebSocketDisconnect`, raised in the `TestClient` if the
```

Kludex (approved):

Kludex:
@kristjanvalur I've invited you to the organization. Thanks for the help on the extension! 🙏 

Please check if you are happy with my last commits, and if so, feel free to merge it. 👍

Kludex:
@paulo-raca Vlw pelo review! 🙏 

English: Thanks for the review! 🙏

tests/test_websockets.py
kristjanvalur: sure, I'll remove it.

starlette/testclient.py
kristjanvalur: it is, previously we used to raise a different exception there, to indicate a special "kind" of close.  Since the distinction is now gone, I can simplify.

kristjanvalur: Ok, I see you already did all that, so never mind.

kristjanvalur:
Thank you, your commits were fine, will do.

Kludex:
This may not have been the most intuitive API. 🤔

Maybe raising `HTTPException` before the acceptance would have been more intuitive?

kristjanvalur:
As an alternative to calling `await websocket.close()` which sends the default rejection response?  That would be rather unintuitive.  This is on the ASGI protocol level where the handling on of the connection is handled by ASGI messages.   I guess an application could use exception handling to affect the same.

Any reason why this comment comes up now?  Are there any conflicts with current development?

Kludex:
> As an alternative to calling `await websocket.close()` which sends the default rejection response?

No. In contrast to raising an exception, like what I intend with this: https://github.com/encode/starlette/pull/2725

Also, don't get me wrong, this PR was great. I'm just unsure about the API on the `websocket.send_denial_response`. It looks a bit weird when coding to construct the `Response` object before calling this.

> Any reason why this comment comes up now? Are there any conflicts with current development?

I was adding this to a presentation I'm going to do, and then I noticed that we don't support raising HTTPException before the `websocket.accept()` is called, so I implemented it here:  https://github.com/encode/starlette/pull/2725. I just noticed that having that would be enough here.

kristjanvalur:
I see.  Well, exceptions are super.  We probably should have considered exceptions, but my thinking was that `close()` was the existing API to reject a connection, and to reject the connection with extra information should be done in a similar imperative way by calling a method.  I think supporting exceptions is a great alternative way to do this.

PR #1308, opened 2021-10-11
Add proper type hints to add_exception_handler

phillipuniverse:
I lost a bit of time due to a missing return statement and type hint in an exception handler. 

In my case I had registered an `HTTPException` exception handler without a return. Here's the exception I ended up with:

```
Traceback (most recent call last):
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/routing.py", line 609, in __call__
    await self.default(scope, receive, send)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/routing.py", line 497, in not_found
    raise HTTPException(status_code=404)
starlette.exceptions.HTTPException

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/middleware/cors.py", line 78, in __call__
    await self.app(scope, receive, send)
  File "/Users/phillip/Library/Caches/pypoetry/virtualenvs/myproj-0dUE00O2-py3.8/lib/python3.8/site-packages/starlette/exceptions.py", line 93, in __call__
    await response(scope, receive, sender)
TypeError: 'NoneType' object is not callable
```

Here's what my exception handler and registration looked like:

```python
...

async def http_exception_handler(request, exc):
   JSONResponse(
        status_code=exc.status_code,
        content=jsonable_encoder(
            {
                "errors": [
                    {
                        "id": uuid4(),
                        "title": "Internal Exception",
                        "detail": exc.detail,
                    }
                ]
            }
        ),
    )

application.add_exception_handler(HTTPException, http_exception_handler)
```

This is clearly an error but it would've been nice to get a type hint error on it before I went through and traced down the implementation.

Thanks for such a great library!

PR #2042, opened 2023-02-19
Add typing test for Websockets exception_handler

kristjanvalur:
Version 0.21.0 added support for handing errors in WebSockets.
However, the signature of `Starlette(exception_handlers)` keyword argument was not updated and there
were no type checks catching the mismatch.  

- We tweak the `test_applications.py` to get mypy to check types
- We update the type signature for the `exception_handlers` keyword.
- We update the in-line documentation for `class Starlette`

Internally, the type hints don't match.  And they don't even allow for async functions.  In light of this, I propose a
different approach, to change the type to `Callable[[Union[Request, Websocket], Any], Any]`.  Using `Exception` as the second arg in the type strictly precludes us from defining handers with a more specific excepty type argument.  And the variable type of the first argument, variable return argument, and allowing both sync and async, well, maybe this is one place where type strictness could be relaxed.


tests/test_applications.py
```diff
@@ -81,7 +81,8 @@ async def websocket_raise_custom(websocket: WebSocket):
     raise CustomWSException()
 
 
-def custom_ws_exception_handler(websocket: WebSocket, exc: CustomWSException):
+def custom_ws_exception_handler(websocket: WebSocket, exc: Exception) -> None:
```
Kludex: Was there an issue with the typing on `exc`?

tests/test_applications.py
```diff
@@ -121,8 +115,13 @@ def custom_ws_exception_handler(websocket: WebSocket, exc: CustomWSException):
...
         Host("{subdomain}.example.org", app=subdomain),
     ],
-    exception_handlers=exception_handlers,
     middleware=middleware,
+    exception_handlers={
+        500: error_500,
+        405: method_not_allowed,
+        HTTPException: http_exception,
+        CustomWSException: custom_ws_exception_handler,
+    },
```
Kludex: Why is this needed?

kristjanvalur: I'm no expert on MyPy, and how to make it test types, but without it, the `exception_handlers` variable was apparently untyped, and mypy didn't complain about the contents of it not matching the specification.  Maybe there is a better way tom make mypy be more _strict_ in checking types in  the unittests.  the --strict mode does exists but that is completely overboard for unittests, IMHO.

By putting the dict in-line, mypy kicks into gear and starts testing to see if the contents of the dict actually matches what the type hints for `Starlette()` specify.

tests/test_applications.py
kristjanvalur: The type was not being checked.  Once type checking was enabled, it was clear that having the `CustomWSException` is incorrect (as the type spec on `Starlette(exception_handlers=...)` specifies that the second argument can be _any_ `Exception` instance, not just CustomWSException.

kristjanvalur:
I think there are two issues that maybe should be solved independently:

1. Make sure that type checking is happening in the unit tests so that we detect when we are violating the type specification.  Not sure how to do that without `--strict` paraemter to MyPy (I'm not an expert)
2. The type specification for the `exception_handlers` should perhaps be relaxed with more `typing.Any`, now that the accepted type signatures are so varied that it becomes cumbersome to use.

tests/test_applications.py
Kludex: what was the mypy issue?

tests/test_applications.py
Kludex: I don't think that makes sense. How did you prove this assumption?

kristjanvalur: by running `scripts/lint`.  

My guess is the following:  MyPy will check that the type of a _literal_ matches the argument list, but it will not automatically and implicitly assign a type to a variable (`exception_handers`) based on the literal used to initialize it.  Since the variable has no type, it won't be checked against the argument list, and mypy isn't in `--strict` mode and doesn't complain about it.

Anyway, you can clearly see that the type of `custom_ws_exception_handler`, does not match the hints, (For one thing, it specifies a `WebSocket` argument where `Starlette()` expects callbacks taking a `Request`) yet mypy does not complain about this dict.

tests/test_applications.py
kristjanvalur: Mypy complains that element 4 of the dict literal is the wrong type.

Look, if you specify a callable which takes a certain type as an argument, you cannot provide a callable which expects a _narrower_ type.  

This here is invalid typing:
```python
from typing import Callable
def set_callback(callback: Callable[[Exception], None]) -> None:
    pass
    
def my_callback(arg: RuntimeError)->None:
    pass
    
set_callback(my_callback)  # errors
```
running mypy on this results in:
```
test.py:8: error: Argument 1 to "set_callback" has incompatible type "Callable[[RuntimeError], None]"; expected "Callable[[Exception], None]"  [arg-type]
```

kristjanvalur:
Ok, I think I have figured out what is happening, why we are not getting warnings.
By adding a `exception_handlers = 1` line, I get the following error from MyPy:
```
tests/test_applications.py:107: error: Incompatible types in assignment (expression has type "int", variable has type "Dict[object, Callable[[Any, Any], Any]]")  [assignment]
```
In other words: MyPy has implicitly typed `exception_handlers` as having `Any` callables, because with the lowest common denomitator.  The callables used for initializing it are untyped.
And a dict, typed that way, is valid as an argument.

By adding type hints to all of the exception handlers, MyPy gets clued in and it fails:
```
tests/test_applications.py:126: error: Argument "exception_handlers" to "Starlette" has incompatible type "Dict[object, function]"; expected "Optional[Mapping[Any, Callable[[Request, Exception], Union[Response, Awaitable[Response]]]]]"  [arg-type]
```

I think the best course of action here is to excplicitly define exception handlers as being of type `Callable[[Any, Any], Any]` and entrust this part of the API to Python's dynamic typing.

kristjanvalur:
This file demonstrates the issue.  I guess I'll need to raise an issue over at MyPy.
Despite being strongly typed, the handler2 is accepted as a callable where a different signature is present.
```python
from typing import Callable, Any, Mapping


def api(handlers: Mapping[Any, Callable[[str], int]]) -> None:
    pass


def handler1(arg):
    return 1

def handler2(arg:int) -> str:
    return "foo"


handlers = {
    1: handler1,
    2: handler2
}

api(handlers)
```

kristjanvalur:
Well, MyPy promptly closed the issue without discussion.  I guess that's that then.
https://github.com/python/mypy/issues/14779

My suggestion, then, is to simply make sure that we don't add callables to intermediate dicts, to ensure that the typing information of the individual handlers is actually checked against the spec.

adriangb:
We've kinda been through this before: https://github.com/encode/starlette/pull/1308.

I think the best thing we could do is something like this:

https://mypy-play.net/?mypy=latest&python=3.11&gist=ace307dc705717b1ef1fedc21708b8f3

Which would be a pure addition, nothing inside `class Starlette` or anywhere else needs to change.

Or we could not inherit from `dict` and just keep it as a private attribute / implementation detail but that would require supporting two code-paths inside of `class Starlette`.

kristjanvalur:
btw, the problem with typechecking can be disstilled to this:

```python
from typing import Callable, Any, List

def handler1(arg):
    return 1

def handler2(arg:int) -> str:
    return "foo"

callabak: Callable[[str],int] = handler1  # ok, Any callable can take str
callback: Callable[[Any],int] = handler2  # not ok, `str` callable can't expect Any
callbacks:List[Callable[[Any], str]] = [handler2] # is ok, but shouldn't be!
```
When we created a dict of handlers, we used an untyped one, and so the callable type of Any snuck in.  MyPy then allows a typed callable with a narrower type to be assigned.

kristjanvalur:
> We've kinda been through this before: #1308.
> 
> I think the best thing we could do is something like this:

Yes, a TypeVar would allow to have typed handlers with their own exception subclasses.
IMHO I think it is overly complex.  We should simply allowe untyped handers with the specified number of arguments.
Sometimes dynamic is best.

starlette/applications.py
```diff
@@ -47,9 +49,15 @@ def __init__(
...
-                    typing.Union[Response, typing.Awaitable[Response]],
+                typing.Union[
+                    typing.Callable[
+                        [Response, Exception],
+                        typing.Union[Response, typing.Awaitable[Response]],
+                    ],
+                    typing.Callable[
+                        [WebSocket, Exception],
+                        typing.Union[None, typing.Awaitable[None]],
+                    ],
```
Kludex: Can we revert everything else in this PR and only maintain this change?

tests/test_applications.py
kristjanvalur: The problem was that some of the exception handlers were untyped (ANY) and then the whole list collapses to that and even your carefully typed exception handlers don't get any testing.

starlette/applications.py
kristjanvalur: sure.  Except, in stead of `Exception` here I think what you really want is `Any`.  Otherwise, you cannot define an exception handler taking an exception class other than Exception without type checking complaining.

PIG208:
> This file demonstrates the issue. I guess I'll need to raise an issue over at MyPy. Despite being strongly typed, the handler2 is accepted as a callable where a different signature is present.

This is the expected behavior because mypy infers the common type that unionizes with all the values of the `dict` for its value. If you don't want permissive type inference, you need to annotate the type of the `dict`.

kristjanvalur:
> This is the expected behavior because mypy infers the common type that unionizes with all the values of the `dict` for its value. If you don't want permissive type inference, you need to annotate the type of the `dict`.

Yes.  This was explained to me over at MyPy too, basically, an untyped method __silently__  poisons the dict.
It's one of those things which catches people which aren't seasoned in Python type hints unawares.  And in the case being
demonstrated here, caused us (and others) to not notice the discrepancy in types.

Kludex:
This is outdated. We have created the `ExceptionHandler` type.

PR #2043, opened 2023-02-20
Update link to jinja2

magnuswatn:
The old link doesn't support https, which makes it not work for https-only users, and it only redirects to this one anyway.


euri10 (approved):

Kludex (approved):
Thanks 🙏

PR #2044, opened 2023-02-20
Docs: Remove SentryMiddleware from third-party list

frankie567:
I was going through the [third-party middlewares list](https://www.starlette.io/middleware/#sentrymiddleware) in the docs and came across [`SentryMiddleware`](https://github.com/encode/sentry-asgi). As stated in the project:

> Update Sentry now includes built-in support for ASGI. (Based on this implementation.)

So I thought it may not be useful to mention it in the docs anymore.

What we could do however (not done, but I would happily do it) is to add `SentryASGIMiddleware` and link to Sentry docs: https://docs.sentry.io/platforms/python/guides/quart/configuration/integrations/asgi/ What do you think?

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -773,10 +773,6 @@ A middleware class for reading/generating request IDs and attaching them to appl
 
 A middleware class for logging exceptions, errors, and log messages to [Rollbar](https://www.rollbar.com).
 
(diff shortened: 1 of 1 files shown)


euri10 (approved):

Kludex:
We should probably add something about Sentry. I think it was even discussed to be on the Exceptions section.

I'll merge this tho, it's simple.

PR #2050, opened 2023-03-01
Allow "name" argument on `url_for()` and `url_path_for()`

### LINKED ISSUE #608: "url_for" signature prevents use of "name" in path arguments

Kludex:
- Closes https://github.com/encode/starlette/issues/608
- Replaces https://github.com/encode/starlette/pull/611

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -108,8 +108,9 @@ class Starlette:
     def routes(self) -> typing.List[BaseRoute]:
         return self.router.routes
 
-    def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
-        return self.router.url_path_for(name, **path_params)
+    # TODO: Make `__name` a positional-only argument when we drop Python 3.7 support.
+    def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath:
+        return self.router.url_path_for(__name, **path_params)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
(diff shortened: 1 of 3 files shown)


Kludex:
@encode/maintainers :eyes: :pray:

adriangb:
This looks good to me. But admittedly I don't really use this functionality so I'd like someone who actually uses this to be the one approving ✅

Kludex:
I want to make a premature release with the lifespan state that was included, but since I included https://github.com/encode/starlette/pull/1385, it may be good to also include this in the release. :sweat_smile: 

@alex-oleshkevich @abersheeran would you mind checking this? :pray:

abersheeran (approved):
I feel very good. 😊

Issue #892, opened 2020-04-06
Unable to get request body inside an exception handler, got RuntimeError: Receive channel has not been made available


sevaho:
Hello

I want to get the request body in an exception handler but I am getting an Error. Is this a bug?

Minimal example

```python
from fastapi import FastAPI
from starlette.requests import Request
from starlette.responses import UJSONResponse

app = FastAPI()


async def http_exception_handler(request: Request, exc: Exception) -> UJSONResponse:
    print(await request.body())  # RuntimeError: Receive channel has not been made available
    return UJSONResponse(content={"msg": f"SERVER_ERROR, {exc.args[0]}", "code": 500}, status_code=500)


app.add_exception_handler(Exception, http_exception_handler)


@app.post("/")
async def index(request: Request):
    print(await request.body())  # b'{"a": "b"}'
    raise Exception("CATCH ME")


# Run http localhost:8000 some=data
```

Error:

```txt
b'{"a": "b"}'
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/routing.py", line 550, in __call__
    await route.handle(scope, receive, send)
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/routing.py", line 227, in handle
    await self.app(scope, receive, send)
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/starlette/routing.py", line 41, in app
    response = await func(request)
  File "/home/sevaho/.local/share/virtualenvs/testje2-JmgYoeLF/lib/python3.8/site-packages/fastapi/routing.py", line 196, in app
    raw_response = await run_endpoint_function(
[...]

PR #2056, opened 2023-03-02
fix empty_receive for exception handler

tonsh:
As describe in #892, I only want to get request body in exception handler but don't want implement it use APIRouter in fastAPI.
I think the simple way is send the `receive` argument replace to the empty_receiver.


Kludex:
Thanks for the PR! :raised_hands:

This is already being worked on https://github.com/encode/starlette/pull/2026.

Issue #2057, opened 2023-03-02
Incorrect typing for Scope and Message in ASGIApp


ods:
- [x] Initially raised as discussion #2040 (but ignored there)

According to [ASGI spec](https://asgi.readthedocs.io/en/latest/specs/main.html#applications) both scope and event message must be dictionaries, not an arbitrary mutable mapping:

> * `scope`: The connection scope information, a **dictionary** that contains at least a `type` key specifying the protocol that is incoming
> * `receive`: an awaitable callable that will yield a new event **dictionary** when one is available
> * `send`: an awaitable callable taking a single event **dictionary** as a positional argument that will return once the send has been completed or the connection has been closed

and in other place:

> `scope` must be a `dict`.

But in starlette [they are declared](https://github.com/encode/starlette/blob/master/starlette/types.py#L3-L4) as `MutableMapping`:

```python
Scope = typing.MutableMapping[str, typing.Any]
Message = typing.MutableMapping[str, typing.Any]

Receive = typing.Callable[[], typing.Awaitable[Message]]
Send = typing.Callable[[Message], typing.Awaitable[None]]

ASGIApp = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
```

This actually breaks interoperability with other libraries that follows the spec. For example, the following code

```python
from starlette.applications import Starlette
from httpx import ASGITransport

ASGITransport(Starlette())
```

gives the following error when type-checked with mypy:

```
error: Argument 1 to "ASGITransport" has incompatible type "Starlette"; expected "Callable[[Dict[str, Any], Callable[[], Awaitable[Dict[str, Any]]], Callable[[Dict[str, Any]], Coroutine[None, None, None]]], Coroutine[None, None, None]]"  [arg-type]
note: "Starlette.__call__" has type "Callable[[Arg(MutableMapping[str, Any], 'scope'), Arg(Callable[[], Awaitable[MutableMapping[str, Any]]], 'receive'), Arg(Callable[[MutableMapping[str, Any]], Awaitable[None]], 'send')], Coroutine[Any, Any, None]]"
```

The [typing of ASGI app in httpx](https://github.com/encode/httpx/blob/master/httpx/_transports/asgi.py#L17-L24) follows the spec:

```python
_Message = typing.Dict[str, typing.Any]
_Receive = typing.Callable[[], typing.Awaitable[_Message]]
_Send = typing.Callable[
[...]

PR #2058, opened 2023-03-02
Fix typing of ASGIApp (#2057)

ods:
According to ASGI spec both scope and event message must be dictionaries, not an arbitrary mutable mapping.

- [x] Initially raised as discussion #2040

### DIFF
--- a/starlette/types.py
+++ b/starlette/types.py
@@ -1,7 +1,7 @@
 import typing
 
-Scope = typing.MutableMapping[str, typing.Any]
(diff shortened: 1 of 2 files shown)


Kludex:
Let's discuss it first, please: https://github.com/encode/starlette/discussions/2059.

ods:
> Let's discuss it first, please: https://github.com/encode/starlette/discussions/2059.

Sure, I've created discussion first.  But it didn't work: https://github.com/encode/starlette/discussions/2040.

PR #2060, opened 2023-03-04
Support lifespan state

Kludex:
- This PR implements the Lifespan state mentioned on the ASGI spec: https://github.com/django/asgiref/pull/354.
- This PR complements https://github.com/encode/uvicorn/pull/1818.

## Changes
- Implement lifespan state.
- Remove _generator function way_ to use lifespan.

## How to review

You need to have both this branch and the one from https://github.com/encode/uvicorn/pull/1818 on your virtual environment, and then you can use this application to test the behavior:

```python
from typing import Any, Dict

# async def startup(state: Dict[str, Any]):
#     print(state)
#     state["potato"] = True
#     print(state)
#     print("startup")


# async def stateless_startup():
#     print("stateless_startup")


@asynccontextmanager
async def state_manager(app, state):
    state["haha"] = True
    print(state)
    yield
    print(state)

@asynccontextmanager
async def stateless_manager(app):
    print("stateless manager")
    yield

async def homepage(request):
    print(request.state)
    return PlainTextResponse("Hello, world!")


app = Starlette(
    routes=[Route("/", homepage)],
    # on_startup=[startup, stateless_startup],
    lifespan=state_manager,
    # lifespan=stateless_manager,
)
```

Play with the comments above. :eyes:

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -679,22 +679,20 @@ class Router:
         """
         started = False
         app = scope.get("app")
+        state = scope.get("state")
         await receive()
         lifespan_needs_state = (
             len(inspect.signature(self.lifespan_context).parameters) == 2
         )
-        server_supports_state = "state" in scope
+        server_supports_state = state is not None
         if lifespan_needs_state and not server_supports_state:
             raise RuntimeError(
-                'This server does not support "state" in the lifespan scope.'
-                " Please try updating your ASGI server."
+                'The server does not support "state" in the lifespan scope.'
             )
         try:
             lifespan_context: Lifespan
             if lifespan_needs_state:
-                lifespan_context = functools.partial(
-                    self.lifespan_context, state=scope["state"]
-                )
+                lifespan_context = functools.partial(self.lifespan_context, state=state)
             else:
                 lifespan_context = typing.cast(StatelessLifespan, self.lifespan_context)
             async with lifespan_context(app):
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -1,12 +1,9 @@
-import contextlib
 import functools
 import inspect
 import re
 import traceback
-import types
 import typing
 import warnings
-from contextlib import asynccontextmanager
 from enum import Enum
 
 from starlette._utils import is_async_callable
@@ -17,7 +14,7 @@ from starlette.exceptions import HTTPException
 from starlette.middleware import Middleware
 from starlette.requests import Request
 from starlette.responses import PlainTextResponse, RedirectResponse
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp, Lifespan, Receive, Scope, Send, StatelessLifespan
 from starlette.websockets import WebSocket, WebSocketClose
 
 
@@ -530,32 +527,7 @@ class Host(BaseRoute):
 _T = typing.TypeVar("_T")
 
 
-class _AsyncLiftContextManager(typing.AsyncContextManager[_T]):
-    def __init__(self, cm: typing.ContextManager[_T]):
-        self._cm = cm
-
-    async def __aenter__(self) -> _T:
-        return self._cm.__enter__()
-
-    async def __aexit__(
-        self,
-        exc_type: typing.Optional[typing.Type[BaseException]],
-        exc_value: typing.Optional[BaseException],
-        traceback: typing.Optional[types.TracebackType],
-    ) -> typing.Optional[bool]:
-        return self._cm.__exit__(exc_type, exc_value, traceback)
-
-
-def _wrap_gen_lifespan_context(
-    lifespan_context: typing.Callable[[typing.Any], typing.Generator]
-) -> typing.Callable[[typing.Any], typing.AsyncContextManager]:
-    cmgr = contextlib.contextmanager(lifespan_context)
-
-    @functools.wraps(cmgr)
-    def wrapper(app: typing.Any) -> _AsyncLiftContextManager:
-        return _AsyncLiftContextManager(cmgr(app))
-
-    return wrapper
+_TDefaultLifespan = typing.TypeVar("_TDefaultLifespan", bound="_DefaultLifespan")
 
 
 class _DefaultLifespan:
@@ -563,12 +535,17 @@ class _DefaultLifespan:
         self._router = router
 
     async def __aenter__(self) -> None:
-        await self._router.startup()
+        await self._router.startup(state=self._state)
 
     async def __aexit__(self, *exc_info: object) -> None:
-        await self._router.shutdown()
+        await self._router.shutdown(state=self._state)
 
-    def __call__(self: _T, app: object) -> _T:
+    def __call__(
+        self: _TDefaultLifespan,
+        app: object,
+        state: typing.Optional[typing.Dict[str, typing.Any]],
+    ) -> _TDefaultLifespan:
+        self._state = state
         return self
 
 
@@ -580,9 +557,7 @@ class Router:
         default: typing.Optional[ASGIApp] = None,
         on_startup: typing.Optional[typing.Sequence[typing.Callable]] = None,
         on_shutdown: typing.Optional[typing.Sequence[typing.Callable]] = None,
-        lifespan: typing.Optional[
-            typing.Callable[[typing.Any], typing.AsyncContextManager]
-        ] = None,
+        lifespan: typing.Optional[Lifespan] = None,
     ) -> None:
         self.routes = [] if routes is None else list(routes)
         self.redirect_slashes = redirect_slashes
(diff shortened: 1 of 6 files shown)


starlette/routing.py
```diff
@@ -639,21 +600,31 @@ def url_path_for(self, name: str, **path_params: typing.Any) -> URLPath:
...
-    async def startup(self) -> None:
+    async def startup(
+        self, state: typing.Optional[typing.Dict[str, typing.Any]]
+    ) -> None:
         """
         Run any `.on_startup` event handlers.
         """
         for handler in self.on_startup:
+            sig = inspect.signature(handler)
+            if len(sig.parameters) == 1 and state is not None:
```
Kludex, line 653: If the state is `None`, it means that the server doesn't support it, so we'll maintain the same error message.

starlette/routing.py
```diff
@@ -591,27 +566,13 @@ def __init__(
         self.on_shutdown = [] if on_shutdown is None else list(on_shutdown)
 
         if lifespan is None:
-            self.lifespan_context: typing.Callable[
-                [typing.Any], typing.AsyncContextManager
-            ] = _DefaultLifespan(self)
-
-        elif inspect.isasyncgenfunction(lifespan):
```
Kludex, line 598: I took the opportunity to remove those... But I can readd them, and remove only on 1.0. Please ask me to readd.

starlette/routing.py
```diff
@@ -530,45 +527,25 @@ def __repr__(self) -> str:
...
-    def __call__(self: _T, app: object) -> _T:
+        await self._router.shutdown(state=self._state)
+
+    def __call__(
+        self: _TDefaultLifespan,
+        app: object,
+        state: typing.Optional[typing.Dict[str, typing.Any]],
+    ) -> _TDefaultLifespan:
+        self._state = state
         return self
```
adriangb, line 580: At first glance I'm a bit confused about this and the need for a TypeVar.

starlette/routing.py
```diff
@@ -591,27 +566,13 @@ def __init__(
...
-            )
-            self.lifespan_context = _wrap_gen_lifespan_context(
-                lifespan,  # type: ignore[arg-type]
+            self.lifespan_context: Lifespan = _DefaultLifespan(self)
+        elif inspect.isasyncgenfunction(lifespan) or inspect.isgeneratorfunction(
+            lifespan
+        ):
+            raise RuntimeError(
+                "Generator functions are not supported for lifespan, "
+                "use an @contextlib.asynccontextmanager function instead."
```
adriangb: If we'ere removing deprecated things, why not just type this as a contextmanager and just hope it quacks like a duck?

starlette/testclient.py
```diff
@@ -243,6 +245,7 @@ def handle_request(self, request: httpx.Request) -> httpx.Response:
                 "client": ["testclient", 50000],
                 "server": [host, port],
                 "subprotocols": subprotocols,
+                "state": self.state,
```
adriangb: I'm not seeing where this dict is copied. Is it being copied before being passed into each request?

starlette/types.py
```diff
@@ -7,3 +7,9 @@
 Send = typing.Callable[[Message], typing.Awaitable[None]]
 
 ASGIApp = typing.Callable[[Scope, Receive, Send], typing.Awaitable[None]]
+
+StatelessLifespan = typing.Callable[[object], typing.AsyncContextManager]
+StateLifespan = typing.Callable[
+    [typing.Any, typing.Dict[str, typing.Any]], typing.AsyncContextManager
+]
+Lifespan = typing.Union[StatelessLifespan, StateLifespan]
```
adriangb: Very nice!

tests/test_routing.py
```diff
@@ -669,6 +669,40 @@ def run_shutdown():
...
+
+    async def run_startup(state):
+        nonlocal startup_complete
+        startup_complete = True
+        state["startup"] = True
+
+    async def run_shutdown(state):
+        nonlocal shutdown_complete
+        shutdown_complete = True
+        assert state["startup"]
```
adriangb: Do we want to support state in `on_startup` and `on_shutdown`? Don't we want to deprecate that in 1.0? If we were not planning on it, we should be. I think we should only support async context manager lifespans in 1.0.

tests/test_routing.py
```diff
@@ -669,6 +669,40 @@ def run_shutdown():
     assert shutdown_complete
 
 
+def test_lifespan_with_state(test_client_factory):
+    startup_complete = False
+    shutdown_complete = False
+
+    async def hello_world(request):
+        assert request.state.startup
+        return PlainTextResponse("hello, world")
```
adriangb: I'd like to see a test that shows that when a request modifies the state: (1) it doesn't get propagated up to the lifespan and (2) it doesn't leak across requests.

adriangb (approved):
Overall looks good! I have some questions about deprecations, unless it makes the PR a lot more complicated I would suggest doing the deprecations in another PR. You could always support this _new_ feature only for async context manager dependencies to avoid having to implement things for all of the other types of lifespans / startup/shutdown events we support. But not sure how that works out code wise.

tests/test_routing.py
Kludex: > Do we want to support state in `on_startup` and `on_shutdown`?

I don't think we discussed it.

What are the problems we have there besides "multiple ways to do things"?

adriangb: I'd say just that: multiple ways to do things.

Kludex: I really like the `on_startup` and `on_shutdown` parameters, and given that it would be one more thing to do for 1.0, I'm not in favor of doing it.

starlette/routing.py
Kludex: You mean just removing the conditional?

tests/test_routing.py
adriangb: Okay fair enough, it's besides the point of this PR. Not sure why you like them but I'm fairly certain the team decided to promote context manager lifespans as the way forward, so can we at least write the tests using that?

Kludex: Do you know where it was discussed?

adriangb: Not specifically, just my recollection. Maybe @graingert does, I know they advocated heavilty for the async context manager approach.

starlette/testclient.py
Kludex: Ups :eyes:

starlette/routing.py
Kludex: I'm going to readd them

tests/test_routing.py
```diff
@@ -669,6 +669,45 @@ def run_shutdown():
...
+        nonlocal startup_complete
+        startup_complete = True
+        state["count"] = 0
+        state["list"] = []
+
+    async def run_shutdown(state):
+        nonlocal shutdown_complete
+        shutdown_complete = True
+        assert state["count"] == 0
+        assert state["list"] == [1, 1]
```
Kludex, line 700: Do want me to add a comment here?

adriangb:
```suggestion
        # modifications made to the state from a request do not leak to the lifespan
        assert state["count"] == 0
        # unless of course the request mutates a mutable object that is referenced via state
        assert state["list"] == [1, 1]
```

tests/test_routing.py
```diff
@@ -669,6 +669,45 @@ def run_shutdown():
     assert shutdown_complete
 
 
+def test_lifespan_with_state(test_client_factory):
+    startup_complete = False
+    shutdown_complete = False
+
+    async def hello_world(request):
+        request.state.count += 1
+        request.state.list.append(1)
```
adriangb:
```suggestion
        # modifications to the state should not leak across requests
        assert request.state.count == 0
        # modify the state, this should not leak to the lifespan or other requests
        request.state.count += 1
        # since state.list is a mutable object this modification _will_ leak across requests and to the lifespan
        request.state.list.append(1)
```

starlette/routing.py
Kludex: The reason for the `TypeVar` is that it needed to be bound, as the `_state` was introduced here, since the `self` is annotated.

adriangb (approved):

adriangb (approved):

adriangb (changes requested):
https://github.com/encode/starlette/pull/2061

starlette/routing.py
```diff
@@ -666,9 +679,23 @@ async def lifespan(self, scope: Scope, receive: Receive, send: Send) -> None:
...
         app = scope.get("app")
+        state = scope.get("state")
         await receive()
+        lifespan_needs_state = (
+            len(inspect.signature(self.lifespan_context).parameters) == 2
+        )
+        server_supports_state = state is not None
+        if lifespan_needs_state and not server_supports_state:
+            raise RuntimeError(
+                'The server does not support "state" in the lifespan scope.'
```
Kludex, line 690: I removed the "please change server", and use the created the `state` variable to save one line below... 😬

adriangb (approved):

adriangb:
Before we release this, just wondering if this API might be better:

```python
Lifespan = Callable[[Starlette], AsyncContextManager[None | Dict[str, Any]]]

class MyStateType(TypedDict):
    client: httpx.AsyncClient

@asynccontextmanager
async def lifespan(app: Starlette) -> MyStateType:
    async with httpx.AsyncClient() as client:
        yield {"client": client}
```

I think it would also be even _less_ changes in Starlette since we could do:

```python
async with lifespan as maybe_state:
    if maybe_state:
        scope["state"].update(maybe_state)
```

And not have to introspect the function signature or anything.

Kludex:
We'd still have to introspect in the `startup` and `shutdown`, no?

adriangb:
Not sure how to handle `shutdown`. `startup` would be easy: just let it return `None | Dict[str, Any]`. But I'd also be okay only supporting this for context manager lifespans.

Kludex:
My question was more about what you said:

> I think it would also be even less changes in Starlette since we could do:

> And not have to introspect the function signature or anything.

Considering that, we'd still need to introspect for `shutdown`, at least.

That said... I think what you propose looks better for `lifespan`, but I'm not sure if it's worth it... Do you want to create a PR, and we discuss there?

empicano:
I discovered this yesterday after upgrading from an older version and just wanted to say that I'm delighted by it. It's so much nicer than using globals! Thanks for the work that went into this 🔥

PR #2062, opened 2023-03-05
(docs) Updating Sentry: 3rd Party Middlewares

smeubank:
Sentry has a dedicated integration for Starlette. Checking the other 3rd part support here, I thought it might be relevant to add the support for performance data collection, including links to the OpenTelemetry support

Discussion comment: https://github.com/encode/starlette/discussions/1787#discussioncomment-3314878

Related PR: https://github.com/encode/starlette/pull/2044

Apologies, I am a bit late to following up from the discussion :)

PR #2063, opened 2023-03-05
(docs) Adding Sentry: Exceptions

smeubank:
[related discussion](https://github.com/encode/starlette/discussions/1787#discussioncomment-3314878)

Making this a separate PR to ensure discussion of how to appropriately add this to the page

related: https://github.com/encode/starlette/pull/2062

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -131,3 +131,10 @@ You can use the `WebSocketException` class to raise errors inside of WebSocket e
 * `WebSocketException(code=1008, reason=None)`
 
 You can set any code valid as defined [in the specification](https://tools.ietf.org/html/rfc6455#section-7.4.1).
+
+## Third party services
+
+#### Sentry for errors, exceptions, and performance issues
+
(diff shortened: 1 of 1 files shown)


Kludex:
This needs to be better structured to get in this section.

smeubank:
> This needs to be better structured to get in this section.

👍  agreed. What type of structure would you suggest? I was thinking to add under "3rd party" similar to the middleware docs page. But that feels a little out of place when it is the only one.

Should I add some example code snippet? I also don't want to be overly verbose or for it to feel like a Sentry add in Starlette docs

Kludex:
I guess something like:
1. Explaining why someone would use Sentry, focused on the error management.
2. A snippet on how to configure it with Starlette, a basic explanation on what happens, and a link to Sentry's documentation for more details.

This was @tomchristie 's idea, it may be better if he shares a bit his thoughts... :eyes:

Kludex:
Closing it as stale. Please, let's discuss how this should look like on the discussion.

PR #2068, opened 2023-03-05
Remove on_startup, on_shutdown, event system and deprecated lifespans

adriangb:
#2067 

Close to 500 LOC removed. That's ~ 1/20th of the codebase.

### DIFF
--- a/docs/applications.md
+++ b/docs/applications.md
@@ -3,6 +3,7 @@ Starlette includes an application class `Starlette` that nicely ties together al
 its other functionality.
 
 ```python
+from contextlib import asyccontextmanager
 from starlette.applications import Starlette
 from starlette.responses import PlainTextResponse
 from starlette.routing import Route, Mount, WebSocketRoute
@@ -25,8 +26,10 @@ async def websocket_endpoint(websocket):
     await websocket.send_text('Hello, websocket!')
     await websocket.close()
 
-def startup():
+@asyccontextmanager
+async def lifespan():
     print('Ready to go')
+    yield None
 
 
 routes = [
@@ -37,7 +40,7 @@ routes = [
     Mount('/static', StaticFiles(directory="static")),
 ]
 
-app = Starlette(debug=True, routes=routes, on_startup=[startup])
+app = Starlette(debug=True, routes=routes, lifespan=lifespan)
 ```
 
 ### Instantiating the application
--- a/docs/database.md
+++ b/docs/database.md
@@ -18,6 +18,7 @@ DATABASE_URL=sqlite:///test.db
 **app.py**
 
 ```python
+import contextlib
 import databases
 import sqlalchemy
 from starlette.applications import Starlette
@@ -43,6 +44,14 @@ notes = sqlalchemy.Table(
 )
 
 database = databases.Database(DATABASE_URL)
+
+@contextlib.asynccontextmanger
+async def lifespan(app):
+    await database.connect()
+    try:
+        yield
+    finally:
+        await database.disconnect()
 
 
 # Main application code.
@@ -77,8 +86,7 @@ routes = [
 
 app = Starlette(
     routes=routes,
-    on_startup=[database.connect],
-    on_shutdown=[database.disconnect]
+    lifespan=lifespan,
 )
 ```
 
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,44 +1,9 @@
 
-Starlette applications can register multiple event handlers for dealing with
+Starlette applications can run a lifespan context manager for dealing with
 code that needs to run before the application starts up, or when the application
 is shutting down.
 
-## Registering events
-
-These event handlers can either be `async` coroutines, or regular synchronous
-functions.
-
-The event handlers should be included on the application like so:
-
-```python
-from starlette.applications import Starlette
-
-
-async def some_startup_task():
-    pass
-
-async def some_shutdown_task():
-    pass
-
-routes = [
-    ...
-]
-
-app = Starlette(
-    routes=routes,
-    on_startup=[some_startup_task],
-    on_shutdown=[some_shutdown_task]
-)
-```
-
-Starlette will not start serving any incoming requests until all of the
-registered startup handlers have completed.
-
-The shutdown handlers will run once all connections have been closed, and
-any in-process background tasks have completed.
-
-A single lifespan `asynccontextmanager` handler can be used instead of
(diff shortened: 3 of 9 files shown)


euri10:
what do you think about adding a little "migration" guide like:

you used to do this:
```
async def startup():
	await open_some_ressource()
async def shutdown():
	await close_some_ressource()
app = Starlette(
	routes=routes,
	on_startup=[startup],
	on_shutdown=[shutdown]
)
```

now do this:
```
@asynccontextmanager
async def lifespan():
	try:
		await open_some_ressource()
		yield
	finally:
		await close_some_ressource()
app = Starlette(
    routes=routes,
	lifespan=lifespan
)
```

or it's too obvious ?

docs/database.md
```diff
@@ -44,6 +45,14 @@ notes = sqlalchemy.Table(
 
 database = databases.Database(DATABASE_URL)
 
+@contextlib.asynccontextmanger
+async def lifespan(app):
+    await database.connect()
+    try:
```
euri10, line 51: shouldn't those 2 lines be interverted ?

adriangb:
I don't think it's too obvious, I agree we should add it. But we should probably do the same for all of the things we are deprecating / removing. Maybe it's time to start a `v1migration.md` doc that I can add to in this PR?

docs/events.md
```diff
@@ -1,44 +1,9 @@
...
-
-Starlette will not start serving any incoming requests until all of the
-registered startup handlers have completed.
-
-The shutdown handlers will run once all connections have been closed, and
-any in-process background tasks have completed.
-
-A single lifespan `asynccontextmanager` handler can be used instead of
-separate startup and shutdown handlers:
+Starlette will not start serving any incoming requests until the lifespan has been run.
```
euri10, line 6: you mean uvicorn right ?

docs/events.md
```diff
@@ -64,38 +29,10 @@ for managing asynchronous tasks.
 
 ## Lifespan State
 
-The event handlers also accept a `state` argument, which is a dictionary
+The lifespan can also accept a `state` argument, which is a dictionary
 that can be used to share the objects between the startup and shutdown handlers,
 and the requests.
```
euri10, line 35: seems to me that this becomes outdated since there wont be startup and shutdown handlers

gnat:
So, what do we do now? This is a breaking change.


> ```
> @asynccontextmanager
> async def lifespan():
> 	try:
> 		await open_some_ressource()
> 		yield
> 	finally:
> 		await close_some_ressource()
> app = Starlette(
>     routes=routes,
> 	lifespan=lifespan
> )
> ```
This is really not obvious.

Is there any way to get the equivalent functionality of on_startup, on_shutdown more intuitively? 

The DX of being able to just specify a function for these important events is a good thing.

gnat:
Side "meta" note, what is with all the breaking PR's this year? Is the plan to force everyone to completely re-learn Starlette and re-write their apps for 1.0... :shrug: What makes this more egregious is the proposed patterns have worse DX

@tomchristie

adriangb:
Your example is very contrived, I seriously doubt anyone is doing that.

gnat:
Not my code, re: https://github.com/encode/starlette/pull/2068#issuecomment-1457667798

adriangb:
Misread. Thought you were yielding a something, which, indeed, would be breaking, but like I said very contrived.

We have documentation for this. And there are good reasons why this approach is much better than two separate functions. I’m afraid I do not agree that it is very non-obvious how to use.

gnat:
>We have documentation for this.

Is this what you're referring to? 
* https://github.com/encode/starlette/pull/2068/files#diff-2cee6a30a4169080ec68af68d0a8555732955725fe34fbca732dd06e3e55a4a5
* https://fastapi.tiangolo.com/advanced/events/#lifespan

docs/applications.md
```diff
@@ -3,6 +3,7 @@ Starlette includes an application class `Starlette` that nicely ties together al
 its other functionality.
 
 ```python
+from contextlib import asyccontextmanager
```
gnat, line 6: `asyccontextmanager` should be `asynccontextmanager`

docs/applications.md
```diff
@@ -25,8 +26,10 @@ async def websocket_endpoint(websocket):
     await websocket.send_text('Hello, websocket!')
     await websocket.close()
 
-def startup():
+@asyccontextmanager
```
gnat, line 29: `asyccontextmanager` should be `asynccontextmanager`

adriangb:
https://www.starlette.io/lifespan/

Kludex:
Decisions can be reversed until 1.0... But a better place to talk about it would be the discussion that was taken. I'm on my phone right now.

Kludex:
The idea of deprecating things now is to avoid doing so on 1.0.

Feedback is always welcome, and if we made a mistake, we can revert decisions.

gnat:
For people reading this PR, FastAPI went official with this in 0.93.0, only a few days ago: https://github.com/encode/starlette/discussions/2067#discussioncomment-5215837

Totally fair with the discussion that happened in: https://github.com/encode/starlette/discussions/2067 That's fine. The cat is out of the bag at this point.

Gonna go make the changes to my Starlette apps now.

Kludex:
Cool.

But in any case, we are always open to take decisions back if they make sense. We don't want to harm users, and I think we try to always be transparent here.

dineshbvadhia:
@Kludex 
As of today, the FastApi/Starlette lifespan feature results in an "async generators are unimplemented" error when compiling with mypyc. Until mypyc supports async generators, the on_startup/on_shutdown feature should be retained as it passes with mypyc.

Kludex:
> @Kludex As of today, the FastApi/Starlette lifespan feature results in an "async generators are unimplemented" error when compiling with mypyc. Until mypyc supports async generators, the on_startup/on_shutdown feature should be retained as it passes with mypyc.

Ref.: https://github.com/mypyc/mypyc/issues/868

Kludex:
I've applied this on https://github.com/encode/starlette/pull/2384, and given co-author to @adriangb there. FYI

graingert:
> @Kludex As of today, the FastApi/Starlette lifespan feature results in an "async generators are unimplemented" error when compiling with mypyc. Until mypyc supports async generators, the on_startup/on_shutdown feature should be retained as it passes with mypyc.

You don't need to use async generators to create an async context manager you can just manually implement `__aenter__` and `__aexit__` on a class

PR #2070, opened 2023-03-09
Deprecate `on_startup` and `on_shutdown` events

Kludex:
- Follow-up on https://github.com/encode/starlette/pull/2065


starlette/routing.py
```diff
@@ -588,6 +588,14 @@ def __init__(
         self.on_startup = [] if on_startup is None else list(on_startup)
         self.on_shutdown = [] if on_shutdown is None else list(on_shutdown)
 
+        if on_startup or on_shutdown:
+            warnings.warn(
+                "The on_startup and on_shutdown parameters are deprecated, and they "
+                "will be removed on version 1.0. Use the lifespan parameter instead. "
+                "See more about it on https://www.starlette.io/events/.",
```
adriangb: Should we rename this markdown file and hence the URL in this PR to avoid the link being broken in the future?

adriangb (approved):

starlette/routing.py
Kludex: Done :+1:

tiangolo (approved):
Nice! :rocket:

PR #2077, opened 2023-03-11
Fix typing of Lifespan to allow subclasses of Starlette

adriangb:
Fixes https://github.com/tiangolo/fastapi/discussions/9241#discussioncomment-5273143


tests/test_applications.py
```diff
@@ -534,3 +534,14 @@ def get_app() -> ASGIApp:
...
+
+def test_lifespan_typing():
+    class App(Starlette):
+        pass
+
+    @asynccontextmanager
+    async def lifespan(app: App) -> AsyncIterator[None]:
+        yield
+
+    App(lifespan=lifespan)
```
adriangb: Not sure if we actually enforce mypy on our tests, but pylance reports no issues

tiangolo (approved):
Awesome, thank you @adriangb! This looks good.

I just tested this branch locally to confirm it works, and I made the corresponding PR in FastAPI here, I'll merge that and release once this is merged and released: https://github.com/tiangolo/fastapi/pull/9245

starlette/applications.py
```diff
@@ -257,3 +257,6 @@ def decorator(func: typing.Callable) -> typing.Callable:
             return func
 
         return decorator
+
+
+AppType = typing.TypeVar("AppType", bound=Starlette)
```
Kludex: Can we move this to the top, or is there an issue with string typing?

starlette/types.py
```diff
@@ -1,7 +1,6 @@
 import typing
 
-if typing.TYPE_CHECKING:
-    from starlette.applications import Starlette
+T = typing.TypeVar("T")
```
```suggestion
AppType = typing.TypeVar("AppType")
```

tests/test_applications.py
```diff
@@ -534,3 +534,14 @@ def get_app() -> ASGIApp:
...
+
+def test_lifespan_typing():
+    class App(Starlette):
+        pass
+
+    @asynccontextmanager
+    async def lifespan(app: App) -> AsyncIterator[None]:  # pragma: no cover
+        yield
+
+    App(lifespan=lifespan)
```
Kludex, line 550: Does this test fails before this PR?

Kludex (changes requested):

Kludex:
The lifespan parameter on `routing.py` doesn't get a parameter inside the `Lifespan`, is it intended? My IDE is more permissive than yours, so I don't have problems there, but knowing your configuration (:laughing:), your IDE is probably complaining over there.

adriangb:
Good point. We probably have to add an Any

tests/test_applications.py
adriangb: No, not really, I’m not sure how to force it to fail since we don’t run mypy on tests right?

Kludex: Maybe assert_type on the constructor? Idk

adriangb: Ok mypy does pick it up:

Before this change:
```
tests/test_applications.py:550: error: Argument "lifespan" to "App" has incompatible type "Callable[[App], _AsyncGeneratorContextManager[None]]"; expected "Union[Callable[[Starlette], AbstractAsyncContextManager[None]], Callable[[Starlette], AbstractAsyncContextManager[Mapping[str, Any]]], None]"  [arg-type]
```

starlette/applications.py
adriangb: No we can't. It needs to reference `Starlette` so it has to come after it.

Kludex: Just tested here, and it works before...

starlette/applications.py
```diff
@@ -11,6 +11,8 @@
 from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, Lifespan, Receive, Scope, Send
 
+AppType = typing.TypeVar("AppType", bound="Starlette")
+
```
adriangb, line 15: Does this work? If so 👍🏻

starlette/applications.py
```diff
@@ -11,6 +11,8 @@
 from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, Lifespan, Receive, Scope, Send
 
+AppType = typing.TypeVar("AppType", bound="Starlette")
```
Kludex, line 14: Check if you are fine with this, please. @adriangb

Kludex (approved):

starlette/applications.py
Kludex: yes

PR #2086, opened 2023-03-17
Add Starlette bridge to packages

tarsil:
* Update Esmerald docs website

### DIFF
--- a/docs/third-party-packages.md
+++ b/docs/third-party-packages.md
@@ -11,6 +11,16 @@ Here are some of those third party packages:
 
 ## Plugins
 
+## Starlette Bridge
+
+<a href="https://github.com/tarsil/starlette-bridge" target="_blank">GitHub</a> |
+<a href="https://starlette-bridge.tarsild.io/" target="_blank">Documentation</a>
+
+With the deprecation of `on_startup` and `on_shutdown`, Starlette Bridge makes sure you can still
+use the old ways of declaring events with a particularity that internally, in fact, creates the
+`lifespan` for you. This way backwards compatibility is assured for the existing packages out there
+while maintaining the integrity of the newly `lifespan` events of `Starlette`.
+
 ### Authlib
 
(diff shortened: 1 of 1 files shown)


docs/third-party-packages.md
```diff
@@ -11,6 +11,16 @@ Here are some of those third party packages:
...
 
+## Starlette Bridge
+
+<a href="https://github.com/tarsil/starlette-bridge" target="_blank">GitHub</a> |
+<a href="https://starlette-bridge.tarsild.io/" target="_blank">Documentation</a>
+
+With the deprecation of `on_startup` and `on_shutdown`, Starlette Bridge makes sure you can still
+use the old ways of declaring events with a particularity that internally, in fact, creates the
+`lifespan` for you. This way backwards compatibility is assured for the existing packages out there
+while maintaining the integrity of the newly `lifespan` events of `Starlette`.
```
Kludex: How do we sort the packages on this list? :rofl:

tarsil: Oh, sorry my bad 🤣. Let me update. Moved to the end of the list

Kludex: It was kind of a question actually, but thanks hahaha

tarsil: Well, I would say that maybe we could sort by alphabetical order. 

This solves the problem I was trying to solve in the closed PR. This way Starlette goes 100% lifespan while the [Starlette Bridge](https://starlette-bridge.tarsild.io/) keeps the old and new ways in sync for those who still want to use the `on_startup` and `on_shutdown` events.
This will make sure that packages can install it, keep the old ways (internally generates the lifespan) and upgrade to Starlette 1.0+ without issues.

Not sure if you want to use it as suggestion in the future

Kludex (approved):

Kludex:
Obg :pray:

tarsil:
De nada :). I hope this can help people out there when the time comes and they will need something like it for backwards compatibility

Issue #1921, opened 2022-10-25
Rewrite databases page


Kludex:
The idea is to rewrite the databases page on our documentation: https://www.starlette.io/database/ .

Instead of recommending `databases`, we should recommend _pure_ `SQLAlchemy`.

cc @zzzeek (just pinging to let you know about our intentions, no need to react)

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/1921">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/1921/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/1921/pledge.svg">
</picture>
</a>

lovelydinosaur:
The `databases` package is also used in [the `config.md` example](https://github.com/encode/starlette/blob/master/docs/config.md).
We should remove it from there too.

I can't see it used anywhere else. https://github.com/encode/starlette/search?q=databases

gnat:
Ah that's unfortunate. Databases was a great concept- SQLAlchemy's documentation is still some of the most verbose and troublesome to navigate in the Python community, and I don't see that being addressed anytime soon- even if the new 1.4 (and 2.0 API) is a good option once you learn it and can treat the SQLAlchemy docs strictly as a reference manual rather than something one reads top to bottom.

gnat:
To your credit, @tomchristie all your documentation can be read from top to bottom, and that's underappreciated and lovely.

onerandomusername:
Does this mean that `databases` is deprecated or will be soon?

lovelydinosaur:
Thank you for your comment, @gnat.
Much appreciated.
Let's treat this as a proposal, rather than as a given.

ni-aas:
I see SQLAlchemy has `AsyncSession` and `create_async_engine`, etc. Is the reason behind `databases` that there was no async capabilities for it at the time? Is it possible to use it directly without `databases` now? (Sry for noob question).

gnat:
I imagine if SQLAlchemy had `asyncio` previously, `databases` would likely have never been created, even with the poor developer docs story.

[...]

PR #2092, opened 2023-03-22
Remove databases package from config docs

aminalaee:
Subset of https://github.com/encode/starlette/issues/1921


Kludex (commented):
If I look for `databases` in the docs, I can still find others :thinking:

aminalaee:
@Kludex You mean in the `config.md` file? Because I wanted to do change the other page in a separate PR.

docs/config.md
```diff
@@ -69,9 +69,7 @@ in their representations.
 ```python
 >>> from myproject import settings
 >>> settings.DATABASE_URL
-DatabaseURL('postgresql://admin:**********@192.168.0.8/my-application')
->>> str(settings.DATABASE_URL)
-'postgresql://admin:Fkjh348htGee4t3@192.168.0.8/my-application'
+'postgresql://user:password@localhost:5432/database'
```
Kludex: Is this right? Isn't password be *******?

docs/config.md
```diff
@@ -225,13 +222,14 @@ def setup_test_database():
...
     Create a clean test database every time the tests are run.
     """
-    url = str(settings.DATABASE_URL)
-    engine = create_engine(url)
-    assert not database_exists(url), 'Test database already exists. Aborting tests.'
-    create_database(url)             # Create the test database.
-    metadata.create_all(engine)      # Create the tables.
-    yield                            # Run the tests.
-    drop_database(url)               # Drop the test database.
+    engine = create_engine(settings.DATABASE_URL)
```
Kludex: What was wrong with this?

docs/config.md
aminalaee: Yes that's correct, databases did that but SQLALchemy engine accepts pure string

docs/config.md
aminalaee: we had to do `url = str(settings.DATABASE_URL)` to convert DatabaseURL from databases to a string for SQLAlchemy create_engine.
That's not needed anymore, we can pass around the string DB URL.

Kludex (approved):

docs/config.md
Kludex: But then this section becomes useless, doesn't it? :sweat_smile:

Kludex (changes requested):

docs/config.md
aminalaee: kind of yes, I thought it's nice to have an example for database URL anyway, so I kept it. But it doesn't do the magic anymore :smile:

Kludex: Then... :eyes:

aminalaee: So is it good to go? I think this is good now @Kludex

Kludex: No... It doesn't make sense to have this code if we are not able to hide "password" component, since the text above says you are able to.

aminalaee: Ah that's what you meant, I'll remove that.

docs/config.md
```diff
@@ -155,9 +144,7 @@ DEBUG = config('DEBUG', cast=bool, default=False)
 TESTING = config('TESTING', cast=bool, default=False)
 SECRET_KEY = config('SECRET_KEY', cast=Secret)
 
-DATABASE_URL = config('DATABASE_URL', cast=databases.DatabaseURL)
-if TESTING:
-    DATABASE_URL = DATABASE_URL.replace(database='test_' + DATABASE_URL.database)
+DATABASE_URL = config('DATABASE_URL')
```
Kludex, line 144: TESTING is now useless...

docs/config.md
```diff
@@ -63,17 +63,6 @@ Secret('**********')
 '98n349$%8b8-7yjn0n8y93T$23r'
 ```
 
-Similarly, the `URL` class will hide any password component
```
Kludex, line 62: Should we instead teach how to create a `URL` class?

aminalaee: We don't have the URL class anymore. I think in the sqlalchemy docs we can have better examples, I tried to keep this minimum.

Kludex: On our last conversation, I think we agreed to show the user how to do this, right?

Kludex (changes requested):

docs/config.md
aminalaee: Yeah, based on our conversation I checked the databases code: https://github.com/encode/databases/blob/ab5eb718a78a27afe18775754e9c0fa2ad9cd211/databases/core.py#L420
The class has about 120 loc which I think is an overkill to include in the Starlette docs. I don't think Starlette's docs is the right place for this.

Kludex: what we talked is that we could just include the link of the code 👀

aminalaee: Yeah I know, we wanted to see if it's small include it here, if it's big then link to it.
But giving it more thought I think it doesn't make sense to do in Starlette docs, another point is that the databases is hardly maintained that's why we are removing it. What do you think?

docs/config.md
Kludex: it was still more cute before

docs/config.md
```diff
@@ -225,13 +198,14 @@ def setup_test_database():
...
-    yield                            # Run the tests.
-    drop_database(url)               # Drop the test database.
+    engine = create_engine(settings.DATABASE_URL)
+    assert not database_exists(settings.DATABASE_URL), 'Test database already exists. Aborting tests.'
+    create_database(settings.DATABASE_URL)  # Create the test database.
+    metadata.create_all(engine)  # Create the tables.
+
+    yield  # Run the tests.
+
+    drop_database(settings.DATABASE_URL)  # Drop the test database.
```
Kludex:
```suggestion
    url = settings.DATABASE_URL
    engine = create_engine(url)
    assert not database_exists(url), 'Test database already exists. Aborting tests.'
    create_database(url)             # Create the test database.
    metadata.create_all(engine)      # Create the tables.
    yield                            # Run the tests.
    drop_database(url)               # Drop the test database.
```

Kludex (changes requested):
Added last bit

docs/config.md
```diff
@@ -17,21 +14,20 @@ from starlette.datastructures import CommaSeparatedStrings, Secret
 config = Config(".env")
 
 DEBUG = config('DEBUG', cast=bool, default=False)
-DATABASE_URL = config('DATABASE_URL', cast=databases.DatabaseURL)
+DATABASE_URL = config('DATABASE_URL')
```
Kludex, line 17: I'm a bit skeptical about this, because if the user prints or logs it, it will show the password, and that's really not something we want to encourage...

aminalaee: Well they shouldn't log it or print it, we can't control what user will do. It's only about the database, what if they log or print passwords, sensitive information, secrets, etc?

Kludex: the current documentation presents a way to manage an environment variable in a way that if the developer makes such a mistake, it will not leak the sensitive information... And we are replacing that by a way to actually leak it...

That's my point

aminalaee: Yeah makes sense, but there's no mechanism available in Starlette to do this. Maybe use `Secret` instead?

Kludex: Can we copy `DatabaseURL` to `Starlette`? 🤔

Kludex: Actually, what's the problem of point out to `databases.DatabaseURL` here? We are not using any database feature there 🤔

aminalaee: Ok, I'll the link to the databases code.

Kludex: Sorry, I didn't express myself very well on the last message. I mean, why not using the `databases.DatabaseURL` import here?

aminalaee: Ah, because then we would still be using `databases`? That will be against the title of this PR.
I think the options you mentioned before make sense, we can either link to the `databases` code or just copy `DatabaseURL` to Starlette.

Kludex: Let's just link

aminalaee: Ok, I'll do that, BTW I think I already mentioned we could be using `Secret` for the Database URL too, it will avoid the leak in the logs and also doesn't need any extra code or library. Just a thought.

Kludex: But the only thing you want to omit is the password field, the rest you actually want to see.

aminalaee: yeah that's true.

aminalaee: Done, please check the wording.

docs/config.md
```diff
@@ -91,6 +76,13 @@ CommaSeparatedStrings(['127.0.0.1', 'localhost'])
 '127.0.0.1'
 ```
 
+!!! tip
+
+    You can use `DatabaseURL` from `databases`
+    package [here](https://github.com/encode/databases/blob/ab5eb718a78a27afe18775754e9c0fa2ad9cd211/databases/core.py#L420)
+    to store database URLs and avoid leaking them in the logs.
```
Kludex: This is supposed to be in the section above, isn't it right? 🤔

aminalaee: I was thinking it should go after Secret and CommaSeparatedStrings (Starlette's types) but it can also go immediately after Secret which is more related.

BilalAlpaslan (approved):

Kludex (approved):

PR #2095, opened 2023-03-28
Fix `docs/staticfiles.md` wrong signature

NoPointExc:
The code:
```
class StaticFiles:
    def __init__(
        self,
        *,
        directory: typing.Optional[PathLike] = None,
        packages: typing.Optional[
            typing.List[typing.Union[str, typing.Tuple[str, str]]]
        ] = None,
        html: bool = False,
        check_dir: bool = True,
        follow_symlink: bool = False,
    ) -> None:
```
I found this while reading the doc: https://www.starlette.io/staticfiles/

### DIFF
--- a/docs/staticfiles.md
+++ b/docs/staticfiles.md
@@ -3,7 +3,7 @@ Starlette also includes a `StaticFiles` class for serving files in a given direc
 
 ### StaticFiles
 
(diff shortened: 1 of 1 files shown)


Kludex (approved):

Kludex:
Thanks @NoPointExc :pray:

PR #2100, opened 2023-03-30
Use a proxy object to modify middleware chain in place

adriangb:
Superseeds #2017 and fixes https://github.com/encode/starlette/pull/2017#issuecomment-1488537008. The approach is similar to inserting into a linked-list in place.


Kludex:
Given discussion on https://github.com/encode/starlette/pull/2017, shall we close this PR?

adriangb:
Yeah let's close this for now. At least we know what a solution would look like!

PR #2107, opened 2023-04-02
TestClient mustn't consider response started until first body chunk

kristjanvalur:
ASGI spec requires that a response may not be begun (sent to client) until the first "body" message is received.
This helps avoid situations where an internal error cannot be sent, because a success response is already in transit on the wire.

This PR enables this funcitionality for `TestClient` and adds a test where a 500 response is received if the application
tries to send two `http.response.start` messages.  Previously this would result in a 200 response being delivered.

See https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -268,6 +268,7 @@ class _TestClientTransport(httpx.BaseTransport):
         }
 
         request_complete = False
+        response_starting = False
         response_started = False
         response_complete: anyio.Event
         raw_kwargs: typing.Dict[str, typing.Any] = {"stream": io.BytesIO()}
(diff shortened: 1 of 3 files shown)


Kludex:
I don't think we should follow ASGI here. For two reasons:
1. It's already the way it is.
2. Performance issues, since we are going to hold the headers.

Those thoughts also apply to similar changes on Uvicorn.

Also, as mentioned somewhere - I think it was @madkinsz who said it - the developer can hold the `http.response.start` until the first `http.response.body` is completed, if they want.

kristjanvalur:
This PR is about enhancing the way that TestClient emulates a _real_ webserver when testing the framework.  Therefore, the "performance" argument does not apply.  We don't touch the ASGI middleware at all, a correctly written _server_ will have already ensured this behaviour.

The spec is that way for a reason.  I can only guess at _why_ it is that way, but my guess is, that it is because errors which occur during the creation of the body message should be able to result in a 500 response.  
If you send away the start of the response immediately, you forego the chance to return a proper error response when an error occurs.

We also shouldn't put unnecessary details in the hands of the developer.  In this case, ASGI specifies _specifically_ that it is the server which is responsible for ensuring this behaviour.  

So, I do disagree, but I don't have any skin in this game, except for having bothered to write a PR.  And I _have_ run into cases where testing of web frameworks using starlette Testclient have been made difficult because of this.  A proper 500 error not showing up at the client side.

euri10:
> The spec is that way for a reason. I can only guess at _why_ it is that way, but my guess is, that it is because errors which occur during the creation of the body message should be able to result in a 500 response.
> If you send away the start of the response immediately, you forego the chance to return a proper error response when an error occurs.

agreed to that, kind of remember this is what we do in uvicorn dont we ? and what does hypercorn do, and httpx asgi transport ?

Kludex:
Uvicorn and the `TestClient` have the same behavior, hypercorn has the behavior this PR suggests, and I don't know about httpx ASGI transport.

Kludex:
Now we comply with the spec. 🤷‍♂️ 

- https://github.com/django/asgiref/pull/387

kristjanvalur:
I guess I'll be closing my uvicorn PR too, then :)  https://github.com/encode/uvicorn/pull/1917

PR #2117, opened 2023-04-10
Support 200.html for single page apps and client-side routing

newinnovations:
Code to support discussion #2116


alex-oleshkevich (approved):

Kludex:
@alex-oleshkevich Was your statement on the discussion not valid?

Kludex (changes requested):
We need a test, and documentation here as well.

newinnovations:
Docs and test as requested

Kludex:
> @alex-oleshkevich Was your statement on the discussion not valid?

@alex-oleshkevich ?

alex-oleshkevich:
> @alex-oleshkevich Was your statement on the discussion not valid?

No, it wasn't. However, as a full stack dev, I understand that need. A custom route is definitely a solution here but is also a piece of code we can avoid by looking for 200.html in the static files directory. Slight improvement but a huge DX win.

tests/test_staticfiles.py
```diff
@@ -338,6 +338,50 @@ def test_staticfiles_html_only_files(tmpdir, test_client_factory):
     assert response.text == "<h1>Hello</h1>"
 
 
+def test_staticfiles_html_200_html(tmp_path: pathlib.Path, test_client_factory):
+    path = tmp_path / "200.html"
+    path.write_text("<h1>Not found page without error</h1>")
+
+    path = tmp_path / "404.html"
+    path.write_text("<h1>Custom not found page</h1>")
```
Kludex, line 345: Why this one is created tho?

newinnovations: To test that 200.html takes precedence over 404.html when both are present.

starlette/staticfiles.py
```diff
@@ -146,17 +146,18 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
-                self.lookup_path, "404.html"
-            )
-            if stat_result and stat.S_ISREG(stat_result.st_mode):
-                return FileResponse(
-                    full_path,
-                    stat_result=stat_result,
-                    method=scope["method"],
-                    status_code=404,
+            # Check for '200.html' and '404.html' if we're in HTML mode.
+            for status in [200, 404]:
```
Kludex, line 150: @graingert @aminalaee any ideas to not do 2 lookups for 404s?

newinnovations: The thinking behind checking for 200.html first and 404.html later, was that handling 200.html should be faster than 404.html. Because serving 200.html is a regular (non-error) operation, the second is an exception.

alex-oleshkevich: It depends. IMO, most users do not need 200.html, 404.html will be used. So checking for 404 first is a more appropriate solution for the library.

Kludex: Alternatively, the logic inside the `if self.html:` in a method, and then people can create their own `StaticFiles` class.

e.g.:
```py
class MyStaticFiles(StaticFiles):
   def html_method(self):
       return FileResponse(200)
```

starlette/staticfiles.py
```diff
@@ -146,17 +146,18 @@ async def get_response(self, path: str, scope: Scope) -> Response:
                 return self.file_response(full_path, stat_result, scope)
 
         if self.html:
```
Kludex, line 148: Is it too crazy to change the type of `html` to `bool | Literal[200, 404]` ?

Or to include a `html_status_code` parameter?

Kludex:
I couldn't find any web framework that allows this kind of built-in configuration. Did you find anything around?

I also don't like the idea of giving precedence to `202.html` - it's a breaking change, and it doesn't look like a good design to "guess what the user has".

For the time being, I'll be closing this PR. If you can provide a reference for the previous question, we can reconsider.

If you really need this feature, consider subclassing, and reimplementing this method.

In any case, thanks for the PR. 🙏

PR #2123, opened 2023-04-14
Fix types in lifespan examples

alex-oleshkevich:
This makes lifespan examples to pass mypy checks.


alex-oleshkevich:
Initially I was misled by `async def lifespan` return type.

Kludex (approved):

adriangb (approved):

PR #2124, opened 2023-04-17
Clarify the origin of disallowed response

shachar-karry:
After spending a few hours trying to understand why my API didn't work after it's initial deployment, I found that this was the origin of the response. Had it been something less generic than "Invalid host header" a google search would have done the job wonderfully.

### DIFF
--- a/starlette/middleware/trustedhost.py
+++ b/starlette/middleware/trustedhost.py
@@ -56,5 +56,5 @@ class TrustedHostMiddleware:
                 redirect_url = url.replace(netloc="www." + url.netloc)
                 response = RedirectResponse(url=str(redirect_url))
             else:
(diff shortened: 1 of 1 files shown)


Kludex:
The user shouldn't be aware of the middlewares you are using.

For that reason, I'll be closing this PR.

Thanks anyway! :pray:

shachar-karry:
Thanks. That makes sense
Is there still a way to make this less generic?
The way it is, when you deploy this in a complex environment, this message
leaves you with no real leads, as it can potentially come from a dozen
causes. Maybe more.
This felt like an unnecessary friction.
I also looked at it after the fact and said to myself, if I do something
similar in 6 months, I'll probably fall in the same trap again.

What do you think?
Shachar

PR #2125, opened 2023-04-18
Fix StaticFiles symlinks on windows

aminalaee:


aminalaee:
I was testing to see if we can reproduce this in GH actions, which apparently we didn't:
https://github.com/encode/starlette/actions/runs/4729453551/jobs/8392015285?pr=2125

@Kludex closing it

PR #2127, opened 2023-04-20
Allow "name" argument on templates url_for()

rosstitmarsh:
# Summary

#2050 fixed the url_for functions not allowing "name" as a parameter, but it missed the templating url_for function

# Checklist


Kludex:
Hmm... Not sure if we should do this... The `__name` should have been the first parameter.

rosstitmarsh:
> Hmm... Not sure if we should do this... The `__name` should have been the first parameter.

How do you mean? The `@pass_context` decorator inserts `context` as the first argument.

aminalaee (commented):
LGTM. I tested manually but maybe worth adding a simple test case.

adriangb:
At this point, why don’t we wait a bit and drop support for 3.7 first? Then we can make  it a runtime enforced positional argument.

Kludex:
> At this point, why don’t we wait a bit and drop support for 3.7 first? Then we can make it a runtime enforced positional argument.

But the context parameter shouldn't be... 🤔

adriangb:
Not sure what you mean. But also re-reading I realized we already merged a sibling PR, so no point in holding off.

rosstitmarsh:
> LGTM. I tested manually but maybe worth adding a simple test case.

This was the test I made to make sure my change worked. I didn't add it to the PR because it seemed a bit trivial for something that will be redundant soon. I will add it if you want.

```python
def test_template_name_parameter(tmpdir, test_client_factory):
    path = os.path.join(tmpdir, "index.html")
    with open(path, "w") as file:
        link = "<a href='{{ url_for('homepage', name=name) }}'>{{ name }}</a>"
        file.write(f"<html>Hello, { link }</html>")

    async def homepage(request):
        return templates.TemplateResponse(
            "index.html", {"request": request, "name": request.path_params["name"]}
        )

    app = Starlette(
        debug=True,
        routes=[Route("/{name}", endpoint=homepage)],
    )
    templates = Jinja2Templates(directory=str(tmpdir))

    response = client.get("/world")
    exptected_html = "<html>Hello, <a href='http://testserver/world'>world</a></html>"
    assert response.text == exptected_html
    assert response.template.name == "index.html"
    assert set(response.context.keys()) == {"request", "name"}
```

Kludex:
My point is... There's a parameter `context` before `name`, and the idea on the other PR was to rename `name` by `__name` because the IDEs ignore it, but also as a first step towards positional-only argument, which is not possible on the function on this PR.

Kludex:
Ah, my bad. I was actually saying nonsense.

Kludex (approved):

PR #2128, opened 2023-04-21
Minify JSON websocket data via `send_json`

dosisod:
# Summary

Similar to how JSON data is minified in the `JSONResponse` class, JSON data sent via websocket should also be minified.

# Checklist


Kludex (approved):
Ok 🤷‍♂️ 

Thanks.

PR #2144, opened 2023-05-10
add: secure cookie middleware

thearchitector:
# Summary

Add https://github.com/thearchitector/starlette-securecookies as another third party middleware.

Not sure what the exact process is for adding to / updating the docs, so let me know if something else is required!

# Checklist


adriangb (approved):
Might want to rewrite it to be a pure ASGI middleware: https://www.starlette.io/middleware/#pure-asgi-middleware

thearchitector:
It's already on the todo list, thanks!

PR #2149, opened 2023-05-17
Include all files in sdist archives

mgorny:
# Summary

Remove the restriction on files inclduded in sdist archives in order to include documentation sources and tests there.  This is the default hatchling behavior and it is helpful to packagers (such as Linux distributions or Conda) as it permits using sdist archives to do packaging (instead of GitHub archives that are not guaranteed to be reproducible).  Most of the ordinary users will not be affected since starlette is a pure Python package and therefore pip will prefer wheels to install it everywhere.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,11 +51,6 @@ Source = "https://github.com/encode/starlette"
 [tool.hatch.version]
 path = "starlette/__init__.py"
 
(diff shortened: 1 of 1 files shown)


Kludex (approved):

Kludex:
Thanks @mgorny 🙏

PR #2154, opened 2023-05-24
Add testclient test for domain restricted cookies

### LINKED ISSUE #2152: Testclient doesn't accept cookies with domain "testserver" in Python < 3.11
As the starlette testclient uses `http://testserver` as its default `base_url`, it is obvious to set the cookie domain to `testserver` in a test setup.

This works as expected in Python 3.11, but fails in Python 3.10 and below due to a change in `http.cookiejar.DefaultCookiePolicy.set_ok_domain` (compare [3.10](https://github.com/python/cpython/blob/3.10/Lib/http/cookiejar.py#L1047) with [3.11](https://github.com/python/cpython/blob/3.11/Lib/http/cookiejar.py#L1046)).

For example, the following test case works fine in Python 3.11, but not in 3.10:

```py


    response = Response("foo")
    response.set_cookie("bar", "baz", domain="testserver")


def test_app():
    client = TestClient(app)
    response = client.get("/")
    assert len(response.cookies) == 1
```

**Workaround:** Use `testserver.local` as cookie domain for testclient. This works fine in both Python 3.10 and Python 3.11.

---

**Is this a starlette issue?** I think so, somewhat, as Starlette uses `testserver` as default `base_url` - which doesn't work "as expected" if a cookie domain is set.

**What could be done about it?** Either use a different `base_url` (e.g. `testserver.local`) (this would certainly break backward compatibility in lots of tests), or just document the issue, e.g. in the [Testclient chapter](https://www.starlette.io/testclient/).

nidico:
# Summary

This adds a test for #2152 which creates responses with domain restricted cookies for different domains.

The test with the `testserver` domain parametrization succeeds in Python 3.11 but fails on Python 3.10 and below.

Note that this PR just adds a partially failing test, but doesn't resolve #2152 (e.g. by changing documentation or changing the test client default `base_url`).

# Checklist

### DIFF
--- a/tests/test_testclient.py
+++ b/tests/test_testclient.py
@@ -273,3 +273,38 @@ def test_query_params(test_client_factory, param: str):
     client = test_client_factory(app)
     response = client.get("/", params={"param": param})
     assert response.text == param
+
+
+@pytest.mark.parametrize(
+    "domain, ok",
+    [
+        ("testserver", True),
+        ("testserver.local", True),
+        ("localhost", False),
+        ("example.com", False),
+    ],
+)
+def test_domain_restricted_cookies(test_client_factory, monkeypatch, domain, ok):
+    """
+    Test that test client discards domain restricted cookies which do not match the
+    base_url of the testclient (`http://testserver` by default).
+
+    The domain `testserver.local` works because the Python http.cookiejar module derives
+    the "effective domain" by appending `.local` to non-dotted request domains
+    in accordance with RFC 2965.
+    """
+
+    async def app(scope, receive, send):
+        response = Response("Hello, world!", media_type="text/plain")
+        response.set_cookie(
+            "mycookie",
+            "myvalue",
+            path="/",
+            domain=domain,
+        )
+        await response(scope, receive, send)
+
+    client = test_client_factory(app)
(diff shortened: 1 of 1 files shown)


Kludex (approved):
I'm going to merge this for the time being. I understand that the problem is not solved, but at least documented on the test.

Let's see if we can solve it as a next step.

Thanks @nidico :)

tests/test_testclient.py
```diff
@@ -273,3 +274,48 @@ def homepage(request):
...
+                    "#2152)",
+                ),
+            ],
+        ),
+        ("testserver.local", True),
+        ("localhost", False),
+        ("example.com", False),
+    ],
+)
+def test_domain_restricted_cookies(test_client_factory, monkeypatch, domain, ok):
```
Kludex: Why is the `monkeypatch` here?
```suggestion
def test_domain_restricted_cookies(test_client_factory, domain, ok):
```

Kludex (changes requested):

tests/test_testclient.py
nidico: Seems to be left over somehow.

Kludex (approved):

PR #2157, opened 2023-05-29
Add missing type hints to Jinja2Templates()

mattaw:
# Summary
[`FileSystemLoader()`](https://github.com/pallets/jinja/blob/953acd65b2be5428482dcb60f5b8481b66252ac9/src/jinja2/loaders.py#L151) from jinja2 (see its comment block) has supported both a single `str` / `PathLike` or a
sequence of them for at least four years. Starlette currently only has typehints for a single `str` / `PathLike`.

Mypy passes. Documented and test created.

# Checklist

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -24,8 +24,11 @@ routes = [
 app = Starlette(debug=True, routes=routes)
 ```
 
-Note that the incoming `request` instance must be included as part of the
-template context.
+Note that:
+
+* The incoming `request` instance must be included as part of the template context.
+
+* `directory=` can accept a sequence, e.g. `Jinja2Templates(directory=['templates', 'more_templates'])`.
 
 The Jinja2 template context will automatically include a `url_for` function,
 so we can correctly hyperlink to other pages within the application.
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -35,8 +35,6 @@ app = Starlette(debug=True, routes=routes)
 Note that the incoming `request` instance must be included as part of the
 template context.
 
-* `directory=` can accept a sequence, e.g. `Jinja2Templates(directory=['templates', 'more_templates'])`.
-
 The Jinja2 template context will automatically include a `url_for` function,
 so we can correctly hyperlink to other pages within the application.
 
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -1,6 +1,14 @@
 Starlette is not _strictly_ coupled to any particular templating engine, but
 Jinja2 provides an excellent choice.
 
+### Jinja2Templates
+
+Signature: `Jinja2Templates(directory, context_processors=None, **env_options)`
+
+* `directory` - A string, [os.Pathlike][pathlike] or a list of strings or [os.Pathlike][pathlike] denoting a directory path.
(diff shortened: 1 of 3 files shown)


Kludex:
Can you add a test that proves it, please?

mattaw:
Currently the patch changes no logic at all and appears thoroughly tested by existing starlette tests. Additionally, mypy verifies the types hints thoroughly from starlette's API through to Jinja2's. Jinja2 has type hints and tests for sequences already which mypy propagates through starlette.

On that basis, what extra tests are needed?

Thanks!

Kludex:
We are interested in making sure that `directory` on `Jinja2Templates` does accept what the type hints are suggesting. The underlying implication doesn't matter here. If Jinja2 makes a release that breaks when you use a sequence (even if the types are correct), or if we change the package we use on `Jinja2Templates` (well... unlikely... but it doesn't matter).

mattaw:
Thanks for the info, please find both a test and a very minor add to the docs. I attempted to copy the existing styles, variable names as closely as I could.

Happy to take any and all feedback.

If you are happy with it, and want me to squash it before committing please let me know.

Kludex:
No squashs are needed from your behalf.

Thanks. 🙏

Kludex:
It's a bit troublesome when you create PRs from the fork's main branch.

I can't push code to this branch. Can you apply these changes, please: https://github.com/encode/starlette/pull/2158/commits/a46a825270ec6c8d1996031f2e505899d53bdada ?

mattaw:
I'm sorry, when I read the development docs I don't remember seeing any guidance.
As this is kind of a 'bug fix' I went ahead and created the PR on master, but I can happily rebase it onto something else? 
(Also I have found, for a first contribution, it's easier to have a discussion around an actual PR as details matter and maintainers time is short. As someone involved in managing a larger project do you think that's a good thought or what a better approach would be?)

Kludex:
I just need the changes on that commit to be here.

> (Also I have found, for a first contribution, it's easier to have a discussion around an actual PR as details matter and maintainers time is short. As someone involved in managing a larger project do you think that's a good thought or what a better approach would be?)

Hmm... It's fine the way you did, it's really just the branch, because it's protected, so I can't push there.

mattaw:
Nice refactor, it is much clearer than my patch. As Jinja can read files from subdir's of template directories I was being overly cautious in asserting that one directory was not the child or the same as the other.

I shortened the html templates so black does not split them into 3 lines due to their line length, and the template text does not have to say Hello :).

I also found that `template_a.html` and `template_b.html` were being created in the same directory due to a typo and fixed it.

docs/templates.md
```diff
@@ -27,6 +35,8 @@ app = Starlette(debug=True, routes=routes)
 Note that the incoming `request` instance must be included as part of the
 template context.
 
+* `directory=` can accept a sequence, e.g. `Jinja2Templates(directory=['templates', 'more_templates'])`.
+
```
Kludex:
```suggestion
```

Kludex (approved):

Kludex:
Thanks @mattaw 🙏

Issue #2134, opened 2023-04-26
When using a custom Jinja loader, directory parameter is unnecessary but still required.


NickColley:
In our project we're using Jinja templates like so:

```python
from jinja2 import ChoiceLoader, PackageLoader, PrefixLoader, FileSystemLoader
from fastapi.templating import Jinja2Templates

template_loader = ChoiceLoader(
    [
        FileSystemLoader("name/web/templates"),
        PrefixLoader({"govuk_frontend_jinja": PackageLoader("govuk_frontend_jinja")}),
    ]
)

# Use a custom template loader and avoid the default directory option.
templates = Jinja2Templates(directory="null", loader=template_loader)
```

At the moment we use a generic "null" which is not used since in the code if you provide your own loader the default file system loader is bypassed.
https://github.com/encode/starlette/blob/882ab6998e9b9ea73b9237552ef295faae3d2c62/starlette/templating.py#L85-L86

With this in mind, if a loader is passed should the directory parameter be entirely optional?

Should the API allow this?

```python
templates = Jinja2Templates(loader=template_loader)
```

euri10:
see this https://github.com/encode/starlette/issues/1214  for the background as to why the API is what is

I agree with @alex-oleshkevich that currently the special case for loader is unsatisfactory if you want fine-grained control.

NickColley:
I think this could also be confusing if people expect the directory to be merged with the loader or something like this. It's not clear what's happening at the moment. But seems like that thread has covered that confusion somewhat.

Kludex:
@alex-oleshkevich wdyt?

alex-oleshkevich:
My opinion: we need only two mutually exclusive parameters for this purpose: `env: jinja2.Environment` and `directory: str|list[str]`. The latter one is a shortcut. This covers all cases I ever faced.

I disagree with having `loader` and `kwargs` as, IMO, it is not a nice API because:
* not obvious
* mixes env-specific and class-specific parameters
* it is not clear what are kwargs about. Do they include env and class arguments or only env?

cc @Kludex

NickColley:
I agree that the environment should be exposed since you can then do whatever you'd like and starlette doesnt have to concern itself with duplicating interfaces.

I've also ran into issues where a third party Jinja project assumed I had access to the environment, which we dont with starlette.

alex-oleshkevich:
I made a PR to demonstrate the changes we discussed - https://github.com/encode/starlette/pull/2159

[...]

PR #2159, opened 2023-05-31
Add env argument to Jinja2Templates, remove **env_options. (ref #2134)

alex-oleshkevich:
This PR adds `env` argument to `Jinja2Templates` class and removes `**env_options`. 

**Motivation**
See this discussion - #2134

1. arguments of `Jinja2Templates` configure only this class and nothing else
2. users can configure `jinja2.Environment` and use it in the app without limits
3. the `Environment` class may be used by other app components, not just by Starlette

This is a  breaking change!

**Example**
```python
import jinja2

env = jinja2.Environment(...)
templates = Jinja2Templates(env=env)
```

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -61,7 +61,7 @@ templates.env.filters['marked'] = marked_filter
 
 ## Using custom jinja2.Environment instance
 
-Starlette also accepts a preconfigured `jinja2.Environment` instance. 
+Starlette also accepts a preconfigured [`jinja2.Environment`](https://jinja.palletsprojects.com/en/3.0.x/api/#api) instance. 
 
 
 ```python
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -58,6 +58,21 @@ templates = Jinja2Templates(directory='templates')
 templates.env.filters['marked'] = marked_filter
 ```
 
+
+## Using custom jinja2.Environment instance
+
+Starlette also accepts a preconfigured `jinja2.Environment` instance. 
+
+
+```python
+import jinja2
+from starlette.templating import Jinja2Templates
+
+env = jinja2.Environment(...)
+templates = Jinja2Templates(env=env)
+```
+
+
 ## Context processors
 
 A context processor is a function that returns a dictionary to be merged into a template context.
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -50,6 +50,21 @@ templates = Jinja2Templates(directory='templates')
 templates.env.filters['marked'] = marked_filter
 ```
 
+
+## Using custom jinja2.Environment instance
+
+Starlette also accepts a preconfigured `jinja2.Environment` instance. 
+
+
+```python
+import jinja2
+from starlette.templating import Jinja2Templates
+
+env = jinja2.Environment(...)
+templates = Jinja2Templates(env=env)
+```
+
+
 ## Context processors
 
 A context processor is a function that returns a dictionary to be merged into a template context.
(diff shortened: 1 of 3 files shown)


NickColley:
I like this, it meets the "starlette has a simple opinionated way" need whilst also allowing the user to do something really complicated without the starlette project having to think about a new API at all - since it falls to a shared idea of the Jinja environment which is referenced and used in many different jinja projects.

Kludex:
Can we deprecate the `**env_options` first, then remove it on v1?

starlette/templating.py
```diff
@@ -64,18 +65,27 @@ class Jinja2Templates:
 
     def __init__(
```
Kludex, line 93: Can we overload this?

alex-oleshkevich: good idea

starlette/templating.py
```diff
@@ -64,18 +65,27 @@ class Jinja2Templates:
...
-        self.env = self._create_env(directory, **env_options)
+        assert directory or env, "either 'directory' or 'env' arguments must be passed"
         self.context_processors = context_processors or []
+        self.env = env or self._create_env(str(directory), **env_options)
 
     def _create_env(
-        self, directory: typing.Union[str, PathLike], **env_options: typing.Any
+        self,
+        directory: typing.Union[str, PathLike],
+        **env_options: typing.Any,
```
```suggestion
        self, directory: typing.Union[str, PathLike], **env_options: typing.Any
```

tests/test_templates.py
```diff
@@ -88,3 +89,36 @@ async def dispatch(self, request, call_next):
...
+
+
+def test_templates_require_directory_or_environment():
+    with pytest.raises(
+        AssertionError, match="either 'directory' or 'env' arguments must be passed"
+    ):
+        Jinja2Templates()
+
+
+def test_templates_with_directory(tmpdir):
```
Kludex, line 137: Use tmp_path please.

Kludex (changes requested):
With the overload and the rebase, we are done here 👍

tests/test_templates.py
Kludex: ah, this whole file uses tmpdir, nvm then

Kludex:
I've rebased it.

alex-oleshkevich:
I had to make all other arguments as kw-only to make overloading work.

docs/templates.md
```diff
@@ -58,6 +58,21 @@ templates = Jinja2Templates(directory='templates')
 templates.env.filters['marked'] = marked_filter
 ```
 
+
+## Using custom jinja2.Environment instance
+
+Starlette also accepts a preconfigured `jinja2.Environment` instance.
```
Kludex: Should we link the `jinja2.Environment` documentation, or no need?

Kludex (approved):
Thanks 👍 

Let's accept the `context_processors` being a kw-only as a breaking change.

docs/templates.md
```diff
@@ -61,7 +61,7 @@ templates.env.filters['marked'] = marked_filter
 
 ## Using custom jinja2.Environment instance
 
-Starlette also accepts a preconfigured `jinja2.Environment` instance. 
+Starlette also accepts a preconfigured [`jinja2.Environment`](https://jinja.palletsprojects.com/en/3.0.x/api/#api) instance.
```
Kludex, line 64: I think this doesn't render very well on mkdocs

alex-oleshkevich: it looks good
image

Kludex: Ah ok. Great 👍

Kludex:
Thanks 👍

PR #2172, opened 2023-06-04
[DRAFT] Preserved previously (causing) exception

Quix0r:
# Summary

Preserved previously (causing) exception so the back track won't get hidden

See https://github.com/encode/starlette/discussions/2170 for details

# Checklist

- [x] I've added a note on the `docs/release-notes.md` about my changes (changes on documentation or type annotations don't need this).

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,3 +1,6 @@
+### Added
+* Preserve causing exceptions when throwing another one (example: `raise FooException("Foo happened!") from exc`)
+
 ## 0.27.0
 
 May 16, 2023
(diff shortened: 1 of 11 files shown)


Quix0r:
I don't know how to write tests. And which documentation do I need to update?

Kludex:
I thought you were saying specifically about the case you mentioned on #2170.

If you have a case on which the information present from the previous exception would be valuable, you can present the case, and we can discuss it.

Given what I said, I'll be closing the PR. Sorry for the confusion, and thanks for the PR. 🙏

Quix0r:
I found more instances of hidden exceptions and also fixed them. The mentioned code was only an example for more occurrences. Making a ticket for each of them was feeling to me bit of an overdose.

Issue #2155, opened 2023-05-26
Limit max request size


Kludex:
### Discussed in https://github.com/encode/starlette/discussions/1516

<div type='discussions-op-text'>

<sup>Originally posted by **aviramha** April  5, 2020</sup>
As discussed in the Gitter, my opinion is that starlette should provide a default limit for request size.
The main reason is that without it, any Starlette application is vulnerable to very easy DoS.
For example, newbie me can write a program as follows:
```
from starlette.requests import Request
from starlette.responses import Response


async def app(scope, receive, send):
    assert scope['type'] == 'http'
    request = Request(scope, receive)
    body = b''
    json = await request.json()
    response = Response(body, media_type='text/plain')
    await response(scope, receive, send)
```
As a malicious user, I could send a 30GB sized JSON and cause the memory to go OOM.
Other frameworks support this also - Django, Quart.
My proposal is to add a default limit which can be overrided in the app configuration.</div>

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/2155">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/2155/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/2155/pledge.svg">
</picture>
</a>

adriangb:
If we're going to make it a configurable middleware it might also make sense to have some sort of timeout for connections and each chunk, maybe infinite by default but definitely tunable.

Another thing to keep in mind is that this is likely something users want to control on a per-endpoint basis. That is, if I have an app that has an upload feature where I'm expecting 1GB files it's likely a single endpoint that expects 1GB files so I'd want to bump up the limits _just for that endpoint_. That makes me think that the best strategy may be a per-endpoint middleware w/ a companion middleware that just tweaks the config by changing it in `scope`. That would allow layering and overriding of these settings. This would be similar to https://github.com/encode/starlette/pull/2026

alex-oleshkevich:
This is a good one! I also agree that we need a global setting and per-route (Route + Mount). 
Use case: global limit - 1mb, photo upload endpoint - 10mb limit.

[...]

PR #2175, opened 2023-06-06
Add request_max_size option

alex-oleshkevich:
This PR adds an option to limit request body size. This is a mitigation of some DoS attacks.

**How it works**
There are two places where you can define the limit: app instance and route. When defined on the app, the limit becomes global and guards all endpoints. The developer still can overwrite this value on per-route bases (for example on file upload routes).

```python
app = Starlette(
    request_max_size=5,  # global
    routes=[Route("/", endpoint=read_view, methods=["POST"], request_max_size=24)], # route level
)

response = client.post("/", data=SOME_LARGE_PAYLOAD) 
assert response.status_code == 413
``` 

Refs #2155


alex-oleshkevich:
This can be a base for the next guard - request timeout.

alex-oleshkevich:
It is still possible to read unlimited data in mounted apps and in the middleware. But, I think, it is a developer's responsibility in this case.

starlette/applications.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         on_startup: typing.Optional[typing.Sequence[typing.Callable]] = None,
         on_shutdown: typing.Optional[typing.Sequence[typing.Callable]] = None,
         lifespan: typing.Optional[Lifespan["AppType"]] = None,
+        request_max_size: int = 2621440,  # 2.5mb
```
alex-oleshkevich, line 65: Like in Django

starlette/routing.py
```diff
@@ -273,7 +275,25 @@ async def handle(self, scope: Scope, receive: Receive, send: Send) -> None:
...
+            async def receive_wrapper() -> Message:
+                nonlocal bytes_read
+                message = await receive()
+                if message["type"] != "http.request" or max_body_size is None:
+                    return message
+
+                body = message.get("body", b"")
+                bytes_read += len(body)
+                if bytes_read > max_body_size:
+                    raise HTTPException(
```
alex-oleshkevich, line 301: Is the exception good enough?

starlette/applications.py
adriangb: Can you permalink where you got this from in a comment below here just for the record? I'd also be okay with a code comment.

starlette/applications.py
```diff
@@ -78,6 +79,7 @@ def __init__(
         )
         self.user_middleware = [] if middleware is None else list(middleware)
         self.middleware_stack: typing.Optional[ASGIApp] = None
+        self.request_max_size = request_max_size
```
adriangb, line 83: It might be a good idea to start making these attributes private.

starlette/applications.py
```diff
@@ -117,6 +119,7 @@ def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath:
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         scope["app"] = self
+        scope["request_max_size"] = self.request_max_size
```
adriangb: We should namespace these things e.g. `scope["starlette._request_max_size"]`

starlette/routing.py
adriangb: Why don't we make this a middleware since we have that capability? It would help keep the routing code more routing focused.

adriangb: Yeah fine by me

adriangb (changes requested):

starlette/applications.py
alex-oleshkevich: https://github.com/django/django/blob/main/django/conf/global_settings.py#L310

starlette/routing.py
alex-oleshkevich: The usual setup is - a global limit and override per route that can accept a large payload. Usually, you don't have lots of them.

Middleware is an option too. However, it requires a Mount route type.

adriangb: Ah right I forgot that we only added middleware to Mount routes. Maybe this is a good argument for adding it to every route.

starlette/applications.py
Kludex: What does flask do?

alex-oleshkevich: @Kludex Flask depends on the configuration option or on 'content-length' header. Otherwise, reads nothing.
https://flask.palletsprojects.com/en/2.3.x/config/#MAX_CONTENT_LENGTH
https://flask.palletsprojects.com/en/2.3.x/patterns/fileuploads/#improving-uploads

starlette/applications.py
alex-oleshkevich: Currently, all members of Starlette are public. I would not like to introduce a new style here.

adriangb: Not true: https://github.com/encode/starlette/blob/2168e47052239da5df35d5353bb986f760c51cef/starlette/datastructures.py#L61

starlette/applications.py
alex-oleshkevich: PHP has 8mb
https://github.com/php/php-src/blob/165a4e53da819d377734eb2286df1b54709d18f2/php.ini-production#L713

starlette/routing.py
alex-oleshkevich: Yes, per-route middleware is good for core reuse but not sure if it is a good DX.

starlette/applications.py
alex-oleshkevich: I am referring to the `Starlette` class itself
https://github.com/encode/starlette/blob/master/starlette/applications.py#L71

adriangb: Fair enough. I still think a lot of those shouldn't have been public in the first place and just because they are doesn't mean we should make this public as well.

adriangb:
A couple things to consider @alex-oleshkevich:
- Can we make the configurator a middleware? That way it can be applied to Mount, etc. It makes sense to keep the parameters to the application object since I imagine a global and a per route override like you laid out here is going to be the most common approach, but it’d be nice to make it a bit more flexible. Also so that it can be used with current versions of FastAPI.
- Can we make the per-route enforcement a middleware that gets applied automatically? My main reasoning for this is to be able to use it in existing FastAPI applications, but I imagine other use cases like enforcing at the router level or on a mounted app may make sense

alex-oleshkevich:
It can be a middleware from https://github.com/encode/starlette/pull/2328 that automatically gets applied globally and per-route. And be reused anywhere what can accept a middleware (like Mount).

adriangb:
That sounds good to me. It’ll be two middleware’s right? One to set the values (or override them), and another to enforce them.

alex-oleshkevich:
Yes, like that. One middleware in the `Starlette.middleware_stack` and another one in the `Route.handle`. However, I am aware that middleware may add a method-call overhead compared to middlewareless implementation.

alex-oleshkevich:
It may be worth to have a shared code to use in app, route and middleware.

adriangb:
Yeah I was thinking we make two middlewares, let's call them `RequestSizeLimitConfigurationMiddleware` and `RequestSizeLimitEnforcementMiddleware`. In `Starlette` we'd do something like:

```python
def __init__(self, ..., request_max_size_bytes: int | None = 1234) -> None:
    if request_max_size_bytes:
        self.app = RequestSizeLimitConfigurationMiddleware(self.app, request_max_size_bytes=request_max_size_bytes)
```

And then in `Route` (or maybe rather in the `request_response` function as pointed out above) we'd do something like:

```python
def request_response(
    func: typing.Callable[[Request], typing.Union[typing.Awaitable[Response], Response]],
    max_request_size: int | None = 1234,
) -> ASGIApp:
    """
    Takes a function or coroutine `func(request) -> response`,
    and returns an ASGI application.
    """

        request = Request(scope, receive, send)

            if is_async_callable(func):
            else:
                response = await run_in_threadpool(func, request)

        await wrap_app_handling_exceptions(app, request)(scope, receive, send)
    if max_request_size is not None:
        app = RequestSizeLimitEnforcementMiddleware(app, max_request_size=max_request_size)
    return app
```

starlette/routing.py
Kludex: Please follow the same logic as above.

```py
        if "app" in scope:
            raise HTTPException(status_code=413, "Request Entity Too Large")
        else:
            response = PlainTextResponse("Request Entity Too Large", status_code=404)
```

alex-oleshkevich:
@Kludex what should we do with this ticket? It is stale and can be closed.

Kludex:
> @Kludex what should we do with this ticket? It is stale and can be closed.

Let's close this, and proceed with the subject on the issue, or on https://github.com/encode/starlette/pull/2328.

Thanks Alex.

PR #2176, opened 2023-06-06
Run background tasks after the middleware stack returns

### LINKED ISSUE #2093: `HTTPMiddleware` breaks `BackgroundTasks` in the same connection
- [x] Initially raised as a discussion in [the FastAPI repo](https://github.com/tiangolo/fastapi/discussions/7168#discussioncomment-5387343).

Consider the following simple example of background tasks through Starlette:
`server.py`
```py
import logging
from time import sleep

from starlette.responses import JSONResponse, Response

logging.basicConfig(
    level=logging.INFO,
    format="[%(asctime)s] [%(levelname)s] %(message)s",
    datefmt="%Y-%m-%d %H:%M:%S",
)


    async def dispatch(
    ) -> Response:


def create_task(request: Request) -> JSONResponse:
    logging.info("Entered POST method")
    tasks = BackgroundTasks()
    tasks.add_task(background_task)
    message = {"status": "Task created"}
    return JSONResponse(message, background=tasks, status_code=201)


def background_task() -> None:
    logging.info("Started background task")
    sleep(10)
    logging.info("Finished background task")


routes = [Route("/task", endpoint=create_task, methods=["POST"])]
middleware = [Middleware(CustomMiddleware)]
```

I start this code with `uvicorn server:app --reload`.

Then I start the task twice in one connection using this code:
`client.py`
```py
from httpx import Client

with Client(base_url="http://localhost:8000") as client:
    client.post("/task")
    client.post("/task")
```

Which results in a sequential execution of background tasks (and read timeout from the client as a consequence):
```
[2023-03-23 19:32:08] [INFO] Entered POST method
INFO:     127.0.0.1:51448 - "POST /task HTTP/1.1" 201 Created
[2023-03-23 19:32:08] [INFO] Started background task
[2023-03-23 19:32:18] [INFO] Finished background task
[2023-03-23 19:32:18] [INFO] Entered POST method
[2023-03-23 19:32:18] [INFO] Started background task
[2023-03-23 19:32:29] [INFO] Finished background task
```

If I just send simple curl requests (`curl -X 'POST' 'http://127.0.0.1:8000/task'`) one after the other (so they are different connections), everything works as expected:
```
[2023-03-23 19:44:35] [INFO] Entered POST method
INFO:     127.0.0.1:55528 - "POST /task HTTP/1.1" 201 Created
[2023-03-23 19:44:35] [INFO] Started background task
[2023-03-23 19:44:37] [INFO] Entered POST method
INFO:     127.0.0.1:55532 - "POST /task HTTP/1.1" 201 Created
[2023-03-23 19:44:37] [INFO] Started background task
[2023-03-23 19:44:45] [INFO] Finished background task
[2023-03-23 19:44:47] [INFO] Finished background task
```

It looks like this regression was introduced in Starlette [0.23.1](https://github.com/encode/starlette/releases/tag/0.23.1), [this PR](https://github.com/encode/starlette/pull/1940). If I pin Starlette to 0.23.0 everything works fine.

adriangb:
Run response background tasks from an outer `BackgroundTaskMiddleware` after the wrapped application returns. Nested applications share one task collection, middleware-attached tasks are deferred too, and reusable responses retain their background tasks.

Standalone responses keep their existing behavior. Document the outer wrapper required for manually wrapped applications, task execution after middleware cleanup, and discarding deferred tasks when the wrapped application fails.

Addresses [#3458](https://github.com/Kludex/starlette/issues/3458) and retains regression coverage for background-task exceptions from [discussion #2640](https://github.com/Kludex/starlette/discussions/2640).

Validation: 1,265 tests passed, with 2 skips and 2 expected failures; 100% line and branch coverage. Ruff, mypy, wheel/sdist checks, the documentation build, and the runnable documentation example passed.

## AI Disclaimer


adriangb:
@tiangolo you mentioned in the FastAPI PR that you’d endorse this change and the consequences for FastAPI. If so, would you like to review here / state that on this PR?

Kludex:
Since #2093 was solved, can we close this?

adriangb:
No I think this is still worth considering for 1.0. It has value outside of solving that one bug. It’s a more sensible execution model, and I believe it will prevent related bugs in the future, eg if someone uses tasks in their own ASGI middleware.

Kludex:
I don't think this should get in before 1.0. We should probably discuss it first.

adriangb:
Why not target 1.0 so that any observable behavior changes are not problematic?

But sure let’s discuss it.

Kludex:
Also, changing implementation for the sake of preventing future unknown bugs (that can never exist), can actually introduce bugs.

adriangb:
It's not just future bugs in Starlette, it's simplifying the execution model to avoid bugs in users' code, making it easier to time requests, adding a feature in the future to make background tasks a protocol so you can seamlessly switch from local execution like Starlette currently does to putting it on a queue, etc.

Kludex:
@adriangb Do you still want this?

adriangb:
I still think this is a good idea. I spoke with @tiangolo and he thought so as well, despite any changes required in FastAPI. However no one has shown a strong opinion in favor or against. What do you think we should do?

UberKitten:
I support this change, I think it better matches up with what users would expect from a feature named "BackgroundTasks". I imagine the current implementation is already creating a lot of subtle bugs that simply don't get noticed most of the time.

Reading the docs for BackgroundTasks and using them a handful of times in production, I honestly thought this behavior was how it already worked. This was mostly an assumption based on the name BackgroundTasks and this statement in the docs: "A background task should be attached to a response, and will run only once the response has been sent."

Recently I ran into some issues related to #2093 and I started digging in to the code. Only then did I discover that this was not the case. An alternate change could be to clarify the execution model in the docs and make it more obvious how they currently work.

Kludex:
> I still think this is a good idea. I spoke with @tiangolo and he thought so as well, despite any changes required in FastAPI. However no one has shown a strong opinion in favor or against. What do you think we should do?

I also think this is a better design, but I'd like to avoid changes that are not a clear win. Even if the current design brought issues in the past, we are not aware of current issues, and trying to predict them doesn't make sense.

I think we should close the PR for the time being, and if some issue related to `BackgroundTasks` appears again, then we can reconsider this.

adriangb:
I think we’ll come to regret it if we go 1.0

Kludex:
We can always release 2.0, if that happens.

wu-clan:
Since this is a good proposal, why not in 1.0? 

1.0 has been implemented a lot, and even if there are problems here, fixes can be released instead of 2.0 at any time.

tiangolo:
Thanks for pinging me! And thanks for the patience with my response.

## About FastAPI

From the point of view of FastAPI, now that in recent versions dependencies are finished before background tasks, this would not affect FastAPI (it would have affected it before, but not anymore 😎).

So, taking this PR or not would be fine for FastAPI (I think).

## Ideas I had

I had some ideas of allowing some execution of code (background tasks-ish) independent from the request/response cycle.

But the ideas I had were even a bit more drastic than this, putting tasks in one single big list/queue for the whole app, and each request would add to that giant list/queue, and it would be executed in its own time, independent of the request/response lifecycle.

## BackgroundTasks in Middleware

About adopting this specific PR or not, I'm personally undecided (just my opinion).

I agree it's a simpler system, and it would allow doing something like having a potential future custom middleware that would put the tasks in some external service or something, so that's a potential plus for the future.

The other thing I see is that this gets closer to my idea of having code execution independent of the requests, but in this case it's independent of the `Request` object, not of the request/response lifecycle, because it still depends on the middlewares, so, a list of background tasks per request, executed after the request... so I see that it's not as far away from what is currently happening. 🤔 

The other interesting thing is that this PR, has a sort of implicit opt-out mechanism, if people want to use the old behavior, they could remove the middleware, and then the regular execution as part of the `Request` object would happen, just like a quick escape hatch in case someone found a problem with this and needed a quick way to solve it.

## To 1.0 or not 1.0, that is the question

About taking this before or after 1.0, if it's agreed by others that this is a better implementation, I personally would prefer to take it before 1.0, I would try to break as many known changes as needed before 1.0, to be able to revert them when necessary before 1.0, instead of releasing 1.0, then 2.0 with this, and then right after 3.0 reverting this because there was an unexpected issue with it. That's the same reason why I'm taking my time to release FastAPI 1.0, I want to make all the needed breaking changes and reverts while in 0.x land.

## Errors

I wonder what would happen if there's an error/exception in a task. I think it would bubble up to the `ServerErrorMiddleware`, or not? Because then it would try to handle it and return a 500, but that was already done, so the `ServerErrorMiddleware` itself would error out. 🤔

adriangb:
As you point out, making this change now allows for future changes. For example, we could add a `run_in_taskgroup=True` parameter or some sort of `runner: ImplementsSomeRunnerProtocol` to run on a job queue. Any similarity to the current implementation is just to maximize backward compatibility and minimize unintended changes for users.

> I wonder what would happen if there's an error/exception in a task. I think it would bubble up to the ServerErrorMiddleware, or not? Because then it would try to handle it and return a 500, but that was already done, so the ServerErrorMiddleware itself would error out. 

That's a good question. I feel like any errors in background tasks should be logged but otherwise ignored. Or we can put this middleware outside of the ServerErrorMiddleware and let them bubble up to the ASGI server. Then there's questions like if one task fails should all of them fail or should the rest execute, etc. But that seems like relatively minor issues and I'm happy to resolve them in whatever direction reviewers prefer. Probably best to keep the behavior the same as the current behavior for now.

adriangb:
@Kludex has the feedback here changed your mind at all?

Kludex:
Any change is susceptible to unpredictable bugs, and we don't have any issues with the current implementation at the moment. I don't think we should make this change.

(Also, the reason for this PR was to fix a bug that was fixed on a different PR.)

alex-oleshkevich:
Unpopular opinion, but what if we drop background tasks from the Starlette itself? Let that feature be in 3rd-party. Maintaining tasks in web framework feels wrong to me and in-house reimplementation is not so hard.

adriangb:
I’d be okay with that

adriangb:
Hi folks, new bugs with BaseHTTPMiddleware keep popping up... this seems to solve #2640

starlette/applications.py
```diff
@@ -96,6 +97,7 @@ def build_middleware_stack(self) -> ASGIApp:
 
         middleware = (
             [Middleware(ServerErrorMiddleware, handler=error_handler, debug=debug)]
+            + [Middleware(BackgroundTaskMiddleware)]
```
Kludex, line 100:
```suggestion
            [
                Middleware(ServerErrorMiddleware, handler=error_handler, debug=debug),
                Middleware(BackgroundTaskMiddleware),
            ]
```

starlette/middleware/background.py
```diff
@@ -0,0 +1,37 @@
...
+        self._app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        tasks: _BackgroundTaskList
+        scope[_SCOPE_KEY] = tasks = []
+        try:
+            await self._app(scope, receive, send)
+        finally:
+            for task in tasks:
+                await task()
```
Kludex, line 37: Background tasks should only run on 2xx.

I guess you'll need a `wrap_send` to get the status code.

starlette/responses.py
```diff
@@ -148,6 +149,12 @@ def delete_cookie(
         )
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if (
+            self.background is not None
+            and background.is_background_task_middleware_installed(scope)
+        ):
+            background.add_tasks(scope, self.background)
+            self.background = None
```
Kludex, line 157: What's the scenario you are trying to prevent when setting background to `None`? What if a user has a global response object?

```py
response = Response(background=BackgroundTasks(...))
```

Kludex: Ah, I see now. You are doing because you didn't remove the logic below where we run the background task.

I think we can remove that logic, and you can avoid this line.

(Is my understanding correct?)

starlette/middleware/background.py
```diff
@@ -0,0 +1,37 @@
...
+
+def is_background_task_middleware_installed(scope: Scope) -> bool:
+    return _SCOPE_KEY in scope
+
+
+def add_tasks(scope: Scope, task: BackgroundTask, /) -> None:
+    if _SCOPE_KEY not in scope:  # pragma: no cover
+        raise RuntimeError(
+            "`add_tasks` can only be used if `BackgroundTaskMIddleware is installed"
+        )
```
Kludex, line 22: Can you trigger this, tho? If we always call `is_background_task_middleware_installed` before the `add_tasks`, then this will never be called. I guess the point you are trying to make here is to make this as public API, but enforce it for others.

We have some scenarios:
1. You are running `Starlette()`, which you always have the `BackgroundTasksMiddleware` - in this case we wouldn't need to check the scopes, because we know the `scope[_SCOPE_KEY]` is there.
2. You are running only parts of Starlette e.g. `Response` - in this case, you don't have the `BackgroundTasksMiddleware`, which means this would be a breaking change - oh... Now I see what you did. You kept the previous behavior where it runs the background task if standalone `Response` with background tasks is used.

---

Ok... As you see, I was figuring stuff as I was writing this to myself.

Kludex: I don't like that now the logic is different on those scenarios. I'm having a hard time figuring out how to avoid it tho...

We are in a similar situation with the exception of middleware.

Kludex: Actually... If the `_SCOPE_KEY` is not present, we can error as above, which will make it clear to users that they need to wrap their responses with `BackgroundTasksMiddleware`.

What I mean is to not check if the `BackgroundTasksMiddleware` is installed, and error if user tries to use it outside the Starlette application.

---

Maybe that's your plan but with a deprecation warning?

Kludex:
I'm not convinced this is the way to go. It seems the only issues with the background tasks are related to `BaseHTTPMiddleware` - which we don't like anyway.

adriangb:
> I'm not convinced this is the way to go. It seems the only issues with the background tasks are related to `BaseHTTPMiddleware` - which we don't like anyway.

I feel like there was something else broken that this fixed but given that it's been years now it's hard to remember.

I'll still make the point that aside from fixing the bugs with BaseHTTPMiddleware this also:
1. Seems like a better pattern than running background tasks inside of the `Response` which seems super wonky.
2. Opens the door to use task groups in the future which not only would give us concurrency for background tasks (less latency to run them) but also moves the execution outside of the reqeust/response ASGI cycle entirely (they run coupled to the ASGI lifespan).

UberKitten:
I agree that moving the task out of `Response` and out of the ASGI lifespan would be ideal. This is what I expected based on the name. `PostRequestTasks` could be a more descriptive name for the current behavior.

PR #2178, opened 2023-06-08
Drop support for Python 3.7

Kludex:
Python 3.7 EOL is coming: https://endoflife.date/python.


adriangb (approved):

starlette/applications.py
```diff
@@ -111,8 +111,7 @@ def build_middleware_stack(self) -> ASGIApp:
     def routes(self) -> typing.List[BaseRoute]:
         return self.router.routes
 
-    # TODO: Make `__name` a positional-only argument when we drop Python 3.7 support.
-    def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath:
+    def url_path_for(self, /, __name: str, **path_params: typing.Any) -> URLPath:
```
adriangb, line 114:
```suggestion
    def url_path_for(self, /, name: str, **path_params: typing.Any) -> URLPath:
```

starlette/requests.py
```diff
@@ -173,7 +173,7 @@ def state(self) -> State:
             self._state = State(self.scope["state"])
         return self._state
 
-    def url_for(self, __name: str, **path_params: typing.Any) -> URL:
+    def url_for(self, /, __name: str, **path_params: typing.Any) -> URL:
```
adriangb, line 176:
```suggestion
    def url_for(self, /, name: str, **path_params: typing.Any) -> URL:
```

alex-oleshkevich (approved):

starlette/routing.py
```diff
@@ -258,7 +258,7 @@ def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
                     return Match.FULL, child_scope
         return Match.NONE, {}
 
-    def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath:
+    def url_path_for(self, /, __name: str, **path_params: typing.Any) -> URLPath:
```
adriangb:
```suggestion
    def url_path_for(self, name: str, /, **path_params: typing.Any) -> URLPath:
```

adriangb: Same for other occurrences

rkazinets (approved):

Kludex:
I'll make the next release with 3.7, then I'll merge this.

starlette/applications.py
```diff
@@ -111,9 +111,8 @@ def build_middleware_stack(self) -> ASGIApp:
     def routes(self) -> typing.List[BaseRoute]:
         return self.router.routes
 
-    # TODO: Make `__name` a positional-only argument when we drop Python 3.7 support.
-    def url_path_for(self, __name: str, **path_params: typing.Any) -> URLPath:
-        return self.router.url_path_for(__name, **path_params)
+    def url_path_for(self, /, name: str, **path_params: typing.Any) -> URLPath:
```
abersheeran: Is there a mistake here?

Kludex: It isn't supposed to be... Why do you think it is?

abersheeran: I recall from our earlier discussion that `name` should be a positional-only parameter.

Maybe it should be changed to `(self, name: str, /, **path_params: typing.Any)`

Kludex: Ah, it's a mistake! 🤣 

Thanks! 🙏

abersheeran (approved):

zanieb (approved):

adriangb:
LGTM

PR #2180, opened 2023-06-10
Use mypy `strict`

### LINKED ISSUE #2153: Use mypy strict setup
Since `httpx` adopted `strict` setup from mypy, Starlette can follow the same path.

Viicos:
Fixes #2153.

`mypy` config updated:
- `no_implicit_optional` is now the default (and seems to have been replaced by `implicit_optional`)
- `disallow_untyped_defs` is included by `strict`

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ combine-as-imports = true
 [tool.mypy]
 strict = true
 ignore_missing_imports = true
+python_version = "3.8"
 
 [[tool.mypy.overrides]]
 module = "starlette.testclient.*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,13 +58,12 @@ select = ["E", "F", "I"]
 combine-as-imports = true
 
 [tool.mypy]
-disallow_untyped_defs = true
+strict = true
 ignore_missing_imports = true
-show_error_codes = true
 
 [[tool.mypy.overrides]]
 module = "starlette.testclient.*"
-no_implicit_optional = false
+implicit_optional = true
 
 [[tool.mypy.overrides]]
 module = "tests.*"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,13 +66,13 @@ python_version = "3.8"
 module = "starlette.testclient.*"
 implicit_optional = true
 
-# Overrides only apply to modules and doesn't work for tests
-# (see https://github.com/python/mypy/issues/10045). In the
-# meanwhile, a separate mypy run is made (see 'check' script).
-[[tool.mypy.overrides]]
-module = "tests.*"
-disallow_untyped_defs = false
-check_untyped_defs = true
+# TODO: Uncomment the following configuration when
+# https://github.com/python/mypy/issues/10045 is solved. In the meantime,
+# we are calling `mypy tests` directly. Check `scripts/check` for more info.
+# [[tool.mypy.overrides]]
+# module = "tests.*"
+# disallow_untyped_defs = false
+# check_untyped_defs = true
 
 [tool.pytest.ini_options]
 addopts = "-rxXs --strict-config --strict-markers"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,9 @@ python_version = "3.8"
 module = "starlette.testclient.*"
 implicit_optional = true
 
+# Overrides only apply to modules and doesn't work for tests
+# (see https://github.com/python/mypy/issues/10045). In the
+# meanwhile, a separate mypy run is made (see 'check' script).
 [[tool.mypy.overrides]]
 module = "tests.*"
 disallow_untyped_defs = false
--- a/scripts/check
+++ b/scripts/check
@@ -4,14 +4,14 @@ export PREFIX=""
 if [ -d 'venv' ] ; then
     export PREFIX="venv/bin/"
 fi
-export ALL_FILES="starlette tests"
-export SOURCE_FILES="starlette"
-export TESTS_FILES="tests"
+export SOURCE_FILES="starlette tests"
 
 set -x
 
 ./scripts/sync-version
-${PREFIX}black --check --diff $ALL_FILES
-${PREFIX}mypy $SOURCE_FILES
-${PREFIX}mypy $TESTS_FILES --disable-error-code no-untyped-def --disable-error-code no-untyped-call
-${PREFIX}ruff check $ALL_FILES
+${PREFIX}black --check --diff $SOURCE_FILES
+# TODO: Use `[[tool.mypy.overrides]]` on the `pyproject.toml` when the mypy issue is solved:
+# github.com/python/mypy/issues/10045. Check github.com/encode/starlette/pull/2180 for more info.
+${PREFIX}mypy starlette
+${PREFIX}mypy tests --disable-error-code no-untyped-def --disable-error-code no-untyped-call
+${PREFIX}ruff check $SOURCE_FILES
--- a/scripts/check
+++ b/scripts/check
@@ -4,11 +4,14 @@ export PREFIX=""
 if [ -d 'venv' ] ; then
     export PREFIX="venv/bin/"
 fi
-export SOURCE_FILES="starlette tests"
+export ALL_FILES="starlette tests"
+export SOURCE_FILES="starlette"
+export TESTS_FILES="tests"
 
 set -x
 
 ./scripts/sync-version
-${PREFIX}black --check --diff $SOURCE_FILES
+${PREFIX}black --check --diff $ALL_FILES
 ${PREFIX}mypy $SOURCE_FILES
-${PREFIX}ruff check $SOURCE_FILES
+${PREFIX}mypy $TESTS_FILES --disable-error-code no-untyped-def --disable-error-code no-untyped-call
+${PREFIX}ruff check $ALL_FILES
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -1,6 +1,6 @@
-import sys
 import functools
 import inspect
+import sys
 import typing
 from urllib.parse import urlencode
 
@@ -12,7 +12,7 @@ else:  # pragma: no cover
 from starlette._utils import is_async_callable
 from starlette.exceptions import HTTPException
 from starlette.requests import HTTPConnection, Request
-from starlette.responses import RedirectResponse, Response
+from starlette.responses import RedirectResponse
 from starlette.websockets import WebSocket
 
 _P = ParamSpec("_P")
--- a/starlette/authentication.py
+++ b/starlette/authentication.py
@@ -1,7 +1,13 @@
+import sys
 import functools
 import inspect
 import typing
 from urllib.parse import urlencode
+
+if sys.version_info >= (3, 10):  # pragma: no cover
+    from typing import ParamSpec
+else:  # pragma: no cover
+    from typing_extensions import ParamSpec
 
 from starlette._utils import is_async_callable
 from starlette.exceptions import HTTPException
@@ -9,7 +15,7 @@ from starlette.requests import HTTPConnection, Request
 from starlette.responses import RedirectResponse, Response
 from starlette.websockets import WebSocket
 
-_CallableType = typing.TypeVar("_CallableType", bound=typing.Callable)
+_P = ParamSpec("_P")
 
 
 def has_required_scope(conn: HTTPConnection, scopes: typing.Sequence[str]) -> bool:
@@ -23,10 +29,14 @@ def requires(
     scopes: typing.Union[str, typing.Sequence[str]],
     status_code: int = 403,
     redirect: typing.Optional[str] = None,
-) -> typing.Callable[[_CallableType], _CallableType]:
+) -> typing.Callable[
+    [typing.Callable[_P, typing.Any]], typing.Callable[_P, typing.Any]
+]:
     scopes_list = [scopes] if isinstance(scopes, str) else list(scopes)
 
-    def decorator(func: typing.Callable) -> typing.Callable:
+    def decorator(
+        func: typing.Callable[_P, typing.Any]
+    ) -> typing.Callable[_P, typing.Any]:
         sig = inspect.signature(func)
         for idx, parameter in enumerate(sig.parameters.values()):
             if parameter.name == "request" or parameter.name == "websocket":
@@ -40,9 +50,7 @@ def requires(
         if type_ == "websocket":
             # Handle websocket functions. (Always async)
             @functools.wraps(func)
-            async def websocket_wrapper(
-                *args: typing.Any, **kwargs: typing.Any
-            ) -> None:
+            async def websocket_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> None:
                 websocket = kwargs.get(
                     "websocket", args[idx] if idx < len(args) else None
                 )
@@ -58,9 +66,7 @@ def requires(
         elif is_async_callable(func):
             # Handle async request/response functions.
             @functools.wraps(func)
-            async def async_wrapper(
-                *args: typing.Any, **kwargs: typing.Any
-            ) -> Response:
+            async def async_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> typing.Any:
                 request = kwargs.get("request", args[idx] if idx < len(args) else None)
                 assert isinstance(request, Request)
 
@@ -80,7 +86,7 @@ def requires(
         else:
             # Handle sync request/response functions.
             @functools.wraps(func)
-            def sync_wrapper(*args: typing.Any, **kwargs: typing.Any) -> Response:
+            def sync_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> typing.Any:
                 request = kwargs.get("request", args[idx] if idx < len(args) else None)
                 assert isinstance(request, Request)
 
@@ -97,7 +103,7 @@ def requires(
 
             return sync_wrapper
 
-    return decorator  # type: ignore[return-value]
+    return decorator
 
 
 class AuthenticationError(Exception):
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -15,7 +15,7 @@ T = typing.TypeVar("T")
 P = ParamSpec("P")
 
 
-async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:  # type: ignore[type-arg]
+async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:  # type: ignore[type-arg]  # noqa: E501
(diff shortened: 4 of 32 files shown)


starlette/config.py
```diff
@@ -12,8 +12,8 @@ class EnvironError(Exception):
     pass
 
 
-class Environ(MutableMapping):
-    def __init__(self, environ: typing.MutableMapping = os.environ):
+class Environ(MutableMapping[typing.Any, typing.Any]):
```
Viicos: Should this be `Any` for both key and value types? `Config` expects a `typing.Mapping[str, str]` for its `environ` argument

starlette/middleware/wsgi.py
```diff
@@ -62,7 +62,7 @@ def build_environ(scope: Scope, body: bytes) -> dict:
 
 
 class WSGIMiddleware:
-    def __init__(self, app: typing.Callable) -> None:
+    def __init__(self, app: typing.Callable[..., typing.Any]) -> None:
```
Viicos, line 66: Went for `Callable[..., Any]` as this is module is deprecated. Let me know if you want to have the complete definition.

starlette/concurrency.py
```diff
@@ -13,9 +13,10 @@
 
 T = typing.TypeVar("T")
 P = ParamSpec("P")
+CoroutineAny = typing.Coroutine[typing.Any, typing.Any, typing.Any]
```
adriangb: Can we make this private `_CoroutineAny`?

starlette/config.py
adriangb: I'm not too familiar with this part of Starlette, but env variables (e.g. `os.environ`) are always string keys and string values

starlette/middleware/__init__.py
```diff
@@ -6,7 +6,7 @@ def __init__(self, cls: type, **options: typing.Any) -> None:
         self.cls = cls
         self.options = options
 
-    def __iter__(self) -> typing.Iterator:
+    def __iter__(self) -> typing.Iterator[typing.Any]:
```
adriangb, line 9: Is there really no better way to type this 😢 ?

starlette/middleware/wsgi.py
adriangb: That makes sense, let's not spend time typing this.

adriangb (changes requested):
Looks good! Just a few small comments

starlette/config.py
Viicos: I went for `Any` as dunder methods were typed with `Any`, but I guess it makes more sense to have `str` instead

starlette/middleware/__init__.py
Viicos: I've opened an issue on `mypy` regarding a related issue: https://github.com/python/mypy/issues/15408
The thing is `mypy` will use a "join" operation, finding the common supertype between `type` and `Dict[str, Any]`, which resolves to `object`. Maybe we could use `Iterator[Union[type, Dict[str, Any]]]` while ignoring the error or using `cast`

adriangb: Ok probably not worth worrying about too much. It seems like we’re just abusing `__iter__` for little benefit, but changing it would be a breaking change or at least larger refactor that’s not worth doing in this PR. Any is okay for now.

adriangb (approved):

adriangb:
Please fix errors

starlette/concurrency.py
Viicos: https://github.com/encode/starlette/pull/2180/commits/b1766d404dbeb6bfb6ba7e99c4bed5ade0683688

Viicos:
> Please fix errors

This is still WIP, I'm doing it in small parts to avoid having a single big commit

adriangb:
Ok no worries, I’ll just remove the approval until you’re ready

adriangb (dismissed):
Removing approval until this is marked as ready for review

starlette/datastructures.py
```diff
@@ -397,8 +397,8 @@ class QueryParams(ImmutableMultiDict[str, str]):
     def __init__(
         self,
         *args: typing.Union[
-            "ImmutableMultiDict",
-            typing.Mapping,
+            "ImmutableMultiDict[typing.Any, typing.Any]",
+            typing.Mapping[typing.Any, typing.Any],
```
Viicos, line 400: According to the signature of the `ImmutableMultiDict`'s `__init__` method:

```python
class ImmutableMultiDict(typing.Mapping[_KeyType, _CovariantValueType]):
    _dict: typing.Dict[_KeyType, _CovariantValueType]

    def __init__(
        self,
        *args: typing.Union[
            "ImmutableMultiDict[_KeyType, _CovariantValueType]",
            typing.Mapping[_KeyType, _CovariantValueType],
            typing.Iterable[typing.Tuple[_KeyType, _CovariantValueType]],
        ],
        **kwargs: typing.Any,
    ) -> None:
```

We should parametrize these ones as `[str, str]`, but the following happens a few lines underneath:

```python
self._list = [(str(k), str(v)) for k, v in self._list]
self._dict = {str(k): str(v) for k, v in self._dict.items()}
```

starlette/middleware/errors.py
```diff
@@ -137,7 +137,7 @@ class ServerErrorMiddleware:
     def __init__(
         self,
         app: ASGIApp,
-        handler: typing.Optional[typing.Callable] = None,
+        handler: typing.Optional[typing.Callable[[Request, Exception], Response]] = None,
```
Viicos: This one is a bit hard to handle: `handler` can be either a sync callable or a coroutine (and pretty much erverywhere in `starlette` we have this issue):

```python
if is_async_callable(self.handler):
else:
```

We could use `Callable[[Request, Exception], Response | Awaitable[Response]]`, and cast `handler` in the `if...else` above, but doesn't look pretty. Any ideas? I can't find anything that would simplify typing a async/sync callable.

For reference, [`ExceptionMiddleware`](https://github.com/encode/starlette/blob/15d9350eed9437bcb0683e2c24cd4cfbf985ed7e/starlette/middleware/exceptions.py#L15) has the same behavior and does not make the distinction between sync/async, even if both are used.

starlette/templating.py
```diff
@@ -117,16 +111,14 @@ def __init__(
 
     def _create_env(
         self,
-        directory: typing.Union[
-            str, PathLike, typing.Sequence[typing.Union[str, PathLike]]
-        ],
+        directory: "typing.Union[str, PathLike[typing.AnyStr], typing.Sequence[typing.Union[str, PathLike[typing.AnyStr]]]]",
```
Viicos: Is `PathLike[AnyStr]` fine? Or we could enforce str paths. I often see `PathLike[Any]` but doesn't feel right to me as `PathLike` is generic wrt `TypeVar("AnyStr_co", str, bytes, covariant=True)`

Kludex:
@Viicos Do you need any help here?

Viicos:
Hi @Kludex thanks for asking. The main blocking point is handling callables and coroutines together, as stated in https://github.com/encode/starlette/pull/2180#discussion_r1225486748. The solution I provide should work but isn't pretty, and will add a lot of overhead to the code (considering this pattern is used a lot in the `starlette` codebase). Otherwise we could use `Any` as a return type, [it seems to work](https://mypy-play.net/?mypy=master&python=3.11&gist=ab9bfc03023b16efda621f88cad8cd9d).

Apart from that I'll be able to fix the remaining `mypy` issues

Kludex:
> Hi @Kludex thanks for asking. The main blocking point is handling callables and coroutines together, as stated in [#2180 (comment)](https://github.com/encode/starlette/pull/2180#discussion_r1225486748). The solution I provide should work but isn't pretty, and will add a lot of overhead to the code (considering this pattern is used a lot in the `starlette` codebase). Otherwise we could use `Any` as a return type, [it seems to work](https://mypy-play.net/?mypy=master&python=3.11&gist=ab9bfc03023b16efda621f88cad8cd9d).
> 
> Apart from that I'll be able to fix the remaining `mypy` issues

`Any` is good for now. If you make the pipeline to pass I can review it. 👍

Viicos:
> `Any` is good for now. If you make the pipeline to pass I can review it. 👍

Will be ready for review by the end of the week, as I can see `1.0.0` is coming soon

starlette/concurrency.py
```diff
@@ -13,9 +13,14 @@
...
 P = ParamSpec("P")
+_CoroutineAny = typing.Coroutine[typing.Any, typing.Any, typing.Any]
 
 
-async def run_until_first_complete(*args: typing.Tuple[typing.Callable, dict]) -> None:
+async def run_until_first_complete(
+    *args: typing.Tuple[
+        typing.Callable[..., _CoroutineAny], typing.Dict[str, typing.Any]
+    ]
+) -> None:
```
Kludex: Please ignore the deprecated functions/methods (`type: ignore` is fine). 👍

Kludex:
Looks like the overrides from mypy is not working properly? 🤔

Viicos:
~~Mypy errors left:~~

```sh
starlette/testclient.py:383: error: Name "httpx._client.CookieTypes" is not defined  [name-defined]
...
starlette/testclient.py:699: error: Name "httpx._client.TimeoutTypes" is not defined  [name-defined]
```

~~I have no idea why they happen~~ -> https://github.com/encode/starlette/pull/2180/commits/8811c1cdf8c09b4064d4704038bcf89b00fa438c

> Looks like the overrides from mypy is not working properly? 🤔

https://github.com/encode/starlette/blob/1be57fdb42522552f1e579912a22faa3088e1981/pyproject.toml#L68-L71

This one has no effect, `module` isn't working with a plain directory, instead it needs a proper module to import. For reference: https://github.com/python/mypy/issues/10045
This doesn't seem trivial

starlette/authentication.py
```diff
@@ -23,10 +29,14 @@ def requires(
     scopes: typing.Union[str, typing.Sequence[str]],
     status_code: int = 403,
     redirect: typing.Optional[str] = None,
-) -> typing.Callable[[_CallableType], _CallableType]:
+) -> typing.Callable[
+    [typing.Callable[_P, typing.Any]], typing.Callable[_P, typing.Any]
```
Kludex, line 33: Any reason for not using the following?

```suggestion
    [typing.Callable[_P, _T]], typing.Callable[_P, _T]
```

starlette/types.py
```diff
@@ -1,5 +1,8 @@
 import typing
 
+from .requests import Request
+from .responses import Response
```
```suggestion
from starlette.requests import Request
from starlette.responses import Response
```

starlette/testclient.py
```diff
@@ -85,8 +85,8 @@ def __init__(
         self.scope = scope
         self.accepted_subprotocol = None
         self.portal_factory = portal_factory
-        self._receive_queue: "queue.Queue[typing.Any]" = queue.Queue()
-        self._send_queue: "queue.Queue[typing.Any]" = queue.Queue()
+        self._receive_queue: "queue.Queue[Message]" = queue.Queue()
+        self._send_queue: "queue.Queue[Message | BaseException]" = queue.Queue()
```
Kludex, line 83: Why does it allow `BaseException`?

starlette/staticfiles.py
```diff
@@ -108,7 +108,9 @@ def get_path(self, scope: Scope) -> str:
         Given the ASGI scope, return the `path` string to serve up,
         with OS specific path separators, and any '..', '.' components removed.
         """
-        return os.path.normpath(os.path.join(*scope["path"].split("/")))
+        return typing.cast(
+            str, os.path.normpath(os.path.join(*scope["path"].split("/")))
+        )
```
Kludex: I think this is missing from https://github.com/python/typeshed/blob/main/stdlib/os/__init__.pyi isn't it?

Can we instead add an `type: ignore` here, and open an issue there or something?

starlette/routing.py
```diff
@@ -773,7 +783,7 @@ def host(
     def add_route(
         self,
         path: str,
-        endpoint: typing.Callable,
+        endpoint: typing.Callable[..., typing.Any],
```
Kludex: Isn't an endpoint always `Callable[[Request], typing.Any]`?

starlette/routing.py
```diff
@@ -788,7 +798,10 @@ def add_route(
         self.routes.append(route)
 
     def add_websocket_route(
-        self, path: str, endpoint: typing.Callable, name: typing.Optional[str] = None
+        self,
+        path: str,
+        endpoint: typing.Callable[..., typing.Any],
```
Kludex: Isn't an endpoint always `Callable[[WebSocket], typing.Any]`?

starlette/routing.py
```diff
@@ -301,7 +301,11 @@ def __repr__(self) -> str:
 
 class WebSocketRoute(BaseRoute):
     def __init__(
-        self, path: str, endpoint: typing.Callable, *, name: typing.Optional[str] = None
+        self,
+        path: str,
+        endpoint: typing.Callable[..., typing.Any],
```
Kludex, line 309: Same as comments below.

starlette/routing.py
```diff
@@ -209,7 +209,7 @@ class Route(BaseRoute):
     def __init__(
         self,
         path: str,
-        endpoint: typing.Callable,
+        endpoint: typing.Callable[..., typing.Any],
```
Kludex, line 215: Same as comments below.

starlette/requests.py
```diff
@@ -147,7 +147,7 @@ def session(self) -> typing.Dict[str, typing.Any]:
         assert (
             "session" in self.scope
         ), "SessionMiddleware must be installed to access request.session"
-        return self.scope["session"]
+        return typing.cast(typing.Dict[str, typing.Any], self.scope["session"])
```
Kludex: Can we just add a type ignore here?

starlette/middleware/errors.py
```diff
@@ -137,7 +137,9 @@ class ServerErrorMiddleware:
     def __init__(
         self,
         app: ASGIApp,
-        handler: typing.Optional[typing.Callable] = None,
+        handler: typing.Optional[
+            typing.Callable[[Request, Exception], typing.Any]
+        ] = None,
```
Kludex, line 142: Isn't this supposed to be `ExceptionHandler` as well? It's a question, not a suggestion.

Kludex (changes requested):

starlette/testclient.py
Viicos: See here:

https://github.com/encode/starlette/blob/2168e47052239da5df35d5353bb986f760c51cef/starlette/testclient.py#L118-L128

starlette/staticfiles.py
Viicos: As `scope["path"]` is `Any`, the signature of `os.path.join` and `os.path.normpath` doesn't match anymore, thus the return type was `Any`. Switched to a `type: ignore` instead.

starlette/routing.py
Viicos: I went for `Callable[..., Any]` as this `endpoint` argument is passed to `routing.Route`, which is doing the following:

https://github.com/encode/starlette/blob/2168e47052239da5df35d5353bb986f760c51cef/starlette/routing.py#L209-L234

So it seems `endpoint` can also be a `ASGIApp` (+ there's a bunch of things related to `functools.partial`)

starlette/middleware/errors.py
Viicos: Yes it is, this was discussed earlier: https://github.com/encode/starlette/pull/2180#discussion_r1225486748. We could keep it as `Any`, or use `Union[Reponse, Awaitable[Response]]` and then add some casts or `type: ignore`

Kludex:
Suggestions on what to do here? 😅

It looks like we are blocked by https://github.com/python/mypy/issues/10045?

starlette/routing.py
Viicos: See https://github.com/encode/starlette/pull/2180#discussion_r1248772844 (same for https://github.com/encode/starlette/pull/2180#discussion_r1247700951, https://github.com/encode/starlette/pull/2180#discussion_r1247701050 if I'm not mistaken)

starlette/authentication.py
Viicos: Couldn't make it work as it can be async or sync (same issue as https://github.com/encode/starlette/pull/2180#discussion_r1225486748) :/

Viicos:
> It looks like we are blocked by https://github.com/python/mypy/issues/10045?

I guess so. Maybe we could edit the `check` scripts to run `mypy` a second time on the `tests` directory, but I don't know if it will work as excepted.

Kludex:
I've asked on the `python/typing` Gitter to see if someone has a good alternative. 👀

Viicos:
In the meanwhile, I'll see if using a `Union` is feasible for https://github.com/encode/starlette/pull/2180#discussion_r1248791074

Viicos:
@Kludex let me know if the check script tweak is ok 🙂

pyproject.toml
```diff
@@ -58,13 +58,13 @@ select = ["E", "F", "I"]
...
 ignore_missing_imports = true
-show_error_codes = true
+python_version = "3.8"
 
 [[tool.mypy.overrides]]
 module = "starlette.testclient.*"
-no_implicit_optional = false
+implicit_optional = true
 
 [[tool.mypy.overrides]]
```
Kludex: So... Can we remove this?

scripts/check
Kludex, line 1: Can we do the following instead?

On the `${PREFIX}mypy $SOURCE_FILES`, let's change that by:
```py
# TODO: Use `[[tool.mypy.overrides]]` on the `pyproject.toml` when the mypy issue is solved:
# github.com/python/mypy/issues/10045. Check github.com/encode/starlette/pull/2180 for more info.
${PREFIX}mypy starlette
${PREFIX}mypy tests --disable-error-code no-untyped-def --disable-error-code no-untyped-call
```

Kludex (changes requested):

pyproject.toml
Kludex: Actually, let's leave this here, and add a note on top of it mentioning the mypy issue, please.

Kludex:
Main branch is broken.

starlette/types.py
```diff
@@ -15,3 +19,7 @@
     [AppType], typing.AsyncContextManager[typing.Mapping[str, typing.Any]]
 ]
 Lifespan = typing.Union[StatelessLifespan[AppType], StatefulLifespan[AppType]]
+
+ExceptionHandler = typing.Callable[
+    ["Request", Exception], typing.Union["Response", typing.Awaitable["Response"]]
+]
```
Kludex: It's also possible to have an exception handler for WebSockets.

Kludex (changes requested):

Kludex (approved):
Thanks @Viicos 🙏

PR #2181, opened 2023-06-11
Fix empty str(exception) when initialized with kwargs

BoWuGit:
Add a `super().__init__` call to sub class of Python standard `Exception`, for two reasons:

* It's convention to call it for sub class
* Like the title says, when `HTTPException` is initialized with kwargs, `str(HTTPException)` is empty, which is just the case I met in this [pytest error case](https://github.com/pytest-dev/pytest/pull/11073).

Thanks for your attention.

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -1,3 +1,4 @@
+import http
 import warnings
 
 import pytest
@@ -144,13 +145,17 @@ def test_force_500_response(test_client_factory):
     assert response.text == ""
 
 
-def test_http_repr():
-    assert repr(HTTPException(404)) == (
-        "HTTPException(status_code=404, detail='Not Found')"
+def test_http_str_and_repr():
+    exception = HTTPException(404)
+    should_detail = http.HTTPStatus(exception.status_code).phrase
+    assert str(exception) == should_detail
+    assert repr(exception) == (
+        f"HTTPException(status_code=404, detail='{should_detail}')"
     )
-    assert repr(HTTPException(404, detail="Not Found: foo")) == (
-        "HTTPException(status_code=404, detail='Not Found: foo')"
-    )
+    detail = "Not Found: foo"
+    exception = HTTPException(status_code=404, detail=detail)
+    assert str(exception) == detail
+    assert repr(exception) == (f"HTTPException(status_code=404, detail='{detail}')")
 
     class CustomHTTPException(HTTPException):
         pass
@@ -160,10 +165,11 @@ def test_http_repr():
     )
 
 
-def test_websocket_repr():
-    assert repr(WebSocketException(1008, reason="Policy Violation")) == (
-        "WebSocketException(code=1008, reason='Policy Violation')"
-    )
+def test_websocket_str_and_repr():
+    reason = "Policy Violation"
+    exception = WebSocketException(code=1008, reason=reason)
+    assert str(exception) == reason
+    assert repr(exception) == (f"WebSocketException(code=1008, reason='{reason}')")
 
     class CustomWebSocketException(WebSocketException):
         pass
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -145,17 +145,23 @@ def test_force_500_response(test_client_factory):
     assert response.text == ""
 
 
-def test_http_str_and_repr():
-    exception = HTTPException(404)
+def test_http_str():
+    code = 404
+    exception = HTTPException(code)
     should_detail = http.HTTPStatus(exception.status_code).phrase
-    assert str(exception) == should_detail
-    assert repr(exception) == (
-        f"HTTPException(status_code=404, detail='{should_detail}')"
-    )
+    assert str(exception) == f"{code}: {should_detail}"
     detail = "Not Found: foo"
-    exception = HTTPException(status_code=404, detail=detail)
-    assert str(exception) == detail
-    assert repr(exception) == (f"HTTPException(status_code=404, detail='{detail}')")
+    kwargs_exception = HTTPException(status_code=404, detail=detail)
+    assert str(kwargs_exception) == f"{code}: {detail}"
+
+
+def test_http_repr():
+    assert repr(HTTPException(404)) == (
+        "HTTPException(status_code=404, detail='Not Found')"
+    )
+    assert repr(HTTPException(404, detail="Not Found: foo")) == (
+        "HTTPException(status_code=404, detail='Not Found: foo')"
+    )
 
     class CustomHTTPException(HTTPException):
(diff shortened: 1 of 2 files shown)


zanieb:
This makes sense to me after reading the upstream issue, thanks for contributing!

It might be nice to just add a new test case for this instead of changing the existing `repr` test case.

tests/test_exceptions.py
```diff
@@ -144,13 +145,17 @@ async def app(scope, receive, send):
...
+    exception = HTTPException(404)
+    should_detail = http.HTTPStatus(exception.status_code).phrase
+    assert str(exception) == should_detail
+    assert repr(exception) == (
+        f"HTTPException(status_code=404, detail='{should_detail}')"
     )
+    detail = "Not Found: foo"
+    exception = HTTPException(status_code=404, detail=detail)
+    assert str(exception) == detail
+    assert repr(exception) == (f"HTTPException(status_code=404, detail='{detail}')")
```
zanieb:
```suggestion
    assert repr(exception) == f"HTTPException(status_code=404, detail='{detail}')"
```

tests/test_exceptions.py
```diff
@@ -160,10 +165,11 @@ class CustomHTTPException(HTTPException):
...
 
-def test_websocket_repr():
-    assert repr(WebSocketException(1008, reason="Policy Violation")) == (
-        "WebSocketException(code=1008, reason='Policy Violation')"
-    )
+def test_websocket_str_and_repr():
+    reason = "Policy Violation"
+    exception = WebSocketException(code=1008, reason=reason)
+    assert str(exception) == reason
+    assert repr(exception) == (f"WebSocketException(code=1008, reason='{reason}')")
```
zanieb:
```suggestion
    assert repr(exception) == f"WebSocketException(code=1008, reason='{reason}')"
```

Kludex (changes requested):
Please implement the `__str__` instead of `super().__init__(self.detail)`.

Also, on the `__str__`, I think we can have the same implementation as the `__repr__`.

On DRF, they use only the detail: https://github.com/encode/django-rest-framework/blob/376a5cbbba3f8df9c9db8c03a7c8fa2a6e6c05f4/rest_framework/exceptions.py#L99C1-L117

On flask, they use all the information: https://github.com/pallets/werkzeug/blob/82b5ac433c54b8d03cb0de9fdc16a4c58b935860/src/werkzeug/exceptions.py#L65-L85

zanieb:
@Kludex for what it's worth, I don't think that `__str__` and `__repr__` implementations should match if you want standard behavior

```
❯ python -c "print(str(ValueError('test')))"
test
❯ python -c "print(repr(ValueError('test')))"
ValueError('test')
```

Using the `__repr__` implementation would break expectations in downstream libraries that format exceptions e.g. https://github.com/PrefectHQ/prefect/blob/92e3e6de417bf542c9fca201fd396dba7301665c/src/prefect/states.py#L121 would display the exception type twice

The linked Flask code omits the type from the `__str__` https://github.com/pallets/werkzeug/blob/82b5ac433c54b8d03cb0de9fdc16a4c58b935860/src/werkzeug/exceptions.py#L164-L170

Kludex:
> The linked Flask code omits the type from the __str__ [pallets/werkzeug@82b5ac4/src/werkzeug/exceptions.py#L164-L170](https://github.com/pallets/werkzeug/blob/82b5ac433c54b8d03cb0de9fdc16a4c58b935860/src/werkzeug/exceptions.py#L164-L170)

Yeah, I pasted the wrong part of the implementation. By "they use all the information" I mean that the `__str__` uses all the arguments from the constructor.

> Using the __repr__ implementation would break expectations in downstream libraries that format exceptions e.g. [PrefectHQ/prefect@92e3e6d/src/prefect/states.py#L121](https://github.com/PrefectHQ/prefect/blob/92e3e6de417bf542c9fca201fd396dba7301665c/src/prefect/states.py#L121) would display the exception type twice

Fair enough.

Is there any literature that has a recommendation on how the `__str__` should look like, or should we just copy flask? 🤔

BoWuGit:
> @Kludex  Is there any literature that has a recommendation on how the __str__ should look like, or should we just copy flask? 🤔

For sub class of Python `Exception`, `str(Exception)` defaults to `str(Exception.args)` if you don't call `super().__init__` and don't override `__str__` method, for example:

```Python
>>> str(HTTPException(404, "Not found"))
"(404, 'Not found')"
```


> On DRF, they use only the detail: [encode/django-rest-framework@376a5cb/rest_framework/exceptions.py#L99C1-L117](https://github.com/encode/django-rest-framework/blob/376a5cbbba3f8df9c9db8c03a7c8fa2a6e6c05f4/rest_framework/exceptions.py#L99C1-L117)

Yes, `str(Exception)` only displays the detail, and they also have a [`get_full_details`](https://github.com/encode/django-rest-framework/blob/376a5cbbba3f8df9c9db8c03a7c8fa2a6e6c05f4/rest_framework/exceptions.py#L127) method, so for simplicity consideration, I think we can refer Flask's implemention.

BoWuGit:
I have re-implemented it referring Flask, https://github.com/pallets/werkzeug/blob/main/src/werkzeug/exceptions.py#L164. And I think it is somehow backward compatible in this way, because `str(Exception)` defaults to `str(Exception.args)` if you don't call `super().__init__` and don't override `__str__` method, so both `code` and `detail` should be outputted.

Kludex:
Why do we need `super().__init__()`?

BoWuGit:
I think it's desired practice in Python to always call `super().__init__`, and this [pytest error case](https://github.com/pytest-dev/pytest/pull/11073) I met with is related with it more or less. Giving two links to explain:

[Is it necessary to call super().__init__() explicitly in python?](https://stackoverflow.com/questions/60015319/is-it-necessary-to-call-super-init-explicitly-in-python)

> If you override the __init__ method of the superclass, then the __init__ method of the subclass needs to explicitly call it if that is the intended behavior, yes.

[Is calling the superclass constructor in a subclass really important?](https://softwareengineering.stackexchange.com/questions/318165/is-calling-the-superclass-constructor-in-a-subclass-really-important)

> If you add variables to self in the constructor of Class and don't call Class.__init__() in the constructor of Subclass, then these variables will not be in your Subclass object.

Kludex:
> this https://github.com/pytest-dev/pytest/pull/11073 I met with is related with it more or less

Doesn't the `__str__` already solves that?

Kludex (approved):

Kludex:
Thanks @BoWuGit 👍

BoWuGit:
Happy to be merged, thanks again.

Issue #2118, opened 2023-04-11
Host() routes not included in OpenAPI Schema


nicolaipre:
Hi.

While adding OpenAPI to my Starlette application I noticed that the SchemaGenerator does not seem to pick up subdomain routes registered with Host() as described in docs here: https://www.starlette.io/routing/#host-based-routing

There seems to be lacking functionality to pick up the routes of a Host group here: https://github.com/encode/starlette/blob/master/starlette/schemas.py#L51

Here is a small application demonstrating the issue: 
**requirements.txt**
```
anyio==3.6.2
apispec==5.2.2
click==8.1.3
h11==0.14.0
idna==3.4
packaging==23.0
python-dotenv==1.0.0
PyYAML==5.4.1
sniffio==1.3.0
starlette==0.26.1
starlette-apispec==1.0.5
uvicorn==0.21.1
```

**example.py**
```python
from starlette.applications import Starlette
from starlette.routing import Host, Mount, Route, Router
from starlette.responses import HTMLResponse, JSONResponse
from starlette.schemas import SchemaGenerator
from pprint import pprint
import uvicorn
import yaml
import sys

spec = {
    "openapi": "3.0.0", 
    "info": {
        "title": "Example API", 
        "version": "1.1"
    }
}
schema = SchemaGenerator(spec)


# /openapi.json schema
async def openapi(request):
    return schema.OpenAPIResponse(request=request)

# Show docs /api
# FIXME: Redoc not showing "Try it out" button for some reason
async def docs(request):
    html = f"""
        <!DOCTYPE html>
        <html>
          <head>
            <title>Redoc</title>
            <meta charset="utf-8"/>
            <meta name="viewport" content="width=device-width, initial-scale=1">
            <link href="https://fonts.googleapis.com/css?family=Montserrat:300,400,700|Roboto:300,400,700" rel="stylesheet">
          </head>
          <body>
            <redoc spec-url="{request.url.scheme}://{request.url.netloc}/openapi.json"></redoc>
            <script src="https://cdn.redoc.ly/redoc/latest/bundles/redoc.standalone.js"> </script>
          </body>
        </html>
    """
    return HTMLResponse(html)

# GET /users
async def list_users(request):
    """
    responses:
      200:
        description: A list of users.
        examples:
          [{"username": "tom"}, {"username": "lucy"}]
    """
    raise NotImplementedError()

# GET /users
async def get_user(request):
    """
    tags:
      - User Management API
    parameters:
      - name: username
[...]

PR #2183, opened 2023-06-12
fix: collect routes from Host(#2118)

magiskboy:
# Summary

When developer defines Host in application, routes of Host cannot be generated to OpenAPI schemas so this PR fixes that.

I collect routes (in get_endpoints) in Host class when generating OpenAPI schemas

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ venv*/
 .python-version
 build/
 dist/
-env*/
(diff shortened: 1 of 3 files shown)


Kludex (approved):

Kludex:
Thanks @magiskboy :)

PR #2185, opened 2023-06-15
Only initialize the GZipResponder gzip buffer when needed

jlrobins:
# Summary
Only initialize self.gzip_buffer / self.gzip_file when going down codepaths where compression will definitely apply. Get those ping responses out faster by making  GzipMiddleware have less per-hit overhead if the hit isn't going to need compression.

I reviewed test_gzip.py and they cover all the codepath variants here already, and pass with 100% coverage over the module.

# Checklist

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -36,7 +36,7 @@ class GZipResponder:
         self.content_encoding_set = False
         self.compresslevel = compresslevel
 
-    def _init_gzip_buffer(self):
+    def _init_gzip_buffer(self) -> None:
         # Only initialize these when going to definitely need.
         self.gzip_buffer = io.BytesIO()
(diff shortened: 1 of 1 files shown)


Kludex:
Do we have numbers on how much does this change improves?

Kludex:
> Do we have numbers on how much does this change improves?

I'm closing this PR for now, since the changes seem to be unnecessary. If you provide data that sustains your argument, I'll be happy to reopen this PR. 🙏 

Thanks for the interest.

jlrobins:
Why initialize datastructures during a request cycle that won't be used, when the codepaths where initialization will be needed are clear and few and there's 100% code coverage?

Just to watch the user's CPUs spin? These middlewares are used all over the place. Making them a little faster would go a long way across all of the installations worldwide and this doesn't appear to be any sort of complexity burden or risky change.

PR #2189, opened 2023-06-18
Add include_router to Starlette and Router

Lancetnik:
# Summary

It's a PR connected with **FastAPI** nested lifespan [bugfix](https://github.com/tiangolo/fastapi/pull/9630).
So, there is a raw `include_router` implementation to inherit it by the **FastAPI** next step.
Also, it merge all nested by `mount` applications' events and lifespans to main application (#649).

# Checklist

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -237,6 +237,21 @@ class Starlette:
 
         return decorator
 
+    def include_router(
+        self,
+        router: Router,
+        *,
+        prefix: str = "",
+        include_in_schema: bool = True,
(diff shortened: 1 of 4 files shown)


Kludex:
That's really not what I mean. `include_router` doesn't make sense for Starlette.

We have this related PR: https://github.com/encode/starlette/pull/1988.

Lancetnik:
@Kludex did you talk about run `Host` routes lifespan calling? Or something else? Can you explain please?

PR #2191, opened 2023-06-22
Add `request` argument to `TemplateResponse`

alex-oleshkevich:
This PR makes `request` a required argument to `TemplateResponse`. 
Also, as the request is now passed explicitly, the template context can be optional.

**Why?** 
The  `request` is mandatory in the template response. However, the class requires the request to be passed via `context` and there are asserts to check it. I found it better to pass the `request` explicitly. It is similar to what [Django does](https://docs.djangoproject.com/en/4.2/ref/template-response/#the-rendering-process).

**Was**
```python
TemplateResponse('index.html', {'request': request})
```

**Become**
```python
TemplateResponse(request, 'index.html')
```

This is a breaking change.
Ref https://github.com/encode/starlette/discussions/2174

EDIT [@Kludex]: This is not a breaking change anymore - It now introduces deprecation warnings instead.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -22,7 +22,7 @@ from starlette.staticfiles import StaticFiles
 templates = Jinja2Templates(directory='templates')
 
 async def homepage(request):
-    return templates.TemplateResponse('index.html', {'request': request})
+    return templates.TemplateResponse(request, 'index.html')
 
 routes = [
     Route('/', endpoint=homepage),
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -206,7 +206,7 @@ class Jinja2Templates:
             if "request" not in kwargs:
                 warnings.warn(
                     "The `TemplateResponse` now requires the `request` argument.\n"
-                    'Replace `TemplateResponse(name, {"context": context})` by `TemplateResponse(request, name)`.",  # noqa: E501 
+                    'Replace `TemplateResponse(name, {"context": context})` by `TemplateResponse(request, name)`.',  # noqa: E501
                     DeprecationWarning,
                 )
                 if "request" not in kwargs.get("context", {}):
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -140,6 +140,7 @@ class Jinja2Templates:
     def get_template(self, name: str) -> "jinja2.Template":
         return self.env.get_template(name)
 
+    @typing.overload
     def TemplateResponse(
         self,
         request: Request,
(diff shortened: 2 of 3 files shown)


Kludex (changes requested):
> This is a breaking change.

Please make both `request` and `context` optional, and raise a deprecation warning when the "request" is in the `context`.

alex-oleshkevich:
> > This is a breaking change.
> 
> Please make both `request` and `context` optional, and raise a deprecation warning when the "request" is in the `context`.

Actually, the request must be the first argument.
Moving it after `name` also is not okay because `name` and `context` both are in the same domain and good to have them together:
```python
TemplateResponse(request, 'index.html', {})
TemplateResponse('index.html', request, {})
```

Kludex:
> Actually, the request must be the first argument.

"Must" is a bit strong, isn't it? 👀 

I don't think the argument here justifies the breaking change. It's just cosmetics.

alex-oleshkevich:
> > Actually, the request must be the first argument.
> 
> "Must" is a bit strong, isn't it? eyes
> 
> I don't think the argument here justifies the breaking change. It's just cosmetics.

Can you suggest an alternative? Seems that I am stuck..

Kludex:
Hmm... I was writing a huge text as to why my last suggestion was the only one... And I actually got an idea... 🤔

You can make the two first arguments as `str | Request | None = None`, and context as `Dict | None = None`. Meaning:

1. Check if the first parameter:
  1.1. `isinstance(request, str)` -> Use it as `name`; raise a `DeprecationWarning` - say it's not the 1st param anymore.
  1.2. `isinstance(request, Request)` -> Work as expected.
2. Check if the second parameter:
  1.1. `isinstance(name, str)` -> Work as expected.
  1.2. `isinstance(name, dict) -> Use it as `context`; raise `DeprecationWarning` - say it's not the 2nd param anymore.
    1.2.1. Check if `request` is in `context`, as it's happening right now, but if it's, raise `DeprecationWarning`.
3. Check if third parameter ???

My guess is this is going to look hideous, but it's not a breaking change, and we can easily remove all this logic on 1.0.

alex-oleshkevich:
Okay, it worked. The code is a bit ugly but it's okay for the transition period.

Kludex:
> The code is a bit ugly

Eh... Yeah... It was within my predictions hahaha

Would you like to open a PR against this already to show how it would look like on 1.0? I think it will make it easier to get this in.

alex-oleshkevich:
@Kludex https://github.com/encode/starlette/pull/2199/files

starlette/templating.py
```diff
@@ -140,19 +140,85 @@ def url_for(context: dict, __name: str, **path_params: typing.Any) -> URL:
...
+        name: str,
+        context: typing.Optional[dict] = None,
         status_code: int = 200,
         headers: typing.Optional[typing.Mapping[str, str]] = None,
         media_type: typing.Optional[str] = None,
         background: typing.Optional[BackgroundTask] = None,
     ) -> _TemplateResponse:
-        if "request" not in context:
-            raise ValueError('context must include a "request" key')
+        # Deprecated usage
```
Kludex, line 166: Unfortunately, we can't use `@typing_extensions.deprecated` without making `typing_extensions` mandatory for Starlette.

Also, the `@deprecated` PEP was still not accepted.

docs/templates.md
```diff
@@ -22,7 +22,7 @@ from starlette.staticfiles import StaticFiles
 templates = Jinja2Templates(directory='templates')
 
 async def homepage(request):
-    return templates.TemplateResponse('index.html', {'request': request})
+    return templates.TemplateResponse(request, 'index.html')
```
Kludex, line 25: Is this the only thing in the documentation that uses `TemplateResponse`?

starlette/templating.py
```diff
@@ -140,19 +140,85 @@ def url_for(context: dict, __name: str, **path_params: typing.Any) -> URL:
...
+
+            context = kwargs.get("context", {})
+            request = kwargs.get("request", context.get("request"))
+            name = typing.cast(str, kwargs["name"])
+            status_code = kwargs.get("status_code", 200)
+            headers = kwargs.get("headers")
+            media_type = kwargs.get("media_type")
+            background = kwargs.get("background")
+
+        context = context or {}
```
Kludex: Is this necessary? `else` has a `kwargs.get("context")` instead of a `kwargs.get("context", {})`... 🤔

starlette/templating.py
```diff
@@ -140,19 +140,85 @@ def url_for(context: dict, __name: str, **path_params: typing.Any) -> URL:
...
+    def TemplateResponse(
+        self, *args: typing.Any, **kwargs: typing.Any
+    ) -> _TemplateResponse:
+        if args:
+            if isinstance(
+                args[0], str
+            ):  # the first argument is template name (old style)
+                warnings.warn(
+                    "Argument 1 of TemplateResponse must be a Request instance.",
+                    DeprecationWarning,
```
Kludex: What do you think about this?
```suggestion
                    "The `name` is not the first parameter anymore. "
                    "The first parameter should be the `Request` instance.\n"
                    'Replace `TemplateResponse(name, {"request": request})` by `TemplateResponse(request, name)`.',  # noqa: E501
                    DeprecationWarning,
```

starlette/templating.py
```diff
@@ -140,19 +140,85 @@ def url_for(context: dict, __name: str, **path_params: typing.Any) -> URL:
...
+                )
+                headers = args[4] if len(args) > 4 else kwargs.get("headers")
+                media_type = args[5] if len(args) > 5 else kwargs.get("media_type")
+                background = args[6] if len(args) > 6 else kwargs.get("background")
+        else:  # all arguments are kwargs
+            if "request" not in kwargs:
+                warnings.warn(
+                    "TemplateResponse requires `request` keyword argument.",
+                    DeprecationWarning,
+                )
```
```suggestion
                warnings.warn(
                    "The `TemplateResponse` now requires the `request` argument.\n"
                    'Replace `TemplateResponse(name, {"context": context})` by `TemplateResponse(request, name)`.",  # noqa: E501 
                    DeprecationWarning,
                )
```

starlette/templating.py
```diff
@@ -140,19 +140,85 @@ def url_for(context: dict, __name: str, **path_params: typing.Any) -> URL:
...
+                if "request" not in kwargs.get("context", {}):
+                    raise ValueError('context must include a "request" key')
```
Kludex, line 213: This error is being raised with the warning above, can we make sure we only have the warning if this doesn't raise?

Kludex (changes requested):
I've added some comments. @alex-oleshkevich Can you check them, please?

starlette/templating.py
alex-oleshkevich: yes, so we have a comment here.

docs/templates.md
alex-oleshkevich: yes

starlette/templating.py
alex-oleshkevich: This preserves current behavior.

starlette/templating.py
Kludex: yeah yeah, just explaining for future reference 👀

Kludex:
Thanks @alex-oleshkevich 🙏 

For the record, this was accepted given that it is not a breaking change, and that we already have a follow-up PR to clean-up everything for V1: https://github.com/encode/starlette/pull/2199. 🙏 

The idea is to warn the users now, and then remove the `DeprecationWarning` for V1.

Kludex (approved):

PR #2193, opened 2023-06-23
Clarify lifespan vs on_startup/on_shutdown priority

stratosgear:
As per discussion (https://github.com/encode/starlette/discussions/2192), I am trying to make it more clear that if both lifespan AND on_startup/on_shutdown are used, regardless of the subtle hint that both should NOT be used, then only lifespan would be used and  on_startup/on_shutdown would silently be ignored.

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -606,15 +606,14 @@ class Router:
                 "See more about it on https://www.starlette.io/lifespan/.",
                 DeprecationWarning,
             )
+            if lifespan:
+                warnings.warn(
+                    "The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`. "
+                    "Both `on_startup` and `on_shutdown` will be ignored."
+                )
 
         if lifespan is None:
             self.lifespan_context: Lifespan = _DefaultLifespan(self)
-
-        if (on_startup or on_shutdown) and lifespan is not None:
-            warnings.warn(
-                "Lifespan cannot be used together with on_startup, on_shutdown.  The "
-                "on_startup, on_shutdown parameters will be ignored!"
-            )
 
         elif inspect.isasyncgenfunction(lifespan):
             warnings.warn(
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -610,6 +610,12 @@ class Router:
         if lifespan is None:
             self.lifespan_context: Lifespan = _DefaultLifespan(self)
 
+        if (on_startup or on_shutdown) and lifespan is not None:
+            warnings.warn(
+                "Lifespan cannot be used together with on_startup, on_shutdown.  The "
+                "on_startup, on_shutdown parameters will be ignored!"
+            )
+
         elif inspect.isasyncgenfunction(lifespan):
             warnings.warn(
                 "async generator function lifespans are deprecated, "
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -608,8 +608,9 @@ class Router:
             )
             if lifespan:
                 warnings.warn(
-                    "The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`. "
-                    "Both `on_startup` and `on_shutdown` will be ignored."
+                    "The `lifespan` parameter cannot be used with `on_startup` or "
+                    "`on_shutdown`. Both `on_startup` and `on_shutdown` will be "
+                    "ignored."
                 )
 
         if lifespan is None:
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -663,7 +663,8 @@ def test_lifespan_with_on_events():
         return PlainTextResponse("hello, world")
 
     with pytest.warns(
-        UserWarning, match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`."
+        UserWarning,
+        match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`.",
     ):
         Router(
             on_startup=[run_startup],
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -648,6 +648,31 @@ def test_lifespan_async(test_client_factory):
     assert shutdown_complete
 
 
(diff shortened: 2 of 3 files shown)


Kludex:
I don't think this is what we meant on the discussion (at least not what I mean). 🤔 

I mean to raise a warning if `lifespan` is used on `Router` if not provided from `Starlette`.

stratosgear:
Something like this?

Kludex:
Yes. Are you able to create tests for it?

stratosgear:
Something like this? Note, locally some existing tests would still fail with:

```
+ venv/bin/mypy starlette tests
starlette/routing.py:638: error: Incompatible types in assignment (expression has type "Union[Callable[[Any], AbstractAsyncContextManager[None]], Callable[[Any], AbstractAsyncContextManager[Mapping[str, Any]]], None]", variable has type "Union[Callable[[Any], AbstractAsyncContextManager[None]], Callable[[Any], AbstractAsyncContextManager[Mapping[str, Any]]]]")  [assignment]
Found 1 error in 1 file (checked 67 source files)
```  

But this has nothing to do with my patch!

stratosgear:
Just noticed that this is the same reason why the online tests fail too, but I have no idea how to fix those...

tests/test_routing.py
```diff
@@ -648,6 +648,31 @@ async def run_shutdown():
...
+        pass
+
+    def run_shutdown():
+        pass
+
+    def hello_world(request):
+        return PlainTextResponse("hello, world")
+
+    with pytest.warns(
+        UserWarning, match="Lifespan cannot be used together with on_startup"
```
Kludex:
```suggestion
        UserWarning, match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`."
```

Kludex:
> Something like this? Note, locally some existing tests would still fail with:
> 
> ```
> + venv/bin/mypy starlette tests
> starlette/routing.py:638: error: Incompatible types in assignment (expression has type "Union[Callable[[Any], AbstractAsyncContextManager[None]], Callable[[Any], AbstractAsyncContextManager[Mapping[str, Any]]], None]", variable has type "Union[Callable[[Any], AbstractAsyncContextManager[None]], Callable[[Any], AbstractAsyncContextManager[Mapping[str, Any]]]]")  [assignment]
> Found 1 error in 1 file (checked 67 source files)
> ```
> 
> But this has nothing to do with my patch!

It does. You added the conditional in a block of other `elseif`s.

tests/test_routing.py
```diff
@@ -648,6 +648,31 @@ async def run_shutdown():
...
+        pass
+
+    def run_shutdown():
+        pass
+
+    def hello_world(request):
+        return PlainTextResponse("hello, world")
+
+    with pytest.warns(
+        UserWarning, match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`."
```
Kludex:
```suggestion
        UserWarning,
        match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`.",
```

starlette/routing.py
```diff
@@ -606,6 +606,11 @@ def __init__(
                 "See more about it on https://www.starlette.io/lifespan/.",
                 DeprecationWarning,
             )
+            if lifespan:
+                warnings.warn(
+                    "The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`. "
+                    "Both `on_startup` and `on_shutdown` will be ignored."
```
Kludex:
```suggestion
                    "The `lifespan` parameter cannot be used with `on_startup` or "
                    "`on_shutdown`. Both `on_startup` and `on_shutdown` will be "
                    "ignored."
```

tests/test_routing.py
```diff
@@ -648,6 +648,32 @@ async def run_shutdown():
...
+
+    def run_shutdown():
+        pass
+
+    def hello_world(request):
+        return PlainTextResponse("hello, world")
+
+    with pytest.warns(
+        UserWarning,
+        match="The `lifespan` parameter cannot be used with `on_startup` or `on_shutdown`.",
```
Kludex:
```suggestion
        match=(
            "The `lifespan` parameter cannot be used with `on_startup` or "
            "`on_shutdown`."
        ),
```

stratosgear:
I am not sure I understand why it still fails. :(

I see your edit of moving my if inside the previous if (makes sense) but still it behaves very weird... 

I'm not sure I can help anymore here... :(

stratosgear:
I now completely understand why so few projects have 100% test coverage.  The amount of time I spend trying to make this trivial fix was close to 10 times more time in test authoring, rather than the actual code change itself (at least!!!). 

I am not sure if you really persist and become more proficient how much more efficient this process ever becomes... So as much as I would like to also have 100% coverage in my projects I do not know if I would ever find the time to actually do it.

This is a task that I would not mind if some "testGPT" would ever take over! :)

Kludex (approved):

Kludex:
> I now completely understand why so few projects have 100% test coverage. The amount of time I spend trying to make this trivial fix was close to 10 times more time in test authoring, rather than the actual code change itself (at least!!!).
> 
> I am not sure if you really persist and become more proficient how much more efficient this process ever becomes... So as much as I would like to also have 100% coverage in my projects I do not know if I would ever find the time to actually do it.
> 
> This is a task that I would not mind if some "testGPT" would ever take over! :)

It is worth it, and it actually saves a lot of time. :)

You should have asked for feedback if you were feeling frustrated. I believe the interpretation of the coverage report on your behalf was not accurate.

stratosgear:
> You should have asked for feedback if you were feeling frustrated. I believe the interpretation of the coverage report on your behalf was not accurate.

I hope you mean the first time only, when I said "I'm not sure I can help anymore here... :face_with_spiral_eyes: ", because after that I could not accept defeat and took a more serious look and finally changed the tests enough to pass the whole test suite. Give me some credit here!  :smile:

PR #2194, opened 2023-06-27
Stop `body_stream` in case `more_body=False`

### LINKED ISSUE #2093: `HTTPMiddleware` breaks `BackgroundTasks` in the same connection

Kludex:
Closes https://github.com/encode/starlette/issues/2093

We need to check how this interacts with FastAPI: https://github.com/encode/starlette/pull/1609#issuecomment-1307600095

I've tested this with the following application:

```py
from asyncio import sleep
from typing import Any, Coroutine

from starlette.responses import JSONResponse, Response, StreamingResponse

n = 0


async def task() -> None:
    global n
    local = n
    print(f"Start {local}.")
    n += 1
    await sleep(3)
    print(f"Finish {local}.")


    async def dispatch(
    ) -> Coroutine[Any, Any, Response]:


async def homepage(request: Request) -> ...:
    tasks = BackgroundTasks()
    tasks.add_task(task)
    return JSONResponse(content={"potato": "potato"}, background=tasks)


def exc_stream(request: Request):
    return StreamingResponse(_generate_faulty_stream())


def _generate_faulty_stream():
    yield b"Ok"
    raise Exception("Faulty Stream")


def empty(request: Request):
    return Response(status_code=200)


app = Starlette(
    routes=[
        Route("/", endpoint=homepage),
        Route("/stream", endpoint=exc_stream),
        Route("/empty", endpoint=empty),
    ],
    middleware=[Middleware(GZipMiddleware), Middleware(CustomMiddleware), Middleware(GZipMiddleware)],
)
```

Run with `uvicorn --http h11 main:app --reload`.

I've used the following client:

```py
from httpx import Client

with Client(base_url="http://localhost:8000") as client:
    print("Calling '/'")
    for n in range(5):
        client.get("/")

    print("Calling '/empty'")
    for n in range(5):
        client.get("/empty")

#     print("Calling '/stream'")
#     for n in range(5):
#         client.get("/stream")
```

Run with `python client.py`.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -170,6 +170,8 @@ class BaseHTTPMiddleware:
                         body = message.get("body", b"")
                         if body:
                             yield body
+                        if not message.get("more_body", False):
+                            break
 
                 if app_exc is not None:
                     raise app_exc
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -276,7 +276,6 @@ async def test_do_not_block_on_background_tasks():
         await anyio.sleep(0.1)
         events.append("Background task finished")
 
-
     async def endpoint_with_background_task(_):
         return PlainTextResponse(
             content="Hello", background=BackgroundTask(sleep_and_set)
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -266,6 +266,72 @@ async def test_run_background_tasks_even_if_client_disconnects():
 
 
 @pytest.mark.anyio
+async def test_do_not_block_on_background_tasks():
+    request_body_sent = False
+    response_complete = anyio.Event()
+    events: List[str | Message] = []
+
+    async def sleep_and_set():
+        events.append("Background task started")
+        await anyio.sleep(0.1)
+        events.append("Background task finished")
+
+
+    async def endpoint_with_background_task(_):
+        return PlainTextResponse(
+            content="Hello", background=BackgroundTask(sleep_and_set)
+        )
+
+    async def passthrough(
+        request: Request, call_next: Callable[[Request], Awaitable[Response]]
+    ) -> Response:
+        return await call_next(request)
+
+    app = Starlette(
+        middleware=[Middleware(BaseHTTPMiddleware, dispatch=passthrough)],
+        routes=[Route("/", endpoint_with_background_task)],
+    )
+
+    scope = {
+        "type": "http",
+        "version": "3",
+        "method": "GET",
+        "path": "/",
+    }
+
+    async def receive() -> Message:
+        nonlocal request_body_sent
+        if not request_body_sent:
+            request_body_sent = True
+            return {"type": "http.request", "body": b"", "more_body": False}
+        await response_complete.wait()
+        return {"type": "http.disconnect"}
+
+    async def send(message: Message):
+        if message["type"] == "http.response.body":
+            events.append(message)
+            if not message.get("more_body", False):
+                response_complete.set()
+
+    async with anyio.create_task_group() as tg:
+        tg.start_soon(app, scope, receive, send)
+        tg.start_soon(app, scope, receive, send)
+
+    # Without the fix, the background tasks would start and finish before the
+    # last http.response.body is sent.
+    assert events == [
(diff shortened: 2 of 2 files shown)


Kludex:
I've tested against FastAPI main branch, and it looks like it doesn't break the dependency injection system, as the previous attempt.

@encode/maintainers Can someone review it, please? I want us to make sure that we don't break FastAPI. If you have a suggestion about how to test this, it would be appreciated as well.

alex-oleshkevich:
That's ok. `break` closes the iterator just like if there are no more data.

adriangb:
On `fastapi==0.100.0` https://github.com/encode/starlette/pull/1609#issuecomment-1306767290 works correctly for me either with the version of starlette it ships with (`0.27.X`) nor `0.28.0`. It also seems to work with this branch. I once again confirmed it's broken on `fastapi==0.85.0`. So it seems that at the very least this change does not break FastAPI in the same way as the previous one did.

As long as we can confirm via a test that this fixes the issues with background tasks this looks good to me 👍🏻

adriangb (approved):

Kludex:
I gave another round of manual test with FastAPI, and it works as expected - not breaking their dependency system. 🙏

b4stien:
Hi, this PR [regressed this issue](https://github.com/tiangolo/fastapi/discussions/8577#discussioncomment-8914357), I've tested manually that commenting those two lines allowed `app_exc` not to be None and to be subsequently raised.

_(I'm no expert in ASGI and starlette codebase, don't understand the root issue that this change was trying to fix and I'm then unable to propose a better fix...)_

PR #2199, opened 2023-06-30
Add request argument to TemplateResponse (v1.x)

alex-oleshkevich:
This PR removes backward compatibility done in https://github.com/encode/starlette/pull/2191
To be merged after 1.0 release.


starlette/templating.py
```diff
@@ -142,17 +142,16 @@ def get_template(self, name: str) -> "jinja2.Template":
 
     def TemplateResponse(
         self,
+        request: Request,
```
aminalaee, line 145: Any reason why this is not an optional argument so it can be released without breaking?

Kludex:
I meant to create a PR so I could see the diff between the PR that adds deprecation and what is supposed to be refactor for v1.

I mean, target the PR you are building from this one.

alex-oleshkevich:
@Kludex this PR is the target repo state after removing all deprecations.

Kludex:
Screenshot 2023-07-03 at 17 16 06

I need instead of "encode:master" to be "alex-oleshkevich:templates-request". 🤔

alex-oleshkevich:
See it here https://github.com/alex-oleshkevich/starlette/pull/56/files

Kludex:
> To be merged after 1.0 release.

Not after, but for 1.0.

starlette/templating.py
Kludex: This was discussed on https://github.com/encode/starlette/pull/2191/.

The diff here is actually incomplete.

Kludex:
I've applied this on https://github.com/encode/starlette/pull/2384, and given co-author to @alex-oleshkevich . FYI

PR #2205, opened 2023-07-03
Update pyproject.toml

aminalaee:
# Summary

Remove mypy `no_implicit_optional=True` as it is True by default.

# Checklist


Kludex (approved):

Kludex:
Thanks Amin 🙏 

Let's hang out this week, if you are in Utrecht. 👍

Kludex:
Thanks Amin 🙏 

Let's go out this week. 👍

PR #2207, opened 2023-07-06
Allow global `follow_redirects` option.

alex-oleshkevich:
When you have many good-old HTML forms and use post-redirect-get pattern, it quickly becomes noisy to add `follow_redirects=False` to each HTTP call if you want to check the redirect location and headers (like 200/400 for form validation errors and 30x for successful submissions).

Setting `follow_redirects` at `TestClient` level makes things simpler and less repeatable.


Kludex (approved):
👍

PR #2209, opened 2023-07-06
Version 0.29.0

Kludex:
The intention, as mentioned before, is for this release to be the last before stable release.

The plan is:
1. Release `0.29.0`.
2. Wait for FastAPI to use this version.
3. Wait a month to see if there's any bug.
4. Either release `0.29.1` if bugs are found, otherwise `1.0.0`.

fyi @encode/maintainers We agreed on trying to release 1.0 as soon as possible in October 2022, so I assume we don't need to discuss the matter once again.

## Checklist

- [x] https://github.com/encode/starlette/pull/2194
- [x] https://github.com/encode/starlette/pull/2191
- [x] ~https://github.com/encode/starlette/pull/2180~ Pipeline is not green - we need to configure mypy properly over there, but let's unblock the release.


Kludex:
I'm going to release `0.29.0`, and then a follow-up with the Python 3.7 drop, and AnyIO bump.

aminalaee (approved):

docs/release-notes.md
```diff
@@ -1,3 +1,14 @@
+## 0.29.0
+
+July 7, 2023
```
aminalaee: Reminder.

Kludex: Thanks Amin 🙏

PR #2211, opened 2023-07-12
Upgraded to AnyIO 4.0

agronholm:
# Summary

This updates Starlette to work with AnyIO 4.0. Note that while this PR maintains compatibility with AnyIO 3.x, some of the new AnyIO behavior is likely to bleed through in some use cases at least.

# Checklist


Kludex:
(I'll rebase this on top of https://github.com/encode/starlette/pull/2178, fyi)

agronholm:
Would you like me to add an extra test job to test against both AnyIO 3.x and 4.x? There is no 4.x release available yet though, hence the git pointer in `requirements.txt`.

tests/middleware/test_wsgi.py
```diff
@@ -66,7 +68,7 @@ def test_wsgi_exception(test_client_factory):
     # The HTTP protocol implementations would catch this error and return 500.
     app = WSGIMiddleware(raise_exception)
     client = test_client_factory(app)
-    with pytest.raises(RuntimeError):
+    with pytest.raises(RuntimeError), convert_excgroups():
```
Kludex: We can use the `ExceptionGroup` here. I'm fine with this break, since we intend to remove `WSGIMiddleware`.

tests/middleware/test_base.py
```diff
@@ -15,6 +15,8 @@
 from starlette.testclient import TestClient
```
Kludex: I'd prefer if we maintain the tests unchanged, and unpack the `ExceptionGroup` on the `BaseHTTPMiddleware` in case we have a single exception.

Kludex (approved):
Thank you @agronholm! 🙏

PR #2214, opened 2023-07-13
Added official Python 3.12 support

agronholm:
# Summary

- Adds Python 3.12 to CI
- Updates `pytest` to a compatible version
- Adds the `Python :: 3.12` classifier

# Checklist

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,13 +14,14 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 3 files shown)


agronholm:
I looked for documentation to update but couldn't find anything relevant.

Kludex (approved):
Thanks 🙏

Kludex:
Don't we need to point at 3.12-dev?

agronholm:
> Don't we need to point at 3.12-dev?

Nope, the `allow-prereleases: true` flag takes care of that.

Kludex:
Thanks 🙏

PR #2218, opened 2023-07-20
Try to make Request generic over State

adriangb:
https://github.com/encode/starlette/pull/2065#issuecomment-1643474342

TODO: how do we let users at runtime customize the state creation function? Currently, Starlette creates the Request object so it can't be as simple as a constructor parameter. For FastAPI I suppose it could introspect the type hints to get the generic argument.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -3,6 +3,7 @@ import typing
 from http import cookies as http_cookies
 
 import anyio
+import typing_extensions
 
 from starlette._utils import AwaitableOrContextManager, AwaitableOrContextManagerWrapper
 from starlette.datastructures import URL, Address, FormData, Headers, QueryParams, State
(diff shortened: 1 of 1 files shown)


adriangb:
@Kludex any idea how to get this to work nicely with mypy?

Kludex:
> @Kludex any idea how to get this to work nicely with mypy?

I think I have some comments around about it. I'll look later, and share it here. (It maybe in our WhatsApp conversation... I'll check.)

Kludex:
I've created https://github.com/python/typing/discussions/1457.

michaeloliverx:
On an semi related side note I hacked around this in FastAPI by doing this:
```python
from collections.abc import AsyncGenerator
from typing import Annotated, TypedDict, cast

from fastapi import Depends, FastAPI, Request


class State(TypedDict):
    foo: str
    bar: str


@asynccontextmanager
async def lifespan(app: FastAPI) -> AsyncGenerator[State, None]:
    ...


async def get_state(request: Request) -> State:
    return cast(State, request.state._state)


StateDependency = Annotated[State, Depends(get_state)]

```

Kludex:
I'll close this, since the Python type system still doesn't support this.

- We can watch https://github.com/python/typing/discussions/1457 and related discussions.

PR #2219, opened 2023-07-20
Don't group duplicated headers on a single string on the `TestClient` 

ooknimm:
# Summary

The testclient handles headers with duplicate field names differently.

when request with headers with duplicate field name, starlette can handles like this 

#### starlette
```python
# request header
[("x-token", "foo"), ("x-token", "bar")]
```

``` python
# in starlette
request.headers.getlist("x-token")
--> ["foo", "bar"]
```

<br>

but testclient(httpx client) not handles like that

### testclient
``` python
# request header
[("x-token", "foo"), ("x-token", "bar")]
```

``` python
# in starlette
request.headers.getlist("x-token")
--> ["foo, bar"]
```

so i fixed the testclient same as starlette to handle headers with duplicate field name

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -223,7 +223,10 @@ class _TestClientTransport(httpx.BaseTransport):
             headers = [(b"host", (f"{host}:{port}").encode())]
 
         # Include other request headers.
-        headers += [(key, value) for _, key, value in request.headers.raw]
+        headers += [
+            (key.lower().encode(), value.encode())
+            for key, value in request.headers.multi_items()
+        ]
 
         scope: typing.Dict[str, typing.Any]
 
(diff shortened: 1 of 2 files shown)


starlette/testclient.py
```diff
@@ -223,10 +223,7 @@ def handle_request(self, request: httpx.Request) -> httpx.Response:
             headers = [(b"host", (f"{host}:{port}").encode())]
 
         # Include other request headers.
-        headers += [
-            (key.lower().encode(), value.encode())
-            for key, value in request.headers.items()
-        ]
+        headers += [(key, value) for _, key, value in request.headers._list]
```
```suggestion
        headers += [(key, value) for _, key, value in request.headers.raw]
```

ooknimm:
`headers.raw` includes upper case key

i found multi_items method that lower case key.

Kludex (approved):
Thanks @ooknimm 🙏

PR #2222, opened 2023-07-21
Gracefully handle OSErrors in StaticFiles

etr2460:
# Summary
Instead of reraising the OSError, handle it and raise an HTTPException with a Bad Request status code response. Arguably this could be a 404 to avoid leaking anything, but a 400 seemed reasonable (and better than not catching the error).

Implements solution 1 to the discussion here: https://github.com/encode/starlette/discussions/2221

# Checklist

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -123,8 +123,8 @@ class StaticFiles:
             )
         except PermissionError:
             raise HTTPException(status_code=401)
-        except OSError:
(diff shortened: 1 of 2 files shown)


Kludex:
@etr2460 Thanks for the PR.

Given what I replied on https://github.com/encode/starlette/discussions/2221#discussioncomment-6515799, I'll be closing this.

PR #2223, opened 2023-07-22
Add `encoder` argument to JSONResponse

alex-oleshkevich:
This small change allows users to serialize a custom types with a little effort compared to subclassing.

```python
import datetime
import json

class MyJSONEncoder(json.JSONEncoder):
    def default(self, value):
        if isinstance(value, datetime.datetime):
            return value.isoformat()
        ...

    response = JSONResponse({'now': datetime.datetime.now()}, encoder=MyJSONEncoder)
```

Similar to Django - https://docs.djangoproject.com/en/4.2/ref/request-response/#jsonresponse-objects

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -116,6 +116,33 @@ In general you *probably* want to stick with `JSONResponse` by default unless
 you are micro-optimising a particular endpoint or need to serialize non-standard
 object types.
 
+
+#### Custom JSON encoders
+
+If you need to encode custom data types, you can use a custom JSON encoder.
+
+For example, if you want to automatically serialize `datetime.datetime` instances to strings:
+
+```python
+import datetime
+import json
+
+from starlette.responses import JSONResponse
+
+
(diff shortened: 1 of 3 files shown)


Kludex:
We already had this discussion before... Need to find the reference.

But... I'm not keen on this change. It's too easy for the user to implement its custom `Response` class.

alex-oleshkevich:
I found it cleaner to share the encoder (when used not only for the API JSON responses) than building a custom response class. Moreover, inheriting the base class requires you to completely override JSON serialization, which may be unwanted in cases when base JSONResponse class does more than just `json.dumps` (not a case for now).

Kludex:
- https://github.com/encode/starlette/discussions/1517
- https://github.com/encode/starlette/discussions/1281

Kludex:
This PR goes against what is written in the "Custom JSON serialization" section in the documentation (the above section).

alex-oleshkevich:
ok, closing it

PR #2229, opened 2023-07-25
Apply AnyIO upgrade changes only on test suite

### LINKED ISSUE #2227: No module named 'exceptiongroup' when upgraded to starlette 0.31.0
After upgrading to starlette 0.31.0 today, we run into the following exception when importing starlette:

```
from starlette import applications, requests, responses, routing
2023-07-24 15:47:43    File "/opt/conda/lib/python3.8/site-packages/starlette/applications.py", line 8, in <module>
2023-07-24 15:47:43      from starlette.middleware.base import BaseHTTPMiddleware
2023-07-24 15:47:43    File "/opt/conda/lib/python3.8/site-packages/starlette/middleware/base.py", line 14, in <module>
2023-07-24 15:47:43      from exceptiongroup import BaseExceptionGroup
2023-07-24 15:47:43  ModuleNotFoundError: No module named 'exceptiongroup'
```

This is likely introduced from https://github.com/encode/starlette/pull/2211/files

Kludex:
- Closes https://github.com/encode/starlette/issues/2227
- Closes https://github.com/encode/starlette/discussions/2228

### DIFF
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -1,6 +1,7 @@
 import contextvars
-from contextlib import AsyncExitStack
-from typing import AsyncGenerator, Awaitable, Callable, List, Union
+import sys
+from contextlib import AsyncExitStack, contextmanager, suppress
+from typing import AsyncGenerator, Awaitable, Callable, Generator, List, Union
 
 import anyio
 import pytest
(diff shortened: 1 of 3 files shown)


tests/middleware/test_base.py
```diff
@@ -210,7 +231,8 @@ async def homepage(request):
     )
 
     client = test_client_factory(app)
-    response = client.get("/")
+    with convert_exc_groups():
+        response = client.get("/")
```
Kludex, line 235: This isn't supposed to raise an error. I'll investigate later.

agronholm:
Are you sure this is the right solution? You told me yourself that you'd prefer the singular exception from the exception group to be extracted and reraised. If you don't want to add `exceptiongroup` as a dependency yet, I think the exceptiongroup import just needs to be wrapped with a `try...except ImportError`. Then we can replace the context manager with a no-op.

PR #2230, opened 2023-07-25
Set `url_for` global for custom jinja envs

alex-oleshkevich:
Starlette defines `url_for` helper in `Jinja2Templates` by default. However, when a custom environment is used, this helper is missing.
This PR fixes this issue.


Kludex (approved):

Kludex:
Thanks @alex-oleshkevich :)

PR #2231, opened 2023-07-26
Fixed import error when exceptiongroup isn't available

### LINKED ISSUE #2227: No module named 'exceptiongroup' when upgraded to starlette 0.31.0

agronholm:
Fixes #2227. Closes #2229. This is an alternate version of the linked PR.

# Summary

Last minute changes to the AnyIO 4.0 compatibility PR inadvertently caused an `ImportError` on Pythons earlier than 3.11 when the `exceptiongroup` back-port wasn't installed. This makes the import optional and thus restores compatibility with AnyIO 3.x.

# Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ types-contextvars==2.4.7.2
 types-PyYAML==6.0.12.10
 types-dataclasses==0.6.6
 pytest==7.4.0
-trio==0.22.1
+trio==0.21.0
 anyio==3.7.1
 
 # Documentation
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ types-PyYAML==6.0.12.10
 types-dataclasses==0.6.6
 pytest==7.4.0
 trio==0.22.1
-anyio@git+https://github.com/agronholm/anyio.git
+anyio==4.0.0rc1
 
 # Documentation
 mkdocs==1.4.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,7 +13,7 @@ types-PyYAML==6.0.12.10
 types-dataclasses==0.6.6
 pytest==7.4.0
 trio==0.22.1
-anyio==4.0.0rc1
+anyio==3.7.1
 
 # Documentation
 mkdocs==1.4.3
--- a/starlette/_utils.py
+++ b/starlette/_utils.py
@@ -2,11 +2,19 @@ import asyncio
 import functools
 import sys
 import typing
+from contextlib import contextmanager
 
 if sys.version_info >= (3, 10):  # pragma: no cover
     from typing import TypeGuard
 else:  # pragma: no cover
     from typing_extensions import TypeGuard
+
+has_exceptiongroups = True
+if sys.version_info < (3, 11):  # pragma: no cover
+    try:
+        from exceptiongroup import BaseExceptionGroup
+    except ImportError:
+        has_exceptiongroups = False
 
 T = typing.TypeVar("T")
 AwaitableCallable = typing.Callable[..., typing.Awaitable[T]]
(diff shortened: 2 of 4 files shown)


agronholm:
I also went ahead and updated the pinned AnyIO version as v4.0.0rc1 is out now.

fluix-dev:
This would also need something like the following patch to fix the tests for anyio < 4 which https://github.com/encode/starlette/discussions/2228 raises.
```patch
From 917d1c135635d6d9e5aec83633673ec21a529933 Mon Sep 17 00:00:00 2001
From: Steven Guikal <void@fluix.one>
Date: Sun, 30 Jul 2023 17:18:40 -0400
Subject: [PATCH] Fix wsgi test for AnyIO versions before 4.0.0

---
 tests/middleware/test_wsgi.py | 15 ++++++++++-----
 1 file changed, 10 insertions(+), 5 deletions(-)

diff --git a/tests/middleware/test_wsgi.py b/tests/middleware/test_wsgi.py
index ad39754..5b2508c 100644
--- a/tests/middleware/test_wsgi.py
+++ b/tests/middleware/test_wsgi.py
@@ -2,6 +2,7 @@ import sys
 
 import pytest
 
+from importlib.metadata import version
 from starlette.middleware.wsgi import WSGIMiddleware, build_environ
 
 if sys.version_info < (3, 11):  # pragma: no cover
@@ -69,11 +70,15 @@ def test_wsgi_exception(test_client_factory):
     # The HTTP protocol implementations would catch this error and return 500.
     app = WSGIMiddleware(raise_exception)
-    with pytest.raises(ExceptionGroup) as exc:
-        client.get("/")
-
-    assert len(exc.value.exceptions) == 1
-    assert isinstance(exc.value.exceptions[0], RuntimeError)
+    if version("anyio") < "4":
+        with pytest.raises(RuntimeError):
+            client.get("/")
+    else:
+        with pytest.raises(ExceptionGroup) as exc:
+            client.get("/")
+
+        assert len(exc.value.exceptions) == 1
+        assert isinstance(exc.value.exceptions[0], RuntimeError)
 
 
 def test_wsgi_exc_info(test_client_factory):
-- 
2.41.0
```

agronholm:
So do you want me to keep 4.0.0rc1 as the pinned AnyIO version? If so, how do I prove that it also passes on 3.x?

fluix-dev:
I'm not a contributor so I'm not sure what compatibility guarantees are made or how tests are run. Applying my patch means distributions that have AnyIO < 4 in their repositories (aka everyone since 4.0.0 is only a pre-release) don't need to modify tests.

requirements.txt
```diff
@@ -13,7 +13,7 @@ types-PyYAML==6.0.12.10
 types-dataclasses==0.6.6
 pytest==7.4.0
 trio==0.22.1
-anyio@git+https://github.com/agronholm/anyio.git
+anyio==4.0.0rc1
```
```suggestion
anyio==3.7.1
```

Kludex: Is this the only change needed that you mentioned on https://github.com/encode/starlette/pull/2246 ?

requirements.txt
```diff
@@ -13,7 +13,7 @@ types-PyYAML==6.0.12.10
 types-dataclasses==0.6.6
 pytest==7.4.0
 trio==0.22.1
```
```suggestion
trio==0.21.0
```

agronholm:
Curious – the coverage fails even on AnyIO 3.7.1. Maybe the test changes themselves caused this?

Kludex:
Closed by mistake.

agronholm:
The only reason coverage fails here is that the code base is only tested against AnyIO 3 now, and tripping the missing line requires AnyIO 4. Should I add a `# pragma: no cover` there?

Ideally we would test against both but that may complicate the test suite too much.

Kludex:
Yes, add the pragma, please.

Kludex (approved):

PR #2232, opened 2023-07-27
Optimize IO Efficiency by Renaming Uploaded Files in the Same Disk

gshmu:
When 'PERFORMANCE_FILE_UPLOAD_IO_TMPDIR' environment variable is not set, no code logic is impacted. 

When set as the target storage disk through the environment variable, the file can be directly saved using the save_to method.


Kludex:
Please open a discussion to discuss about this.

PR #2234, opened 2023-07-28
Remove useless brackets from code example in documentation.

proofit404:
# Summary

Code example contains optional brackets which makes `yield` keyword looks like a function call. That may confuse newcomers.

# Checklist

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -143,11 +143,11 @@ import asyncio
 
 
 async def slow_numbers(minimum, maximum):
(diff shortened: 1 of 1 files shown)


proofit404:
CI is failing for the same reason main branch is failing.

See https://github.com/encode/starlette/commit/a8b8856ce393a82ab4a714131085dbc4d658e34d for more details.

Kludex (approved):

Kludex:
Thanks @proofit404 :)

PR #2245, opened 2023-08-13
Treat repeated query param names as list

Matthew-Burkard:
# Summary

An array may be represented as repeated keys in query params.

If a value is already present for a key, append to what is already there rather than overwriting it.
e.g.
```
a=123&a=456
```

With the current behavior of `QueryParams` repeated key names will override previous values.
This change converts the value to a list in the event of repeated key names.

I ran into this using `hx-vals` in htmx like so:
```html
hx-vals='{"params": [{"name": "a"}, {"name": "b"}]}'
```

The resulting `QueryParams._dict` was:

```python
{'params': '{"name":"b"}'}
```

# Checklist

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -388,7 +388,7 @@ class MultiDict(ImmutableMultiDict[typing.Any, typing.Any]):
         self._dict.update(value)
 
 
-class QueryParams(ImmutableMultiDict[str, str]):
+class QueryParams(ImmutableMultiDict[str, typing.Union[str, typing.List[str]]]):
     """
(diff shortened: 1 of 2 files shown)


alex-oleshkevich:
You should use `QueryParams.getlist(keyname)` to retrieve a list.
An automatic change from `str` to `list[str]` may break your app at runtime.

alex-oleshkevich:
Moreover, there is no a common standard. For example, PHP automatically parse `variable[]=1&variable=[]=2`  to array.
So adding it to Startette may be a source of confusion.

PR #2246, opened 2023-08-18
Revert "Upgraded to AnyIO 4.0 (#2211)"

Kludex:
@agronholm Are you fine with this? It will give me a bit more time, and it will unblock you, since I'm pinning back the to <4.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "anyio>=3.4.0,<4",
+    "anyio>=3.4.0,<5",
     "typing_extensions>=3.10.0; python_version < '3.10'",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "anyio>=3.4.0,<5",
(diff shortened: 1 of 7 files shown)


agronholm:
I'd rather you kept the memory object stream related changes, as reverting those would cause Starlette to not work with AnyIO 4 at all.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
-    "anyio>=3.4.0,<5",
+    "anyio>=3.4.0,<4",
```
Kludex: Do I also remove this, then? @agronholm

agronholm: I think that the only change you should make is to pin AnyIO to 3.* (and trio to 0.21) in your test requirements. The other changes don't hurt compatibility. But given that tests demonstrably do pass with AnyIO 4, users should be given the option of running Starlette against it.

Kludex: > But given that tests demonstrably do pass with AnyIO 4, users should be given the option of running Starlette against it.

I didn't get this... The tests are failing on main.

agronholm: Not the tests themselves, only the coverage part where 1 code line is missed, and looking at the code, it's not clear if that is a problem at all. That's the thing I asked you to investigate.

Kludex:
```suggestion
    "anyio>=3.4.0,<5",
```

PR #2251, opened 2023-08-24
Add example on how to use `SessionMiddleware`

Bernardoow:
# Summary

Add example to how to use session middleware. I made this PR because I found the documentation but the path to middleware is not present. So I had to search on code.

# Checklist


docs/middleware.md
```diff
@@ -107,6 +107,20 @@ The following arguments are supported:
...
 
+```python
+from starlette.applications import Starlette
+from starlette.middleware import Middleware
+from starlette.middleware.sessions import SessionMiddleware
+
+routes = ...
+
+middleware = [
+    Middleware(SessionMiddleware, https_only=True)
```
aminalaee: I think you need the `secret_key` argument for `SessionMiddleware`

Bernardoow: @aminalaee you're right.

image

I'm going to fix it now.. sorry. !!

docs/middleware.md
```diff
@@ -115,7 +115,7 @@ from starlette.middleware.sessions import SessionMiddleware
 routes = ...
 
 middleware = [
-    Middleware(SessionMiddleware, https_only=True)
+    Middleware(SessionMiddleware, secret_key="R4ND0M_S3CR3T_K3Y", https_only=True)
```
```suggestion
    Middleware(SessionMiddleware, secret_key=..., https_only=True)
```

Bernardoow: @Kludex done here 8e1128d

Kludex (approved):

PR #2255, opened 2023-08-26
Add JwtSessionMiddleware Class

lannuttia:
# Summary

This PR adds a JwtSessionMiddleware class. This class functions much like the existing SessionMiddleware except instead of sigining and unsigning data using itsdangerous, it encodes and decodes JWTs using python-jose. This will simplify interactions with third-party components.

# Checklist


Kludex:
As a start, I recommend creating a third party package, or check the already existing ones - it's likely something like this already exists.

Other than that, if you don't find anything, please create a discussion.

lannuttia:
I see that there are third-party JWT session middleware available but I personally have concerns about leveraging them for my purposes. I have created #2256 to lay out some ideas that I have for more flexible session middleware that might be able to alleviate the need for largely similar third-party session middleware packages.

PR #2264, opened 2023-09-01
Change Response.render content argument typing

### LINKED ISSUE #2242: Liskov substitution problems in new `Response.render` type signature
I recently updated to 0.31.0 and found I had a new type checking error (I run with `mypy --strict`).

```
myfile.py:28: error: Argument 1 of "render" is incompatible with supertype "Response"; supertype defines the argument type as "str | bytes | None"  [override]
        def render(self, model: Model) -> bytes:
                         ^~~~~~~~~~~~
src/ibeks/server/response.py:28: note: This violates the Liskov substitution principle
src/ibeks/server/response.py:28: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

This is due to the base class `Response` having the type signature for `render` being constrained to `str | bytes | None`.  From this PR: #2183 @Viicos @Kludex 

Here is my derived class using starlette 0.28.0:

```python
class JSONSerializedResponse(Response):
    charset = CHARSET

    def render(self, model: str | bytes | None) -> bytes:
        # unstructure the model to JSON using its registered converter, then
        # encode the payload
        return model.to_json().encode(self.charset)
```

Now I could put the encoding from my arbitrary Python object `Model` into the application code, but I quite like it being part of the response. In any case this is how it is advertised in the documentation and indeed the derived class `JSONResponse` in that same module re-types it to `Any`: https://github.com/encode/starlette/blob/master/starlette/responses.py#L180.

Whatever the final decision is I think this contradiction needs to be clarified.

Possible solutions would be (in order of preference):

- create a new generic base class that that uses a type variable (see below for example)
- revert `Response.render` type signature to accept `Any` (Any is typically the wrong solution for most issues and especially this, but it at least works with subtyping to more specific types)

---

```python
from typing import Generic, TypeVar

AnyContent = TypeVar("AnyContent")
class AbstractResponse(Generic[AnyContent]):

    def render(self, content: AnyContent) -> bytes:
        ...


class Response(AbstractResponse[str | bytes | None]):
    def render(self, content: str | bytes | None) -> bytes:
        ...

```

And my class could be:

```python
class JSONSerializedResponse(AbstractResponse[Model):
    charset = CHARSET

    def render(self, model: Model) -> bytes:
        # unstructure the model to JSON using its registered converter, then
        # encode the payload
        return model.to_json().encode(self.charset)

```

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/2242">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/2242/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

aminalaee:
# Summary

Fixes https://github.com/encode/starlette/issues/2242
According to https://github.com/encode/starlette/issues/2242#issuecomment-1669153640

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -37,12 +37,12 @@ class Response:
         self.body = self.render(content)
         self.init_headers(headers)
 
-    def render(self, content: typing.Union[str, bytes, None]) -> bytes:
+    def render(self, content: typing.Any) -> bytes:
         if content is None:
             return b""
         if isinstance(content, bytes):
             return content
-        return content.encode(self.charset)
+        return content.encode(self.charset)  # type: ignore
 
     def init_headers(
         self, headers: typing.Optional[typing.Mapping[str, str]] = None


starlette/responses.py
```diff
@@ -37,11 +37,12 @@ def __init__(
...
         self.init_headers(headers)
 
-    def render(self, content: typing.Union[str, bytes, None]) -> bytes:
+    def render(self, content: typing.Any) -> bytes:
         if content is None:
             return b""
         if isinstance(content, bytes):
             return content
+        assert isinstance(content, str)
         return content.encode(self.charset)
```
aminalaee: Maybe we can change it to this?

```py
    if isinstance(content, bytes):
        return content
    if isinstance(content, str)
         return content.encode(self.charset)
    return b""
```

It shouldn't have any breaking behaviour as we had limited content to be `str | bytes | None`.

Kludex: No. This can potentially hide issues for our users.

starlette/responses.py
```diff
@@ -37,11 +37,12 @@ def __init__(
         self.body = self.render(content)
         self.init_headers(headers)
 
-    def render(self, content: typing.Union[str, bytes, None]) -> bytes:
+    def render(self, content: typing.Any) -> bytes:
         if content is None:
             return b""
         if isinstance(content, bytes):
             return content
+        assert isinstance(content, str)
```
Kludex: Can we just add the `type: ignore` on line 46 for now? If someone complains, we change it.

starlette/responses.py
aminalaee: yeah, maybe.

Kludex (approved):

PR #2274, opened 2023-09-07
Inherit from HTMLResponse instead of Response on _TemplateResponse (#2272)

sponsfreixes:
That way we can get rid of setting the media_type. Also allows using a more specific type (HTMLResponse) on applications/libraries using type annotations.

See discussion #2272.


Kludex (approved):

Kludex:
Thanks :)

PR #2280, opened 2023-09-13
feat: add domain argument to SessionMiddleware

zllvm:
# Summary

This PR introduces an optional `domain` parameter to the SessionMiddleware constructor. Users can include this parameter when initializing SessionMiddleware to specify the desired `domain` for cookie scoping. More information about it can be found, for example, [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute).

The feature has been discussed in [Discussions #1603](https://github.com/encode/starlette/discussions/1603)

## Impact

- **Minimal Impact**: The changes in this PR are backward-compatible and should not disrupt existing implementations of SessionMiddleware.
- **Opt-In**: Users can choose to utilize the new `domain` parameter when initializing SessionMiddleware, ensuring that it does not affect those who do not use it.

# Checklist


abersheeran (approved):
Great improvement!

starlette/middleware/sessions.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         path: str = "/",
         same_site: typing.Literal["lax", "strict", "none"] = "lax",
         https_only: bool = False,
+        domain: str = "",
```
Kludex:
```suggestion
        domain: typing.Optional[str] = None,
```

Kludex (approved):

PR #2284, opened 2023-09-16
Ensure event ordering in `test_app_receives_http_disconnect_while_sending_if_discarded`

jhominal:
# Summary

Because of the way test was written, there was no guarantee that the `response_sent` event would be set after the downstream app calls `receive` to listen for disconnection.

Because of that, it was possible for one of the "if response_sent.is_set()" blocks to stay uncovered, which happened regularly when:
 * running the tests under `trio`;
 * running the tests using anyio 4.0.0 and `asyncio`;

# Checklist

- [*] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [*] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [*] I've updated the documentation accordingly.

### DIFF
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -396,7 +396,18 @@ async def test_run_context_manager_exit_even_if_client_disconnects():
 def test_app_receives_http_disconnect_while_sending_if_discarded(test_client_factory):
     class DiscardingMiddleware(BaseHTTPMiddleware):
         async def dispatch(self, request, call_next):
-            await call_next(request)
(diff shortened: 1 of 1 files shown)


adriangb (approved):

Kludex:
cc @agronholm

Kludex:
@jhominal Does this mean that we can run Starlette against anyio 4.0 now?

agronholm:
*Running* has worked at least since the last release. The only problem was that missing coverage line, so switching to AnyIO 4 for *testing* is what this PR enables.

Issue #1724, opened 2022-07-02
Custom `CapacityLimiter`


Kludex:
Starlette is a web framework that supports both async and sync functions. The sync part code runs in a threadpool.

The threadpool contains a maximum number of threads: 40.

https://github.com/agronholm/anyio/blob/4f3a8056a8b14dbe43c95039a0d731ede1083cb7/src/anyio/_backends/_asyncio.py#L2071-L2077 

The concern on this issue is that the threads are shared between the ones that handle the endpoint, and the background tasks.

Assume we have a simple application:
```python
from time import sleep

from starlette.applications import Starlette
from starlette.background import BackgroundTasks
from starlette.requests import Request
from starlette.responses import JSONResponse
from starlette.routing import Route

num = 0


def count_sleep():
    global num
    num += 1
    print(f"Running number {num}.")
    sleep(10)


def endpoint(request: Request) -> JSONResponse:
    tasks = BackgroundTasks()
    tasks.add_task(count_sleep)
    return JSONResponse({"message": "Hello, world!"}, background=tasks)


app = Starlette(routes=[Route("/", endpoint)])
```
Running it with `uvicorn`:
```
uvicorn main:app
```
And performing some requests (using [httpie](https://httpie.io/docs/cli)):
```bash
for run in {1..100}; do
  http :8000 &
done
```
We can observe that:
1. We can see `Running number 40.`.
2. Wait 10 seconds...
3. We can see `Running number 80.`.
4. Wait 10 seconds...
5. We can see `Running number 100.`.

I'm just bringing this up, so people are aware.

@agronholm proposed on Gitter that we create a separated `CapacityLimiter` dedicated only for handling the application (i.e. `request_response()`). This means that `n` (depending on the number of tokens we choose) number of threads would be dedicated for `request_response()`.

> [!IMPORTANT]
> - We're using [Polar.sh](https://polar.sh/encode) so you can upvote and help fund this issue.
> - We receive the funding once the issue is completed & confirmed by you.
> - Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/encode/starlette/issues/1724">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/1724/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/1724/pledge.svg">
</picture>
</a>

jhominal:
[...]

PR #2288, opened 2023-09-19
Add AnyIO CapacityLimiter docs

aminalaee:
Related to https://github.com/encode/starlette/issues/1724

# Summary

# Checklist

### DIFF
--- a/docs/threadpools.md
+++ b/docs/threadpools.md
@@ -2,8 +2,7 @@ Starlette supports both `async` and `sync` routes and background tasks.
 The sync part is running inside a threadpool by using [`AnyIO`](https://github.com/agronholm/anyio).
 
 The `AnyIO` library by default starts a threadpool with a fixed number of threads, which works for most cases.
-But since the threadpool is shared by both routes and background tasks,
-it might be useful to control the number of threads.
+But since the threadpool is shared by both incoming requests, background tasks, and any other AnyIO thread pool usage it might be useful to control the number of threads.
 
 
 ```py
--- a/docs/threadpools.md
+++ b/docs/threadpools.md
@@ -1,0 +1,34 @@
+Starlette supports both `async` and `sync` routes and background tasks.
+The sync part is running inside a threadpool by using [`AnyIO`](https://github.com/agronholm/anyio).
+
+The `AnyIO` library by default starts a threadpool with a fixed number of threads, which works for most cases.
+But since the threadpool is shared by both routes and background tasks,
+it might be useful to control the number of threads.
+
+
+```py
+import anyio
+from starlette.applications import Starlette
+from starlette.background import BackgroundTasks
+from starlette.request import Request
+from starlette.responses import JSONResponse
+from starlette.routing import Route
+
+# Set number of threads in threadpool to 100
+anyio.to_thread.current_default_thread_limiter().total_tokens = 100
+
+
+def do_task():
+    ...
+
+
+def endpoint(request: Request) -> JSONResponse:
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -47,6 +47,7 @@ nav:
   - Exceptions: 'exceptions.md'
   - Configuration: 'config.md'
   - Test Client: 'testclient.md'
+  - Sync Threadpools: 'threadpools.md'
```
aminalaee, line 50: I think it's not that easy to find this in the docs, if you don't know that you should look for the threadpool docs.

docs/threadpools.md
```diff
@@ -0,0 +1,34 @@
+Starlette supports both `async` and `sync` routes and background tasks.
+The sync part is running inside a threadpool by using [`AnyIO`](https://github.com/agronholm/anyio).
+
+The `AnyIO` library by default starts a threadpool with a fixed number of threads, which works for most cases.
+But since the threadpool is shared by both routes and background tasks,
+it might be useful to control the number of threads.
```
adriangb:
```suggestion
But since the threadpool is shared by both incoming requests, background tasks, and any other AnyIO thread pool usage it might be useful to control the number of threads.
```

docs/threadpools.md
```diff
@@ -0,0 +1,34 @@
...
+```py
+import anyio
+from starlette.applications import Starlette
+from starlette.background import BackgroundTasks
+from starlette.request import Request
+from starlette.responses import JSONResponse
+from starlette.routing import Route
+
+# Set number of threads in threadpool to 100
+anyio.to_thread.current_default_thread_limiter().total_tokens = 100
```
adriangb, line 17:
```suggestion
# Set number of threads in threadpool to 100
# Note that this increases the limit for anything using AnyIO, not just Starlette
anyio.to_thread.current_default_thread_limiter().total_tokens = 100
```

adriangb: Devil's advocate: if we document this would it then be breaking change if we start using our own capacity limiter(s), meaning this stops working? I think not, it's okay to break docs to some extent.

mkdocs.yml
adriangb: Agreed but can't think of a better name right now. @Kludex any ideas?

docs/threadpools.md
aminalaee: We can update the docs when we add the CapacityLimiter to explain this.

mkdocs.yml
Kludex: Concurrency?

Kludex: Concurrency.

docs/threadpools.md
```diff
@@ -0,0 +1,33 @@
...
```
Kludex, line 17: Doesn't this needs to be in the lifespan if using multiple workers?

aminalaee: I think each worker will do this, so there’s no difference.

but I can do a quick test to confirm

mkdocs.yml
agronholm: "Sync" threadpools, as opposed to what?

docs/threadpools.md
agronholm: Each worker runs in a separate process, right?

Kludex:
@aminalaee Do you still want to work on this?

Kludex:
I'm closing this as stale. Please reopen if you still want to work on it, or if someone else wants to take it over, go ahead.

PR #2301, opened 2023-10-14
Fix bug in routing when using Mount and "fixes" path expected by Uvicorn logger

nicolaipre:
# Summary

Solves a bug in Starlettes routing when using Mounts and a dirty fix to provide the full path to the Uvicorn logger for paths to be correctly logged. 

Solves the problems in discussion: https://github.com/encode/starlette/discussions/2299

# Checklist

### DIFF
--- a/starlette/convertors.py
+++ b/starlette/convertors.py
@@ -38,6 +38,16 @@ class PathConvertor(Convertor[str]):
         return str(value)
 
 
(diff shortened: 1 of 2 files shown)


nicolaipre:
Found another bug which probably is the reason behind the failed tests. Will update once fixed.

Kludex:
This should be fixed in Starlette 0.33.0.

PR #2316, opened 2023-10-31
Added `allow_path_regex` to the `SessionMiddleware`

hasansezertasan:
# Summary

SessionMiddleware updates (re assignes) session data even if it's not updated. This causes problems when we use a sub application and update that session data in it, in my case it's a Flask Application (which uses another way to sign session data) mounted to a Starlette Application.

With `allow_path_regex` I can restrict `SessionMiddleware` from doing what its doing for a given regex like: `'/api/.*'` or `'/admin/.*'` and partially solves #2018 

# Checklist

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -106,7 +106,7 @@ The following arguments are supported:
 * `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to `None` then the cookie will last as long as the browser session.
 * `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
 * `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to `False`.
-* `domain` - Domain of the cookie used to share cookie between subdomains or cross-domains. The browser defaults the domain to the same host that set the cookie, excluding subdomains [refrence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute). 
+* `domain` - Domain of the cookie used to share cookie between subdomains or cross-domains. The browser defaults the domain to the same host that set the cookie, excluding subdomains [refrence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute).
 
 
 ## HTTPSRedirectMiddleware
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -107,7 +107,7 @@ The following arguments are supported:
 * `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
 * `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to `False`.
 * `domain` - Domain of the cookie used to share cookie between subdomains or cross-domains. The browser defaults the domain to the same host that set the cookie, excluding subdomains [refrence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute).
-
+* `allow_path_regex`: A regex string to match against paths that should be permitted to use sessions. eg. `'/api/.*'`. Defaults to `None` (all routes).
 
 ## HTTPSRedirectMiddleware
 
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -45,7 +45,7 @@ class SessionMiddleware:
             return
         if self.allow_path_regex is not None and self.allow_path_regex.match(
             scope["path"]
-        ):
+        ):  # pragma: no cover
             await self.app(scope, receive, send)
             return
 
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -1,4 +1,5 @@
 import json
+import re
 import typing
 from base64 import b64decode, b64encode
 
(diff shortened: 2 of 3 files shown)


hasansezertasan:
In addition, I was thinking to open a PR that modifies the session data in a way that Flask does to sign it, I have referenced the thing here at #1922 but I believe it'll break some things for many users.

abersheeran:
We already have the ability to use middleware on a specific route. It seems unnecessary to add this parameter in Session Middleware. https://www.starlette.io/middleware/#applying-middleware-to-groups-of-routes

alex-oleshkevich:
@abersheeran yes, I agree and this is a way I do it.

hasansezertasan:
Thank you @alex-oleshkevich and @abersheeran for your honest reviews. 🙏 

This one is mostly inspired by `CORSMiddleware` which has a parameter called `allow_origin_regex`. I'm using Starlette with FastAPI and doing `.include_router()` to add routers and `.mount()` to add sub applications, which `.mount` skippes `middlewares` parameter.

https://github.com/encode/starlette/blob/966f0fc85764bd478d7c582fc8e2e5f91dbebae6/starlette/middleware/cors.py#L13-L24

> We already have the ability to use middleware on a specific route. It seems unnecessary to add this parameter in Session Middleware. https://www.starlette.io/middleware/#applying-middleware-to-groups-of-routes

I don't think it's the same. I don't want to use middleware on a specific route, I want it not to work on a specific route, I want to use middleware on application level but skip some routes.

On one hand you can add a middleware on application level and add one line to that middleware to skip some of the routes or path regex.

On the other hand you can add same middleware for 25 routes (which probably 50 to 75 lines of code) and skip the 5 routes you want your middleware to skip. (or maybe just one or two, you just want to skip two routes but you need to write 50 to 75 lines more now...) 😞 

I'm not insisting on PR to be merged but I believe if `CORSMiddleware` has this feature, `SessionMiddleware` also deserves it, since it keeps re-signing the session data even if its not updated. I'mean what's the point of "re-signing" a data if there is no change? 😕

alex-oleshkevich:
@hasansezertasan here is a code snippet that solves your need
```python
routes = [
   Route('/'), # no mw
   Mount('/api', routes[ # mounted route set w\o mw
        Route(), Route()
    ]),  
   Mount('/admin', # enable session only to routes starting with '/admin' 
      routes[ # mounted route set w\o mw
        Route(), Route()
     ],
    middleware=[Middleware(SessionMiddleware)]
  ),  
]
```

hasansezertasan:
> @hasansezertasan here is a code snippet that solves your need
> 
> ```python
> routes = [
>    Route('/'), # no mw
>    Mount('/api', routes[ # mounted route set w\o mw
>         Route(), Route()
>     ]),  
>    Mount('/admin', # enable session only to routes starting with '/admin' 
>       routes[ # mounted route set w\o mw
>         Route(), Route()
>      ],
>     middleware=[Middleware(SessionMiddleware)]
>   ),  
> ]
> ```

Thank you @alex-oleshkevich for your answer but that's not the case...

> I don't want to use middleware on a specific route, I want it not to work on a specific route, I want to use middleware on application level but skip some routes.

As I said...

Here is an example of what I need:

```python
from a2wsgi import WSGIMiddleware
from bottle import Bottle
from flask import Flask

flask_app = Flask(__name__)
botte_app = Bottle()


@flask_app.route("/")
def flask_hello_world():
    return "I'm a Flask app!"


@botte_app.route("/")
def bottle_hello_world():
    return "I'm a Bottle app!"


def starlette_hello_world(request):
    return PlainTextResponse("I'm a Starlette app!")


def starlette_dummy_route(request):
    return PlainTextResponse("Dummy Route!")


sapp = Starlette(
    routes=[
        Route(
            "/",
            starlette_hello_world,
        ),
        Route(
            "/route-a",
            starlette_dummy_route,
        ),
        Route(
            "/route-b",
            starlette_dummy_route,
        ),
        Route(
            "/route-c",
            starlette_dummy_route,
        ),
        Route(
            "/route-n",
            starlette_dummy_route,
        ),
        Mount(
            path="/flask-app",
            app=WSGIMiddleware(flask_app),
        ),
        Mount(
            path="/bottle-app",
            app=WSGIMiddleware(botte_app),
        ),
    ],
    middleware=[
        # App level middleware to apply to all routes
        Middleware(
            SessionMiddleware,
            secret_key="secret-key",
            https_only=False,
            # Allow any route but /flask-app and /bottle-app and /route-c
            allow_path_regex="^(?!/flask-app|/bottle-app|/route-c).*$",
        ),
    ],
)
```

Why? Because the way of session signing on Flask, Starlette and Bottle are different and there is a collision.

alex-oleshkevich:
why does this don't work?

```python
middleware = [
        Middleware(
            SessionMiddleware,
            secret_key="secret-key",
            https_only=False,
        ),
]
routes = [
		Mount('', 
   			routes=[
	        	Route(
	            	"/route-c",
		            starlette_dummy_route,
		        ),
			],
		    middleware=middleware,
	),
       Mount(
            path="/flask-app",
            app=WSGIMiddleware(flask_app),

        ),
        Mount(
            path="/bottle-app",
            app=WSGIMiddleware(botte_app),

        ),
]
```

alex-oleshkevich:
This is rather a question of route organization, not about Starlette.
As the last resort, you can extend SessionMiddleware and apply filtering in it.

PR #2328, opened 2023-11-06
Add RequestSizeLimitMiddleware and RequestTimeoutMiddleware

### LINKED ISSUE #2155: Limit max request size

adriangb:
Closes #2155

### DIFF
--- a/starlette/middleware/limits.py
+++ b/starlette/middleware/limits.py
@@ -1,0 +1,183 @@
+from __future__ import annotations
+
+from typing import ClassVar
+
+import anyio
+
+from starlette.exceptions import HTTPException
+from starlette.responses import PlainTextResponse
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+DEFAULT_MAX_REQUEST_SIZE = 2_621_440  # 2.5MB, same as Django (https://docs.djangoproject.com/en/1.11/ref/settings/#data-upload-max-memory-size)
+
+
+class _TooLarge(HTTPException):
+    msg: ClassVar[str]
+
+    def __init__(self, limit_bytes: int | None) -> None:
+        self.limit = limit_bytes
+        super().__init__(
+            status_code=413,
+            detail=(
+                self.msg + f" Max allowed size is {limit_bytes} bytes."
+                if limit_bytes
+                else self.msg
+            ),
+        )
+
+
+class RequestTooLarge(_TooLarge):
+    """The request body exceeded the configured limit."""
+
+    msg = "Request body is too large."
+
+
+class ChunkTooLarge(_TooLarge):
+    """A chunk exceeded the configured limit."""
+
+    msg = "Chunk size is too large."
+
+
+class RequestSizeLimitMiddleware:
+    def __init__(
+        self,
+        app: ASGIApp,
+        *,
+        max_request_size: int | None = DEFAULT_MAX_REQUEST_SIZE,
+        max_chunk_size: int | None = None,
+        include_limits_in_error_responses: bool = True,
+    ) -> None:
+        self.app = app
+        self.max_request_size = max_request_size
+        self.max_chunk_size = max_chunk_size
+        self.include_limits_in_error_responses = include_limits_in_error_responses
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        total_size = 0
+
+        async def rcv() -> Message:
+            nonlocal total_size
+            message = await receive()
+            chunk_size = len(message.get("body", b""))
+            if self.max_chunk_size is not None and chunk_size > self.max_chunk_size:
+                raise ChunkTooLarge(
+                    self.max_chunk_size
+                    if self.include_limits_in_error_responses
+                    else None
+                )
+            total_size += chunk_size
+            if self.max_request_size is not None and total_size > self.max_request_size:
+                raise RequestTooLarge(
+                    self.max_request_size
+                    if self.include_limits_in_error_responses
+                    else None
+                )
+            return message
+
+        await self.app(scope, rcv, send)
+
+
+class _Timeout(HTTPException):
+    msg: ClassVar[str]
+
+    def __init__(self, limit_seconds: float | None) -> None:
+        self.limit = limit_seconds
+        super().__init__(
+            status_code=408,
+            detail=(
+                self.msg + f" Max allowed time is {limit_seconds} seconds."
+                if limit_seconds
+                else self.msg
+            ),
+        )
+
+
+class ReceiveTimeout(_Timeout):
+    """The receive exceeded the configured limit."""
+
+    msg = "Client was too slow sending data."
+
+
+class SendTimeout(_Timeout):
+    """The send exceeded the configured limit."""
+
+    msg = "Client was too slow receiving data."
+
+
+class RequestTimeoutMiddleware:
+    def __init__(
+        self,
+        app: ASGIApp,
+        request_timeout_seconds: float | None = None,
+        receive_timeout_seconds: float | None = None,
+        send_timeout_seconds: float | None = None,
+        include_limits_in_error_responses: bool = True,
+    ) -> None:
+        self.app = app
+        self.timeout_seconds = request_timeout_seconds
+        self.receive_timeout_seconds = receive_timeout_seconds
+        self.send_timeout_seconds = send_timeout_seconds
+        self.include_limits_in_error_responses = include_limits_in_error_responses
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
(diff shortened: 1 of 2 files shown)


adriangb:
TODOs:
- Docs
- What do we want to do for websockets? Lifespans?

adriangb:
A couple of things to consider:
- Dealing with SSE requests
- Overriding instead of adding. As this PR currently stands you could set a 10s timeout for your entire app and a stricter timeout for individual routes, but not the other way around (a 1s timeout for all routes except some specific routes which get not timeout, e.g. because they stream data). To get that working I think we'd need to do the enforcing at the route level and have a middleware that simply sets some markers on the `scope` object to customize the limits. That would be more invasive of a change than this.

Kludex:
Both middlewares can be on a third party packages, and we can close the issue by just pointing at them.

Can we separate the two middlewares in different PRs? The initial discussion was about the request size one.

adriangb:
> Both middlewares can be on a third party packages, and we can close the issue by just pointing at them.

I disagree. It's an important feature to have readily available to all Starlette users, it doesn't require 3rd party packages, the complexity isn't massive and I believe we should make some limits the default in 1.0 like Django and other more mature frameworks do.

> Can we separate the two middlewares in different PRs? The initial discussion was about the request size one.

Yes I'm happy to do that if it'd make it easier to discuss.

alex-oleshkevich:
for reference: my approach which works on per-route basis
https://github.com/encode/starlette/pull/2175

alex-oleshkevich:
My main concern about middleware based approach is that it is very limited. Say, your global middleware will always have a large value. Because if just one route needs to handle 1gb uploads, the global limit will be 1gb. 
Per route limits does not have this flaw, however it is still possible to read unlimited data in custom middlewares. But per-route approach solves the majority of cases.

I see it as app-level limit (the global one that covers all middlewares also) + option to override per a route. In result, we can get this example setup: 8mb global and 24mb for `/user/photo-upload`.

adriangb:
Yeah, I brought that up above.

I'll rework this to satisfy that use case, although it will be a good chunk more invasive as I said above.

adriangb:
Also sorry I did not consider your PR, I had lost track / forgotten about it.

abersheeran:
I have a small question, why not just read the request content-length to limit the size.

adriangb:
Clients can easily lie about that

abersheeran:
> Clients can easily lie about that

But proxy server or asgi
 server should not receive data more than content-length?

adriangb:
I'm not saying we shouldn't also limit the length to the reported content length, I'm just saying that it doesn't serve a security purpose.

abersheeran:
> I'm not saying we shouldn't also limit the length to the reported content length, I'm just saying that it doesn't serve a security purpose.

Maybe my previous words caused some misunderstanding. What I meant was: the forward reverse proxy server or ASGI server will not pass data to the ASGI application that exceeds the content-length length. Therefore, forging a content-length length on the client side will only result in truncation when the server receives data.
For example, h11 used by Uvicorn will check whether the Content-Length in the Reader matches the actual size of the received data.

adriangb:
Then I’m confused as to what you are suggesting: are you saying that we should also enforce that limit, that that limit is already enforced elsewhere, and this PR (or #2174) is not needed, or something else?

starlette/middleware/limits.py
```diff
@@ -0,0 +1,183 @@
...
+                    else None
+                )
+            total_size += chunk_size
+            if self.max_request_size is not None and total_size > self.max_request_size:
+                raise RequestTooLarge(
+                    self.max_request_size
+                    if self.include_limits_in_error_responses
+                    else None
+                )
+            return message
```
abersheeran, line 79: > Then I’m confused as to what you are suggesting: are you saying that we should also enforce that limit, that that limit is already enforced elsewhere, and this PR (or #2174) is not needed, or something else?

Would it be more concise and efficient to change here to judge the request content-length?

adriangb: Instead of the user-defined limits? Or in addition to?

Kludex: What @abersheeran meant is to check the `Content-Length` instead of having the logic of adding up `chunk_size`s - which is actually what Django does: https://github.com/django/django/blob/6daf86058bb6fb922eb3fe3abae6f5c0e645020c/django/http/request.py#L323-L347.

Django also has this logic on the multipart parser: https://github.com/django/django/blob/594873befbbec13a2d9a048a361757dd3cf178da/django/http/multipartparser.py#L241-L248.

Kludex:
> What do we want to do for websockets?

1. Tornado WebSockets have a limitation by default.
    1.  https://www.tornadoweb.org/en/stable/releases/v4.5.0.html
    2. https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler
2. Uvicorn already limits the size of each WebSocket message via `—ws-max-size` configuration - it only applies to websockets.
    1.  https://www.uvicorn.org/settings/#implementation.

I don't think WebSockets should be a point of concern here.

> Lifespans?

Nothing to do for lifespans.

> Dealing with SSE requests

No special logic is required for SSE.

## Notes/Questions

1. Is the argumentation of having resource limits on the ASGI application, and not ASGI server because you can apply on individual endpoints/have more granularity?
2. We need to check the `Content-Length` (single body event) on the requests that are not `Transfer-Encoding: chunked` (streaming).

adriangb:
> I don't think WebSockets should be a point of concern here.

The configurations offered by Uvicorn don't cover all of the use cases here. You can't configure it on a per route basis for starters.

> No special logic is required for SSE.

What I mean is that if you install a blanket timeout or request size limit it might make sense to say "SSE requests don't have a limit". That would require special logic.

> Is the argumentation of having resource limits on the ASGI application, and not ASGI server because you can apply on individual endpoints/have more granularity?

Yes.

> We need to check the Content-Length (single body event) on the requests that are not Transfer-Encoding: chunked (streaming).

Agreed, we can check it and if it is larger than the limit error immediately without streaming anything. But we can't "trust" it if it's smaller.

Kludex:
We can trust the content length header. The server errors otherwise.

adriangb:
But Starlette doesn't depend on the server. We shouldn't make assumptions like this about the server's behavior unless it's part of the ASGI spec, which to my knowledge this is not. Besides, the only advantage of trusting the `content-length` header when it says it's smaller than the user defined limit is that we do one less ASGI middleware wrapping on `receive`. It's not expensive to keep track of the actual streamed size, I don't think it justifies the potential footgun.

adriangb:
@Kludex I’m not sure if it was on purpose or not but I noticed you closed #2175 and not this one. I don’t think there was any concensus on how to implement this feature, nor do I think this PR is intrinsically better in significant ways, so I just wanted to check what your intention was.

ofer-pd:
It's a shame the momentum on this issue has dissipated.  I need this feature for my current project.

In the meantime, I guess my best option is to review the code in PRs #2175, #2328, and #2350 and cobble together my solution from the bits that make the most sense to me, or try out this project that looks promising but is five-years-stale:  https://github.com/steinnes/content-size-limit-asgi

---

**UPDATE:**  The code from the "content-size-limit-asgi" repo worked great.  I copied it into my project and restyled it to my taste.  I also used the recommended modern way to add ASGI middleware to Starlette:

```python
from mylib import ContentSizeLimitMiddleware

middleware = [
    Middleware( ContentSizeLimitMiddleware, limit = 12_000_000 ),
]

app = Starlette( routes = routes, middleware = middleware )
```

Kludex:
I think this feature should be on the server side.

adriangb:
> I think this feature should be on the server side.

The issue with doing this at the server level, and the only reason it's not trivial to do at the application level, is that you'll often want some routes (e.g. a file upload route) to have a large limit (10MB / 5s) and other routes (e.g. a CRUD route) to have a low limit (100kB / 500ms).

Kludex:
Starlette now supports request body size limits through `RequestBodyLimitMiddleware` and the `max_body_size` parameter on `Starlette`, `Route`, `Router`, and `Mount`, added in [#3431](https://github.com/Kludex/starlette/pull/3431).

About the timeout, I think that should be handled by the ASGI server, not the ASGI framework.

---

Sorry the delay Adrian 😞

PR #2329, opened 2023-11-07
Add option `timeout` to `WebSocketTestSession::receive`

mbergen:
# Summary

Allows to pass a timeout to various receive methods of `WebSocketTestSession` which is passed into `Queue::get`.
This allows to write tests that raise an exception (which can be expected as well) instead of blocking forever in case the tested app does not send a message on the websocket.

It is a potential fix for #2195, but allows finer control about which call to receive should have a timeout, and which should block instead of setting a generic timeout option to the Session.

# Checklist

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -156,6 +156,7 @@ May raise `starlette.websockets.WebSocketDisconnect` if the application does not
 * `.receive_bytes()` - Wait for incoming bytestring sent by the application and return it.
 * `.receive_json(mode="text")` - Wait for incoming json data sent by the application and return it. Use `mode="binary"` to receive JSON over binary data frames.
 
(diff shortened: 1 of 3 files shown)


Kludex:
Thanks for the PR, but this was already rejected on: https://github.com/encode/starlette/pull/1140.

mbergen:
> Thanks for the PR, but this was already rejected on: #1140.

Sorry, that PR didn't show up in my search (only the referenced issue/discussion).
Nontheless, the contributors of that PR as well as I don't think the reasoning for closing that PR doesn't really address the issue we experience, having an option like timeout would improve test development a lot.
Could you elaborate why you are against it? Do you want to support every parameter of the underlying `receive` before considering it?

nextmat:
I agree. This would be very helpful, I don't really understand the resistance to allowing it as an option?

When refactoring a failing test with details about the failure is generally preferable to the test suite hanging indefinitely.

PR #2330, opened 2023-11-08
Replace black with ruff formatter

aminalaee:
# Summary

Ruff stable release supports both formatter and linter which is compatible with `Black` and can replace it the same way it did for `isort`.
Open to ideas and concerns.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,10 +51,10 @@ Source = "https://github.com/encode/starlette"
 [tool.hatch.version]
 path = "starlette/__init__.py"
 
-[tool.ruff]
+[tool.ruff.lint]
 select = ["E", "F", "I"]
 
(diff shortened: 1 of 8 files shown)


aminalaee:
List of `Black` deviations: https://github.com/astral-sh/ruff/blob/main/docs/formatter/black.md

Kludex (changes requested):
Please set line limit to 88 to match black.

aminalaee:
@Kludex I think that's the case by default, the difference is how `ruff` ignores pragma comments when computing line length: https://github.com/astral-sh/ruff/blob/main/docs/formatter/black.md#pragma-comments-are-ignored-when-computing-line-width

Kludex (approved):

PR #2334, opened 2023-11-12
Take weak ETags in consideration on `StaticFiles`

### LINKED ISSUE #2298: Strong and weak etags for StaticFiles
### Discussed in https://github.com/encode/starlette/discussions/2297

<sup>Originally posted by **markus1978** October  6, 2023</sup>
tl;tr: It seems that `starlette.staticfiles.StaticFiles` does not use "proper" HTTP etags which causes issues with reverse proxies and browsers.

The `starlette.staticfiles.StaticFiles` class supports etags, e.g. it creates a `Etag` header in all responses and considers `If-None-Match` headers in its `is_not_modified` method. However, it only puts the plain hash values in the response `Etag` header and also only interprets `If-None-Match` values as plain hash values . HTTP Etags are either weak or strong and the values are supposed to be wrapped in `W/"..."` or `"..."` respectively (https://datatracker.ietf.org/doc/html/rfc7232#section-2.3.1).

This causes two problems for me. First, nginx as a reverse proxy is considering all starlette etag headers as weak etags and the gzip module removes. Secondly, when the browser sends a strong or weak etag in the `If-None-Match` header, it never matches because `'...some-tag...' != '"...some-tag..."'`. 

I currently use this work arround  to solve my issues:
```python
class StaticFiles(StarletteStaticFiles):

    etag_re = r'^(W/)?"?([^"]*)"?$'

    def is_not_modified(
        self, response_headers: Headers, request_headers: Headers
    ) -> bool:
        try:
            if_none_match = request_headers["if-none-match"]
            # I remove all weak/strong etag syntax, basically doing a weak match 
            match = re.match(StaticFiles.etag_re, if_none_match)
            if_none_match = match.group(2)
            etag = response_headers["etag"]
            if if_none_match == etag:
                return True
        except KeyError:
            pass

        return super().is_not_modified(response_headers, request_headers)

    def file_response(
        self, full_path: PathLike, stat_result: os.stat_result, scope: Scope, 
        status_code: int = 200,
    ) -> Response:
        response = super().file_response(full_path, stat_result, scope, status_code)
        # I add strong etag syntax, basically wrapping the etag value in "..."
        if 'etag' in response.headers:
            response.headers['etag'] = f'"{response.headers.get("etag")}"'
        return response
```

I am not an expert in etags, just trying to make it work in our app. I am glad for any hints, if i am getting it wrong or missing anything here.

For reference, the relevant part from the current starlette.
From `staticfiles.py::StaticFiles`
```python
    def is_not_modified(
        self, response_headers: Headers, request_headers: Headers
    ) -> bool:
        """
        Given the request and response headers, return `True` if an HTTP
        "Not Modified" response could be returned instead.
        """
        try:
            if_none_match = request_headers["if-none-match"]
            etag = response_headers["etag"]
            if if_none_match == etag:
                return True
        except KeyError:
            pass

        try:
            if_modified_since = parsedate(request_headers["if-modified-since"])
            last_modified = parsedate(response_headers["last-modified"])
            if (
                if_modified_since is not None
                and last_modified is not None
                and if_modified_since >= last_modified
            ):
                return True
        except KeyError:
            pass

        return False
```

From `responses.py::FileResponse`
```python
    def set_stat_headers(self, stat_result: os.stat_result) -> None:
        content_length = str(stat_result.st_size)
        last_modified = formatdate(stat_result.st_mtime, usegmt=True)
        etag = md5_hexdigest(etag_base.encode(), usedforsecurity=False)

        self.headers.setdefault("content-length", content_length)
        self.headers.setdefault("last-modified", last_modified)
        self.headers.setdefault("etag", etag)
```</div>

<!-- POLAR PLEDGE BADGE START -->
> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/2298">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/2298/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/encode/starlette/issues/2298/pledge.svg">
</picture>
</a>
<!-- POLAR PLEDGE BADGE END -->

### PR DESCRIPTION
**Pull request by neolooong**
# Summary

fix #2298.

# Checklist

### CODE REVIEW

File: starlette/staticfiles.py [hunk 1]
```diff
@@ -226,8 +228,9 @@ def is_not_modified(
         try:
             if_none_match = request_headers["if-none-match"]
             etag = response_headers["etag"]
-            if if_none_match == etag:
-                return True
+            for match in ETAG_PATTERN.finditer(if_none_match):
+                if match.groups()[1] == etag:
+                    return True
```
karpetrosyan: Looks good to me, though I prefer to avoid regexes where they aren't required, so we can possibly simplify this part.

```suggestion
            if etag in [tag.strip(r" W/") for tag in if_none_match.split(",")]:
                return True
```

neolooong: Good suggestion. This looks like more simplify. May I know why this string prefix with `r`. Is it to have some special meaning here?

karpetrosyan: It's a typo, sorry :smile:

karpetrosyan (approved):

starlette/staticfiles.py
```diff
@@ -229,7 +229,7 @@ def is_not_modified(
         try:
             if_none_match = request_headers["if-none-match"]
             etag = response_headers["etag"]
-            if if_none_match == etag:
+            if etag in [text.strip(" W/") for text in if_none_match.split(",")]:
```
Kludex:
```suggestion
            if etag in [tag.strip(" W/") for tag in if_none_match.split(",")]:
```

PR #2341, opened 2023-11-22
Set `ensure_ascii=False` on `json.dumps()` for `WebSocket.send_json()`

xiayouran:
# Summary

When the data output by the `send_json` function is in Chinese, it will force Chinese characters to be converted to Unicode, so I fixed this issue.(The most fundamental reason is that the value of the `ensure_ascii` parameter was not modified when using the `json.dumps` function)

# Checklist

- [√] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [√] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [√] I've updated the documentation accordingly.

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -147,7 +147,7 @@ class WebSocketTestSession:
 
     def send_json(self, data: typing.Any, mode: str = "text") -> None:
         assert mode in ["text", "binary"]
-        text = json.dumps(data, separators=(",", ":"))
+        text = json.dumps(data, separators=(",", ":"), ensure_ascii=False)
         if mode == "text":
             self.send({"type": "websocket.receive", "text": text})
         else:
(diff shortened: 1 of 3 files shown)


Kludex:
I think we have some issues about this around. Can we link them here?

abersheeran:
I'm not sure what problem you encountered that requires adding this option. Sending Chinese should still work fine without using this option. Can you provide a more detailed description of what kind of problem this solves? (Converting Chinese to Unicode encoding is not a problem, it is standard-compliant.)

xiayouran:
Yes, from the perspective of the code workflow, there is no problem. I am building a Chinese question and answer system, but when using `send_json` function sends Chinese characters, my front-end display is Unicode characters instead of Chinese characters, so I want to solve this problem. The problem lies in the fact that the `json.dumps` function did not set the `ensure_ascii`  parameter to `False`.

abersheeran (approved):
I verified it, and it was

Kludex (changes requested):
Can we add a test case?

Kludex (approved):
Yeah, according to the spec (https://asgi.readthedocs.io/en/latest/specs/www.html#send-send-event), there was no need to `ensure_ascii` here. 👍

PR #2350, opened 2023-11-27
Add `LimitBodySizeMiddleware`

### LINKED ISSUE #2155: Limit max request size

Kludex:
- Alternative to #2328.
- Closes #2155.

## Q&A

1. Why not also limit chunks?

Because the chunks the application receive are not the chunks the server is receiving from the client.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -183,7 +183,7 @@ The following arguments are supported:
 
 The middleware won't GZip responses that already have a `Content-Encoding` set, to prevent them from being encoded twice.
 
-## LimitRequestMiddleware
+## LimitBodySizeMiddleware
 
 Limits the body size of incoming requests. If the incoming request has a body
 larger than the limit, then a `413 Content Too Large` response will be sent.
@@ -191,13 +191,13 @@ larger than the limit, then a `413 Content Too Large` response will be sent.
 ```python
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
-from starlette.middleware.limits import LimitRequestMiddleware
+from starlette.middleware.limits import LimitBodySizeMiddleware
 
 
 routes = ...
 
 middleware = [
-    Middleware(LimitRequestMiddleware, max_body_size=1024)
+    Middleware(LimitBodySizeMiddleware, max_body_size=1024)
 ]
 
 app = Starlette(routes=routes, middleware=middleware)
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -106,8 +106,7 @@ The following arguments are supported:
 * `max_age` - Session expiry time in seconds. Defaults to 2 weeks. If set to `None` then the cookie will last as long as the browser session.
 * `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
 * `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to `False`.
-* `domain` - Domain of the cookie used to share cookie between subdomains or cross-domains. The browser defaults the domain to the same host that set the cookie, excluding subdomains [refrence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute). 
-
+* `domain` - Domain of the cookie used to share cookie between subdomains or cross-domains. The browser defaults the domain to the same host that set the cookie, excluding subdomains [refrence](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#domain_attribute).
 
 ## HTTPSRedirectMiddleware
 
@@ -183,6 +182,31 @@ The following arguments are supported:
 * `minimum_size` - Do not GZip responses that are smaller than this minimum size in bytes. Defaults to `500`.
 
 The middleware won't GZip responses that already have a `Content-Encoding` set, to prevent them from being encoded twice.
+
+## LimitRequestMiddleware
+
+Limits the body size of incoming requests. If the incoming request has a body
+larger than the limit, then a `413 Content Too Large` response will be sent.
+
+```python
+from starlette.applications import Starlette
+from starlette.middleware import Middleware
+from starlette.middleware.limits import LimitRequestMiddleware
+
+
+routes = ...
+
+middleware = [
+    Middleware(LimitRequestMiddleware, max_body_size=1024)
+]
+
+app = Starlette(routes=routes, middleware=middleware)
+```
+
+The following arguments are supported:
+
+* `max_body_size` - Send a "413 - Content Too Large" on requests that surpass the maximum allowed body size.
+  Defaults to `2.5 * 1024 * 1024` bytes (2.5MB).
 
 ## BaseHTTPMiddleware
 
@@ -573,7 +597,7 @@ import time
 class MonitoringMiddleware:
     def __init__(self, app):
         self.app = app
-    
+
     async def __call__(self, scope, receive, send):
         start = time.time()
         try:
--- a/starlette/middleware/limits.py
+++ b/starlette/middleware/limits.py
@@ -1,0 +1,66 @@
+"""Middleware that limits the body size of incoming requests."""
+from starlette.datastructures import Headers
+from starlette.responses import PlainTextResponse
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+DEFAULT_MAX_BODY_SIZE = 2_621_440  # 2.5MB
+
+
+class ContentTooLarge(Exception):
+    def __init__(self, max_body_size: int) -> None:
+        self.max_body_size = max_body_size
+
+
+class LimitRequestMiddleware:
+    def __init__(
+        self, app: ASGIApp, max_body_size: int = DEFAULT_MAX_BODY_SIZE
+    ) -> None:
+        self.app = app
+        self.max_body_size = max_body_size
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":  # pragma: no cover
+            return await self.app(scope, receive, send)
+
+        headers = Headers(scope=scope)
+        content_length = headers.get("content-length")
+        if content_length is not None:
+            if int(content_length) > self.max_body_size:
+                return await _content_too_large_app(scope)(scope, receive, send)
+
+            # NOTE: The server makes sure that the content-length header sent by the
+            #   client is the same as the length of the body.
+            #   Ref.: https://github.com/django/asgiref/issues/422
+            return await self.app(scope, receive, send)
+
+        total_size = 0
+        response_started = False
+
+        async def wrap_send(message: Message) -> None:
+            nonlocal response_started
+            if message["type"] == "http.response.start":
+                response_started = True
+            await send(message)
+
+        async def wrap_receive() -> Message:
+            nonlocal total_size
+            message = await receive()
+            if message["type"] == "http.request":
+                chunk_size = len(message["body"])
+                total_size += chunk_size
+                if total_size > self.max_body_size:
+                    raise ContentTooLarge(self.max_body_size)
+            return message
+
+        try:
+            await self.app(scope, wrap_receive, wrap_send)
+        except ContentTooLarge as exc:
+            # NOTE: If response has already started, we can't return a 413, because the
+            #   headers have already been sent.
+            if not response_started:
+                return await _content_too_large_app(scope)(scope, receive, send)
+            raise exc
+
+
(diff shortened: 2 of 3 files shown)


starlette/middleware/limits.py
```diff
@@ -0,0 +1,66 @@
...
+            #   Ref.: https://github.com/django/asgiref/issues/422
+            return await self.app(scope, receive, send)
+
+        total_size = 0
+        response_started = False
+
+        async def wrap_send(message: Message) -> None:
+            nonlocal response_started
+            if message["type"] == "http.response.start":
+                response_started = True
```
Kludex, line 46: If the response has started, the status code was already sent. We don't want to try to send another headers, that's useless.

In this case, we'll just raise the exception.

starlette/middleware/limits.py
```diff
@@ -0,0 +1,66 @@
...
+        except ContentTooLarge as exc:
+            # NOTE: If response has already started, we can't return a 413, because the
+            #   headers have already been sent.
+            if not response_started:
+                return await _content_too_large_app(scope)(scope, receive, send)
+            raise exc
+
+
+def _content_too_large_app(scope: Scope) -> PlainTextResponse:
+    return PlainTextResponse("Content Too Large", status_code=413)
```
Kludex: This is not needed. When I started, I thought there will be more logic here. I can remove it.

The body, and the status code match the spec.

starlette/middleware/limits.py
```diff
@@ -0,0 +1,66 @@
...
+        headers = Headers(scope=scope)
+        content_length = headers.get("content-length")
+        if content_length is not None:
+            if int(content_length) > self.max_body_size:
+                return await _content_too_large_app(scope)(scope, receive, send)
+
+            # NOTE: The server makes sure that the content-length header sent by the
+            #   client is the same as the length of the body.
+            #   Ref.: https://github.com/django/asgiref/issues/422
+            return await self.app(scope, receive, send)
```
Kludex: I still need to do more thinking/research here. I don't think it makes sense for the middleware to do this validation.

On the other hand... Is just 1 conditional...

starlette/middleware/limits.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class ContentTooLarge(Exception):
+    def __init__(self, max_body_size: int) -> None:
+        self.max_body_size = max_body_size
+
+
+class LimitRequestMiddleware:
+    def __init__(
+        self, app: ASGIApp, max_body_size: int = DEFAULT_MAX_BODY_SIZE
```
Kludex: It's called `max_body_size` and not `max_request_size` because the request also contain headers, and the request line.

starlette/middleware/limits.py
```diff
@@ -0,0 +1,66 @@
...
+from starlette.datastructures import Headers
+from starlette.responses import PlainTextResponse
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+DEFAULT_MAX_BODY_SIZE = 2_621_440  # 2.5MB
+
+
+class ContentTooLarge(Exception):
+    def __init__(self, max_body_size: int) -> None:
+        self.max_body_size = max_body_size
```
Kludex, line 12: This shouldn't be an `HTTPException`. The `user_middleware`s run after the `ExceptionMiddleware`.

But also, I think it should be just this, as we do in other parts of the code.

alex-oleshkevich:
Can user set a low body size globally and enlarge it per route? It seems to me that app-level middleware returns 413 response as fast as limit reached.

```python
Starlette(
	middleware=[
		Middleware(LimitRequestMiddleware, max_body_size=1024) # 1024b globally
	],
    routes=[
        Route('/'), # it's ok to use a global limit here
		Route('/upload', middleware=[
			Middleware(LimitRequestMiddleware, max_body_size=1024 ** 3 ), # i want a greater limit here
		])
    ],
)
```

Kludex:
> Can user set a low body size globally and enlarge it per route? It seems to me that app-level middleware returns 413 response as fast as limit reached.
> 
> ```python
> Starlette(
> 	middleware=[
> 		Middleware(LimitRequestMiddleware, max_body_size=1024) # 1024b globally
> 	],
>     routes=[
>         Route('/'), # it's ok to use a global limit here
> 		Route('/upload', middleware=[
> 			Middleware(LimitRequestMiddleware, max_body_size=1024 ** 3 ), # i want a greater limit here
> 		])
>     ],
> )
> ```

Yes. That should be possible. I'm going to add a test for it.

docs/middleware.md
```diff
@@ -184,6 +183,31 @@ The following arguments are supported:
 
 The middleware won't GZip responses that already have a `Content-Encoding` set, to prevent them from being encoded twice.
 
+## LimitRequestMiddleware
```
adriangb: I’d put Size in the name

docs/middleware.md
```diff
@@ -184,6 +183,31 @@ The following arguments are supported:
...
+middleware = [
+    Middleware(LimitRequestMiddleware, max_body_size=1024)
+]
+
+app = Starlette(routes=routes, middleware=middleware)
+```
+
+The following arguments are supported:
+
+* `max_body_size` - Send a "413 - Content Too Large" on requests that surpass the maximum allowed body size.
```
adriangb, line 208: Are you skipping a limit on chunk sizes in favor of setting that on the server?

starlette/middleware/limits.py
adriangb: Are you sure? There’s currently two places where exceptions are caught. Once in the route and once at the app middleware level. I think we’ll remove the latter eventually.

starlette/middleware/limits.py
adriangb: Doesn’t hurt to do!

adriangb:
> Yes. That should be possible. I'm going to add a test for it.

I think that to make have to split it into two middlewares and pass the max size around in the scope like #2175 does

docs/middleware.md
Kludex: The reason for this is written in the PR description:

> 1. Why not also limit chunks?
> Because the "chunks" the application receive are not the chunks the server is receiving from the client.

starlette/middleware/limits.py
Kludex: Right. What I said doesn't make sense.

I still don't want to use the `HTTPException`, because we'll need to check if `app` is in the scope in any case.

Kludex:
Ah sorry, I thought you meant to increase the constraint, not to relax it. I don't think we should allow that.

One could always mount the applications as they want.

starlette/middleware/limits.py
Kludex: Done.

docs/middleware.md
Kludex: Yeah, my thoughts were to make it possible to add more stuff on this middleware, but I've added the "Size".

Kludex: Ah no. It can't be `RequestSize`... Either `LimitRequestBodySizeMiddleware`, or `LimitRequestMiddleware`. Or anything that says "payload" or "body".

adriangb:
> Ah sorry, I thought you meant to increase the constraint, not to relax it. I don't think we should allow that.
> 
> One could always mount the applications as they want.

I disagree and I think @alex-oleshkevich will as well. Small limits by default and larger limits on certain endpoints seems like a very important use case.

docs/middleware.md
adriangb: I like `LimitRequestBodySizeMiddleware`

starlette/middleware/limits.py
adriangb: > I still don't want to use the HTTPException, because we'll need to check if app is in the scope in any case.

Why? My idea was that one of two things would happen:
- `BodySizeLimitMiddleware -> ExceptionMiddleware -> App`: `App` raises a `ContentTooLarge` exception. That gets caught by `ExceptionMiddleware` which returns the 413 status code and message. `BodySizeLimitMiddleware` does nothing and never catches `ContentTooLarge`. This would be the scenario where you wrap an individual route.
- `BodySizeLimitMiddleware -> SomeOtherMiddleware -> ExceptionMiddleware -> App`. Where `SomeOtherMiddleware` might read the body (e.g. `BaseHTTPMiddleware`). In this case, there may be no `ExceptionMiddleware` to catch the exception (currently there is but as per above we may remove it in the future, also we shouldn't make assumptions about other middleware) so `BodySizeLimitMiddleware` itself catches the exception and returns the 413 response. This is the scenario where an entire app or router is wrapped, or the middleware is being used outside of a Starlette context.

docs/middleware.md
adriangb: > Yeah, my thoughts were to make it possible to add more stuff on this middleware, but I've added the "Size".

I'd rather have more smaller middleware than one big one

starlette/middleware/limits.py
Kludex: I was wrong on my last message. It does need the `if 'app' in scope:`.

Kludex: I'll push the code to show you.

Kludex: This is the commit: https://github.com/encode/starlette/pull/2350/commits/c4e7aad4c5beae118890348af0fcb9d434ef7368

I'll remove the `HTTPException` inheritance.

starlette/middleware/limits.py
adriangb: To be clear, I think the advantage of this is that if the client says the body is 1GB and our limit is 1MB we can return an error without reading 1MB of the body into memory. If the client says the body is 1KB we should still verify that's the case. So it's a fail-fast path.

starlette/middleware/limits.py
adriangb: Look as long as it works both with and without `ExceptionMiddleware` I'm happy

Kludex:
> > Ah sorry, I thought you meant to increase the constraint, not to relax it. I don't think we should allow that.
> > One could always mount the applications as they want.
> 
> I disagree and I think @alex-oleshkevich will as well. Small limits by default and larger limits on certain endpoints seems like a very important use case.

I've complied with this.

Kludex:
I feel like this middleware will give the false impression of security. The moment the application calls `receive()` is when we are going to do the appropriate checks on the middleware. The thing is that by that moment, the server could already have stored the whole request body in memory.

`receive()` just gets what is already in memory.

I think a more correct approach would be to restrict this from the server side, and allow customizing per path e.g.:
`uvicorn.run(app, limit_max_body_size=2*MB, limit_max_body_size_per_path={'/potato': 3*MB})`.

alex-oleshkevich:
IMO servers should (and do?) stream data and not buffer it for too long. So server is also not aware of the payload size until EOF.

abersheeran:
> I feel like this middleware will give the false impression of security. The moment the application calls is when we are going to do the appropriate checks on the middleware. The thing is that by that moment, the server could already have stored the whole request body in memory.`receive()`
> 
> `receive()` just gets what is already in memory.
> 
> I think a more correct approach would be to restrict this from the server side, and allow customizing per path e.g.: .`uvicorn.run(app, limit_max_body_size=2*MB, limit_max_body_size_per_path={'/potato': 3*MB})`

I agree this. Add limit body size config in uvicorn/nginx is better.

Kludex:
After a lot of thinking, I don't think Starlette should have this level of granularity on those limits. I can be convinced on the global level, but not on _per endpoint_ basis. Besides, there's no record of a popular web framework doing this per basis (unless you can prove me wrong?).

We can document that any ASGI web server (Uvicorn and Gunicorn) has global limits set by default.

Also, I do believe it's a nice contribution to have this middleware implemented as a third party. @adriangb Since the last commit is pretty much your implementation, do you mind if I create this package, or do you want to create it yourself?

PR #2351, opened 2023-11-28
Add middleware per `Router`

Kludex:


starlette/routing.py
```diff
@@ -650,6 +652,11 @@ def __init__(
         else:
             self.lifespan_context = lifespan
 
+        self.middleware_stack = self.app
```
Kludex, line 673: I know you'll comment about the names, but I prefer them.

iudeen (approved):
Looks good.

Do we need to update docs?

Kludex:
@encode/maintainers someone else wants to review this?

starlette/routing.py
```diff
@@ -615,6 +615,8 @@ def __init__(
         # the generic to Lifespan[AppType] is the type of the top level application
         # which the router cannot know statically, so we use typing.Any
         lifespan: typing.Optional[Lifespan[typing.Any]] = None,
+        *,
+        middleware: typing.Optional[typing.Sequence[Middleware]] = None,
```
abersheeran, line 627: Maybe use name `middlewares`?

Kludex: But we use `middleware` everywhere already...

abersheeran: 😂 Oh, I didn't even notice that.

abersheeran (approved):
Simple and elegant modification

PR #2352, opened 2023-11-29
Do not overwrite `"path"` and `"root_path"` scope keys

### LINKED ISSUE #1336: Handling of `scope["root_path"]` and `scope["path"]` differs from ASGI spec

Kludex:
- Closes #1336
- Completes #1774

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -252,7 +252,11 @@ class Route(BaseRoute):
 
     def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
         if scope["type"] == "http":
-            match = self.path_regex.match(scope["path"])
+            root_path = scope.get("current_root_path", scope.get("root_path", ""))
+            path = scope.get(
+                "current_path", re.sub(r"^" + root_path, "", scope["path"])
+            )
+            match = self.path_regex.match(path)
             if match:
                 matched_params = match.groupdict()
                 for key, value in matched_params.items():
@@ -339,7 +343,11 @@ class WebSocketRoute(BaseRoute):
 
     def matches(self, scope: Scope) -> typing.Tuple[Match, Scope]:
         if scope["type"] == "websocket":
-            match = self.path_regex.match(scope["path"])
+            root_path = scope.get("current_root_path", scope.get("root_path", ""))
+            path = scope.get(
+                "current_path", re.sub(r"^" + root_path, "", scope["path"])
+            )
+            match = self.path_regex.match(path)
             if match:
                 matched_params = match.groupdict()
                 for key, value in matched_params.items():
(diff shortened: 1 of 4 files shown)


Kludex:
I'm going to try to make this PR cleaner, but it already solves the problem.

starlette/routing.py
```diff
@@ -402,21 +410,23 @@ def routes(self) -> typing.List[BaseRoute]:
...
-                matched_path = path[: -len(remaining_path)]
+                matched_path = current_path[: -len(remaining_path)]
                 path_params = dict(scope.get("path_params", {}))
                 path_params.update(matched_params)
                 root_path = scope.get("root_path", "")
                 child_scope = {
                     "path_params": path_params,
                     "app_root_path": scope.get("app_root_path", root_path),
-                    "root_path": root_path + matched_path,
-                    "path": remaining_path,
```
frankie567, line 429: Don't we need to keep the original `root_path` and `path` in the child scope?

frankie567 (commented):
Apart from my question above, LGTM! You have filled the holes I left in my previous PR :) 

Still not convinced about the naming `current_root_path` and `current_path`, but that's minor 😄 Suggestion (not ideal): `route_root_path` / `route_path`.

starlette/routing.py
Kludex: Aren't they always there anyway? Those lines were just modifying them, no?

frankie567: Yes, you're right. My point was if we were using this `child_scope` as the new scope for inner routers, but it seems actually that it's always merged with the root `scope`.

frankie567 (approved):

Kludex:
@encode/maintainers can someone review this, please?

abersheeran (approved):
It looks very nice, I like the regular expressions here.

PR #2362, opened 2023-12-06
Use `Iterable` instead `Iterator` on `iterate_in_threadpool`

jamesbraza:
# Summary

Having `AsyncContentStream` match typing of `SyncContentStream`, they should both be `Iterator` or `Iterable`, not separate.

~Given `iterate_in_threadpool` is typed with `Iterator`, and returns an `AsyncIterator`, I think then `AsyncContentStream` should be an `Iterator` as well~

From `mypy` inferred types in `test_streaming_response_custom_iterable`, it seems `AsyncIterable` is required. So, this PR standardizes on `Iterable`/`AsyncIterable`

# Checklist

### DIFF
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -55,8 +55,8 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterator: typing.Iterator[T],
-) -> typing.AsyncIterator[T]:
+    iterator: typing.Iterable[T],
+) -> typing.AsyncIterable[T]:
     while True:
         try:
             yield await anyio.to_thread.run_sync(_next, iterator)
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -55,8 +55,9 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterator: typing.Iterable[T],
+    iterable: typing.Iterable[T],
 ) -> typing.AsyncIterable[T]:
+    iterator = iter(iterable)
     while True:
         try:
             yield await anyio.to_thread.run_sync(_next, iterator)
--- a/starlette/concurrency.py
+++ b/starlette/concurrency.py
@@ -55,11 +55,11 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterable: typing.Iterable[T],
-) -> typing.AsyncIterable[T]:
-    iterator = iter(iterable)
+    iterator: typing.Iterable[T],
+) -> typing.AsyncIterator[T]:
+    as_iterator = iter(iterator)
     while True:
         try:
-            yield await anyio.to_thread.run_sync(_next, iterator)
+            yield await anyio.to_thread.run_sync(_next, as_iterator)
         except _StopIteration:
             break
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -220,7 +220,7 @@ class StreamingResponse(Response):
         media_type: typing.Optional[str] = None,
         background: typing.Optional[BackgroundTask] = None,
     ) -> None:
-        if isinstance(content, typing.AsyncIterable):
+        if isinstance(content, typing.AsyncIterator):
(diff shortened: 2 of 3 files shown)


jamesbraza:
Okay, nevermind, per type checks looks like `iterate_in_threadpool` and everything else should be updated to `AsyncIterable`. Will revisit this tomorrow

jamesbraza:
Cc @tomchristie if you got time to take a look at this one, it's another bug I encountered recently

starlette/concurrency.py
```diff
@@ -50,8 +50,8 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterator: typing.Iterator[T],
-) -> typing.AsyncIterator[T]:
+    iterator: typing.Iterable[T],
+) -> typing.AsyncIterable[T]:
     while True:
         try:
             yield await anyio.to_thread.run_sync(_next, iterator)
```
Kludex: But `_next()` receives an Iterator... 👀

Kludex (changes requested):

lovelydinosaur:
@jamesbraza you probably want to check in with other encode maintainers for this project. (I'm focused on `httpx`, `httpcore` at the moment.)

starlette/concurrency.py
jamesbraza: I guess `mypy` didn't realize this, great catch! Looks like https://github.com/encode/starlette/pull/2357 updates to latest `mypy`, but unfortunately `scripts/check` still doesn't catch this

jamesbraza: Alright the fix was simple, having `iterate_in_threadpool` call `iter` before using `_next`. PR is ready for review again

starlette/concurrency.py
```diff
@@ -50,8 +50,9 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterator: typing.Iterator[T],
-) -> typing.AsyncIterator[T]:
+    iterable: typing.Iterable[T],
```
Kludex: This is a breaking change.

starlette/concurrency.py
```diff
@@ -50,8 +50,9 @@ def _next(iterator: typing.Iterator[T]) -> T:
...
+) -> typing.AsyncIterable[T]:
+    iterator = iter(iterable)
```
Kludex: Can you create a test case that fails without this?

Kludex (changes requested):

starlette/concurrency.py
jamesbraza: Okay gotchu. Would you like me to leave the name as `iterator`, and just update the type hint?

starlette/concurrency.py
jamesbraza: Sure, I can cook something up 👍

jamesbraza: This test fails on `master` but passes here

```python
@pytest.mark.anyio
async def test_iterate_in_threadpool() -> None:
    class CustomIterable:
        def __iter__(self):
            yield from range(3)

    assert [v async for v in iterate_in_threadpool(CustomIterable())] == [0, 1, 2]
```

Let me know if you were hoping for something different

starlette/concurrency.py
Kludex: Yes, please.

starlette/concurrency.py
```diff
@@ -50,8 +50,9 @@ def _next(iterator: typing.Iterator[T]) -> T:
 
 
 async def iterate_in_threadpool(
-    iterator: typing.Iterator[T],
-) -> typing.AsyncIterator[T]:
+    iterable: typing.Iterable[T],
+) -> typing.AsyncIterable[T]:
```
Kludex: Isn't the return an `AsyncIterator`?

jamesbraza: Good catch, thanks. Implemented this just now

starlette/concurrency.py
jamesbraza: Alright, renamed back to `iterator`

Kludex (approved):

PR #2364, opened 2023-12-07
Make Mount work like before

abersheeran:
# Summary

This should solve https://github.com/encode/starlette/discussions/2361

# Checklist


Kludex:
What do you want to do here @abersheeran ?

abersheeran:
https://github.com/django/asgiref/issues/424 I want to wait for the final result of this discussion before deciding how to do this PR.

tarsil:
@abersheeran I can see here and the https://github.com/django/asgiref/issues/424 are closed. What are the conclusions to make the Mount work as before?

PR #2366, opened 2023-12-10
Deprecate `FileResponse(method=...)` parameter

Kludex:
The server drops the body on `HEAD` anyway, so even if we don't have this logic at all, the client still receives an empty body.

The idea of still having to check the method is to avoid opening the file.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -3,6 +3,7 @@ import json
 import os
 import stat
 import typing
+import warnings
 from datetime import datetime
 from email.utils import format_datetime, formatdate
 from functools import partial
@@ -10,6 +11,7 @@ from mimetypes import guess_type
 from urllib.parse import quote
 
 import anyio
+import anyio.to_thread
 
 from starlette._compat import md5_hexdigest
 from starlette.background import BackgroundTask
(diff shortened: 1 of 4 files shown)


starlette/responses.py
```diff
@@ -280,7 +282,11 @@ def __init__(
         self.path = path
         self.status_code = status_code
         self.filename = filename
-        self.send_header_only = method is not None and method.upper() == "HEAD"
+        if method is not None:
+            warnings.warn(
+                "The 'method' parameter is not used, and it will be removed.",
```
Kludex, line 287: The `method` parameter becomes a no-op with this PR, and it's "scheduled" to be removed in 1.0.

tests/test_responses.py
```diff
@@ -245,6 +246,27 @@ async def app(scope, receive, send):
...
+def test_file_response_on_head_method(
+    tmpdir: Path, test_client_factory: typing.Callable[..., TestClient]
+):
+    path = os.path.join(tmpdir, "xyz")
+    content = b"<file content>" * 1000
+    with open(path, "wb") as file:
+        file.write(content)
+
+    app = FileResponse(path=path, filename="example.png")
+    client = test_client_factory(app)
```
Kludex: This test is actually useless, since the `TestClient` drops the body on HEAD methods.

I'll create a new test later, but if you want to approve, it will speed up the process here. 👀

adriangb (approved):

iudeen (approved):

PR #2372, opened 2023-12-13
Using `raise from` in `concurrency._next`

jamesbraza:
# Summary

Minor style change using `raise from` in `_next`

# Checklist


Kludex:
Thanks, but that information is not used anywhere.

PR #2378, opened 2023-12-17
Allow multiple Mount objects with the same prefix (fixes #2374)

nicolaipre:
# Summary

# Checklist

This is a proposed fix to solve #2374, similar to the proposed fix for Host() in #2263.

This allows users to have multiple Mount() objects with the same path/prefix set, but use a different name. In essence it allows users to have named route groups, allowing the Mount names to be used directly in the OpenAPI spec, etc...

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -421,9 +421,12 @@ class Mount(BaseRoute):
             path = scope["path"]
             root_path = scope.get("route_root_path", scope.get("root_path", ""))
             route_path = scope.get("route_path", re.sub(r"^" + root_path, "", path))
-            match = self.path_regex.match(route_path)
-            if match:
-                matched_params = match.groupdict()
+            mount_match = self.path_regex.match(route_path)
+            path_match = self.routes == [] or any(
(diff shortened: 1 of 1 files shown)


sb1-nicolai:
After thinking this through with a clear mind it actually isn't necessary at all. I got so stuck in trying to figure out the other routing issues related to Host that I completely lost track of what I was actually trying to solve here. 

It is good that it is the way it is now, because if you notice a route not resolving after putting it in a Mount() it is most likely that you have a duplicate Mount, which will let you know and keep your code clean.

The PR can just be closed and I will close the related discussion.

nicolaipre:
After thinking this through with a clear mind it actually isn't necessary at all. I got so stuck in trying to figure out the other routing issues related to Host that I completely lost track of what I was actually trying to solve here.

It is good that it is the way it is now, because if you notice a route not resolving after putting it in a Mount() it is most likely that you have a duplicate Mount, which will let you know and keep your code clean.

Closing the PR and related discussion.

PR #2381, opened 2023-12-18
Add `*args` to `Middleware` and improve its type hints

pawelrubin:
Use ParamSpec to provide concrete type annotations for middleware's parameters.

# Summary

See https://github.com/encode/starlette/discussions/2380

# Checklist

- [ ] **N/A** ~I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.~
- [ ] **N/A** ~I've updated the documentation accordingly.~

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 import typing
 import warnings
 
@@ -14,7 +15,14 @@ from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, ExceptionHandler, Lifespan, Receive, Scope, Send
 from starlette.websockets import WebSocket
 
+if sys.version_info >= (3, 10):  # pragma: no cover
+    from typing import Concatenate, ParamSpec
+else:  # pragma: no cover
+    from typing_extensions import Concatenate, ParamSpec
+
+
 AppType = typing.TypeVar("AppType", bound="Starlette")
+P = ParamSpec("P")
 
 
 class Starlette:
@@ -124,10 +132,15 @@ class Starlette:
     def host(self, host: str, app: ASGIApp, name: str | None = None) -> None:
         self.router.host(host, app=app, name=name)  # pragma: no cover
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(
+        self,
+        middleware_class: typing.Callable[Concatenate[ASGIApp, P], typing.Any],
+        *args: P.args,
+        **options: P.kwargs,
+    ) -> None:
         if self.middleware_stack is not None:  # pragma: no cover
             raise RuntimeError("Cannot add middleware after an application has started")
-        self.user_middleware.insert(0, Middleware(middleware_class, **options))
+        self.user_middleware.insert(0, Middleware(middleware_class, *args, **options))
 
     def add_exception_handler(
         self,
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
-import sys
 import typing
 import warnings
+
+from typing_extensions import Concatenate, ParamSpec
 
 from starlette.datastructures import State, URLPath
 from starlette.middleware import Middleware
@@ -14,12 +15,6 @@ from starlette.responses import Response
 from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, ExceptionHandler, Lifespan, Receive, Scope, Send
 from starlette.websockets import WebSocket
-
-if sys.version_info >= (3, 10):  # pragma: no cover
-    from typing import Concatenate, ParamSpec
-else:  # pragma: no cover
-    from typing_extensions import Concatenate, ParamSpec
-
 
 AppType = typing.TypeVar("AppType", bound="Starlette")
 P = ParamSpec("P")
@@ -106,8 +101,8 @@ class Starlette:
         )
 
         app = self.router
-        for cls, options in reversed(middleware):
-            app = cls(app=app, **options)
+        for cls, args, kwargs in reversed(middleware):
+            app = cls(app=app, *args, **kwargs)
         return app
 
     @property
@@ -136,11 +131,11 @@ class Starlette:
         self,
         middleware_class: typing.Callable[Concatenate[ASGIApp, P], typing.Any],
         *args: P.args,
-        **options: P.kwargs,
+        **kwargs: P.kwargs,
     ) -> None:
         if self.middleware_stack is not None:  # pragma: no cover
             raise RuntimeError("Cannot add middleware after an application has started")
-        self.user_middleware.insert(0, Middleware(middleware_class, *args, **options))
+        self.user_middleware.insert(0, Middleware(middleware_class, *args, **kwargs))
 
     def add_exception_handler(
         self,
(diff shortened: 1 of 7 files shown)


starlette/applications.py
```diff
@@ -124,10 +132,15 @@ def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
     def host(self, host: str, app: ASGIApp, name: str | None = None) -> None:
         self.router.host(host, app=app, name=name)  # pragma: no cover
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(
+        self,
+        middleware_class: typing.Callable[typing.Concatenate[ASGIApp, P], typing.Any],
+        *args: P.args,
```
Kludex, line 133: Now without the args. 👀

Kludex (changes requested):

starlette/applications.py
pawelrubin: This is quite unfortunate as ParamSpec **requires** both `args` and `kwargs` to be present in the signature. This is not a breaking change, but certainly redundant since the underlying API does not accept `args`.

I figured this is _good enough_ and a small price for better type annotations that guard against potential issues regarding missing arguments to a Middleware class. 

What's your take on this?

pawelrubin:
Regarding failing checks - since the `Middleware` and `add_middleware` API is now typed, some typing issues have been detected. I'll have a look at those and fix them in a separate commit.

starlette/applications.py
adriangb: Can we just make `args` work? if so I wouldn't mind just adding it.

pawelrubin: Should the API remain intact, the ideal type annotations would be something along these lines:

```python
OptionsT = TypeVar("OptionsT", bound=TypedDict)


class MiddlewareT(Protocol[OptionsT]):
    def __call__(self, app: ASGIApp, **options: Unpack[OptionsT]) -> Any:
        ...


class Middleware:
    def __init__(self, cls: MiddlewareT[OptionsT], **options: Unpack[OptionsT]) -> None:
        self.cls = cls
        self.options = options
```

hypothetically - as `Unpack` does not currently allow "higher order" generic arguments.

Using `(app: ASGIApp, **options: P.kwargs) -> Any` won't work as well. 

Until the type system supports such a use case, we have to accept trade-offs or change the API accordingly (e.g. by passing down `args` as @adriangb suggested).

pawelrubin: Passing down `args` could definitely work, although I'm not sure if this is a desired change. I guess that the `kwargs`-only approach was very much intentional as it forces the user to be explicit. What do you think?

adriangb: Maybe being explicit was more important when there was no type checking?

starlette/applications.py
```diff
@@ -124,10 +132,15 @@ def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
     def host(self, host: str, app: ASGIApp, name: str | None = None) -> None:
         self.router.host(host, app=app, name=name)  # pragma: no cover
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(
+        self,
+        middleware_class: typing.Callable[Concatenate[ASGIApp, P], typing.Any],
```
```suggestion
        middleware_class: typing.Callable[Concatenate[ASGIApp, P], None],
```

starlette/middleware/__init__.py
```diff
@@ -1,12 +1,28 @@
...
+
+
+P = ParamSpec("P")
 
 
 class Middleware:
-    def __init__(self, cls: type, **options: typing.Any) -> None:
+    def __init__(
+        self,
+        cls: Callable[Concatenate[ASGIApp, P], Any],
```
Kludex:
```suggestion
        cls: Callable[Concatenate[ASGIApp, P], None],
```

starlette/applications.py
pawelrubin: I'm afraid this suggestion is incorrect. 

Consider the following minimal example
```python
from typing import Callable


class Foo: pass

foo: Callable[..., None] = Foo
```

This will raise the following error in **mypy**:
```shell
 error: Incompatible types in assignment (expression has type "Type[Foo]", variable has type "Callable[..., None]")  [assignment]
```

`Class`es can be considered `Callable`s since they returns instances when called.

starlette/applications.py
pawelrubin: Good point. I'll get `args` working and we'll continue the discussion from there.

starlette/middleware/__init__.py
```diff
@@ -1,12 +1,28 @@
...
+        *args: P.args,
```
Kludex, line 22: We either need to make it work with `args` or raise a `RuntimeError`.

adriangb: I vote make it work. As per above I'm _guessing_ the main reason to not allow args is that it's more error prone than kwargs but that's very much mitigated by enforcing typing. And args can be more convenient in simple cases!

pawelrubin: I agree. I've just pushed changes regarding making it work with `*args`. Please have another look.

One thought comes to mind - enabling the usage of `*args` is somewhat a breaking change if one is directly using `Middleware`'s `__iter__` or `__repr__`. What's your take on this? Do you consider this a breaking change?

Kludex: It's ok.

Kludex (approved):

starlette/applications.py
```diff
@@ -124,10 +127,15 @@ def mount(self, path: str, app: ASGIApp, name: str | None = None) -> None:
     def host(self, host: str, app: ASGIApp, name: str | None = None) -> None:
         self.router.host(host, app=app, name=name)  # pragma: no cover
 
-    def add_middleware(self, middleware_class: type, **options: typing.Any) -> None:
+    def add_middleware(
+        self,
+        middleware_class: typing.Type[MiddlewareClass[P]],
```
pawelrubin: I figured that we could achieve more precise annotation via `typing.Protocol`. Moreover, this ensures that the middleware instance does adhere to the ASGI protocol.

starlette/middleware/__init__.py
```diff
@@ -1,17 +1,38 @@
...
+from typing import Any, Iterator, Protocol, Type
+
+from typing_extensions import ParamSpec
+
+from starlette.types import ASGIApp, Receive, Scope, Send
+
+P = ParamSpec("P")
+
+
+class MiddlewareClass(Protocol[P]):  # pragma: no cover
```
adriangb: Can we somehow make this private?

adriangb (approved):
Looks good to me other than a slight preference for not adding new public symbols

starlette/middleware/__init__.py
pawelrubin: Sure.

adriangb: Also why is there pragma here? Doesn't it get run just by being imported?

pawelrubin: This particular line - as well as the class body - does get run when imported. However, the empty body (`...`) of `__init__` and `__call__` does not, since those are not being called. Putting pragma at the top of the class was merely a shortcut - not to put pragma in both of those functions.

Though it's probably best to be explicit. Changed.

adriangb: Makes total sense! It just confused me.

pawelrubin: Thanks! :)

PR #2384, opened 2023-12-23
Version 1.0.0

### LINKED ISSUE #51: Version 1.0

Kludex:
- Closes https://github.com/encode/starlette/issues/51

I've tried to be careful on the commit sequence here - each commit passes the pipeline by itself.

- This PR follows the same strategy as `httpcore` on https://github.com/encode/httpcore/pull/809.

@encode/maintainers Does someone want to add a breaking change on Starlette before we go to 1.0.0? Please, let's focus on breaking changes, if there are features that can be added on minor, let's not bother with them here.


starlette/applications.py
```diff
@@ -1,13 +1,11 @@
...
 import typing
-import warnings
 
 from typing_extensions import ParamSpec
 
 from starlette.datastructures import State, URLPath
 from starlette.middleware import Middleware, _MiddlewareClass
-from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.middleware.exceptions import ExceptionMiddleware
```
adriangb, line 14: We need to remove ExceptionMiddleware from here. And maybe delete that file as well.

docs/applications.md
```diff
@@ -25,8 +27,11 @@ async def websocket_endpoint(websocket):
     await websocket.send_text('Hello, websocket!')
     await websocket.close()
 
-def startup():
+@asynccontextmanager
+async def lifespan(app: Starlette):
     print('Ready to go')
+    yield
```
graingert, line 33: Does this need to be a

```
try:
    yield
finally:
    print("shutting down")
```
?

starlette/applications.py
Kludex: Where should we register the `starlette.exception_handlers` now?

docs/applications.md
Kludex: I don't think so.

starlette/applications.py
adriangb: Sorry, I guess we need to get rid of is this part: https://github.com/encode/starlette/blob/966f0fc85764bd478d7c582fc8e2e5f91dbebae6/starlette/middleware/exceptions.py#L62

Kludex: Ah! Just to confirm: you mean that the `ExceptionMiddleware` should just register the `exception_handlers` in the scope, but the `wrap_app_handling_exceptions` should NOT be called there, only on the other places we have them (`request_response` and `websocket_session`). Correct?

adriangb: Yep!

Kludex: We have a problem. We had an oversight when implementing the _per route exception handler_.

There are places we have a `raise HTTPException` that don't follow the `request_response` or `websocket_session` flow, and we can't just remove/replace by `PlainTextResponse` because the user may be already be changing the behavior of those `HTTPException`.

If you just remove the `wrap_app_handling_exceptions` you mentioned above, you'll see the test failures that express what I'm saying.

adriangb: I can try after the holidays but I’d say we should change those and users shouldn’t expect any request/response stuff in middleware (aside from BaseHTTPMiddleware which is beyond changing)

Kludex: Based on what you said, I don't think I made myself clear with my comment above.

I'll wait for later then. Thanks! 🙏

Kludex: I cannot come up with a solution for this, FYI.

Kludex: @adriangb how do you want to proceed here? We talked in PyCon US a bit, but I don't recall how we left this.

This is the only blocker for 1.0.

The only thing is that I don't want to add any breaking change.

adriangb: If you want to do this without any breaking change we can just leave `ExceptionMiddleware ` where it is and ignore this thread.

But conceptually I think the thing is that:
1. After routing once we have a Request/Response exception handling can be done via a Request/Response (the current API).
2. Outside of that (e.g. in other middleware or routing) we're in pure ASGI land. We could still let you catch `HTTPException`'s but I think the API should essentially be an ASGI app `HTTPException | int, Scope, Receive, Send` or something like that. But that would be a breaking change. Otherwise I fear we're going to get bugs like "I wanted to catch the `404` HTTPException and log the request body but when I called `.body()` I got a `StreamConsumed` error".

To do that we'd have to replace `ExceptionMiddleware` with `AppExceptionMiddleware` or something like that with the new API and have two different arguments to `Starlette` for each type of exception handler.

abersheeran: Leaving the pure ASGI realm, I have a minimalistic and effective design. https://github.com/abersheeran/baize/blob/master/baize/asgi/shortcut.py#L59

adriangb: Very nice :)

lovelydinosaur:
I reckon Starlette 1.0 needs a clear resolution to https://github.com/encode/starlette/issues/2673.

Kludex:
> I reckon Starlette 1.0 needs a clear resolution to #2673.

I've been trying to push this forward for 2 years.

lovelydinosaur:
The simplest approach fro your side is https://github.com/encode/starlette/pull/2534#issuecomment-1979174688.

PR #2385, opened 2023-12-23
Add metadata to route classes.

alex-oleshkevich:
This PR adds metadata to route classes.
Metadata is an arbitrary data that can be associated with the route.

The primary audience of this PR is library/framework developers.
One of the use cases can be Open API integration. User can define schema on route and some tool generates Open API docs by scanning available routes.

```
Route('/', index, metadata={'description': 'some route info'})
```

My use case: I am building a decorator that provides dependency injection and generates API docs. It makes sense to me to do that in one pass:
```python
def create_route_from_endpoint(route_spec: RouteSpec, endpoint: Endpoint) -> BaseRoute:
    resolvers = create_dependency_resolvers(endpoint)

   # scan request, response and function parameters for open api components
    components = ....

    async def real_endpoint(request: Request) -> Response:
        dependencies = await resolve_dependencies(request, resolvers)
        return await endpoint(**dependencies)

    return Route(
        path=route_spec.path,
        name=route_spec.name,
        methods=list(route_spec.methods),
        endpoint=real_endpoint,
        metadata={'openapi':  components}, # usage here
        include_in_schema=route_spec.include_in_schema,
    )

```

### DIFF
--- a/docs/routing.md
+++ b/docs/routing.md
@@ -331,3 +331,17 @@ The `endpoint` argument can be one of:
 
 * An async function, which accepts a single `websocket` argument.
 * A class that implements the ASGI interface, such as Starlette's [WebSocketEndpoint](endpoints.md#websocketendpoint).
+
+
+## Routing metadata
+
+You can attach arbitrary metadata to route instances.
+Metadata allows user-defined data to be associated with this route instance.
+
+
+```python
+routes = [
+    Route("/", endpoint=websocket_index, metadata={"foo": "bar"}),
+    WebSocketRoute("/{name}", endpoint=websocket_user, metadata={"foo": "bar"}),
+]
+```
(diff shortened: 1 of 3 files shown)


adriangb:
Can't frameworks just subclass and add a typed attribute?

What I've found lacking for this previously is a function that can visit all routes keeping track of the paths and taking into account `Mount`, etc. That needs to be in Starlette because it needs to handle all routing primitives current and future.

alex-oleshkevich:
Subclassing is like a sledgehammer in this context.

hasansezertasan:
> Subclassing is like a sledgehammer in this context.

I totally agree. Think of it this way, FastaAPI using a sublcass and instead of 'metadata' it uses 'mdata', another implementation, let's call it QuickAPI, uses 'meta', when we wrap these together on a top level application and try to do something with some metadata, with no standard approach... It sounds like a nightmare.

One question for @alex-oleshkevich and others, why not just use '**kwargs' and pass the values to 'metadata'?

Edit: The example abowe might be problematic because metadata accepts a mapping of string as key and any as value, which is still a nightmare because no standard in it, but being able to pass some extra metadata is still quite a good idea.

Kludex:
This can be solved subclassing. The recommendation was always to not include changes in the package if you can do it easily by yourself.

I'll close this PR because this is not a clear win, and we can start with a discussion, if you want.

PR #2387, opened 2023-12-28
Fix: Caching of request body for passing down to middlewares

ayushbisht2001:
# Summary
Hey all, I was trying to figure out a way to get the request body inside a middleware that I need for logging the audit logs. But seems like, we don't have that much flexibility for getting the `multipart/form-data` #495 .

I tried to make it work by caching the data in the `state` itself. I don't know whether it is the best way to fix this thing. I'm quite new to this community, please assist me for fixing this issue.

Thanks

# Checklist 

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -25,7 +25,7 @@ class _CachedRequest(Request):
     """
 
     def __init__(self, scope: Scope, receive: Receive):
-        super().__init__(scope, receive)
+        super().__init__(scope, receive, enable_request_caching=True)
         self._wrapped_rcv_disconnected = False
         self._wrapped_rcv_consumed = False
         self._wrapped_rc_stream = self.stream()
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -189,7 +189,8 @@ class Request(HTTPConnection):
     _form: typing.Optional[FormData]
 
     def __init__(
-        self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send
+        self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send,
+        enable_request_caching: bool = False
     ):
         super().__init__(scope)
         assert scope["type"] == "http"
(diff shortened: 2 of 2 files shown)


adriangb:
This has been discussed extensively. Putting the body in state breaks things in multiple ways. Consider for example a middleware that modifies the request body, e.g. unzips it. If a middleware upstream put it in state then the request handler won't see the unzipped body, it will see the original body.

That said, I'm not sure what your actual use case is but there are safe ways to access the request body in a middleware and in the downstream app. Take a look at https://github.com/encode/starlette/blob/483849a466a2bfc121f5a367339e1aa3ed20344b/starlette/middleware/base.py#L19-L91. Maybe you can do something similar?

If you're writing your own middleware I'd recommend you read this section of the docs. At the very least it will give you a better understanding of what's going on under the hood: https://www.starlette.io/middleware/#pure-asgi-middleware

ayushbisht2001:
> That said, I'm not sure what your actual use case is but there are safe ways to access the request body in a middleware and in the downstream app. Take a look at
> 
> https://github.com/encode/starlette/blob/483849a466a2bfc121f5a367339e1aa3ed20344b/starlette/middleware/base.py#L19-L91
I've look into this file, but my concern is for the `multipart/form-data`, How can I get the form data values ? Other json based request are properly fetched, while with stream request, it showing `stream consumed`.
Right now, I'm trying some hacky way to parse the chunks to some custom `MutliPartParser` to create `FormData`. Is there a better way to get the `multipart/form-data` record.

ayushbisht2001:
@adriangb Also, about the issue with Gzip, I guess it will break, if the we place the cache middleware before the Gzip middleware. Creating a separate middleware  would be better way to do this I guess, with an instructions to always keep it at the lowest stream. 
WHAT U THINK ?

adriangb:
I don't think we should move forward with this or similar changes. Please look at the links in my previous comment and see if you can solve the problem that way or report back why that won't work.

ayushbisht2001:
@adriangb Actually, for my usecase I was able to solve this issue. But what I feel is , it would be really helpful if the caching of request body is supported by library ( more specifically  `multipart/form-data` which throws `stream consumed error` ).
BTW thanks @adriangb  for all your responses.

Kludex:
As Adrian said, this was discussed extensively. Also, you solved your use case, so let's close this.

If you need more guidance, please open a discussion.

Issue #2019, opened 2023-01-27
SessionMiddleware sends a new set-cookie for every request, with unintended results


Kludex:
### Discussed in https://github.com/encode/starlette/discussions/2018

<div type='discussions-op-text'>

<sup>Originally posted by **sm-Fifteen** January 27, 2023</sup>
The way SessionMiddleware currently works is that it encodes session data in a signed base64 object with timestamp, in a format not entirely dissimilar to (but incompatible with) JWT. Because of the way it works, being run right before the headers for an HTTP response are sent, and because it doesn't perform many checks besides whether or not there is session data to encode, it will try to update the session cookie whenever a route is called. This is somewhat wasteful, as session data is not usually updated on every request, but it leads to interesting issues as well.

I have a FastAPI application where login is handeled via a session cookie with a separate login process. Several routes are unlocked based on whether or not the cookie is present and based on the user authorizations recorded in said session cookie. When the frontend fetches API data, if the user is logged in, every response contains a new `Set-Cookie` header, which *Firefox ignores, but Chrome acknowledges*. None of the API routes alter session data, so the cookie value is always the same, save for the timestamp and signature at the end. This actually runs into [an interesting Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=696204#c107) where responses to fetch requests that arrive after the page has been navigated away from still alter the session cookie, even if if the new page changed the session data on load. This means I have this strange, Chrome-exclusive, race-condition-y bug that matches lepture/authlib#334, where the session state data setup by the Oauth flow before redirecting to the third party login form may get clobbered if some `/api/slow_route` query was pending when the user clicked the login button.

It's unclear if it's actually a chrome bug or unintended behavior from everything working as intended. If Starlette wasn't updating cookies for every request, and the ``Set-Cookie`` was actually important here, it's difficult to argue whether or not Chrome should have ignored it.

See here the Starlette test case from the Chromium bug report, which doesn't use SessionMiddleware, but shows the Chrome behavior in action.

```py
from starlette.applications import Starlette
from starlette.responses import HTMLResponse, RedirectResponse, JSONResponse, PlainTextResponse
from starlette.routing import Route
import uvicorn
import datetime
from asyncio import sleep

html_content = """
    <!DOCTYPE html>
    <html>
        <head>
            <meta http-equiv="refresh" content="2;url=/set_cookie_and_redirect" />
            <script>fetch("/set_cookie_on_fetch")</script>
        </head>
        <body>
[...]

PR #2401, opened 2024-01-07
Session cookie fix

ToasterChicken:
# Summary

This is regarding Issue #2019  

It adds persistent session features, and also the option for refreshing the current session before it expires while it's still valid (refresh window)  and if it's beyond that window the session is invalid.

The default behavior without the new features should also remain the same was it was prior to the changes, every single requests results in a new session cookie.

I feel like there's possibly a much better way of doing this like being able to handle it all with in the __call__ before the send_wrapper.  However this is my first time delving into the ASGI framework so I'm not sure if I am able to obtain the new session data and compare it to the old session data only within the __call__ and not have to make one cookie in __call__ and another one with in send_wraper to resolve the current cookie and compare it to the new data.

Haven't added any tests to test_session.py yet, since I think this is a pretty large change and I can go ahead and work on that if this appears to be something that maybe accepted. The existing test suite passes however, and manually testing individual behaviors from the update work as expected.

There's no single change here, its a pretty big overall change to the middleware.

# Checklist

- [x] Passes existing tests.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -106,7 +106,7 @@ The following arguments are supported:
 * `same_site` - SameSite flag prevents the browser from sending session cookie along with cross-site requests. Defaults to `'lax'`.
 * `https_only` - Indicate that Secure flag should be set (can be used with HTTPS only). Defaults to `False`.
 * `persist_session` - Sets the session cookie that's created to persist between connections. Defaults to `False`.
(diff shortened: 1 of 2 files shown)


ToasterChicken:
Not sure what it's failing on.  There's another test that's unrelated that's throwing XFAIL, but everything else seems to be passing when I run the tests locally. Though this is also my first adventure with running tests....  I've tested with 3.10, 3.11 and 3.12.

scripts/test shows no errors.
tests/middleware/test_session.py ..............                                                                                                                  [ 98%]

running the coverage report
tests/middleware/test_session.py            104      0   100%
(edit: had to redo my notes, it did something weird when submitting and didn't notice)

PR #2405, opened 2024-01-10
Documenting how `StreamingResponse` adheres to ASGI spec

jamesbraza:
# Summary

I was reading the source code how `StreamingResponse` works, and found myself in-depth researching `more_body`.

This PR just upstreams that knowledge (and links the ASGI spec) in the docs.

# Checklist

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -159,6 +159,11 @@ async def app(scope, receive, send):
 
 Have in mind that <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> objects (like those created by `open()`) are normal iterators. So, you can return them directly in a `StreamingResponse`.
 
+For reference, the way `StreamingResponse` works under the hood is
+the `more_body` parameter within the response body,
+as described in the
+[ASGI specification](https://asgi.readthedocs.io/en/latest/specs/www.html).
+
 ### FileResponse
 
 Asynchronously streams a file as the response.
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -159,10 +159,11 @@ async def app(scope, receive, send):
 
 Have in mind that <a href="https://docs.python.org/3/glossary.html#term-file-like-object" target="_blank">file-like</a> objects (like those created by `open()`) are normal iterators. So, you can return them directly in a `StreamingResponse`.
 
-For reference, the way `StreamingResponse` works under the hood is
-the `more_body` parameter within the response body,
-as described in the
-[ASGI specification](https://asgi.readthedocs.io/en/latest/specs/www.html).
+!!! info
+    If you are curious how the streaming works under the hood,
+    the streaming protocol comes from the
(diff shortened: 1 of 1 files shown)


Kludex:
I don't find this clear. 🤔

jamesbraza:
What aspect do you find unclear? Also, feel free to make commits to this PR directly

My motivation here is helping others get a better head-start on understanding `StreamingResponse`'s inner workings

hasansezertasan (changes requested):
This is unclear because there is no mention of "more_body" in the file @jessebraham edited.

I saw your [PR](https://github.com/tiangolo/fastapi/pull/10932) on FastAPI repo. It's good for you, to search in depth... But I think we need to KISS the docs. I see what you are trying to do but developers use all kinda technologies and we need to leave some space for devs to search/discover. No documentation can have all answers (or refs) and IMO it's a good thing.

In the end, there is no mention of "more_body" in the docs. It's just confusing me and there is already a reference for ASGI Specs in the `index.md` file.

jamesbraza:
I have just tweaked the wording and moved it to an `info` section for visual separation.  @Kludex let me know if there are parts still confusing you

Kludex:
I'm not convinced. We can wait for other member to comment.

jamesbraza:
Sounds good. To follow up on a few comments:

> I think we need to KISS the docs.

Yep sounds good. My last commit tries to further simplify and be less technical, while still containing relevant information.

> In the end, there is no mention of "more_body" in the docs.

There's no other mentions of `more_body` because it's specific to streaming, and this is the streaming docs. So I am not sure if it would be documented elsewhere.

> It's just confusing me and there is already a reference for ASGI Specs in the `index.md` file.

As far as there being other links to the ASGI spec in the docs, I think it's fine to link it several times, I don't think many people ever read full documentation, just relevant portions

Thanks again all

tiangolo:
Thanks!

But I don't really think it's a good idea to link to the ASGI spec in the docs and point newcomers learning about the framework to the dry and long specification, it's much more complicated and not really necessary for most people.

As Starlette already mentions in several places it is an ASGI framework, I would think people who wanted to know more about the details of each specific part would be able to go and read the spec, and it also applies to many other parts, not just `StreamingResponse`... So I would be more inclined to not take this one. 😅

tiangolo:
Sorry, closed instead of commenting. 😅

jamesbraza:
My motivation here is I use `StreamingResponse` in a somewhat performant `FastAPI` system to stream JSON data, and had to create a subclass for this. So I wanted to know what mechanism it used, if the connection was long-lived, etc.

Understanding the inner workings for me was somewhat nontrivial, involving lots of IDE debugging (probably because I am new to ASGI). So I tried to leave behind a paper trail to help "the next guy" in figuring out how the streaming actually works, without resorting to the debugger.

I think the thing with PRs like this is, once you know it, you forget what it's like not to know.

Anyways, I can let others make the call here, thanks for participation!

Kludex:
I agree with @tiangolo .

Thanks for the PR, but I don't think this should be included.

PR #2406, opened 2024-01-11
♻️ Do not use the deprecated `method` parameter in `FileResponse` inside of `StaticFiles`

tiangolo:
# Summary

♻️ Do not use the deprecated `method` parameter in `FileResponse` inside of `StaticFiles`

This is a continuation of https://github.com/encode/starlette/pull/2366

# Checklist


Kludex:
I saw this the other day... I thought I had included this somewhere. 🤔

Thanks. 👍

Kludex (approved):

PR #2422, opened 2024-01-15
Raise `FileNotFoundError` when the `env_file` parameter on `Config` is not valid

mathause:
# Summary

Addresses #2101

# Checklist

-> I think this if ok without doc update. Let me know otherwise.


Kludex (approved):

rambo:
oof, all my docker containers using the old *documented* pattern are now failing after routine dependency updates.
Maybe the PR checklist should contain checks about breaking things, though in this case OP obviously thought nothing would break. Ref https://github.com/encode/starlette/discussions/2446

PR #2425, opened 2024-01-17
Raise `WebSocketDisconnect` when `WebSocket.send()` excepts `IOError`

Kludex:
- I don't think I can test this can be tested without fixing https://github.com/encode/starlette/issues/947.

I think we should fix the above issue before merging this.

Uvicorn PR: https://github.com/encode/uvicorn/pull/2218

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -61,36 +61,34 @@ class WebSocket(HTTPConnection):
         """
         Send ASGI websocket messages, ensuring valid state transitions.
         """
-        try:
-            if self.application_state == WebSocketState.CONNECTING:
-                message_type = message["type"]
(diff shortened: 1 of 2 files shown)


Kludex:
- I've created https://github.com/encode/starlette/pull/2427. 🙏

Kludex:
It turns out that now it's impossible to create a test with the `TestClient` with the changes I've implemented on #2427. 😆

It's because the `TestClient` will exit on client disconnect. I think the behavior is fine tho.

Kludex:
@encode/maintainers someone? 🙏

michaeloliverx (approved):
Don't have a lot of context here but it looks fine at first glance and there is a test.

PR #2430, opened 2024-01-20
Add `scope_client` parameter to `TestClient`

pythonweb2:
# Summary

A breaking change was made when the `TestClient`'s `scope["client"]` was set to None. In cases where an application uses the scope["client"] in places like custom middleware, logging, etc, upgrading to 0.35 becomes a very hard task. Testing middleware/logging/business logic that relies on the client being populated becomes hard if not impossible. Middleware that worked before breaks, and now we need to add checks in application code to disable certain things based on whether or not the test suite is running.

This PR adds the ability for users to specify what the `scope["client"]` should be when using the test client. This way, when using fixtures in a framework like pytest (which is the recommended setup), a one line change is all that is required.

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -215,12 +215,14 @@ class _TestClientTransport(httpx.BaseTransport):
         root_path: str = "",
         *,
         app_state: dict[str, typing.Any],
+        client: typing.Optional[typing.List[typing.Union[str, int]]],
     ) -> None:
         self.app = app
         self.raise_server_exceptions = raise_server_exceptions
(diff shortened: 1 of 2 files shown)


Kludex:
Why should we revert this?

The new parameter is unnecessary.

pythonweb2:
I made it so instead of reverting the change, there is a new option to keep the old functionality, which should be less controversial (if some have already change their code for the 0.35.0 release). Would this be OK to add?

@Kludex

Kludex:
Let's discuss it first.

PR #2435, opened 2024-01-22
Add support for ASGI `pathsend` extension

gi0baro:
# Summary

As per title, this adds support for [ASGI pathsend](https://asgi.readthedocs.io/en/latest/extensions.html#path-send) extension on `FileResponse` class for servers implementing it.

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -337,6 +337,8 @@ class FileResponse(Response):
         )
         if scope["method"].upper() == "HEAD":
             await send({"type": "http.response.body", "body": b"", "more_body": False})
+        elif "http.response.pathsend" in scope["extensions"]:
+            await send({"type": "http.response.pathsend", "path": str(self.path)})
         else:
             async with await anyio.open_file(self.path, mode="rb") as file:
                 more_body = True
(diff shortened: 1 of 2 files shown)


Kludex (changes requested):
We need tests.

@abersheeran is this the only thing needed on the implementation side?

abersheeran:
> @abersheeran is this the only thing needed on the implementation side?

Yes, its actual logic implementation is all on the server side.

gi0baro:
@Kludex I added a test for this, but it only test the correctness of the messages as `TestClient` doesn't support `pathsend` directly. Let me know if this is ok for you.

Kludex:
Reference of the previous PR regarding this extension: https://github.com/encode/starlette/pull/1288/

gi0baro:
> Reference of the previous PR regarding this extension: #1288

🤔 `http.response.zerocopysend` ([ASGI ref](https://asgi.readthedocs.io/en/latest/extensions.html#zero-copy-send)) != `http.response.pathsend` ([ASGI ref](https://asgi.readthedocs.io/en/latest/extensions.html#path-send))

Kludex:
OHHHH, this is the one you created. Sorry!

Kludex (approved):

Kludex:
Thanks @gi0baro 🙏

PR #2445, opened 2024-01-26
Add type hints in `test_base.py`

TechNiick:
# Summary

This is part of a bigger Task. The idea is to remove those flags on mypy check command on: `scripts/check`.

**Before**
`${PREFIX}mypy starlette`
`${PREFIX}mypy tests --disable-error-code no-untyped-def --disable-error-code no-untyped-call`

**After**
`${PREFIX}mypy starlette tests`

In order to do that, we need to fix all issues after deleting those flags. After deleting those, there are more than 730 issues, in 34 files. I will make separate PRs for each file in order to have atomic changes.

Things done in this PR:
 * Added type annotations to function parameters and return type that were requested by mypy in test_base.py file.
 * Added used types in the types.py files.
 * Changed types that were already in types.py in places where the types were used directly
 
NOTES:
 * There are some places where i have added "Any" type since i couldn't figure out what types should be there. If anyone has any idea what types should be there, please let me know.
 
# Checklist


adriangb (approved):
Looks great, thank you!

starlette/types.py
Kludex, line 1: The types only used on the test suite shouldn't be added here. You can add them in the `test_base.py` itself for now, and if reused on multiple files, we can have them added in a module on the `tests/` directory.

Kludex (changes requested):
Besides this, looks good. Thanks! 🙏

starlette/types.py
TechNiick: Thank you for the review and the guidance!

I've noticed that certain types, such as `TestClientFactory`, are indeed reused across multiple test files (e.g., `test_routing.py`, `test_applications.py`, `test_background.py`, etc.). To confirm my understanding and plan of action:

1. Create a New Module for Shared Test Types: I will create a new module within the tests/ directory to house these shared types. I'm considering naming it `tests.test_types.types`. Does this naming convention work for you, or would you suggest a different structure?

2. Revert Changes to `starlette.types`: I understand that the `starlette.types` module is intended for application-level types. Therefore, I'll move the test-specific types that I have added out of this module and into the newly created module for shared test types.

3. Handling Types Used in a Single File: For types that are currently only used in one test file, I plan to keep them within their respective files to maintain simplicity. However, now that we'll have a dedicated module for shared test types (`test_types.types`), would you prefer moving all test-specific types there for consistency, or should they remain in their individual test files until needed elsewhere?

Kludex: No. For now, just the `test_base.py`. Create the types you need there. We'll talk if they are needed or not individually, and if we should move them somewhere.

Kludex: Let's go step by step. 👍

TechNiick: Got it. Thank you!

I have done the requested changes, you can check.

tests/middleware/test_base.py
```diff
@@ -14,44 +23,59 @@
 from starlette.routing import Route, WebSocketRoute
 from starlette.testclient import TestClient
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
+from starlette.websockets import WebSocket
+
+BytesGenerator = Generator[bytes, None, None]
+AwaitableGenerator = Generator[Any, None, None]
+TestClientFactory = Callable[[ASGIApp], TestClient]
+MiddlewareClass = Type[_MiddlewareClass[Any]]
```
Kludex: Can you instead not create those aliases, please? You can leave the `TestClientFactory` here, but the ones that are used only once, I'd prefer to avoid.

TechNiick: Sure, deleted them.

Kludex (approved):

Kludex:
Thanks! :)

PR #2447, opened 2024-01-28
Implement MutableMapping in the State

HHongSeungWoo:
# Summary
#2389 

# Checklist


Kludex:
Would you mind reopening this?

PR #2450, opened 2024-01-29
Fix the memory issue in WSGI middleware

peterzaccha:
# Summary

This PR attempts to fix the WSGIMiddleware, which has a serious memory consumption issue when handling large request and response sizes by:

- Limiting memory usage by reducing the max buffer size in `create_memory_object_stream` from `math.inf` to zero.
- Introducing `WsgiRequestBody` class for incremental handling of the request body, replacing the previous byte array loading approach.

The implementation for WSGIMiddleware was 

# Checklist

### DIFF
--- a/starlette/middleware/wsgi.py
+++ b/starlette/middleware/wsgi.py
@@ -16,7 +16,37 @@ warnings.warn(
 )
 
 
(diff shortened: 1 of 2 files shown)


Kludex:
The middleware is deprecated. Use the `a2wsgi` package instead.

PR #2459, opened 2024-02-04
Create `anyio.Event` using a blocking portal

Kludex:
This fixes Starlette for anyio < 4.2.0.

We need a way to test both low and max version on anyio on Starlette.


starlette/testclient.py
```diff
@@ -95,7 +96,6 @@ def __init__(
     def __enter__(self) -> WebSocketTestSession:
         self.exit_stack = contextlib.ExitStack()
         self.portal = self.exit_stack.enter_context(self.portal_factory())
-        self.should_close = anyio.Event()
```
Kludex, line 98: `anyio.Event` cannot be instantiated on non-async context before anyio 4.2.0.

tiangolo (approved):

PR #2469, opened 2024-02-04
Add type hints to `conftest.py`

TechNiick:
# Summary

Type annotation added to `conftest.py`

# Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,12 +1,18 @@
 import functools
+from typing import Any, Callable, Dict
 
 import pytest
 
 from starlette.testclient import TestClient
 
+TestClientFactory = Callable[..., TestClient]
+
 
 @pytest.fixture
-def test_client_factory(anyio_backend_name, anyio_backend_options):
+def test_client_factory(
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -1,12 +1,18 @@
...
 from starlette.testclient import TestClient
 
+TestClientFactory = Callable[..., TestClient]
+
 
 @pytest.fixture
-def test_client_factory(anyio_backend_name, anyio_backend_options):
+def test_client_factory(
+    anyio_backend_name: str,
+    anyio_backend_options: Dict[str, Any],
```
Kludex, line 14: Is the type available in AnyIO?

TechNiick: As far as I can see, it looks like it uses the same str and Dict[str, Any] in AnyIO too.

Example [here](https://github.com/agronholm/anyio/blob/137de708f40262e959d9d43178dda600432da56b/tests/test_debugging.py#L29)

Kludex (approved):

PR #2470, opened 2024-02-04
Add type hints to `test__utils.py`

TechNiick:
# Summary

Type annotation added to `test__utils.py`

# Checklist

### DIFF
--- a/tests/test__utils.py
+++ b/tests/test__utils.py
@@ -25,7 +25,7 @@ def test_async_partial() -> None:
     partial = functools.partial(async_func, 1)
     assert is_async_callable(partial)
 
-    assert not is_async_callable(partial)
+    partial = functools.partial(func, 1)
     assert not is_async_callable(partial)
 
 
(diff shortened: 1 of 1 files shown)


tests/test__utils.py
```diff
@@ -1,77 +1,99 @@
...
     partial = functools.partial(async_func, 1)
     assert is_async_callable(partial)
 
-    partial = functools.partial(func, 1)
+    partial = functools.partial(
+        cast(
+            Callable[..., Coroutine[Any, Any, None]], func
+        ),
+        1,
+    )
```
Kludex: What's the error here? Can we just have a type ignore?

TechNiick: errors:
`tests/test__utils.py:78: error: Argument 1 to "partial" has incompatible type "Sync"; expected "Callable[..., Coroutine[Any, Any, None]]"  [arg-type]`

`tests/test__utils.py:78: note: "Sync.__call__" has type "Callable[[Arg(Any, 'a'), Arg(Any, 'b')], None]"`

Yes, we can just ignore the type here. Thank you!

Kludex (approved):

PR #2472, opened 2024-02-04
Add type hints to `test_authentication.py`

TechNiick:
# Summary

Type annotation added to `test_authentication.py`

# Checklist


tests/test_authentication.py
```diff
@@ -1,9 +1,16 @@
 import base64
 import binascii
+import sys
+from typing import Any, Awaitable, Callable, Optional, Tuple
```
Kludex, line 3: Let's avoid use `Any`, please.

tests/test_authentication.py
```diff
@@ -15,14 +22,23 @@
...
 from starlette.middleware.authentication import AuthenticationMiddleware
-from starlette.requests import HTTPConnection
+from starlette.requests import HTTPConnection, Request
 from starlette.responses import JSONResponse
 from starlette.routing import Route, WebSocketRoute
-from starlette.websockets import WebSocketDisconnect
+from starlette.testclient import TestClient
+from starlette.websockets import WebSocket, WebSocketDisconnect
+
+P = ParamSpec("P")
```
Kludex: Is it being used?

Kludex (changes requested):

tests/test_authentication.py
TechNiick: Yes, I have added it on `**kwargs: P.kwargs`.

tests/test_authentication.py
```diff
@@ -152,9 +172,9 @@ def decorated_sync(request, additional):
     )
 
 
-def ws_inject_decorator(**kwargs):
-    def wrapper(endpoint):
-        def app(websocket):
+def ws_inject_decorator(**kwargs: P.kwargs) -> Callable[..., AsyncEndpoint]:  # type: ignore
```
Kludex:
```suggestion
def ws_inject_decorator(**kwargs: Any) -> Callable[..., AsyncEndpoint]:
```

Kludex: P is wrong.

Kludex (approved):

PR #2478, opened 2024-02-04
Add type hints to `test_endpoints.py`

TechNiick:
# Summary

Type annotation added to `test_endpoints.py`

# Checklist


Kludex (approved):

Kludex (approved):

tests/test_endpoints.py
```diff
@@ -1,12 +1,19 @@
+from typing import Callable, Generator
```
Kludex:
```suggestion
from typing import Callable, Iterator
```

tests/test_endpoints.py
```diff
@@ -19,33 +26,33 @@ async def get(self, request):
 
 
 @pytest.fixture
-def client(test_client_factory):
+def client(test_client_factory: TestClientFactory) -> Generator[TestClient, None, None]:
```
Kludex:
```suggestion
def client(test_client_factory: TestClientFactory) -> Iterator[TestClient]:
```

Kludex (changes requested):

Kludex (approved):

PR #2485, opened 2024-02-06
warn instead of raise for missing env file

mathause:
# Summary

Adresses #2422

Let me know if you want a custom Warning and/ or any doc changes.

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.

PR #2486, opened 2024-02-06
Restore old documented functionality, add optional warning, document the warning

rambo:
# Summary

As per https://github.com/encode/starlette/discussions/2446 the new "raise by default" broke a ton of containerized apps where environment is used in the container but developers use .env files locally.

This will restore old documented functionality to a pattern that is just about everywhere but also adds an option to issue warning if the file is missing.

IDK why coverage on CI is being a silly bugger, my [local coverage report](https://imgur.com/a/ZSdjPvN) says all modified code paths are covered

# Checklist

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -11,7 +11,8 @@ from starlette.config import Config
 from starlette.datastructures import CommaSeparatedStrings, Secret
 
 # Config will be read from environment variables and/or ".env" files.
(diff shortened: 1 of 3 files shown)


Kludex:
I prefer the approach on https://github.com/encode/starlette/pull/2485.

Thanks for the PR.

PR #2487, opened 2024-02-07
Add '.idea' Pycharm folder for project setting in .gitignore

0xsirsaif:
# Summary

Simply, Add `.idea` Pycharm related folder that contains project specific settings files
# Checklist


Kludex:
We do not accept PRs for specific environments. Add `.idea` to your global `.gitignore`.

PR #2493, opened 2024-02-07
Add type hints to `test_testclient.py`

TechNiick:
# Summary

Type annotation added to `test_testclient.py`

# Checklist

### DIFF
--- a/tests/test_testclient.py
+++ b/tests/test_testclient.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import itertools
 import sys
 from asyncio import Task, current_task as asyncio_current_task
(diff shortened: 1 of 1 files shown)


tests/test_testclient.py
```diff
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import itertools
 import sys
-from asyncio import current_task as asyncio_current_task
+from asyncio import Task, current_task as asyncio_current_task
 from contextlib import asynccontextmanager
-from typing import Callable
+from typing import Any, AsyncGenerator, Awaitable, Callable, Union
```
Kludex: Remove Any, please.

Use `|` instead of `Union`.

Kludex (changes requested):

tests/test_testclient.py
TechNiick: In this file, Any is used only here:

`def current_task() -> Task[Any] | trio.lowlevel.Task:`

I couldn't figure it out what type should be there instead of Any. Could you help me on this one please?

Kludex (approved):

PR #2471, opened 2024-02-04
Add type hints to `test_applications.py`

TechNiick:
# Summary

Related to [this](https://github.com/encode/starlette/pull/2445)
Type annotation added to `test_applications.py`

# Checklist

- [ ] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.

PR #2496, opened 2024-02-09
Skip magic trailing comma

jonathanberthias:
# Summary

There have been many comments asking contributors to merge lines where possible such as [this one](https://github.com/encode/starlette/pull/2471#discussion_r1477417360). There is an option available in the Ruff formatter to do this automatically, at the expense of ignoring magic trailing commas: [`skip-magic-trailing-comma`](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma), and a sibling `isort` option for imports: [`split-on-trailing-comma`](https://pycqa.github.io/isort/docs/configuration/options.html#split-on-trailing-comma).

This change reformats a lot of code, so I would understand if you prefer not merging this.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,10 @@ select = ["E", "F", "I"]
 
 [tool.ruff.lint.isort]
 combine-as-imports = true
+split-on-trailing-comma = false
+
+[tool.ruff.format]
+skip-magic-trailing-comma = true
 
 [tool.mypy]
 strict = true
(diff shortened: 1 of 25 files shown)


Kludex:
We can merge this and, add a `.git-blame-ignore-revs` in a subsequent commit.

Kludex:
Yeah, let's not merge this. 👍 

Thanks for letting me know about those flags, anyway.

PR #2501, opened 2024-02-10
feat: add partitioned attribute to cookie

Flecart:
# Summary

Added the `partitioned` attribute.

See https://github.com/encode/starlette/discussions/2500.

# Checklist

# Why was this draft?

Wait for https://github.com/python/cpython/pull/112714 to get merged. (it was merged in Jan 2025)

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -31,7 +31,7 @@ async def app(scope, receive, send):
 
 Starlette provides a `set_cookie` method to allow you to set cookies on the response object.
 
-Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/", domain=None, secure=False, httponly=False, samesite="lax")`
+Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/", domain=None, secure=False, httponly=False, samesite="lax", partitioned=False)`
 
 * `key` - A string that will be the cookie's key.
 * `value` - A string that will be the cookie's value.
@@ -42,6 +42,7 @@ Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/"
 * `secure` - A bool indicating that the cookie will only be sent to the server if request is made using SSL and the HTTPS protocol. `Optional`
 * `httponly` - A bool indicating that the cookie cannot be accessed via JavaScript through `Document.cookie` property, the `XMLHttpRequest` or `Request` APIs. `Optional`
 * `samesite` - A string that specifies the samesite strategy for the cookie. Valid values are `'lax'`, `'strict'` and `'none'`. Defaults to `'lax'`. `Optional`
+* `partitioned` - A bool indicates to user agents that these cross-site cookies should only be available in the same top-level context that the cookie was first set in. `Optional`
 
 #### Delete Cookie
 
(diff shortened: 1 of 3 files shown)


Kludex:
Closing this since `SimpleCookie` doesn't support it. When that is supported, PR welcome.

Flecart:
@Kludex can we keep this open as draft like https://github.com/python/cpython/pull/112714 in the meantime?

Kludex:
I prefer to keep it closed for organization. When that is merged, we can continue the work here.

estyrke:
The CPython PR is merged now (targeted at 3.14 as far as I can tell). Can this be reopened or do we need a new PR?

Flecart:
> The CPython PR is merged now (targeted at 3.14 as far as I can tell). Can this be reopened or do we need a new PR?

I'd be happy if this is reopened :)

aaronted009:
> > The CPython PR is merged now (targeted at 3.14 as far as I can tell). Can this be reopened or do we need a new PR?
> 
> I'd be happy if this is reopened :)

@Flecart , wish granted 😅 . I guess conflicts could be resolved here so this can move forward?

Flecart:
> > > The CPython PR is merged now (targeted at 3.14 as far as I can tell). Can this be reopened or do we need a new PR?
> > 
> > 
> > I'd be happy if this is reopened :)
> 
> @Flecart , wish granted 😅 . I guess conflicts could be resolved here so this can move forward?

Sry for the delay, coming!

Flecart:
Hello people, 
I added the change for partitioned cookies.

This is a **breaking change**, I currently do not know a cleaner version other than checking the version we are in currently, if you have better ways, that'll be well accepted.

Ifrunning set cookie for python version earlier than 3.14 just returns an error like:

```python
    def __setitem__(self, K, V):
        K = K.lower()
        if not K in self._reserved:
>           raise CookieError("Invalid attribute %r" % (K,))
E           http.cookies.CookieError: Invalid attribute 'partitioned'
```

I have tested locally, it passes tests with python version 3.14.0a7, and returns that error for python 3.11.8

starlette/responses.py
```diff
@@ -122,6 +124,9 @@ def set_cookie(
                 "none",
             ], "samesite must be either 'strict', 'lax' or 'none'"
             cookie[key]["samesite"] = samesite
+        if sys.version_info >= (3, 14) and partitioned:
+            cookie[key]["partitioned"] = True  # pragma: no cover
```
Kludex: We want an error if `partitioned` is used on <3.14.

Flecart: Can you help me point what line is not coveredby the tests?

Flecart:
Hello people.
I have added no cover for the whole test that checks for raising errors. I've looked up that in theory we should separate which tests are run for each version and then combine the coverage reports, but this is not compatible with the current testing format. The easiest solution I got was just to ignore coverage so that it works for all versions. Better ideas are welcome.

Flecart:
@Kludex reminder to check this when you have time. Thank you very much!

Kludex (approved):

docs/responses.md
```diff
@@ -42,6 +42,7 @@ Signature: `Response.set_cookie(key, value, max_age=None, expires=None, path="/"
 * `secure` - A bool indicating that the cookie will only be sent to the server if request is made using SSL and the HTTPS protocol. `Optional`
 * `httponly` - A bool indicating that the cookie cannot be accessed via JavaScript through `Document.cookie` property, the `XMLHttpRequest` or `Request` APIs. `Optional`
 * `samesite` - A string that specifies the samesite strategy for the cookie. Valid values are `'lax'`, `'strict'` and `'none'`. Defaults to `'lax'`. `Optional`
+* `partitioned` - A bool indicates to user agents that these cross-site cookies should only be available in the same top-level context that the cookie was first set in. `Optional`
```
Kludex: We should mention that it's only available on Python 3.14+.

Kludex (changes requested):

Kludex (approved):

PR #2502, opened 2024-02-10
Create types module inside tests

TechNiick:
# Summary

This PR is related to [this](https://github.com/encode/starlette/pull/2445) issue. 
Moved all types used inside tests module, in a types module.

# Checklist


tests/test_types/types.py
Kludex, line 1: We don't need the folder `test_types`.

tests/types.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Awaitable, Callable, Dict, Iterable
+
+from starlette.responses import Response
+from starlette.testclient import TestClient
+
+TestClientFactory = Callable[..., TestClient]
```
Kludex: Can this be a protocol?

Kludex (changes requested):

tests/types.py
```diff
@@ -0,0 +1,13 @@
...
+
+WSGIResponse = Iterable[bytes]
+StartResponse = Callable[..., Any]
+Environment = Dict[str, Any]
+
+AsyncEndpoint = Callable[..., Awaitable[Response]]
+SyncEndpoint = Callable[..., Response]
```
Kludex: Can those types not be in `types.py`? They are only used once, right?

TechNiick: Yes, that is right, I have moved them back in their files.

tests/types.py
TechNiick: I have made the changes but I had to change the naming to `ClientFactoryProtocol` because using `TestClientFactory` leads to another issue.

If i use the name starting with `Test` pytest is trying to collect it as if it were a test case or test class, but it is not intended to be either.

I tried two things for this:
1. add `# pytest: no collect` but this did not fix the issue
2. add `__test__ = False` but this led to some other errors:
`All protocol members must have explicitly declared types`
`Incompatible return value type (got "partial[TestClient]", expected "TestClientFactory")  [return-value]` 
`"partial" is missing following "TestClientFactory" protocol member` inside `conftest.py`

Is that naming ok or should I try another fix or keep it without protocol?

tests/types.py
```diff
@@ -0,0 +1,16 @@
...
+    from typing_extensions import ParamSpec
+
+from starlette.testclient import TestClient
+
+P = ParamSpec("P")
+
+
+class ClientFactoryProtocol(Protocol):  # pragma: no cover
+    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> TestClient:
+        ...
```
Kludex: I meant this:

```suggestion
class TestClientFactory(Protocol):  # pragma: no cover
    def __call__(self, app: AppType, etc...etc...) -> TestClient:
        ...
```

tests/types.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TestClientFactory(Protocol):  # pragma: no cover
+    __test__ = False  # type: ignore
+
+    def __call__(
+        self,
+        app: AppType,
+        *args: P.args,
+        **kwargs: P.kwargs,
```
Kludex: I think you should rewrite all parameters TestClient receives instead.

Kludex: Without `backend` and `backend_options`.

Kludex (approved):

PR #2525, opened 2024-02-22
Revert "Turn `scope["client"]` to `None` on `TestClient` (#2377)"

aminalaee:
This reverts commit 483849a466a2bfc121f5a367339e1aa3ed20344b.

# Summary

# Checklist


Kludex:
- I've reopened https://github.com/encode/starlette/issues/2295.

Let's try to find another solution for that problem. It looks like removing the client brought more unsatisfaction than happiness.

Kludex (approved):

PR #2529, opened 2024-02-29
Use try/except instead of checking environment version.

JustinBacher:
# Summary

Instead calling `sys.version_info` and checking the version before importing it's better to just use a try/except block. Saves having to even check the version altogether.

# Checklist

### DIFF
--- a/starlette/_utils.py
+++ b/starlette/_utils.py
@@ -3,23 +3,21 @@ from __future__ import annotations
 import asyncio
 import functools
 import re
-import sys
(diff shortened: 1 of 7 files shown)


JustinBacher:
Nevermind I just took a look at [mypy's docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module). That must be why you did that. Just seemed that using the try/except would be better.

PR #2533, opened 2024-03-03
Version 0.37.2

Kludex:
https://github.com/encode/starlette/compare/0.37.1...master


karpetrosyan:
Maybe we should pin the upper version of httpx to the current version?
Take a look at #2534

tarsil (approved):

Kludex:
Since we can't save previous starlette versions, I don't think we should pin the max version.

docs/release-notes.md
```diff
@@ -1,3 +1,16 @@
+## 0.37.2
+
+March 3, 2024
```
Kludex:
```suggestion
March 5, 2024
```

PR #2542, opened 2024-03-14
Remove processed background task objects

jainal09:
Remove already processed background task objects once a task is processed

# Summary

When a background task is processed it is not removed from the array and it is very difficult to implement features in which we want to find length of queued background tasks

This can be used to freeze shutdowns when handling OS signals for enabling graceful shutdowns. Thus, if there are already enqueued background tasks wait and only then stop the server once all the tasks are finished.

# Checklist

### DIFF
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -41,5 +41,7 @@ class BackgroundTasks(BackgroundTask):
         self.tasks.append(task)
 
     async def __call__(self) -> None:
-        for task in self.tasks:
+        for task in self.tasks.copy():
             await task()
+            if task in self.tasks:
+                self.tasks.remove(task)


starlette/background.py
```diff
@@ -41,5 +41,7 @@ def add_task(
         self.tasks.append(task)
 
     async def __call__(self) -> None:
-        for task in self.tasks:
+        for task in self.tasks.copy():
             await task()
+            if task in self.tasks:
+                self.tasks.remove(task)
```
adriangb, line 47: Why would the task ever not be in tasks?

jainal09: The reason, for this is because once the task is processed in `await task()`, why is there a need to keep it in the list. I am removing it because, I want to achieve something like this

If I want to fetch the current task queue length

using something like below

```python
class QueueOut(BaseModel):
    task_lenth: int

@app.get("/get_queue_length/", response_model=QueueOut)
async def read_task():
    return QueueOut(task_lenth=len(tasks.tasks))
    
  ```
This is helpful for graceful shutdown of the application in scenarios where a lots of bg task are enqueued and now you intercepted a SIGTERM or SIGKILL which sends a termination signal but, the server will not be stopped until all the enqueued tasks are processed.

Kludex:
Please create a discussion.

jainal09:
Sure Done: https://github.com/encode/starlette/discussions/2546

PR #2545, opened 2024-03-15
refactor: Use @staticmethod decorator to optimize performance

0xsirsaif:
refactor: Use @staticmethod decorator to optimize performance by avoiding instantiation of bound methods for each object

- Bound methods incur overhead as they are objects themselves

# Summary

However, using the @staticmethod decorator does not have obvious preferable use cases. But I find that as long as methods that do not use self in their body create a bound object, that we could avoid by using @staticmethod

# Checklist

### DIFF
--- a/starlette/config.py
+++ b/starlette/config.py
@@ -117,7 +117,8 @@ class Config:
             return self._perform_cast(key, default, cast)
         raise KeyError(f"Config '{key}' is missing, and has no default.")
 
-    def _read_file(self, file_name: str | Path) -> dict[str, str]:
+    @staticmethod
+    def _read_file(file_name: str | Path) -> dict[str, str]:
         file_values: dict[str, str] = {}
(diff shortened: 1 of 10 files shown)


Kludex:
This doesn't look like it will bring much benefit.

alex-oleshkevich:
I agree with @Kludex. 
See https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators
> Never use @staticmethod unless forced to in order to integrate with an API defined in an existing library. Write a module level function instead.

See https://mail.python.org/pipermail/python-ideas/2016-July/041189.html 
> Honestly, staticmethod was something of a mistake – I was trying to do something like Java class methods but once it was released I found what was really needed was classmethod. But it was too late to get rid of staticmethod.

0xsirsaif:
Ok, no problem. Thanks @alex-oleshkevich @Kludex

PR #2547, opened 2024-03-19
Add `ContentType` generic to `Response`

Kludex:
- Retry of https://github.com/encode/starlette/pull/1534.

At the time that PR was implemented the `default=` parameter was not available on `TypeVar`, and mypy didn't support it.

Now it's possible to achieve what that PR intended.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=3.4.0,<5",
-    "typing_extensions>=3.10.0; python_version < '3.10'",
+    "typing_extensions>=4.4.0; python_version < '3.13'",
 ]
 
 [project.optional-dependencies]
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 # Testing
 coverage==7.4.3
 importlib-metadata==7.0.1
-mypy==1.8.0
+mypy==1.9.0
 ruff==0.1.15
 typing_extensions==4.10.0
 types-contextvars==2.4.7.3
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -4,6 +4,7 @@ import http.cookies
 import json
 import os
 import stat
+import sys
 import typing
 import warnings
 from datetime import datetime
@@ -21,14 +22,21 @@ from starlette.concurrency import iterate_in_threadpool
 from starlette.datastructures import URL, MutableHeaders
 from starlette.types import Receive, Scope, Send
 
+if sys.version_info >= (3, 13):  # pragma: no cover
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
 
-class Response:
+Content = TypeVar("Content", default=typing.Any)
+
+
+class Response(typing.Generic[Content]):
     media_type = None
     charset = "utf-8"
 
     def __init__(
         self,
-        content: typing.Any = None,
+        content: Content | None = None,
         status_code: int = 200,
         headers: typing.Mapping[str, str] | None = None,
         media_type: str | None = None,
@@ -41,7 +49,7 @@ class Response:
         self.body = self.render(content)
         self.init_headers(headers)
 
-    def render(self, content: typing.Any) -> bytes:
+    def render(self, content: Content | None) -> bytes:
         if content is None:
             return b""
         if isinstance(content, bytes):
@@ -170,12 +178,12 @@ class PlainTextResponse(Response):
     media_type = "text/plain"
 
 
-class JSONResponse(Response):
+class JSONResponse(Response[Content]):
     media_type = "application/json"
 
     def __init__(
         self,
-        content: typing.Any,
+        content: Content,
         status_code: int = 200,
         headers: typing.Mapping[str, str] | None = None,
         media_type: str | None = None,
@@ -183,7 +191,7 @@ class JSONResponse(Response):
     ) -> None:
         super().__init__(content, status_code, headers, media_type, background)
 
-    def render(self, content: typing.Any) -> bytes:
+    def render(self, content: Content | None) -> bytes:
         return json.dumps(
             content,
             ensure_ascii=False,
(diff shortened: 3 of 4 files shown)


pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=3.4.0,<5",
-    "typing_extensions>=3.10.0; python_version < '3.10'",
+    "typing_extensions>=4.4.0; python_version < '3.13'",
```
Kludex, line 30: `default=` was introduced in this version: https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeVar

requirements.txt
```diff
@@ -4,7 +4,7 @@
 # Testing
 coverage==7.4.3
 importlib-metadata==7.0.1
-mypy==1.8.0
+mypy==1.9.0
```
Kludex, line 7: Last release supports `default=`.

starlette/responses.py
```diff
@@ -41,7 +49,7 @@ def __init__(
         self.body = self.render(content)
         self.init_headers(headers)
 
-    def render(self, content: typing.Any) -> bytes:
+    def render(self, content: Content | None) -> bytes:
```
alex-oleshkevich, line 52: `|` was introduced in 3.10.

Kludex: We use `__future__.annotations`, and the pipeline passes. 👀

alex-oleshkevich (approved):

starlette/responses.py
```diff
@@ -207,9 +215,9 @@ def __init__(
         self.headers["location"] = quote(str(url), safe=":/%#?=@[]!$&'()*+,;")
 
 
-Content = typing.Union[str, bytes]
-SyncContentStream = typing.Iterable[Content]
-AsyncContentStream = typing.AsyncIterable[Content]
+_Content = typing.Union[str, bytes]
```
adriangb, line 218: Just to confirm, this is a breaking change right?

adriangb (commented):
I think JsonResponse and others don’t actually accept Any, they accept a limited set of types. Is that codified in this PR? If it is I’m missing it.

starlette/responses.py
Kludex: Oh yeah, it is...

Kludex:
> I think JsonResponse and others don’t actually accept Any, they accept a limited set of types. Is that codified in this PR? If it is I’m missing it.

You are right, but I'm not changing that in this PR. And I think there was an attempt to change the type, but it failed, and we had to revert.

starlette/responses.py
```diff
@@ -21,14 +22,21 @@
...
+if sys.version_info >= (3, 13):  # pragma: no cover
+    from typing import TypeVar
+else:
+    from typing_extensions import TypeVar
 
-class Response:
+Content = TypeVar("Content", default=typing.Any)
+
+
+class Response(typing.Generic[Content]):
```
adriangb, line 33: It seems to me like the reality is that the default should be `str | bytes` given what `Response` accepts. Then subclasses like `StreamingResponse` should override this generic with whatever they accept. If we make the default `typing.Any` then `Response` is not correctly typed. Since we're setting the default in this PR we might as well choose an appropriate value.

Kludex: I'm not changing this in this PR. There was a previous attempt to change the type to `str | bytes | None`, and it was reverted - I don't remember the reason.

adriangb: Ok. It would be good to find that reason.

Kludex: https://github.com/encode/starlette/issues/2242

adriangb: Ok I see. So yes the right thing to do would be to make a `BaseResponse` and make `Response` just one of the many concrete classes:

```python

_ResponseDataT = TypeVar('_ResponseDataT')

class BaseResponse(Generic[_ResponseDataT]):
    def render(self, data: _ResponseDataT | None) -> bytes:
        raise NotImplementedError

class Response(BaseResponse[str | bytes]):
    ...

_JsonData = str | bytes | float | int | None | ...

class JsonResponse(BaseResponse[_JsonData]):
    ...
```

abersheeran: The JSONResponse should be of type Any, because the user may have customized the serializer to handle other types of data.

Also, I don't think that using "render" as a method for BaseResponse is a good idea. Methods like StreamResponse and FileResponse will not use this method at all. In Baize, I have designed corresponding classes for inheritance, which may bring some inspiration to everyone. https://github.com/abersheeran/baize/blob/master/baize/asgi/responses.py

adriangb: Thank you for sharing @abersheeran !

Kludex: > Also, I don't think that using "render" as a method for BaseResponse is a good idea.

What do you propose I do here @abersheeran ?

abersheeran: > > Also, I don't think that using "render" as a method for BaseResponse is a good idea.
> 
> What do you propose I do here @abersheeran ?

I think we should derive a SmallContentResponse from BaseResponse and include the `render` method in this class. Similar to HTMLResponse, JSONResponse inherits it, while other Responses that do not require render directly inherit BaseResponse.

Kludex: But the `Response` class does have the `.render()` method... So I don't see how I'd that? I can't make the classes stop inheriting `Response`, because that would be a breaking change.

abersheeran: Yes, this is also what I think is unreasonable about Starlette. For example, StreamResponse and FileReponse do not use render, but they inherit this method.

Kludex: Are you suggesting adding a breaking change then?

abersheeran: Yes. 😊

adriangb: Can we do some naming to make it a non-breaking change or minimize the breakage? We could leave the render() method on the base class but mark it as deprecated vi a warning. Adding a BaseResponse class is not a breaking change. I think removing the render() method from eg StreamingResponse (because it inherits from BaseResponse and not Response) would also be okay.

PR #2553, opened 2024-03-25
Add missing compresslevel parameter on docs

junah201:
# Summary

The compresslevel parameter was added in #1128.
But, it was not added to the document, so I added an explanation on docs.

# Checklist

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -194,7 +194,7 @@ app = Starlette(routes=routes, middleware=middleware)
 The following arguments are supported:
 
 * `minimum_size` - Do not GZip responses that are smaller than this minimum size in bytes. Defaults to `500`.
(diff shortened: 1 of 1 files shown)


docs/middleware.md
```diff
@@ -194,6 +194,7 @@ app = Starlette(routes=routes, middleware=middleware)
 The following arguments are supported:
 
 * `minimum_size` - Do not GZip responses that are smaller than this minimum size in bytes. Defaults to `500`.
+* `compresslevel` - will be used during gzip compression. It is an integer ranging from 1 to 9. Defaults to `9`. Lower value results in faster compression but larger file sizes, while higher value results in slower compression but smaller file sizes.
```
```suggestion
* `compresslevel` - Used during GZip compression. It is an integer ranging from 1 to 9. Defaults to `9`. Lower value results in faster compression but larger file sizes, while higher value results in slower compression but smaller file sizes.
```

Kludex (approved):

Kludex (approved):

Kludex:
Thanks @junah201

PR #2568, opened 2024-04-12
Fail fast on invalid Jinja2Template instantiation parameters

pdelagrave:
# Summary
Calling `Jinja2Template()` with both `directory` and `env` shouldn't be allowed. When both parameters were used, the passed `env` was silently ignored in favor of creating a new one with the provided `directory` and the deprecated `env_options`.

# Checklist


Kludex (approved):
This reflects the `override`s. 👍 

I think we should probably remove this assertion at some point... The type checker is able to catch this anyway.

PR #2572, opened 2024-04-15
Update SessionMiddleware and TrustedHostMiddleware Documentation

soulee-dev:
# Summary
This PR updates the documentation for `SessionMiddleware` and `TrustedHostMiddleware` to reflect the current code more accurately.

## Changes
1. SessionMiddleware:

- Added documentation for the path parameter which specifies the path set for the session cookie. It defaults to "/".

2. TrustedHostMiddleware:
- Added documentation for the www_redirect parameter. When set to True, this option redirects requests to non-www versions of allowed hosts to their www counterparts. This parameter defaults to True.

# Checklist


Kludex (approved):

Kludex:
Thanks! :)

PR #2576, opened 2024-04-22
Allow returning memoryview in StreamingResponse

bschoenmaeckers:
# Summary

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -207,7 +207,7 @@ class RedirectResponse(Response):
         self.headers["location"] = quote(str(url), safe=":/%#?=@[]!$&'()*+,;")
 
 
-Content = typing.Union[str, bytes]
+Content = typing.Union[str, bytes, memoryview]
 SyncContentStream = typing.Iterable[Content]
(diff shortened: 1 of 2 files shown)


adriangb (approved):
Makes sense to me!

Kludex:
What's the use case? To create the `memoryview` you need `bytes` or `bytearray` anyway...

bschoenmaeckers:
Memoryview allows you to efficiently return a slice of the original bytes value without copying. This is especially convenient when you want to strip some prefix or header of some value in memory you don’t control.

adriangb:
Yup I'd say the use cases are varied, it allows moving around bytes with zero copying.

Kludex:
This is a PR without description, and there's a single discussion that mentioned this feature request (that is 4 years old) without reactions.

Also, this can be easily done by overriding the `stream_response` method. And that's what we've always recommended.

adriangb:
Feel free to revert it if you don’t agree. To me it made sense and was simple.

danielballan:
Thanks for this! Useful when transmitting buffers from numpy array without copying.

PR #2580, opened 2024-04-24
Fix documentation on client address type

julien4215:
# Summary

Updates the documentation to match the [ASGI specs](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) and the type in the [source code](https://github.com/encode/starlette/blob/f305f003c27b768ae1f5cf417b158791b10b10dc/starlette/requests.py#L147). I also changed the comment in the source code because I think there was a mistake.

This was changed in the PR https://github.com/encode/starlette/pull/1462 but the documentation was not updated.

This was also previously discussed here https://github.com/encode/starlette/discussions/2244.

# Checklist

### DIFF
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -61,7 +61,6 @@ For example: `request.path_params['username']`
 #### Client Address
 
 The client's remote address is exposed as a named two-tuple `request.client` (or `None`).
-`request.client` may be `None` but each item in the tuple will not be `None`.
 
 The hostname or IP address: `request.client.host`
 
--- a/docs/requests.md
+++ b/docs/requests.md
@@ -60,7 +60,7 @@ For example: `request.path_params['username']`
 
 #### Client Address
 
-The client's remote address is exposed as a named two-tuple `request.client`.
+The client's remote address is exposed as a named two-tuple `request.client` (or `None`).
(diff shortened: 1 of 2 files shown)


docs/requests.md
```diff
@@ -61,7 +61,7 @@ For example: `request.path_params['username']`
 #### Client Address
 
 The client's remote address is exposed as a named two-tuple `request.client`.
```
Kludex: I think the sentence got too complex below... Can we just do this, or something?
```suggestion
The client's remote address is exposed as a named two-tuple `request.client` (or `None`).
```

docs/requests.md
```diff
@@ -60,8 +60,8 @@ For example: `request.path_params['username']`
 
 #### Client Address
 
-The client's remote address is exposed as a named two-tuple `request.client`.
-Either item in the tuple may be `None`.
+The client's remote address is exposed as a named two-tuple `request.client` (or `None`).
+`request.client` may be `None` but each item in the tuple will not be `None`.
```
Kludex: I meant with this being removed
```suggestion
```

julien4215: My bad, I forgot that I had still to remove the line

Kludex (approved):

PR #2583, opened 2024-04-30
Send 404 instead of 500 when filename requested is too long on `StaticFiles`

Jaza:
# Summary

If a static file with a really long name is requested, then it causes an `OSError` to be raised saying "Filename too long". This error currently gets propagated, resulting in a 500 response. In such a case, the name is too long for the underlying OS to handle, therefore a file with that name couldn't possibly exist, therefore it would be more appropriate to not propagate the error and to return a 404 response.

# Checklist

Note: I feel that the existing documentation is adequate, it already states:

> Static files will respond with "404 Not found" or "405 Method not allowed" responses for requests which do not match.

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
+import errno
 import importlib.util
 import os
-import re
 import stat
 import typing
 from email.utils import parsedate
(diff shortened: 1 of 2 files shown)


Kludex (approved):
Thanks @Jaza !

Relying on regex is usually not a good idea for exceptions. I've used the `errno` instead.

I've also changed the test to use the max path on each OS that runs the test.

Jaza:
Thanks @Kludex ! TIL you can be more specific about `OSError` s with `errno` . And I guess your google-fu (or your background knowledge of such things) is superior to mine, I did search for a way to programatically get the max path size, but I didn't end up finding `PC_PATH_MAX` , so TIL++.

PR #2598, opened 2024-05-21
Adding timeout to WebSocket session receive on test client

serans1:
# Summary

# Checklist

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -152,9 +152,9 @@ May raise `starlette.websockets.WebSocketDisconnect` if the application does not
 
 #### Receiving data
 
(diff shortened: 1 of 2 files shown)


serans1:
Hi, 
Can you share why this was closed ? 
Testing WebSocket currently hangs if there is no response

serans1:
Hi,
Can you share why this was closed ?
Testing WebSocket currently hangs if there is no response

Kludex:
You didn't provide any description or look for previous PRs and issues about the subject.

serans1:
I get your point, I see it here
https://github.com/encode/starlette/discussions/2195
for reference, this PR is right and timeout should be added without pytest
extensions.
No intention in getting into holy wars, just yet another take on the
matter.

PR #2608, opened 2024-06-02
add ETagMiddleware

keakon:
# Summary

Implement the ETagMiddleware to drop the response body if the content not changed.

# Checklist


Kludex:
Hi @keakon ,

Thanks for the PR, but for this kind of middleware, we prefer to have the author create their own package, and if they want, they can create a PR on our documentation to add that on the "Third party packages" section. 🙏

PR #2616, opened 2024-06-10
fix: pathsend handling in middlewares

### LINKED ISSUE #2613: pathsend causing issues with BaseHTTPMiddleware
Hi guys

This is a follow up of https://github.com/emmett-framework/granian/issues/320 

I'm not entirely sure if this is really a bug of starlette, or more of a feature that's missing or so.

On my staging environment, I did receive an assertion error:

```
 File "/tmp/8dc89326e92b51b/antenv/lib/python3.12/site-packages/starlette/middleware/base.py", line 173, in body_stream
2024-06-10T09:58:53.052818649Z     assert message["type"] == "http.response.body"
``` 

As I could not reproduce it locally, I monkey-patched starlette and printed an error message with details:

```
2024-06-10T09:58:53.052814649Z   File "/tmp/8dc89326e92b51b/antenv/lib/python3.12/site-packages/starlette/middleware/base.py", line 173, in body_stream
2024-06-10T09:58:53.052818649Z     assert message["type"] == "http.response.body", f"Unexpected message of type {message['type']} from class {type(self)}: {message}"  # noqa: E501
2024-06-10T09:58:53.052821649Z     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-06-10T09:58:53.052824449Z AssertionError: Unexpected message of type http.response.pathsend from class <class 'starlette.middleware.base.BaseHTTPMiddleware'>: {'type': 'http.response.pathsend', 'path': 'frontend_app/output/index.html'}
```

I think this happens because my server does support the pathsend extension, however middlware that act's on it and just derives from BaseHTTPMiddleware will fail, since it cannot handle this. 

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/2613">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/2613/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

gi0baro:
# Summary

Closes #2613 

It seems that when I introduced support for pathsend in #2435 I didn't check the included middlewares to also support the relevant ASGI messages. Affected Starlette versions 0.36.0 and onwards when running on a server supporting pathsend.

<s>Note: this is still in draft since I'd like to plan some tests for the involved code; also the proposed solution isn't necessarily the best one, probably someone like @Kludex might review this and give his feedbacks before proceeding further.</s>
Update: I added tests just to verify the code. In case you're unhappy with the implementation, we can refactor that and keep the tests.

# Checklist

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -170,6 +170,10 @@ class BaseHTTPMiddleware:
             async def body_stream() -> typing.AsyncGenerator[bytes, None]:
                 async with recv_stream:
                     async for message in recv_stream:
+                        if message["type"] == "http.response.pathsend":
+                            # with pathsend we don't need to stream anything
+                            yield message
+                            break
                         assert message["type"] == "http.response.body"
                         body = message.get("body", b"")
                         if body:
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -15,6 +15,9 @@ RequestResponseEndpoint = typing.Callable[[Request], typing.Awaitable[Response]]
 DispatchFunction = typing.Callable[
     [Request, RequestResponseEndpoint], typing.Awaitable[Response]
 ]
+BodyStreamGenerator = typing.AsyncGenerator[
+    typing.Union[bytes, typing.MutableMapping[str, typing.Any]], None
+]
 T = typing.TypeVar("T")
 
 
@@ -167,7 +170,7 @@ class BaseHTTPMiddleware:
 
             assert message["type"] == "http.response.start"
 
-            async def body_stream() -> typing.AsyncGenerator[bytes, None]:
+            async def body_stream() -> BodyStreamGenerator:
                 async with recv_stream:
                     async for message in recv_stream:
                         if message["type"] == "http.response.pathsend":
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -207,7 +207,7 @@ class RedirectResponse(Response):
         self.headers["location"] = quote(str(url), safe=":/%#?=@[]!$&'()*+,;")
 
 
-Content = typing.Union[str, bytes, memoryview]
+Content = typing.Union[str, bytes, memoryview, typing.MutableMapping[str, typing.Any]]
 SyncContentStream = typing.Iterable[Content]
 AsyncContentStream = typing.AsyncIterable[Content]
 ContentStream = typing.Union[AsyncContentStream, SyncContentStream]
@@ -254,7 +254,7 @@ class StreamingResponse(Response):
                 should_close_body = False
                 await send(chunk)
                 break
-            if not isinstance(chunk, (bytes, memoryview)):
+            if isinstance(chunk, str):
                 chunk = chunk.encode(self.charset)
             await send({"type": "http.response.body", "body": chunk, "more_body": True})
 
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -247,12 +247,19 @@ class StreamingResponse(Response):
                 "headers": self.raw_headers,
             }
         )
+        should_close_body = True
         async for chunk in self.body_iterator:
+            if isinstance(chunk, dict):
+                # We got an ASGI message which is not response body (eg: pathsend)
+                should_close_body = False
+                await send(chunk)
+                break
             if not isinstance(chunk, (bytes, memoryview)):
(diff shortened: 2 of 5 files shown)


starlette/middleware/base.py
```diff
@@ -167,9 +170,13 @@ async def coro() -> None:
 
             assert message["type"] == "http.response.start"
 
-            async def body_stream() -> typing.AsyncGenerator[bytes, None]:
+            async def body_stream() -> BodyStreamGenerator:
                 async with recv_stream:
                     async for message in recv_stream:
+                        if message["type"] == "http.response.pathsend":
```
aersam, line 176: Hi there! wouldn't it make sense to just read the file and stream it through here? Would be better for compatibility, no? Or maybe have some flag like "supports_message_pass_through"?

gi0baro: Reading the file from Python would be just the standard ASGI behaviour and defeat the whole `pathsend` idea (see for ref https://asgi.readthedocs.io/en/latest/extensions.html#path-send).

About the flag: I'm not sure I get the point around it. How should we evaluate such flag? `pathsend` messages will be returned by Starlette only if the relevant extension is present in the ASGI scope as the server supports it.

aersam: I was thinking about random middlewares out there, deriving from this BaseMiddleware that do not know pathsend. To keep those running, returning the message instead of bytes could be opt-in via a flag/class property

gi0baro: @aersam but even if you subclass `BaseMiddleware`, there's no way you can interact with that code. If the middleware completely overrides `__call__` then it should be compliant with ASGI messages IMHO.

gi0baro:
@Kludex any updates on this?

starlette/responses.py
Kludex, line 1: Are the changes here just because the `StreamingResponse` is used in the `BaseHTTPMiddleware`?

gi0baro: Yes. A Better implementation would probably be to refactor `BaseHTTPMiddleware`, but it is something I am not confident to do by myself given my knowledge of the codebase is still limited.

Kludex: We are not making changes to the rest of the code source to make `BaseHTTPMiddleware` happy.

My recommendation for Granian users, and/or any server supporting `http.response.pathsend` is to replace `BaseHTTPMiddleware` for pure ASGI middleware. I don't want us to add more complexity to that code. That piece of code had too many issues over time.

gi0baro: I understand your position.

I think there are 2 major themes here though:
- `StreamingResponse` doesn't support any message type other than `http.response.body`, which make sense given the intent of streaming responses, so unless an additional *body* message type is added to ASGI this is future-proof. But, since `BaseHTTPMiddleware` depends on that, this also means any application using that middleware or a derivate middleware won't support any other ASGI message in responses, which is something you might want to change to be future-proof on ASGI spec changes.
- As of today Starlette states (due to the changes I made in #2435) to support `http.response.pathsend`, which is only partially true: as soon as the route involves a `BaseHTTPMiddleware`, `GZIPMiddleware` or a derivate of those in the middleware chain, then Starlette doesn't actually support pathsend.

The rationale behind this PR is not to address the first point, but rather the 2nd; as today experience with `pathsend` supportive servers with any Starlette/FastAPI application is practically dependant on which middlewares are implemented on the route returning a file response.

So given all of the above, if we can't find a proper way to address the pathsend issue, I'd rather suggest to revert #2435 and remove pathsend support completely until we find a proper way to support that ASGI extension in all the usages of Starlette and its dependants. This would sacrifice the performance gain of `pathsend` but at least we'd have the same DX across all the Starlette apps, regardless of the middlewares used, and thus prevent the whole bunch of issues related to this opened in Granian repo.

Kludex: Yeah, we can revert that one. It will be easier.

gi0baro: 👍 I'll open a different PR later today and close this one

gi0baro:
Closing this in favour of #2649, as discussed.

PR #2618, opened 2024-06-11
Refactoring code related to on_headers_finished

TonyMistark:
# Summary

# Checklist

Refactoring the code for the function:

1. Extract the logic related to building the temp_file to the _handle_upload_file function, making the logic clearer
2. Add a new init_upload_file method, which makes it easier to instantiate the UploadFile, and at the same time, a minimal code overwrite in the inherited class can achieve the best effect.

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -195,32 +195,43 @@ class MultiPartParser:
             self._current_part.field_name = _user_safe_decode(
                 options[b"name"], self._charset
             )
-        except KeyError:
+        except KeyError as e:
             raise MultiPartException(
                 'The Content-Disposition header field "name" must be ' "provided."
-            )
+            ) from e
         if b"filename" in options:
-            self._current_files += 1
-            if self._current_files > self.max_files:
-                raise MultiPartException(
-                    f"Too many files. Maximum number of files is {self.max_files}."
-                )
-            filename = _user_safe_decode(options[b"filename"], self._charset)
-            tempfile = SpooledTemporaryFile(max_size=self.max_file_size)
-            self._files_to_close_on_error.append(tempfile)
-            self._current_part.file = UploadFile(
-                file=tempfile,  # type: ignore[arg-type]
-                size=0,
-                filename=filename,
-                headers=Headers(raw=self._current_part.item_headers),
-            )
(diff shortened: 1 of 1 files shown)


Kludex:
I don't see a clear advantage of doing this... But thanks!

TonyMistark:
> I don't see a clear advantage of doing this... But thanks!

If I want to rewrite the logic of instantiating UploadFile here, I need to rewrite too many things, and the functions coupled with on_headers_finished are too many. Separating each function into a separate method makes the rewriting process much simpler, and you don't need to rewrite too much logic.

```python
class MyMultiPartParser(MultiPartParser):

    def on_headers_finished(self) -> None:
        disposition, options = parse_options_header(self._current_part.content_disposition)
        try:
            self._current_part.field_name = _user_safe_decode(options[b"name"], self._charset)
        except KeyError:
            raise MultiPartException(
                'The Content-Disposition header field "name" must be ' "provided."
            )
        if b"filename" in options:
            self._current_files += 1
            if self._current_files > self.max_files:
                raise MultiPartException(
                    f"Too many files. Maximum number of files is {self.max_files}."
                )
            filename = _user_safe_decode(options[b"filename"], self._charset)
            tempfile = SpooledTemporaryFile(max_size=self.max_file_size)
            self._files_to_close_on_error.append(tempfile)
            self._current_part.file = MyUploadFile(
                file=tempfile,  # type: ignore[arg-type]
                size=0,
                filename=filename,
                headers=Headers(raw=self._current_part.item_headers),
            )
        else:
            self._current_fields += 1
            if self._current_fields > self.max_fields:
                raise MultiPartException(
                    f"Too many fields. Maximum number of fields is {self.max_fields}."
                )
            self._current_part.file = None
```

After refactoring, you only need to rewrite the init_upload_file function, isn't that much simpler? The original on_headers_finished was too tightly coupled with too much logic, and it did not follow the single responsibility principle.

```python
class MyMultiPartParser(MultiPartParser):

    def init_upload_file(
        self,
        file: typing.BinaryIO,
        *,
        size: int | None = None,
        filename: str | None = None,
        headers: Headers | None = None,
    ):
        return MyUploadFile(
            file=file,
            size=size,
            filename=filename,
            headers=headers,
        )
```

PR #2621, opened 2024-06-16
Add Private Network Access support in CORSMiddleware

dm1sh:
# Summary

Chromium-based browsers now enforce requests from websites in public network to localhost or private network endpoints and from private to localhost to "ask for permission" from server with `OPTIONS` pre-flight request. See more in their [blog post](https://developer.chrome.com/blog/private-network-access-preflight#what_to_do_if_your_website_is_affected). It is also applied to navigation fetches (e.g. iframes, popups): [blog post](https://developer.chrome.com/blog/private-network-access-update-2024-03)

For it I added `allow_private_network` field in `CORSMiddleware` constructor as discussed in https://github.com/encode/starlette/discussions/2509. Also, I renamed former `preflight_response` to `cors_preflight_response` to add `pna_preflight_response` method to support pre-flight request in no-cors mode. Tests were also updated to cover new functionality.

I invite other contributors and maintainers to work on this pull request to be merged in future. Google is going to enforce various restrictions in their platform, so our instruments must be ready to deal with it!

# Checklist


Impro02 (approved):

Kludex (changes requested):
Is there any reason why you ignore the `access-control-request-method` on the pna logic?

Also... Do we need this separation on `pna` and `cors`? The logic is very similar.

Kludex:
Close as stale.

PR #2639, opened 2024-07-09
test: refine the test of client disconnection after reading request body

iamgodot:
# Summary

The test's original name and doc aren't really describing what it does so I refined it.

# Checklist

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -269,8 +269,8 @@ def test_request_disconnect(
 
 def test_request_is_disconnected(test_client_factory: TestClientFactory) -> None:
     """
-    If a client disconnect occurs while reading request body
-    then ClientDisconnect should be raised.
+    If a client disconnect occurs after reading request body
+    then request will be set disconnected properly.
     """
     disconnected_after_response = None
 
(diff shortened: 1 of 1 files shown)


iamgodot:
@aminalaee @Kludex Appreciate if u can take a look sometime 🙏

tests/test_requests.py
```diff
@@ -267,26 +267,26 @@ async def receiver() -> Message:
         )
 
 
-def test_request_is_disconnected(test_client_factory: TestClientFactory) -> None:
```
Kludex, line 270: The name of the test reflects that we are testing `request.is_disconnected()`.

tests/test_requests.py
```diff
@@ -267,26 +267,26 @@ async def receiver() -> Message:
...
+def test_request_disconnect_after(test_client_factory: TestClientFactory) -> None:
     """
-    If a client disconnect occurs while reading request body
-    then ClientDisconnect should be raised.
+    If a client disconnect occurs after reading request body
+    then ClientDisconnect should not be raised.
```
Kludex: This makes sense to change, I think it was copy-paste from the above tests.

Kludex: I think we can improve it tho, because it doesn't make sense to say what would not happen, we should explain what will happen succinctly.

Kludex (changes requested):

tests/test_requests.py
iamgodot: Alright, I'll keep it as before

tests/test_requests.py
iamgodot: Got ur point, I'll update

iamgodot:
@Kludex Hi i've updated the changes, let me know if anything else, thanks

Kludex (approved):

PR #2648, opened 2024-07-21
Fix `routing.get_name()` not to assume all routines have `__name__`

### LINKED ISSUE #2638: Python 3.13.0b3: AttributeError: 'functools.partial' object has no attribute '__name__'
```
$ gh repo clone encode/starlette
$ cd starlette
$ python3.13 --version
Python 3.13.0b3
$ python3.13 -m venv _e
$ . _e/bin/activate
(_e) $ pip install -r requirements.txt
(_e) $ pip install -e .
(_e) $ python -m pytest
_e/lib64/python3.13/site-packages/trio/_path.py:164: in generate_forwards
    raise TypeError(attr_name, type(attr))
E   TypeError: ('parser', <class 'module'>)
(_e) $ pip install --upgrade trio
(_e) $ python -m pytest
==================================================================================================== ERRORS ====================================================================================================
____________________________________________________________________________________ ERROR collecting tests/test_routing.py ____________________________________________________________________________________
tests/test_routing.py:135: in <module>
    Route("/", endpoint=functools.partial(partial_endpoint, "foo")),
starlette/routing.py:227: in __init__
    self.name = get_name(endpoint) if name is None else name
starlette/routing.py:103: in get_name
    return endpoint.__name__
E   AttributeError: 'functools.partial' object has no attribute '__name__'. Did you mean: '__ne__'?
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=============================================================================================== 1 error in 0.64s ===============================================================================================
```

This appears to be a new regression in Python 3.13.0b3 (vs. 3.13.0b2).

---

- [x] Initially raised as discussion #2610 – “Please file an issue in this repo for any incompatibility you encounter with 3.13.”

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/2638">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/2638/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

mgorny:
# Summary

Fix `routing.get_name()` to use the `__name__` attribute only if it is actually present, rather than assuming that all routine and class types have it, and use the fallback to class name otherwise.  This is necessary for `functools.partial()` that doesn't have a `__name__` attribute, but is recognized as a routine starting with Python 3.13.0b3. Since for older versions of Python, it would have used the fallback anyway, this doesn't really change the behavior there.

Fixes #2638

# Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. ← this is covered by existing tests already
- [ ] I've updated the documentation accordingly. ← I don't think any changes are necessary

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -99,9 +99,7 @@ def websocket_session(
 
 
 def get_name(endpoint: typing.Callable[..., typing.Any]) -> str:
-    if inspect.isroutine(endpoint) or inspect.isclass(endpoint):
-        return endpoint.__name__
-    return endpoint.__class__.__name__
+    return getattr(endpoint, "__name__", endpoint.__class__.__name__)
 
 
 def replace_params(


Kludex:
Can we link the CPython PR here?

I don't understand why change this behavior.

Kludex (dismissed):
Oh well... Seems okay anyway. If it has a `__name__`, use it, otherwise use `__class__.__name__`.

mgorny:
> Can we link the CPython PR here?

Do you mean the commit that changed `partial()` behavior?

mgorny:
> > Can we link the CPython PR here?
> 
> Do you mean the commit that changed `partial()` behavior?

The change in question is python/cpython@49e5740135670e04ae6da7e6f52dbe380655e0f1. Actually, I'm not sure if `inspect.isroutine()` change was intentional or an accidental side-effect, I'll file a bug there too, just in case.

> We should add some tests anyway.

What kind of tests do you have in mind?

Kludex (dismissed):
There's actually a test called `test_route_name` that tests this code.

All good here. Thanks @mgorny :)

Kludex:
But... To be honest... If it was unintended on CPython's side, I'd prefer to wait for them to revert the unintended changes.

musicinmybrain:
Thanks for this PR. It works nicely as a downstream patch in Fedora Rawhide, which allows me to keep building and updating Starlette. I’ll monitor https://github.com/python/cpython/issues/122087 and drop the downstream patch if this ends up being handled in CPython 3.13.0rc1.

mgorny:
@Kludex, upstream confirmed that:

> Yes, not all method descriptors (and therefore not all routines) have the `__name__` attribute. This always was so, you always should have a fallback.

So I think we should merge this anyway.

Kludex (approved):

Kludex:
> @Kludex, upstream confirmed that:
> 
> > Yes, not all method descriptors (and therefore not all routines) have the `__name__` attribute. This always was so, you always should have a fallback.
> 
> So I think we should merge this anyway.

Well... But they are fixing it for `functools.partial`, right?

mgorny:
> Well... But they are fixing it for `functools.partial`, right?

Yes, they're adding a workaround for Python 3.13. However, it will fail again in 3.14.

Kludex:
> > Well... But they are fixing it for `functools.partial`, right?
> 
> Yes, they're adding a workaround for Python 3.13. However, it will fail again in 3.14.

Well... Then let's go. 😅 

Thanks @mgorny 🙏

mgorny:
Thanks a lot!

Issue #2615, opened 2024-06-10
Python 3.13.0b2: `test_gzip_ignored_for_responses_with_encoding_set[trio]` fails with a `ValueError: I/O operation on closed file`


befeleme:
Since this is a different failure than in #2614, I decided to go for another issue.

```
___________ test_gzip_ignored_for_responses_with_encoding_set[trio] ____________

cls = <class '_pytest.runner.CallInfo'>
func = <function call_runtest_hook.<locals>.<lambda> at 0x7f0fd00a2840>
when = 'call'
reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
        cls,
        func: "Callable[[], TResult]",
        when: "Literal['collect', 'setup', 'call', 'teardown']",
        reraise: Optional[
            Union[Type[BaseException], Tuple[Type[BaseException], ...]]
        ] = None,
    ) -> "CallInfo[TResult]":
        """Call func, wrapping the result in a CallInfo.
    
        :param func:
            The function to call. Called without arguments.
        :param when:
            The phase in which the function is called.
        :param reraise:
            Exception or exceptions that shall propagate if raised by the
            function, instead of being wrapped in the CallInfo.
        """
        excinfo = None
        start = timing.time()
        precise_start = timing.perf_counter()
        try:
>           result: Optional[TResult] = func()

/usr/lib/python3.13/site-packages/_pytest/runner.py:341: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.13/site-packages/_pytest/runner.py:262: in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
/usr/lib/python3.13/site-packages/pluggy/_hooks.py:493: in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
/usr/lib/python3.13/site-packages/pluggy/_manager.py:115: in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
/usr/lib/python3.13/site-packages/_pytest/unraisableexception.py:88: in pytest_runtest_call
    yield from unraisable_exception_runtest_hook()
[...]

Issue #2614, opened 2024-06-10
Python 3.13.0b2: KeyError: 'content-type' in `test_debug_html[asyncio]` and `test_debug_html[trio]`


befeleme:
Two tests fail with Python 3.13.0b2 on a missing key 'content-type'. I run then during the RPM build on version 0.37.2.
@musicinmybrain reports it can be reproduced via a virtual environment: `pip install -e .[full]; pip install -r requirements.txt; pip install --update trio; python -m pytest`)

```
___________________________ test_debug_html[asyncio] ___________________________

test_client_factory = functools.partial(<class 'starlette.testclient.TestClient'>, backend='asyncio', backend_options={})

    def test_debug_html(test_client_factory: TestClientFactory) -> None:
        async def app(scope: Scope, receive: Receive, send: Send) -> None:
            raise RuntimeError("Something went wrong")
    
        app = ServerErrorMiddleware(app, debug=True)
        client = test_client_factory(app, raise_server_exceptions=False)
        response = client.get("/", headers={"Accept": "text/html, */*"})
        assert response.status_code == 500
>       assert response.headers["content-type"].startswith("text/html")

tests/middleware/test_errors.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Headers({}), key = 'content-type'

    def __getitem__(self, key: str) -> str:
        """
        Return a single header value.
    
        If there are multiple headers with the same key, then we concatenate
        them with commas. See: https://tools.ietf.org/html/rfc7230#section-3.2.2
        """
        normalized_key = key.lower().encode(self.encoding)
    
        items = [
            header_value.decode(self.encoding)
            for _, header_key, header_value in self._list
            if header_key == normalized_key
        ]
    
        if items:
            return ", ".join(items)
    
>       raise KeyError(key)
E       KeyError: 'content-type'

/usr/lib/python3.13/site-packages/httpx/_models.py:228: KeyError
____________________________ test_debug_html[trio] _____________________________

test_client_factory = functools.partial(<class 'starlette.testclient.TestClient'>, backend='trio', backend_options={})

    def test_debug_html(test_client_factory: TestClientFactory) -> None:
        async def app(scope: Scope, receive: Receive, send: Send) -> None:
            raise RuntimeError("Something went wrong")
    
        app = ServerErrorMiddleware(app, debug=True)
[...]

PR #2662, opened 2024-08-06
Support Python 3.13

Kludex:


starlette/middleware/gzip.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 9) -> N
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         self.send = send
         await self.app(scope, receive, self.send_with_gzip)
+        self.gzip_file.close()
```
Kludex: @graingert Is this good enough?

graingert: You might want to do it in a finally, or even better use `with`

```python
with BytesIO() as buffer, GzipFile(..., buffer) as gzip_file:
    await self.app(scope, receive, functools.partial(self.send_with_gzip, buffer, gzip_file))
```

musicinmybrain:
I tried this PR as a patch for [`python-starlette`](https://src.fedoraproject.org/rpms/python-starlette) in Fedora Rawhide, and I can confirm that it appears to fix all of the Python 3.13 test regressions from #2614 and #2615.

I do like @graingert’s suggestion in https://github.com/encode/starlette/pull/2662#discussion_r1705310571, though.

adriangb (approved):
Approving pending https://github.com/encode/starlette/pull/2662#discussion_r1705310571 being applied

graingert (approved):

PR #2668, opened 2024-08-12
Add most used objects to `__init__.py`

Kludex:
@tomchristie Is there a reason for not doing this?

### DIFF
--- a/starlette/__init__.py
+++ b/starlette/__init__.py
@@ -1,1 +1,31 @@
 __version__ = "0.38.2"
+
+from .applications import Starlette
+from .middleware import Middleware
+from .responses import (
+    HTMLResponse,
+    JSONResponse,
+    PlainTextResponse,
+    RedirectResponse,
+    Response,
+    StreamingResponse,
+)
+from .routing import Mount, Route, Router, WebSocketRoute
+from .websockets import WebSocket
+
+__all__ = (
+    "__version__",
+    "Starlette",
+    "Middleware",
+    "Response",
+    "HTMLResponse",
+    "JSONResponse",
+    "PlainTextResponse",
+    "RedirectResponse",
+    "StreamingResponse",
+    "Route",
+    "Router",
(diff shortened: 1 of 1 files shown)


lovelydinosaur:
Held back on that as wanted to encourage a consistent import style, the concise imports are nice tho. 👍

starlette/__init__.py
```diff
@@ -1 +1,31 @@
...
+    "JSONResponse",
+    "PlainTextResponse",
+    "RedirectResponse",
+    "StreamingResponse",
+    "Route",
+    "Router",
+    "Mount",
+    "WebSocketRoute",
+    "WebSocket",
+)
```
lovelydinosaur, line 31: Reviewing the complete set of documented imports here's what I've got...

```python
from starlette.authentication import requires, AuthCredentials, AuthenticationBackend, AuthenticationError, SimpleUser
from starlette.config import environ, Config
from starlette.datastructures import Headers, MutableHeaders, Secret, URL
from starlette.exceptions import HTTPException, WebSocketException
from starlette.responses import Response, FileResponse, HTMLResponse, JSONResponse, PlainTextResponse, RedirectResponse, StreamingResponse
from starlette.routing import Route, Mount, WebSocketRoute
from starlette.schemas import SchemaGenerator

from starlette.middleware.httpsredirect import HTTPSRedirectMiddleware
from starlette.middleware.trustedhost import TrustedHostMiddleware
```

Does it make sense to include the whole set of imports here, or just the subset above?

adriangb (commented):
The only con of this is that it increases import time for cases where you only use a subset of the types/features which I feel like goes against Starlette's positioning as a toolking as opposed to a framework.

Kludex:
> The only con of this is that it increases import time for cases where you only use a subset of the types/features which I feel like goes against Starlette's positioning as a toolking as opposed to a framework.

That's a good point.

PR #2671, opened 2024-08-22
Add support for ASGI `pathsend` extension

gi0baro:
# Summary

As per title, this adds support for [ASGI pathsend](https://asgi.readthedocs.io/en/latest/extensions.html#path-send) extension on `FileResponse` class for servers implementing it.

This is a re-work of #2435 combined with parts of #2616, as discussed in #2613 with @Kludex.

# Checklist


docs/middleware.md
```diff
@@ -264,6 +264,7 @@ around explicitly, rather than mutating the middleware instance.
 Currently, the `BaseHTTPMiddleware` has some known limitations:
 
 - Using `BaseHTTPMiddleware` will prevent changes to [`contextlib.ContextVar`](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar)s from propagating upwards. That is, if you set a value for a `ContextVar` in your endpoint and try to read it from a middleware you will find that the value is not the same value you set in your endpoint (see [this test](https://github.com/encode/starlette/blob/621abc747a6604825190b93467918a0ec6456a24/tests/middleware/test_base.py#L192-L223) for an example of this behavior).
+- Using `BaseHTTPMiddleware` will prevent [ASGI pathsend extension](https://asgi.readthedocs.io/en/latest/extensions.html#path-send) to work properly. Thus, if you run your Starlette application with a server implementing this extension, routes returning [FileResponse](responses.md#fileresponse) should avoid the usage of this middleware.
```
adriangb: FWIW I think https://github.com/encode/starlette/pull/2616/files#r1684224798 would be made easier / resolved by https://github.com/encode/starlette/pull/2620 because then we can change it only for `BaseHTTPMiddleware`

gi0baro: @adriangb I think it still really depends on wether `BaseHTTPMiddleware` only supports `response.body` messages or not..

Kludex: @gi0baro I think you can change the `BaseHTTPMiddleware` now, can't you?

adriangb: Yep my point was that since it how has it's own Response class it should be easier to edit :)

gi0baro: 👍 gonna rebase and edit `BaseHTTPMiddleware`

gi0baro:
@Kludex I just rebased from current master for your convenience

gi0baro:
@Kludex what's your current take on this? Should I rebase so it can be merged, or should we just forget about it?

Kludex:
> @Kludex what's your current take on this? Should I rebase so it can be merged, or should we just forget about it?

I want this to get in, I want to get back to this soon. I can rebase myself.

Sorry the delay here.

starlette/middleware/base.py
Kludex, line 1: Is altering the `scope["extensions"]` in the application side allowed? Would it be bad to remove the `http.response.httpsend` from the `BaseHTTPMiddleware`? It's a question. I'm not sure what would be best.

I would like to avoid adding complexity to the `BaseHTTPMiddleware`.

starlette/responses.py
```diff
@@ -364,10 +364,12 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         if self.background is not None:
             await self.background()
 
-    async def _handle_simple(self, send: Send, send_header_only: bool) -> None:
+    async def _handle_simple(self, scope: Scope, send: Send, send_header_only: bool) -> None:
         await send({"type": "http.response.start", "status": self.status_code, "headers": self.raw_headers})
         if send_header_only:
             await send({"type": "http.response.body", "body": b"", "more_body": False})
+        elif "http.response.pathsend" in scope["extensions"]:
+            await send({"type": "http.response.pathsend", "path": str(self.path)})
```
Kludex: nitpick

```suggestion
    async def _handle_simple(self, send: Send, send_header_only: bool, send_pathsend: bool) -> None:
        await send({"type": "http.response.start", "status": self.status_code, "headers": self.raw_headers})
        if send_header_only:
            await send({"type": "http.response.body", "body": b"", "more_body": False})
        elif send_pathsend:
            await send({"type": "http.response.pathsend", "path": str(self.path)})
```

Kludex (commented):
This has been going for long - Sorry about it.

The only blocker here is deciding what to do about the changes on the `BaseHTTPMiddleware` - but I guess that's really the point of this PR. Would it make sense to remove the extension from the scope?

starlette/responses.py
gi0baro: Done in last rebase

starlette/middleware/base.py
gi0baro: I'm not sure how altering the scope would help in avoiding changes on `BaseHTTPMiddleware`.

To my perspective, the only way to avoid these lines https://github.com/encode/starlette/pull/2671/files#diff-931c53548e8908544bd0818ee1a8d5f1ada4f07430a439e9322ab369eb40a8acR161-R165 is to re-implement the whole _streaming wrapper_, which doesn't sound less complex to me tbh..
Maybe we could try to avoid the response re-wrap in `_StreamingResponse` depending on the context? Probably I don't have the needed to knowledge of Starlette internals to propose such a change though :(

gi0baro:
@Kludex not sure why the coverage fails after the last rebase..

Kludex:
> @Kludex not sure why the coverage fails after the last rebase..

Because we recently set `branch = True` on coverage.

gi0baro:
@Kludex sorry for the long delay.

This is now rebased on latest master, and tests/CI is finally green. Let me know if any additional edits are required for merge.

Kludex (approved):

Kludex:
Sorry the delay. Gradly you caught me at PyCon IT. 😅

PR #2672, opened 2024-08-25
Support `request.url_for` when only "app" scope is avaialable

Jdsleppy:
# Summary

Resolve the bug (or limitation) in [this related discussion](https://github.com/encode/starlette/discussions/2522) where `Request.url_for()` cannot be called in a middleware inheriting from `BaseHTTPMiddleware` before the `Router`/`call_next` is invoked.

# Checklist

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -176,7 +176,7 @@ class HTTPConnection(typing.Mapping[str, typing.Any]):
         return self._state
 
     def url_for(self, name: str, /, **path_params: typing.Any) -> URL:
-        url_path_provider: Starlette | Router | None = self.scope.get("router") or self.scope.get("app")
+        url_path_provider: Router | Starlette | None = self.scope.get("router") or self.scope.get("app")
         if url_path_provider is None:
             raise RuntimeError("The `url_for` method can only be used inside a Starlette application or with a router.")
         url_path = url_path_provider.url_path_for(name, **path_params)
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -19,6 +19,7 @@ except ModuleNotFoundError:  # pragma: no cover
 
 
 if typing.TYPE_CHECKING:
+    from starlette.applications import Starlette
     from starlette.routing import Router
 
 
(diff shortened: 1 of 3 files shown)


starlette/requests.py
```diff
@@ -183,8 +184,8 @@ def state(self) -> State:
         return self._state
 
     def url_for(self, name: str, /, **path_params: typing.Any) -> URL:
-        router: Router = self.scope["router"]
-        url_path = router.url_path_for(name, **path_params)
+        url_pather: Starlette | Router = self.scope.get("app") or self.scope["router"]
+        url_path = url_pather.url_path_for(name, **path_params)
```
Jdsleppy: I needed to fall back to using `scope["router"]` to support the case where a `Router` is directly fed to a `TestClient`, as happens in `test_routing.py`.

tests/middleware/test_base.py
```diff
@@ -1033,3 +1033,35 @@ async def endpoint(request: Request) -> Response:
...
+) -> None:
+    class CallsRequestUrlForMiddleware(BaseHTTPMiddleware):
+        async def dispatch(
+            self,
+            request: Request,
+            call_next: RequestResponseEndpoint,
+        ) -> Response:
+            if request.url == request.url_for("special"):
+                return PlainTextResponse("Special")
+            return await call_next(request)
```
Jdsleppy: This shows the main motivation for supporting `request.url_for()` before `call_next`: it allows you to handle authentication or other concerns differently depending on the URL without hardcoding paths or path prefixes.

Kludex:
Well, actually this is not only for `BaseHTTPMiddleware`, but for any ASGI middleware.

Jdsleppy:
> Well, actually this is not only for `BaseHTTPMiddleware`, but for any ASGI middleware.

@Kludex Let me know if you'd like the commit message changed or anything, that's no problem.

Otherwise, it's rebased and ready if you're interested in the change.

Kludex:
As a note, we do the `build_middleware_stack` on the `Starlette`, but we don't do this "build" on the `Router`, and that's why calling the `scope["app"]` works here.

Kludex (approved):

Jdsleppy:
Thank you!

Issue #2452, opened 2024-01-31
Enable `branch = true` for `coverage.run`


Kludex:
The goal here is to enable `branch = true` on `tool.coverage.run`. We want to be able to cover all logic paths.

ShreySinha02:
#2512

ShreySinha02:
opened #2513

ShreySinha02:
can not pass on --fail-under=100 but on 99 it can pass the the coverage

Kludex:
Well, the goal is exactly to add tests...

ShreySinha02:
which means i have to write test for every file which have not full coverage

TechNiick:
There was already a PR open for this issue [here](https://github.com/encode/starlette/pull/2503).
I have started to work on test coverage. The first PR was opened [here](https://github.com/encode/starlette/pull/2505).
After merging this first one I will continue with the rest of them.

hoenie-ams:
> [ ![Fund with Polar](https://camo.githubusercontent.com/9597ae75678dcf7fd4201b195661a74f581e81b8216a8bfb06ba7cff306507bd/68747470733a2f2f706f6c61722e73682f6170692f6769746875622f656e636f64652f737461726c657474652f6973737565732f323435322f706c656467652e737667) ](https://polar.sh/encode/starlette/issues/2452)

FYI @Kludex : The link doesn't work (404)

Kludex:
> > [ ![Fund with Polar](https://camo.githubusercontent.com/9597ae75678dcf7fd4201b195661a74f581e81b8216a8bfb06ba7cff306507bd/68747470733a2f2f706f6c61722e73682f6170692f6769746875622f656e636f64652f737461726c657474652f6973737565732f323435322f706c656467652e737667) ](https://polar.sh/encode/starlette/issues/2452)
> 
> FYI @Kludex : The link doesn't work (404)

We are not using polar. They don't remove the links after you remove the integration.

lealre:
Putting the table to make it easier to identify:

| Name                               | Branch   | BrPart   | Cover   | Missing                         |
| ---------------------------------- | -------- | -------- | ------- | ------------------------------- |
| starlette/\_utils.py               | 14       | 1        | 98%     | 78->82                          |
| starlette/middleware/gzip.py       | 22       | 2        | 98%     | 16->22, 91->exit                |
| starlette/requests.py              | 96       | 3        | 99%     | 114->116, 233->225, 253->278    |
[...]

PR #2676, opened 2024-08-31
test: add tests in `test_routing`

Orenoid:
# Summary

Add tests for some of the uncovered branches in `starlette.routing`, related to [this issue](https://github.com/encode/starlette/issues/2452)

# Checklist

- [x] I've updated the documentation accordingly. (It seems no documentation needs update)


tests/test_routing.py
```diff
@@ -516,6 +524,8 @@ def test_host_reverse_urls() -> None:
         )
         == "https://port.example.org:3600/"
     )
+    with pytest.raises(NoMatchFound):
+        mixed_hosts_app.url_path_for("api", path="whatever", foo="bar")
```
Orenoid, line 491: Testing the circumstance where there are some remaining parameters and an exception is expected.
https://github.com/encode/starlette/blob/master/starlette/routing.py#L532

tests/test_routing.py
```diff
@@ -417,6 +417,14 @@ def test_reverse_mount_urls() -> None:
         mounted.url_path_for("users", subpath="test", path="/tom") == "/test/users/tom"
     )
 
+    mounted = Router([Mount("/users", ok, name="users")])
+    with pytest.raises(NoMatchFound):
+        mounted.url_path_for("users", path="/a", foo="bar")
+
+    mounted = Router([Mount("/users", ok, name="users")])
+    with pytest.raises(NoMatchFound):
+        mounted.url_path_for("users")
```
Orenoid, line 399: Testing the cases where the if conditions in these two logic blocks are not true.
https://github.com/encode/starlette/blob/master/starlette/routing.py#L456
https://github.com/encode/starlette/blob/master/starlette/routing.py#L458

Kludex:
Can you explain me what behavior is being tested?

Orenoid:
> Can you explain me what behavior is being tested?

I've provided the code lines above. According to the coverage report, those if statements are always True with the other tests. The tests I added check if the logic can correctly jump to another branch, which should raise an exception in these cases.

tests/test_routing.py
Orenoid: I passed an extra parameter `foo` in this case. This will make the `if` statement become `False` and jump to this line:
```
raise NoMatchFound(name, path_params)
```
This behavior was not tested and covered in the other tests before.

Kludex (approved):
Thanks, and sorry the delay. All good.

PR #2677, opened 2024-08-31
test: add tests in `test_requests`

Orenoid:
# Summary

Add tests for some of the uncovered branches in `starlette.requests`, related to [this issue](https://github.com/encode/starlette/issues/2452)

# Checklist

- [x] I've updated the documentation accordingly. (It seems no documentation needs update)


tests/test_requests.py
```diff
@@ -425,7 +453,7 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
         # Browsers don't send extra whitespace or semicolons in Cookie headers,
         # but cookie_parser() should parse whitespace the same way
         # document.cookie parses whitespace.
-        # ("  =  b  ;  ;  =  ;   c  =  ;  ", {"": "b", "c": ""}),
+        ("  =  b  ;  ;  =  ;   c  =  ;  ", {"": "b", "c": ""}),
```
Orenoid, line 428: Testing [this](https://github.com/encode/starlette/blob/master/starlette/requests.py#L55), when the key and value both are empty, this pair should be ignored.

I don't know why this parameter was commented out before. However, to meet the condition where both the key and value are empty, this parameter with an empty chunk ("; ;") is what we need.

tests/test_requests.py
```diff
@@ -78,6 +106,17 @@ def test_request_client(scope: Scope, expected_client: Address | None) -> None:
...
 
+@pytest.mark.anyio
+async def test_request_close() -> None:
+    request = Request({"type": "http", "method": "GET", "path": "/foo/"})
+    assert request._form is None
+    # When request.`_form` is None, close() should do nothing.
+    # To verify this behavior, we could just call request.close() here.
+    # If request.`_form`.close is called,
+    # an Exception will be raised and the test will fail.
+    await request.close()
```
Kludex: We can add a `pragma: no branch` on the conditional branch that this is trying to cover. I don't think there's a need for this test.

tests/test_requests.py
Kludex: This may have been on purpose? When was introduced?

tests/test_requests.py
```diff
@@ -28,6 +28,34 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+        response = JSONResponse(data)
+        await response(scope, receive, send)
+
+    client = test_client_factory(app)
+    response = client.post("/42?foo=bar", json={"baz": "qux"})
+    assert response.json() == {
+        "url": "http://testserver/42?foo=bar",
+        "query_params": {"foo": "bar"},
+        "json": {"baz": "qux"},
+    }
```
Kludex: Which lines are this trying to cover?

Orenoid: https://github.com/encode/starlette/blob/master/starlette/requests.py#L96
https://github.com/encode/starlette/blob/master/starlette/requests.py#L125
https://github.com/encode/starlette/blob/master/starlette/requests.py#L240

tests/test_requests.py
Orenoid: Yeah, agreed. That makes more sense.

tests/test_requests.py
Orenoid: Or should we add `pragma: no branch` to these lines too?

tests/test_requests.py
Orenoid: The test and `cookie_parser` were introduced in this [PR](https://github.com/encode/starlette/pull/900).
It looks like these tests were inspired by tornado test suite ( [relevant commit](https://github.com/encode/starlette/pull/900/commits/b548645a33569af52613e6ec3fdaf4ff61f553b6) ).
And I didn't see any discussion related to this specific parameter. It was just commented out at the beginning.

My thought is that enabling the parameter shouldn't be a problem. `cookie_parser` does support this case.

Orenoid: Just did more searching in Tornado's code and found [this](https://github.com/tornadoweb/tornado/blob/master/tornado/test/httputil_test.py#L561). They did have tested the case.

Orenoid:
@Kludex I'm sorry but may I ask what I should do with this PR now, as well as [this one](https://github.com/encode/starlette/pull/2676)?

tests/test_requests.py
Kludex: The thing is that this test is doing too much, maybe we can divide it a bit? Or maybe just the `no branch`... I don't want us to try to satisfy the coverage just because we want to satisfy it, but actually create meaningful tests (naming wise as well).

Orenoid: > Or maybe just the `no branch`

No problem, let's go with this.

Kludex (approved):

Kludex:
Thanks @Orenoid ! 🙏

PR #2680, opened 2024-09-01
Change documentation layout

Kludex:
@tomchristie I'm making some changes to the Starlette documentation to make it more _modern_ and appealing.

I intend to do the same with Uvicorn, and I intend to add a `sponsors` tab.


docs/img/logo.svg
lovelydinosaur, line 1: 🤔Let’s not use this short logo

Kludex: https://github.com/encode/starlette/pull/2682 👍

PR #2687, opened 2024-09-06
Streaming response early disconnect mode

dmitry-mli:
# Summary

Starlette 0.38.4 introduced a bugfix that also introduced a backward incompatible change. This PR keep both the original fix as well as recovers the backward compatibility through parameterization.

## https://github.com/encode/starlette/pull/2620 solved issue

https://github.com/encode/starlette/pull/2620 delivered an important fix for `BaseHTTPMiddleware` which relied on `StreamingResponse` for chained upstream and downstream communication. Unfortunately, `StreamingResponse` optimized for early disconnect which resulted into unexpected termination for some of middleware handlers, in case the communication did not manage to propagate through a longer stack (a stack of 4 was enough to reproduce).

## https://github.com/encode/starlette/pull/2620 created issue

Change https://github.com/encode/starlette/pull/2620 introduced a backward incompatible change for streaming response from `BaseHTTPMiddleware` (background tasks were dropped). This impacted software that relied async post-processing.

## This PR content

This PR extends `StreamingResponse` with `early_disconnect` flag (default `True`) that controls the early disconnect behavior that is undesirable for the middleware case `BaseHTTPMiddleware`. The flag is backward compatible for clients of `StreamingResponse`. The `_StreamingResponse` is reverted back to inherit from `StreamingResponse` but this time suppressing the early disconnect mode. Unit tests extended to cover the case.

# Checklist

- [X] [N/A]  I've updated the documentation accordingly.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -6,8 +6,9 @@ import anyio
 from anyio.abc import ObjectReceiveStream, ObjectSendStream
 
 from starlette._utils import collapse_excgroups
+from starlette.background import BackgroundTask
 from starlette.requests import ClientDisconnect, Request
-from starlette.responses import AsyncContentStream, Response
+from starlette.responses import AsyncContentStream, Response, StreamingResponse
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 RequestResponseEndpoint = typing.Callable[[Request], typing.Awaitable[Response]]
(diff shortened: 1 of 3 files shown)


dmitry-mli:
@Kludex @adriangb please review, thank you! This fixes backward incompatible change made in https://github.com/encode/starlette/pull/2620, details in the description.

@dzhulgakov FYI

adriangb:
@Kludex another reason we should do https://github.com/encode/starlette/pull/2176

adriangb:
Thank you for proposing a fix!

Can we go with https://github.com/encode/starlette/pull/2688 instead? It's a much smaller diff / code change.

dmitry-mli:
Sure thing, approved (with comments)!

PR #2688, opened 2024-09-06
Fix BackgroundTasks with BaseHTTPMiddleware

adriangb:
Alternative to https://github.com/encode/starlette/pull/2176

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -6,8 +6,9 @@ import anyio
 from anyio.abc import ObjectReceiveStream, ObjectSendStream
 
 from starlette._utils import collapse_excgroups
+from starlette.background import BackgroundTask
 from starlette.requests import ClientDisconnect, Request
-from starlette.responses import AsyncContentStream, Response
+from starlette.responses import AsyncContentStream, Response, StreamingResponse
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 RequestResponseEndpoint = typing.Callable[[Request], typing.Awaitable[Response]]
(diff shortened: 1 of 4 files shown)


dmitry-mli (approved):

dmitry-mli:
FYI, The streaming response used to also do:
```
            if not isinstance(chunk, (bytes, memoryview)):
                chunk = chunk.encode(self.charset)
```
It itsn't covered so when it was removed in https://github.com/encode/starlette/pull/2620/files it didn't break

starlette/middleware/base.py
```diff
@@ -222,3 +222,6 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             await send({"type": "http.response.body", "body": chunk, "more_body": True})
 
         await send({"type": "http.response.body", "body": b"", "more_body": False})
+
+        if self.background:
+            await self.background()
```
dmitry-mli, line 228: For full backward compatibility add it to constructor also:

```
class _StreamingResponse(Response):
    def __init__(
        self,
        content: AsyncContentStream,
        status_code: int = 200,
        headers: typing.Mapping[str, str] | None = None,
        media_type: str | None = None,
        background: BackgroundTask | None = None,                <<<
        info: typing.Mapping[str, typing.Any] | None = None,
    ) -> None:
        self.info = info
        self.body_iterator = content
        self.media_type = media_type
        self.background = background                             <<<
        self.init_headers(headers)
```

dmitry-mli: This is because some middleware may be referring to that field. Or below will work also

```
self.background = None
```

adriangb: No users should be calling the constructor unless they do something like `type(response)(...)` which is not something we need to support.

dmitry-mli: Agree, this aligns with the class being marked as private. Then please include
```
self.background = None
```

For full backward compatibility, any response object injected into `BaseHTTPMiddleware` based middleware needs to have the `background` field (default None)

adriangb: Ok I see your point now with the edit. Good point. Added.

@Kludex this is another reason to move the background task logic out of responses _and_ a good reason to rework our Response inheritance so that `Response` and `StreamingResponse` both inherit from a `BaseResponse` that has the API and initialization of common bits.

Kludex (changes requested):
Can we add a test here?

adriangb:
There is a test already, am I missing something?

Kludex (approved):

PR #2693, opened 2024-09-17
Close unclosed `MemoryObjectReceiveStream` in `TestClient`

Moataz-E:
# Summary

`TestClient` isn't closing `stream_receive: StreamObjectStream` correctly during shutdown. This has been confirmed by multiple people as part of discussion [#2603](https://github.com/encode/starlette/discussions/2603). [@markwaddle](https://github.com/markwaddle) has identified that the issue is due to `stream_receive` not being disposed of correctly in the shutdown function. Adding `stream_receive` to the shutdown function's context manager resolved the issue as confirmed by myself and others in the discussion.

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -784,7 +784,7 @@ class TestClient(httpx.Client):
                 self.task.result()
             return message
 
-        async with self.stream_send:
+        async with self.stream_send, self.stream_receive:
             await self.stream_receive.send({"type": "lifespan.shutdown"})
             message = await receive()
             assert message["type"] in (


adriangb (approved):
Looks good to me!

Kludex:
We have an ignore of resource warning on the pyproject. Is this PR enough to remove that?

abdullahsumsum:
@Kludex Any idea when this will be merged? We are currently working around this issue by pinning an older version of anyio.

Kludex:
It would be cool to get my question replied...

Moataz-E:
> It would be cool to get my question replied...

I removed the ignore filter and re-ran the tests and can see 7 tests still fail because of this warning. So this only prevents the issue from propagating to people who are using TestClient.

@Kludex  If you would like these issues resolved first before the PR is merged, let me know and I can take a look this evening.

Kludex:
> > It would be cool to get my question replied...
> 
> I removed the ignore filter and re-ran the tests and can see 7 tests still fail because of this warning. So this only prevents the issue from propagating to people who are using TestClient.
> 
> @Kludex If you would like these issues resolved first before the PR is merged, let me know and I can take a look this evening.

Ideally, it would be cool to solve everything that was introduced by the bump of anyio on the project before merging this.

But... It's not a blocker...

Moataz-E:
> > > It would be cool to get my question replied...
> > 
> > 
> > I removed the ignore filter and re-ran the tests and can see 7 tests still fail because of this warning. So this only prevents the issue from propagating to people who are using TestClient.
> > @Kludex If you would like these issues resolved first before the PR is merged, let me know and I can take a look this evening.
> 
> Ideally, it would be cool to solve everything that was introduced by the bump of anyio on the project before merging this.
> 
> But... It's not a blocker...

No worries, I will take a look when I get back on my computer.

Kludex:
Thanks.

Kludex:
I'll make a release with this PR, but I think this will lower the motivation to find out why we have more unclosed memory objects... 😢

alvindera97:
I had more warnings by a factor of about 4X...


**Edit:** 
My mistake. I had to update `starlette` manually to the latest version (as of time of writing) `0.39.2` and now I don't have any warnings anymore.

Moataz-E:
> I had more warnings by a factor of about 4X

Which version were you running before and after? Can you provide more details about the warning numbers.

alvindera97:
> > I had more warnings by a factor of about 4X
> 
> Which version were you running before and after? Can you provide more details about the warning numbers.

Updating `anyio` to the latest version (as of writing) `4.6.0` was what increased the warning by a factor of 4. I was previously on `anyio` version `4.4.0`. 

So updating to the latest version `starlette` got rid of all the warnings.

Moataz-E:
@alvindera97 if the issue is related to `anyio` and not this change then please create a separate issue / discussion.

juliangilbey:
I'm still seeing this error when testing fastapi using the following package versions:
* fastapi version 0.115.5
* starlette version 0.41.2
* anyio version 4.6.2
So there still seem to be cases where this PR is insufficient to close the streams, unfortunately.  But I have no idea how to track down the cause.

Kludex:
If you read this PR, I mention this in the beginning.

PR #2700, opened 2024-09-23
perf: avoid regex re-compile

trim21:
# Summary

found `re.findall` and `re.sub` usage in starlette module, we should always pre-compile regex pattern if it's possible.

Althrough `re` has a built-in cache, but it has size limit and for project using many regex it still may be invalidated and pattern get unnecessary re-compile.

There is still one case in https://github.com/encode/starlette/blob/1131b3cbcd6f9e2aa3500fd4d7bf3cd200fa2532/starlette/_utils.py#L88

It's not a simple case so I didn't touch it.

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -56,8 +56,7 @@ class Response:
             populate_content_length = True
             populate_content_type = True
         else:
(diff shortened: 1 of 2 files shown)


Kludex (approved):

Kludex:
Thanks! 🙏

PR #2702, opened 2024-09-23
generate boundary with token_hex

trim21:
# Summary

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -404,7 +404,9 @@ class FileResponse(Response):
         file_size: int,
         send_header_only: bool,
     ) -> None:
-        boundary = token_hex(20)
+        # in firefox and chrome, they use boundary with 95-96 bits entropy,
+        # that's roughly 13 bytes
+        boundary = token_hex(13)
         content_length, header_generator = self.generate_multipart(
             ranges, boundary, file_size, self.headers["content-type"]
         )
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -11,7 +11,6 @@ from datetime import datetime
 from email.utils import format_datetime, formatdate
 from functools import partial
 from mimetypes import guess_type
-from random import choices as random_choices
 from secrets import token_hex
 from urllib.parse import quote
 
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -12,6 +12,7 @@ from email.utils import format_datetime, formatdate
 from functools import partial
 from mimetypes import guess_type
 from random import choices as random_choices
+from secrets import token_hex
 from urllib.parse import quote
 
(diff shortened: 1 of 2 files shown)


starlette/responses.py
```diff
@@ -404,7 +404,7 @@ async def _handle_multiple_ranges(
         file_size: int,
         send_header_only: bool,
     ) -> None:
-        boundary = "".join(random_choices("abcdefghijklmnopqrstuvwxyz0123456789", k=13))
+        boundary = token_hex(20)
```
Kludex: Why was it increased to 20?

trim21: just A random number I picked. len(boundary) is actually 40 now

abersheeran: According to the specification, this value can be between 1 and 70. But is 40 too long?

abersheeran (approved):
Elegant modification.

starlette/responses.py
graingert: \>96 bits of entropy is recommended so only 13 bytes is needed

abersheeran: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#directives

Kludex: Can we drop a bit the number then?

Kludex: @trim21 Can we drop the number a bit so I can include this on the next release?

trim21: to what? 13?

Kludex: Yep. Maybe we can add a comment about the choice, to avoid future confusion/questions?

trim21: > >96 bits of entropy is recommended so only 13 bytes is needed

where does this recommendation comes from?

graingert: https://github.com/whatwg/html/issues/6424#issuecomment-826036018
https://andreubotella.github.io/multipart-form-data/#serializing
https://github.com/andreubotella/multipart-form-data/blob/main/index.bs#L32

Also I misremembered the recommendation it's 95 bits, so 12 bytes would be sufficient

graingert (approved):

starlette/responses.py
```diff
@@ -404,7 +404,9 @@ async def _handle_multiple_ranges(
         file_size: int,
         send_header_only: bool,
     ) -> None:
-        boundary = "".join(random_choices("abcdefghijklmnopqrstuvwxyz0123456789", k=13))
+        # in firefox and chrome, they use boundary with 95-96 bits entropy,
+        # that's roughly 13 bytes
```
```suggestion
        # In firefox and chrome, they use boundary with 95-96 bits entropy (that's roughly 13 bytes).
```

PR #2703, opened 2024-09-23
Consider `FileResponse.chunk_size` when handling multiple ranges

Kludex:
cc @frostming

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -414,8 +414,8 @@ class FileResponse(Response):
             async with await anyio.open_file(self.path, mode="rb") as file:
                 for start, end in ranges:
                     await send({"type": "http.response.body", "body": header_generator(start, end), "more_body": True})
+                    await file.seek(start)
                     while start < end:
-                        await file.seek(start)
                         chunk = await file.read(min(self.chunk_size, end - start))
                         start += len(chunk)
                         await send({"type": "http.response.body", "body": chunk, "more_body": True})
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -420,7 +420,7 @@ class FileResponse(Response):
                         start += len(chunk)
                         await send({"type": "http.response.body", "body": chunk, "more_body": True})
                         if len(chunk) < self.chunk_size or start >= end:
-                            break
+                            break  # pragma: no cover
                     await send({"type": "http.response.body", "body": b"\n", "more_body": True})
                 await send(
                     {
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -374,13 +374,7 @@ class FileResponse(Response):
                 while more_body:
                     chunk = await file.read(self.chunk_size)
                     more_body = len(chunk) == self.chunk_size
-                    await send(
-                        {
-                            "type": "http.response.body",
-                            "body": chunk,
-                            "more_body": more_body,
-                        }
-                    )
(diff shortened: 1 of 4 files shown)


starlette/responses.py
```diff
@@ -419,10 +413,14 @@ async def _handle_multiple_ranges(
...
-                    await file.seek(start)
-                    chunk = await file.read(min(self.chunk_size, end - start))
                     await send({"type": "http.response.body", "body": header_generator(start, end), "more_body": True})
-                    await send({"type": "http.response.body", "body": chunk, "more_body": True})
+                    while True:
+                        await file.seek(start)
+                        chunk = await file.read(min(self.chunk_size, end - start))
+                        start += len(chunk)
+                        await send({"type": "http.response.body", "body": chunk, "more_body": True})
+                        if len(chunk) < self.chunk_size or start >= end:
```
Kludex: Is the `len(chunk) < self.chunk_size` necessary?

I'm too tired, I'll check tomorrow. Just leaving my own comment here.

Kludex:
How do I have a different behavior in Python versions with this diff? 🤔

Kludex:
> How do I have a different behavior in Python versions with this diff? 🤔

Coverage is not able to understand that the `break` was reached... @graingert do you know why? 🤔

starlette/responses.py
```diff
@@ -419,10 +413,14 @@ async def _handle_multiple_ranges(
...
+                            break
```
```suggestion
                            break  # pragma: no cover
```

starlette/responses.py
frostming: it can be mathematically deduced that is not required

abersheeran: Yes, there is no need to check once `len(chunk) < self.chunk_size`

starlette/responses.py
```diff
@@ -419,10 +413,14 @@ async def _handle_multiple_ranges(
...
-                    chunk = await file.read(min(self.chunk_size, end - start))
                     await send({"type": "http.response.body", "body": header_generator(start, end), "more_body": True})
-                    await send({"type": "http.response.body", "body": chunk, "more_body": True})
+                    while True:
+                        await file.seek(start)
+                        chunk = await file.read(min(self.chunk_size, end - start))
+                        start += len(chunk)
+                        await send({"type": "http.response.body", "body": chunk, "more_body": True})
+                        if len(chunk) < self.chunk_size or start >= end:
+                            break  # pragma: no cover
```
```suggestion
                    while start < end:
                        await file.seek(start)
                        chunk = await file.read(min(self.chunk_size, end - start))
                        start += len(chunk)
                        await send({"type": "http.response.body", "body": chunk, "more_body": True})
```

starlette/responses.py
```diff
@@ -414,13 +414,11 @@ async def _handle_multiple_ranges(
             async with await anyio.open_file(self.path, mode="rb") as file:
                 for start, end in ranges:
                     await send({"type": "http.response.body", "body": header_generator(start, end), "more_body": True})
-                    while True:
+                    while start < end:
                         await file.seek(start)
```
frostming:
```suggestion
                    await file.seek(start)
                    while start < end:
```
Don't need to seek in every loop

Kludex:
@abersheeran @frostming anything else here, or we are good?

frostming:
lgtm

PR #2708, opened 2024-09-29
Support python-multipart 0.0.12

Kludex:
Now `python-multipart` has the `py.typed`, so I had to fix our `formparsers` module.

I've also taken the opportunity to replace `nocover` to `no cover`. Sorry for the noise with it, but it was bothering me for some time.


starlette/formparsers.py
Kludex, line 1: This file contains the relevant changes.

PR #2709, opened 2024-09-29
Use less private imports on the testclient module

### LINKED ISSUE #2673: Private imports from `httpx`.
Prompted by https://github.com/encode/httpx/discussions/3287

---

`starlette` is currently using private imports from `httpx`...

https://github.com/encode/starlette/blob/c78c9aac17a4d68e0647252310044502f1b7da71/starlette/testclient.py#L500-L513

These are broken against the latest minor point release, which includes this (internal) refactoring... https://github.com/encode/httpx/pull/3245

---

It's possible that the path of least resistance for the short term is for `httpx` to re-include `URLTypes`, tho really we need to be making sure that downstream packages are not using private imports.

Some prior discussion of this also at #2534.

Kludex:
- Closes https://github.com/encode/starlette/issues/2673


starlette/testclient.py
```diff
@@ -37,16 +38,29 @@
...
+URL: TypeAlias = "httpx.URL | str"
+Timeout: TypeAlias = "float | httpx.Timeout | tuple[float | None, float | None, float | None, float | None] | None"
+Headers = typing.Union[
+    httpx.Headers,
+    typing.Mapping[bytes, bytes],
+    typing.Mapping[str, str],
+    typing.Sequence[typing.Tuple[bytes, bytes]],
+    typing.Sequence[typing.Tuple[str, str]],
+    None,
+]
```
Kludex, line 55: I pretty much have to do what httpx is doing on the `_client` module because I think we want to support the type possibilities that are there...

Kludex:
@tomchristie Can we discuss/have a plan with this before httpx 1.0?

Kludex:
@zanieb @tomchristie was this discussed in the meeting?

lovelydinosaur:
To an extent... we did talk about stricter typing.
(Which would resolve this, by removing the need for type synonyms)

I'll see about spinning out a design discussion.

Kludex:
What is expected from Starlette?

lovelydinosaur:
> What is expected from Starlette?

* Don't access private implementation details.
* Currently, pin to [minor point releases](https://semver.org/spec/v2.0.0.html), eg. `>=0.27,<0.28`.
* Once we're at 1.0 you'l be able to pin to major point releases, eg. `>=1.0,<2.0`.

starlette/testclient.py
```diff
@@ -468,19 +474,19 @@ def _choose_redirect_arg(
...
+        files: typing.Any = None,
         json: typing.Any = None,
-        params: httpx._types.QueryParamTypes | None = None,
-        headers: httpx._types.HeaderTypes | None = None,
-        cookies: httpx._types.CookieTypes | None = None,
-        auth: httpx._types.AuthTypes | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT,
+        params: QueryParams | None = None,
+        headers: Headers | None = None,
+        cookies: Cookies | None = None,
+        auth: Auth | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT,
```
lovelydinosaur, line 482: Possibly...

```python
UseClientDefault = type(httpx.USE_CLIENT_DEFAULT)

...

auth: Auth | UseClientDefault = USE_CLIENT_DEFAULT
```

(I don't much like this style and I'd enjoy cleaning it up, tho would that be sufficient for now?)

Kludex: I don't think it works for the type checkers.. Does it?

Kludex: That doesn't work.

Kludex:
I can't really subclass `httpx.Client` if the types are not exposed by `httpx`. 😢

PR #2718, opened 2024-10-06
feat: Optional max_file_size for parsing form files

khadrawy:
# Summary

Limiting the size of file being uploaded in case the storage is limited or some business logic can't handle certain size.

# Checklist


Kludex:
If we instead make the `max_part_size` configurable, would that already be good enough here?

khadrawy:
> If we instead make the `max_part_size` configurable, would that already be good enough here?

But from what I see [here](https://github.com/encode/starlette/blob/master/starlette/formparsers.py#L157) is that this is only applied to non file parts.

Kludex:
> > If we instead make the `max_part_size` configurable, would that already be good enough here?
> 
> But from what I see [here](https://github.com/encode/starlette/blob/master/starlette/formparsers.py?rgh-link-date=2024-12-03T10%3A44%3A28Z#L157) is that this is only applied to non file parts.

Hmmm... Does it make sense to be only on non file parts tho? 🤔

khadrawy:
@Kludex I think we need different limits for parts vs files. As files are spooled, we could allow larger size but for parts we want to have a stricter limit as it is stored in memory.
I can re-work the PR to make it configurable for both.

khadrawy:
@Kludex what do you think about making all of limits configurable? https://github.com/encode/starlette/pull/2798/files
I see that you are going to rename the max_file_size here https://github.com/encode/starlette/pull/2780 I can adjust naming if we can go forward with making all of the size limits configurable.

Kludex:
We can make the limits configurable. We just need to adjust the naming, yep.

Let's start by making the `spool` one configurable, and then we add 1 by one. 🙏 

Can you help?

starlette/requests.py
```diff
@@ -255,6 +255,7 @@ async def _get_form(
         max_files: int | float = 1000,
         max_fields: int | float = 1000,
         max_part_size: int = 1024 * 1024,
+        max_file_spool_size: int = 1024 * 1024,
```
Kludex:
```suggestion
        max_spool_size: int = 1024 * 1024,
```

Kludex (changes requested):

PR #2732, opened 2024-10-18
Support for `send` client disconnect to HTTP

Kludex:
After this, we can implement on Uvicorn side.

We may have too much noise from other responses, tho? We should probably supress the others.


adriangb (approved):
Looks good to me!

PR #2734, opened 2024-10-20
Bump minimum `python-multipart` version to 0.0.13

Kludex:
I forgot this...


mweinelt:
https://pypi.org/project/python-multipart/#history

0.0.13 and 0.0.14 were yanked

PR #2736, opened 2024-10-24
Version 0.41.1

Kludex:


bdura:
Hey @Kludex, this version might need to be yanked as well now that `python-multipart@0.0.14` [is no more](https://github.com/Kludex/python-multipart/issues/170)

PR #2751, opened 2024-11-14
Modify application_state at _raise_on_disconnect

osangu:
# Summary
In my opinion, even when the websocket is closed by the client, the state should be changed to DISCONNECTED. 
This way, it can be handled more easily by users.

# Checklist

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -110,6 +110,8 @@ class WebSocket(HTTPConnection):
 
     def _raise_on_disconnect(self, message: Message) -> None:
         if message["type"] == "websocket.disconnect":
+            self.application_state = WebSocketState.DISCONNECTED
+
             raise WebSocketDisconnect(message["code"], message.get("reason"))
 
     async def receive_text(self) -> str:


Kludex:
But we do have `client_state`...

Kludex:
Also, why are you using those states in your application?

osangu:
Oh.. I didn't know about `client_state`, so using like below code .
I apologize for making PR with light knowledge and finally waste your time.

Once again, I'm really sorry.

```python
@socket_router.websocket('/ws')
async def knowledge_chat(
        websocket: WebSocket
):
    already_disconnected = False
    try:
        await service(websocket)

    except fe.WebSocketDisconnect:
        already_disconnected = True

    finally:
        still_connected = websocket.application_state == WebSocketState.CONNECTED

        if not already_disconnected and still_connected:
            await fe_websocket.close()
```

Kludex:
I don't recommend using those states. You should rely on the `WebSocketDisconnect`.

osangu:
May I kindly ask if the reason you recommend manage through exception handling is to enable immediate processing of the connection? If that’s not the case, I would be truly grateful if you could explain the other reasons.

Kludex:
The reason is that `application_state` and `client_state` are meant to be private, even if there's no underscore as prefix. 😅

osangu:
I understood!
Really appreciate for you kind comment and sorry again : )
Have a nice day!!

Kludex:
No need to apologize, all good. :)

Thanks for closing the PR.

PR #2752, opened 2024-11-14
Fix issue with middleware args passing

uriyyo:
# Summary

A feature introduced in #2381 doesn't allow to pass args inside middleware, here is an example:
```py
from starlette.types import ASGIApp, Receive, Scope, Send


class Middleware:
    def __init__(self, app: ASGIApp, arg: str) -> None:
        self.app = app
        self.arg = arg

        self.app(scope, receive, send)


app = Starlette()
app.add_middleware(Middleware, "foo")

client = TestClient(app)
response = client.get("/")
```
```py
TypeError: Middleware.__init__() got multiple values for argument 'app'
```

Also, make type annotation for middleware less strict, remove restriction for middleware to be class, and make it callable that returns ASGI application, it's a fully backward-compatible change.

```py
from starlette.types import ASGIApp, Receive, Scope, Send


def forbid_http_middleware(app: ASGIApp) -> ASGIApp:
    async def middleware(scope: Scope, receive: Receive, send: Send) -> None:
        if scope["type"] == "http":
            _response = JSONResponse({"detail": "Forbidden"}, status_code=403)
            await _response(scope, receive, send)
        else:

    return middleware


app = Starlette()
app.add_middleware(forbid_http_middleware)

client = TestClient(app)
response = client.get("/")
```

# Checklist

### DIFF
--- a/starlette/middleware/__init__.py
+++ b/starlette/middleware/__init__.py
@@ -8,21 +8,19 @@ if sys.version_info >= (3, 10):  # pragma: no cover
 else:  # pragma: no cover
     from typing_extensions import ParamSpec
 
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp
 
 P = ParamSpec("P")
 
 
-class _MiddlewareClass(Protocol[P]):
-    def __init__(self, app: ASGIApp, *args: P.args, **kwargs: P.kwargs) -> None: ...  # pragma: no cover
-
-    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None: ...  # pragma: no cover
+class _MiddlewareFactory(Protocol[P]):
+    def __call__(self, app: ASGIApp, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover
 
 
 class Middleware:
     def __init__(
         self,
-        cls: type[_MiddlewareClass[P]],
+        cls: _MiddlewareFactory[P],
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> None:
@@ -38,5 +36,6 @@ class Middleware:
         class_name = self.__class__.__name__
         args_strings = [f"{value!r}" for value in self.args]
         option_strings = [f"{key}={value!r}" for key, value in self.kwargs.items()]
-        args_repr = ", ".join([self.cls.__name__] + args_strings + option_strings)
+        cls_name = self.cls.__name__  # type: ignore[attr-defined]
+        args_repr = ", ".join([cls_name] + args_strings + option_strings)
         return f"{class_name}({args_repr})"
--- a/starlette/middleware/__init__.py
+++ b/starlette/middleware/__init__.py
@@ -36,6 +36,6 @@ class Middleware:
         class_name = self.__class__.__name__
         args_strings = [f"{value!r}" for value in self.args]
         option_strings = [f"{key}={value!r}" for key, value in self.kwargs.items()]
-        cls_name = self.cls.__name__  # type: ignore[attr-defined]
-        args_repr = ", ".join([cls_name] + args_strings + option_strings)
+        name = getattr(self.cls, "__name__", "")
+        args_repr = ", ".join([name] + args_strings + option_strings)
         return f"{class_name}({args_repr})"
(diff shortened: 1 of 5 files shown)


starlette/middleware/__init__.py
```diff
@@ -38,5 +36,6 @@ def __repr__(self) -> str:
         class_name = self.__class__.__name__
         args_strings = [f"{value!r}" for value in self.args]
         option_strings = [f"{key}={value!r}" for key, value in self.kwargs.items()]
-        args_repr = ", ".join([self.cls.__name__] + args_strings + option_strings)
+        cls_name = self.cls.__name__  # type: ignore[attr-defined]
```
adriangb: Is this safe to do? I guess the point is that the `self.cls` isn't necessarily a class, it may be a function, but even functions have `__name__` defined. Maybe we should do `getattr(self.cls, "__name__", "")` or something to make sure we don't error in any case?

adriangb: (alternatively we could add `__name__` to the protocol above 😆 )

uriyyo: @adriangb Thanks for suggestion, changed it to be `getattr` with default value call.

adriangb: Thanks! Are there any actual changes to the reprs? I'm not seeing why that code had to change (aside from what I pointed out in this comment)

uriyyo: There were no actual changes to the repr. I changed it only because of mypy error.

adriangb (approved):

adriangb:
Thanks!

arni-xalgo-io:
Upgrading from 0.41.2 to 0.41.3 gives me the following type error
```
REDACTED/api.py:104: error: Missing positional arguments "scope", "receive", "send" in call to "add_middleware" of "Starlette"  [call-arg]
REDACTED/api.py:104: error: Argument 1 to "add_middleware" of "Starlette" has incompatible type "Type[ExceptionHandlerMiddleware]"; expected "_MiddlewareFactory[[MutableMapping[str, Any], Callable[[], Awaitable[MutableMapping[str, Any]]], Callable[[MutableMapping[str, Any]], Awaitable[None]]]]"  [arg-type]
REDACTED/api.py:104: note: Following member(s) of "ExceptionHandlerMiddleware" have conflicts:
REDACTED/api.py:104: note:     Expected:
REDACTED/api.py:104: note:         def __call__(app: Callable[[MutableMapping[str, Any], Callable[[], Awaitable[MutableMapping[str, Any]]], Callable[[MutableMapping[str, Any]], Awaitable[None]]], Awaitable[None]], scope: MutableMapping[str, Any], receive: Callable[[], Awaitable[MutableMapping[str, Any]]], send: Callable[[MutableMapping[str, Any]], Awaitable[None]]) -> Callable[[MutableMapping[str, Any], Callable[[], Awaitable[MutableMapping[str, Any]]], Callable[[MutableMapping[str, Any]], Awaitable[None]]], Awaitable[None]]
REDACTED/api.py:104: note:     Got:
REDACTED/api.py:104: note:         def __call__(self: ExceptionHandlerMiddleware, scope: MutableMapping[str, Any], receive: Callable[[], Awaitable[MutableMapping[str, Any]]], send: Callable[[MutableMapping[str, Any]], Awaitable[None]]) -> Coroutine[Any, Any, None]
Found 2 errors in 1 file (checked 44 source files)

✕ mypy failed.
```
for the following line of code
```python
app.add_middleware(ExceptionHandlerMiddleware)
```
where `app` is a FastAPI app.

The middleware is like this
```python
from starlette.types import ASGIApp, Receive, Scope, Send

class ExceptionHandlerMiddleware:
    def __init__(self, app: ASGIApp):
        self.app = app

        # Implementation here
        ...
```

I'm guessing it's a regression resulting from this PR.

uriyyo:
@arni-xalgo-io There is discussion about this issue - https://github.com/encode/starlette/discussions/2757

Maybe14:
> it's a fully backward-compatible change

After upgrading to 0.41.3, I've been getting this error:
```
TypeError: MyMiddleware.__init__() takes 1 positional argument but 2 were given
```

The issue was the change from `cls(app=app, ...)` to `cls(app, ...)` because of my use of `args` and `kwargs`

PR #2761, opened 2024-11-21
Use ETag from headers when parsing If-Range in FileResponse

viccie30:
# Summary

`FileResponse.set_stat_headers()` takes care not to overwrite any headers set on the response, but `FileResponse._should_use_range()` ignores any set headers and unconditionally uses its own ETag and Last-Modified calculation. This commit rectifies that.

I have chosen to code `self.headers[`HEADER`]` instead of `self.headers.get(`HEADER`)`, because `FileResponse.__call__()` always calls `set_stat_headers()` and therefore the headers should always be set.

# Checklist


Kludex (approved):
Makes sense. :)

Good clean-up as well. Thanks!

Issue #2766, opened 2024-11-25
Replace `RuntimeError` with `WebSocketDisconnected` once the socket has been disconnected


ruitcatarino:
### Discussed in https://github.com/encode/starlette/discussions/2762

proto-atlas:
I can take a look at this. From the discussion, it sounds like the intended path is to use the existing disconnect exception rather than add another state flag. I'd update the WebSocket.send() disconnected-state branch to raise WebSocketDisconnect() instead of RuntimeError, update test_duplicate_close to expect it, and leave the receive-after-disconnect branch unchanged since that is a separate path. Does that match what you had in mind?

PR #2767, opened 2024-11-25
Replace RuntimeError with WebSocketDisconnected

ruitcatarino:
# Summary
Change the exception raised if we attempt to use a disconnected WebSocket from `RuntimeError` to `WebSocketDisconnected`.
Discussed in https://github.com/encode/starlette/discussions/2762
Issue https://github.com/encode/starlette/issues/2766

# Checklist

I haven't changed the documentation, if there is a need for it I will gladly change it.


starlette/websockets.py
```diff
@@ -22,6 +22,14 @@ def __init__(self, code: int = 1000, reason: str | None = None) -> None:
         self.reason = reason or ""
 
 
+class WebSocketDisconnected(RuntimeError):
+    """
+    Raised when attempting to use a disconnected WebSocket.
+    """
+
+    pass
```
Kludex, line 31: That's not really what I meant... The exception already exists.

Kludex (changes requested):

starlette/websockets.py
ruitcatarino: Well, reading the code, `WebSocketDisconnect` seems to be called when you want to disconnect a socket, in these cases the socket is already disconnected. In my opinion, it would make more sense to create a new exception (`WebSocketDisconnected`). Tell me if you still want to use `WebSocketDisconnect` and if so I will make such changes.

dym-ok:
The change suggested by @ruitcatarino looks meaningful to me.
It would allow precise handling of situations like the one mentioned in https://github.com/encode/starlette/discussions/2543.

What stands in the way of accepting this change, can we somehow help?

ruitcatarino:
After the revival of this MR, which, to be honest, I have revisited sporadically in the hopes of getting an update, since my team and I at work are still affected by the original issue that led me to create this pull request, I realized that this newly created exception could also be used in the following cases: `raise RuntimeError('WebSocket is not connected. Need to call "accept" first.')` → `raise WebSocketDisconnected('WebSocket is not connected. Need to call "accept" first.')` in the methods `receive_bytes`, `receive_text` and `receive_json`.

ruitcatarino:
I have added the suggested changes I made in https://github.com/encode/starlette/pull/2767#issuecomment-2738015722 and added more tests. However, despite increasing the number of tests, the pipeline is still failing due to coverage. 🤔 Any ideas?

tests/test_websockets.py
```diff
@@ -602,6 +638,19 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+    async def app(scope: Scope, receive: Receive, send: Send) -> None:
+        websocket = WebSocket(scope, receive=receive, send=send)
+        await websocket.accept()
+        websocket.client_state = WebSocketState.DISCONNECTED
+        await websocket.receive()
+
+    client = test_client_factory(app)
+    with pytest.raises(RuntimeError):
+        with client.websocket_connect("/") as websocket:
+            websocket.send({"type": "websocket.send"})
```
Kludex: This line is never reached.

ruitcatarino: Thank you, I was indeed doing the wrong thing. Fixed it.

starlette/websockets.py
Kludex: Sorry, you are right.

Kludex: No, actually, I understand what I had in mind: my idea was to raise `WebSocketDisconnect` to avoid the user from seeing the exception. The purpose of the `RuntimeError` here is that it's an issue with the code.

ruitcatarino: I'm confused, raising `WebSocketDisconnect` instead of `RuntimeError `or `WebSocketDisconnected` when attempting to explicitly call `close()` on a closed socket would still raise the exception `WebSocketDisconnect` and it would not be handled automatically. The user would have to handle it.
My thought process is by explicitly calling `close()` on a closed socket it will eventually call `send()` that would currently raise a `RuntimeError`. With my changes it would raise a specific exception `WebSocketDisconnected`. If we changed it to `WebSocketDisconnect` that exception would be raised from `send()` and `close()` never handles it.

And even if it was handled automatically and avoided the user from seeing the exception, the idea that I got from our discussion was that making `close()` idempotent was not the approach to follow (https://github.com/Kludex/starlette/discussions/2762#discussioncomment-11367437).

Maybe I'm not understanding your request.

Kludex (approved):

PR #2768, opened 2024-11-26
Make _MiddlewareFactory compatible with Callable

eltoder:
# Summary

Make `app` a positional-only argument so that a `Callable[[ASGIApp, ...], ASGIApp]` can be passed in place of `_MiddlewareFactory`.

We can also consider making `_MiddlewareFactory` and moving it into `starlette.types`. This will make it easier to type arguments for `add_middleware()`.

# Checklist

### DIFF
--- a/starlette/middleware/__init__.py
+++ b/starlette/middleware/__init__.py
@@ -14,7 +14,7 @@ P = ParamSpec("P")
 
 
 class _MiddlewareFactory(Protocol[P]):
-    def __call__(self, app: ASGIApp, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover
+    def __call__(self, app: ASGIApp, /, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover
 
 
 class Middleware:
--- a/tests/test_applications.py
+++ b/tests/test_applications.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import os
 from contextlib import asynccontextmanager
 from pathlib import Path
-from typing import AsyncGenerator, AsyncIterator, Generator
+from typing import AsyncGenerator, AsyncIterator, Callable, Generator
 
 import anyio.from_thread
 import pytest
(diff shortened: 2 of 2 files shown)


starlette/middleware/__init__.py
```diff
@@ -14,7 +14,7 @@
 
 
 class _MiddlewareFactory(Protocol[P]):
-    def __call__(self, app: ASGIApp, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover
+    def __call__(self, app: ASGIApp, /, *args: P.args, **kwargs: P.kwargs) -> ASGIApp: ...  # pragma: no cover
```
adriangb, line 17: I find specifying positional only parameters in Protocols confusion, however I checked this case and I think it's correct:

```python
import typing

class Plain(typing.Protocol):
    def __call__(self, x: int, y: str) -> None: ...

class Positional(typing.Protocol):
    def __call__(self, a: int, /, y: str) -> None: ...


def foo(x: int, y: str) -> None:
    ...


def bar(x: int, /, y: str) -> None:
    ...


a: Plain = foo  # ok
b: Plain = bar # err
c: Positional = foo  # ok
d: Positional = bar # ok
```

Passes on https://mypy-play.net/?mypy=1.4.1&python=3.13&gist=8972af53af6b9d67e6c255a5de32db18

adriangb:
> We can also consider making `_MiddlewareFactory` and moving it into `starlette.types`. This will make it easier to type arguments for `add_middleware()`

This part I did not understand

adriangb (approved):

eltoder:
@adriangb thank you for merging!

>This part I did not understand

Sorry, missed a word:

We can also consider making `_MiddlewareFactory` *public*.

The rationale is that `add_middleware()` is a public function and occasionally I need to refer to the type of its argument. For example, I have a function that returns which middleware to use based on environment. I need to write its return type. With this PR I can write it as `Callalble[]`, which is nice. But it may be even more clear and explicit to write it as `MiddlewareFactory`. However, currently the latter is not public. Here's a sketch of how this can be used: https://mypy-play.net/?mypy=1.13.0&python=3.13&gist=e0652be937c8cfc1d56a5d4dafa0bd89

PR #2777, opened 2024-11-30
Fix 'ResourceWarning's about unclosed 'anyio.streams.memory.MemoryObjectReceiveStream'

nikitagashkov:
Fixes https://github.com/encode/starlette/discussions/2772.

# Summary

This PR addresses the cleanup of memory object streams, particularly focusing on ensuring that streams are properly closed to prevent resource warnings:

1. Close `recv_stream` in `BaseHTTPMiddleware` upon exception within the route to prevent dangling streams.

# Checklist

↑ Looks like no changes in docs needed, this is just a bugfix.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -170,7 +170,7 @@ class BaseHTTPMiddleware:
             assert message["type"] == "http.response.start"
 
             async def body_stream() -> typing.AsyncGenerator[bytes, None]:
-                # NB! `recv_stream` will be left dangling if exception occurs
+                # FIXME: `recv_stream` will be left dangling if exception occurs
                 # before we got a chance to stream response.
                 async with recv_stream:
                     async for message in recv_stream:
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -144,6 +144,9 @@ class BaseHTTPMiddleware:
             async def coro() -> None:
                 nonlocal app_exc
 
+                # XXX: Enter `recv_stream` here as well? If sender is closed, we
+                # should close receiver, streams are unbuffered. Listening for
+                # `http.response.start` should be adjusted though.
                 async with send_stream:
                     try:
                         await self.app(scope, receive_or_disconnect, send_no_error)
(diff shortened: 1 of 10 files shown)


starlette/middleware/base.py
```diff
@@ -144,7 +144,7 @@ async def send_no_error(message: Message) -> None:
             async def coro() -> None:
                 nonlocal app_exc
 
-                async with send_stream:
+                async with send_stream, recv_stream:
```
nikitagashkov: Since streams are unbuffered, it probably makes sense to close `recv` if we closed `send` 🤔

nikitagashkov: Moved this change (as more invasive) to https://github.com/encode/starlette/pull/2778

tests/middleware/test_base.py
```diff
@@ -210,24 +211,17 @@ async def dispatch(
     ) -> Response:
         ctxvar.set("set by middleware")
         resp = await call_next(request)
-        assert ctxvar.get() == "set by endpoint"
+        # BaseHTTPMiddleware creates a TaskGroup which copies the context
+        # and erases any changes to it made within the TaskGroup
+        assert ctxvar.get() == "set by middleware"
```
Kludex: Well, yeah, that's why the `pytest.mark.xfail` is there... 🤔

Kludex:
I'm very confused about the comments in this PR. I'll review the other PR.

Adding tons of comments in unrelated code is not helpful.

tests/middleware/test_base.py
nikitagashkov: `xfail` complicates testing for a warning, which is why I had to make adjustments.

An `AssertionError` within the middleware triggers the same `ResourceWarning` (refer to issue #2778). However, since the warning is emitted only after the stream is deleted, it appears some time later, specifically after an explicit `gc.collect()` in `test_anyio_streams_cleanup_exc_in_route`. This causes the test to fail due to a leftover stream
from another test.

Perhaps there's no need for this test at all, as it is currently quite brittle. 🤷‍♂️ It might be better to stop ignoring the warning from anyio once all streams are properly closed.

nikitagashkov:
> I'm very confused about the comments in this PR. I'll review the other PR.
> 
> Adding tons of comments in unrelated code is not helpful.

I believe comments are useful for highlighting certain parts of the code, but I have removed them as requested.

Additionally, I made the patchset smaller by removing fixes for `TestClient` warnings and un-`ignore` of `anyio`'s `ResourceWarning`, since the original discussion focused on a specific issue. It's probably better to keep these changes separate as well.

Kludex:
Let's continue on the other PR.

PR #2780, opened 2024-12-03
Rename `max_file_size` to `spool_max_size`

Kludex:
@defnull is this what you mentioned some months ago?


adriangb (approved):
I think this makes sense. It's a much better description.

Kludex:
I tried to add a deprecation warning, but the effort didn't seem to pay off.

PR #2789, opened 2024-12-10
Add `pragma: no branch` in `middleware/exceptions.py`

lealre:
# Summary

Trying to help with issue #2452 

Changing the tests to cover the cases where `handlers=None` in `ExceptionMiddleware` to handle the cases where the `if` statement in [line 31](https://github.com/encode/starlette/blob/master/starlette/middleware/exceptions.py#L31) is skipped.

Isolates the tests that have a route with an exception handler (`BadBodyException`) from the ones that do not have one in [test_exceptions.py](https://github.com/encode/starlette/blob/master/tests/test_exceptions.py), so it can deal with both cases separately.

NOTE: When an app is instantiated directly by Starlette, it passes by default to ExceptionMiddleware an empty dictionary, which makes `if handlers is not None` always evaluate to `True` and the for loop iterates over an empty dict. Another option could be to change [line 31](https://github.com/encode/starlette/blob/master/starlette/middleware/exceptions.py#L31) to `if handlers:` and keep the tests as they are.

# Checklist

### DIFF
--- a/starlette/middleware/exceptions.py
+++ b/starlette/middleware/exceptions.py
@@ -28,7 +28,7 @@ class ExceptionMiddleware:
             HTTPException: self.http_exception,
             WebSocketException: self.websocket_exception,
         }
-        if handlers is not None:
+        if handlers is not None:  # pragma: no branch
             for key, value in handlers.items():
                 self.add_exception_handler(key, value)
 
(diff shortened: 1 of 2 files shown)


Kludex:
> Another option could be to change [line 31](https://github.com/encode/starlette/blob/master/starlette/middleware/exceptions.py?rgh-link-date=2024-12-10T19%3A50%3A54Z#L31) to if handlers: and keep the tests as they are.

I think here we can just add `# pragma: no branch` on the `if handlers it not None:`. That would be good enough in this case.

Kludex (approved):

PR #2792, opened 2024-12-12
Resolve the anyio `ResourceWarning` in production code

kurtmckee:
> [!Note]
>
> This addresses one of the same issues that #2778 addresses, but changes fewer lines to do so. This may mean it's more correct or less correct, so please review.
>
> This PR intentionally addresses only the warnings encountered by downstream projects using starlette. It does not address the remaining `ResourceWarning`s caused by starlette's own test code.

This resolves the `ResourceWarning`s caused by a failure to close the receive stream opened at line 112.

Identified by setting pytest `filterwarnings` to "all" instead of "error", and running pytest with tracemalloc set:

```
PYTHONTRACEMALLOC=10 pytest
```

Fixes #2772

# Summary

# Checklist

- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change. _Existing tests already triggered the `ResourceWarning`_
- [ ] I've updated the documentation accordingly. _Doesn't appear to be relevant_

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -131,8 +131,11 @@ class BaseHTTPMiddleware:
                 return message
 
             async def close_recv_stream_on_response_sent() -> None:
-                await response_sent.wait()
-                recv_stream.close()
+                # The stream must always be closed, even if an exception occurs.
+                try:
+                    await response_sent.wait()
+                finally:
+                    recv_stream.close()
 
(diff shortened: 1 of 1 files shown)


Kludex:
- This is being worked on here: https://github.com/encode/starlette/pull/2778

kurtmckee:
Okay, I'll comment there. :+1:

PR #2796, opened 2024-12-14
tests: branches coverage in `test_authentication`, `test_background`, `test_responses`, `test_routing`

lealre:
# Summary

Related to issue #2452

This PR addresses the cases in the test files below and essentially adds `#pragma: no branch` to all the cases, except for `test_authentication`, where I changed the `#pragma: no cover` line to cover the function that is not used.

From what I can see, there are no branch options to be covered, as the code relies on `elif` conditions or depends on a condition to pass the test.

| **File**                                                                                                       | **Branches**         |
| -------------------------------------------------------------------------------------------------------------- | -------------------- |
| [`tests/test_authentication.py`](https://github.com/encode/starlette/blob/master/tests/test_authentication.py) | 213->210             |
| [`tests/test_background.py`](https://github.com/encode/starlette/blob/master/tests/test_background.py)         | 76->exit             |
| [`tests/test_responses.py`](https://github.com/encode/starlette/blob/master/tests/test_responses.py)           | 268->exit, 752->exit |
| [`tests/test_routing.py`](https://github.com/encode/starlette/blob/master/tests/test_routing.py)               | 800->802, 874->exit  |

The only one I don't fully understand is `test_routing` ([line 874](https://github.com/encode/starlette/blob/master/tests/test_routing.py#L874)), where it's marking the @pytest.parametrized test as having uncovered branches.

# Checklist


Kludex (approved):

tests/test_routing.py
```diff
@@ -871,7 +871,7 @@ async def my_staticmethod(request: Request) -> None: ...  # pragma: no cover
     def __call__(self, request: Request) -> None: ...  # pragma: no cover
 
 
-@pytest.mark.parametrize(
+@pytest.mark.parametrize(  # pragma: no branch
```
Kludex: I don't know this one as well... 👀

Kludex: Can you try to find out why this works?

Kludex: We can merge the others here first, if you want.

lealre: Sure, I can try to find out as I will be working on the other cases.

I think we can merge the other ones if it's not a problem.

Kludex (approved):

tests/test_routing.py
Kludex: Thanks.

Kludex:
```suggestion
@pytest.mark.parametrize(
```

PR #2804, opened 2024-12-16
tests: branches coverage in `responses.py`, `staticfiles.py`, `templating.py`, `middleware/wsgi.py` and `endpoints.py`

lealre:
# Summary

Related to issue #2452.

### `responses.py` - [117→124](https://github.com/encode/starlette/blob/master/starlette/responses.py#L117-L124)
Test with `samesite = None` added.

### `staticfiles.py` - [207→212](https://github.com/encode/starlette/blob/master/starlette/staticfiles.py#L207-L212)
Test with `etag` not matching the `headers["if-none-match"]`.

### `templating.py` - [46→56](https://github.com/encode/starlette/blob/master/starlette/templating.py#L46-L56)
It seems that `TestClient` adds `{'http.response.debug': {}}` to the scope on every HTTP request ([line 310](https://github.com/encode/starlette/blob/master/starlette/testclient.py#L310)), making the statement always `True`.  
Added `pragma: no branch` and needs confirmation.

### `middleware/wsgi.py` - [124→exit](https://github.com/encode/starlette/blob/master/starlette/middleware/wsgi.py#L124-L139)
`self.response_started` is initialized as `False` in `WSGIResponder` and is only used in `WSGIMiddleware` ([line 79](https://github.com/encode/starlette/blob/master/starlette/middleware/wsgi.py#L79)). Removing the `if` statement does not affect the tests.  
Added `pragma: no branch` and needs confirmation.

### `endpoints.py` - [77→72](https://github.com/encode/starlette/blob/master/starlette/endpoints.py#L77-L72)
It seems there are only two possible options inside the `while` loop.
Added `pragma: no branch` and needs confirmation.

# Checklist

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -74,7 +74,7 @@ class WebSocketEndpoint:
                 if message["type"] == "websocket.receive":
                     data = await self.decode(websocket, message)
                     await self.on_receive(websocket, data)
-                elif message["type"] == "websocket.disconnect":
+                elif message["type"] == "websocket.disconnect":  # pragma: no branch
                     close_code = int(message.get("code") or status.WS_1000_NORMAL_CLOSURE)
                     break
         except Exception as exc:
--- a/starlette/middleware/wsgi.py
+++ b/starlette/middleware/wsgi.py
@@ -121,7 +121,7 @@ class WSGIResponder:
         exc_info: typing.Any = None,
     ) -> None:
         self.exc_info = exc_info
-        if not self.response_started:
+        if not self.response_started:  # pragma: no branch
             self.response_started = True
             status_code_string, _ = status.split(" ", 1)
             status_code = int(status_code_string)
--- a/starlette/templating.py
+++ b/starlette/templating.py
@@ -43,7 +43,7 @@ class _TemplateResponse(HTMLResponse):
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         request = self.context.get("request", {})
         extensions = request.get("extensions", {})
-        if "http.response.debug" in extensions:
+        if "http.response.debug" in extensions:  # pragma: no branch
             await send(
                 {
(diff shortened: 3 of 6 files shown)


starlette/templating.py
```diff
@@ -43,7 +43,7 @@ def __init__(
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         request = self.context.get("request", {})
         extensions = request.get("extensions", {})
-        if "http.response.debug" in extensions:
+        if "http.response.debug" in extensions:  # pragma: no branch
```
Kludex, line 46: This can be tested without the `TestClient`.

We have some tests that don't use the `TestClient`.

starlette/middleware/wsgi.py
```diff
@@ -121,7 +121,7 @@ def start_response(
         exc_info: typing.Any = None,
     ) -> None:
         self.exc_info = exc_info
-        if not self.response_started:
+        if not self.response_started:  # pragma: no branch
```
Kludex, line 124: This is okay.

starlette/endpoints.py
```diff
@@ -74,7 +74,7 @@ async def dispatch(self) -> None:
                 if message["type"] == "websocket.receive":
                     data = await self.decode(websocket, message)
                     await self.on_receive(websocket, data)
-                elif message["type"] == "websocket.disconnect":
+                elif message["type"] == "websocket.disconnect":  # pragma: no branch
```
Kludex, line 77: This is also okay.

starlette/templating.py
lealre: That's right, I didn't think about it. 

Added a test to the case based on other similar tests

Kludex: I put this back. I think this is enough for the `"http.response.debug"`.

Kludex (approved):

Issue #2783, opened 2024-12-04
Elif branch with error raise in TestClient _choose_redirect_arg never triggers


t-ded:
In here, the elif branch should likely be the first condition to trigger only then followed by check of the individual parameters - I can do a PR for this if accepted

https://github.com/encode/starlette/blob/d40bc9e5e12d686542a180368bc78402e9501efd/starlette/testclient.py#L458C1-L462C24

Kludex:
Makes sense. PR welcome.

Kludex:
We dropped the `allow_redirects` from the `TestClient`.

Ref.: https://github.com/encode/starlette/pull/2808

PR #2805, opened 2024-12-16
fix: change conditional logic in `_choose_redirect_arg` for `TestClent` (#2783).

lealre:
# Summary

Related to issue #2783

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -450,9 +450,7 @@ class TestClient(httpx.Client):
     ) -> bool | httpx._client.UseClientDefault:
         redirect: bool | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT
         if allow_redirects is not None and follow_redirects is not None:
-            raise RuntimeError(  # pragma: no cover
-                "Cannot use both `allow_redirects` and `follow_redirects`."
-            )
+            raise RuntimeError("Cannot use both `allow_redirects` and `follow_redirects`.")
         if allow_redirects is not None:
             message = "The `allow_redirects` argument is deprecated. Use `follow_redirects` instead."
             warnings.warn(message, DeprecationWarning)
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -449,16 +449,16 @@ class TestClient(httpx.Client):
         self, follow_redirects: bool | None, allow_redirects: bool | None
     ) -> bool | httpx._client.UseClientDefault:
         redirect: bool | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT
+        if allow_redirects is not None and follow_redirects is not None:
+            raise RuntimeError(  # pragma: no cover
+                "Cannot use both `allow_redirects` and `follow_redirects`."
+            )
         if allow_redirects is not None:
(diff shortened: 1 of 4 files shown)


starlette/testclient.py
```diff
@@ -449,16 +449,16 @@ def _choose_redirect_arg(
         self, follow_redirects: bool | None, allow_redirects: bool | None
     ) -> bool | httpx._client.UseClientDefault:
         redirect: bool | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT
+        if allow_redirects is not None and follow_redirects is not None:
+            raise RuntimeError(  # pragma: no cover
```
Kludex: We can test it now.

starlette/testclient.py
```diff
@@ -449,16 +449,16 @@ def _choose_redirect_arg(
...
+                "Cannot use both `allow_redirects` and `follow_redirects`."
+            )
         if allow_redirects is not None:
             message = "The `allow_redirects` argument is deprecated. Use `follow_redirects` instead."
             warnings.warn(message, DeprecationWarning)
```
Kludex, line 456: When did we add this line? Maybe we should just remove this logic... 🤔

Kludex (commented):
I forgot to press "comment". I reviewed yesterday.

starlette/testclient.py
lealre: It seems that it was added [here](https://github.com/encode/starlette/commit/6765502c1e5a418c16bb379c54b8ea706b0444cf#diff-aca25e5f16c4fd49338ccdf3631f72455309335fee1e27f3d3b6016fa94ecedfR401).

I will work on the test, and if you decide that it should be removed, I will remove it.

lealre: I saw that the `allow_redirects` was being covered in tests because it was being used in `test_responses.py` and in `tests/middleware/test_https_redirect.py`.

I changed it to `follow_redirects` as `allow_redirects` is deprecated and added tests to cover all the cases in `test_testclient`.

Kludex (changes requested):
I've been thinking about this...

We added the `RuntimeError` logic to make sure people would not use the two parameters. At the time, it was a good idea. Nowadays, it's been two years since we added this function. I think we should just drop the `_choose_redirect_arg`.

The deprecation warning has been there for long, so I think it's fine.

@lealre Would you like to create a PR with it?

lealre:
> I think we should just drop the `_choose_redirect_arg`.
> 
> The deprecation warning has been there for long, so I think it's fine.
> 
> @lealre Would you like to create a PR with it?

Yes, I would like to work on it. 

So, just to confirm, it will remove the `_choose_redirect_arg` and keep only the `follow_redirects` argument in `TestClient`.

Kludex:
No, the goal is to remove the deprecated parameter. The function doesn't make sense without the parameter.

lealre:
> So, just to confirm, it will remove the `_choose_redirect_arg` and keep only the `follow_redirects` argument in `TestClient`.

Yes, I meant to keep the `follow_redirects` and remove the deprecated `allow_redirects`.

EDIT: Makes more sense to pass `httpx._client.USE_CLIENT_DEFAULT` as a default argument.

Kludex:
Closing this in favor of https://github.com/encode/starlette/pull/2808

PR #2806, opened 2024-12-16
Make UUID path parameter conversion more flexible

edthrn:
# Summary

Based on @adriangb suggestion made in this [discussion thread](https://github.com/encode/starlette/discussions/2802#discussioncomment-11583842), here is a more flexible conversion of UUID path parameters. 

UUID representation may vary depending on the system from which it is generated. Some systems may return an hex representation with no hyphens, while some others use upper-case letters.

From Python perspective, they are all valid:
image

Not handling those cases may yield to hard-to-debug 404 errors.

# Checklist

### DIFF
--- a/tests/test_convertors.py
+++ b/tests/test_convertors.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from typing import Iterator
+from uuid import UUID
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


adriangb (approved):
looks good to me :)

tests/test_convertors.py
```diff
@@ -58,15 +59,30 @@ def test_datetime_convertor(test_client_factory: TestClientFactory, app: Router)
...
+        ("uuid", "00000000aaaaffff9999000000000000", 200),
+        ("uuid", "00000000-AAAA-FFFF-9999-000000000000", 200),
+        ("uuid", "00000000AAAAFFFF9999000000000000", 200),
+        ("uuid", "not-a-uuid", 404),
+    ],
+)
+def test_default_convertor(
+    test_client_factory: TestClientFactory, param_type: str, param_value: str, status_code: int
+) -> None:
+    expected_types = {"float": float, "uuid": UUID}
```
Kludex: Can we avoid this by leaving the `test_default_float_convertor` unchanged and create a new one for `uuid`? 🙏

Kludex (changes requested):
This is fine. 🙏 

Just a minor nitpick, as I think our test suite focus a lot on readability.

tests/test_convertors.py
edthrn: Done :heavy_check_mark:

Kludex (approved):

Kludex:
Thanks @edthrn 🙏

PR #2808, opened 2024-12-22
Remove deprecated `allow_redirects` argument from `TestClient`

lealre:
# Summary

As discussed in #2805 and related to #2783 .

Remove the deprecated `allow_redirects` argument from `TestClient`, and update the tests and `pyproject.toml`.

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -8,7 +8,6 @@ import math
 import queue
 import sys
 import typing
-import warnings
 from concurrent.futures import Future
 from functools import cached_property
 from types import GeneratorType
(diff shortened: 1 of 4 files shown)


starlette/testclient.py
```diff
@@ -398,6 +397,7 @@ class TestClient(httpx.Client):
     __test__ = False
     task: Future[None]
     portal: anyio.abc.BlockingPortal | None = None
+    redirect: bool | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT
```
Kludex: Why this? There's a parameter in the init

lealre: My bad, I forgot it there

Kludex (approved):

Kludex:
Thanks!

PR #2813, opened 2024-12-26
Fix unclosed 'MemoryObjectReceiveStream' upon exception in 'BaseHTTPMiddleware' children

Kludex:
I open a different PR to remove the `close_recv_stream_on_response_sent`. I'm not sure that's why the test is failing...


starlette/middleware/base.py
```diff
@@ -130,10 +127,6 @@ async def wrap(func: typing.Callable[[], typing.Awaitable[T]]) -> T:
 
                 return message
 
-            async def close_recv_stream_on_response_sent() -> None:
-                await response_sent.wait()
-                recv_stream.close()
-
```
Kludex, line 136: Remove this as per @graingert comment on https://matrix.to/#/!JfFIjeKHlqEVmAsxYP:gitter.im/$S0X8-1wH1qscoq6FQjVCVDgBL4Qk-5BcTnaQfqBCkGI?via=gitter.im&via=matrix.org&via=matrix.freyachat.eu

Kludex:
Well... I'm very confused as to why the current test is failing. 😆

graingert:
hmm, 3.8 is EOL so probably getting an old anyio?

edit 1: I ran the test suite locally and got the failure once, running the suite again 3 more times did not reproduce it

edit 2: I deleted my pyc files, ran and it passed. I deleted my pyc files again and recreated my venv and it passed

Kludex:
The flakiness worries me. Do you know why? Even if it's an old version of anyio.

starlette/middleware/base.py
```diff
@@ -182,11 +173,15 @@ async def body_stream() -> typing.AsyncGenerator[bytes, None]:
             response.raw_headers = message["headers"]
             return response
 
-        with collapse_excgroups():
+        send_stream: MemoryObjectSendStream[Message]
+        recv_stream: MemoryObjectReceiveStream[Message]
+        send_stream, recv_stream = anyio.create_memory_object_stream()
```
graingert:
```suggestion
        send_stream, recv_stream = anyio.create_memory_object_stream[Message]()
```

graingert:
I'm going to try running it 1000x

graingert:
for future reference in case the CI logs expire, in python3.8 CI the failure is:

```pytb

=================================== FAILURES ===================================
_______________ test_websocket_endpoint_on_receive_text[asyncio] _______________

test_client_factory = functools.partial(<class 'starlette.testclient.TestClient'>, backend='asyncio', backend_options={})

    def test_websocket_endpoint_on_receive_text(
        test_client_factory: TestClientFactory,
    ) -> None:
        class WebSocketApp(WebSocketEndpoint):
            encoding = "text"
    
            async def on_receive(self, websocket: WebSocket, data: str) -> None:
                await websocket.send_text(f"Message text was: {data}")
    
        client = test_client_factory(WebSocketApp)
        with client.websocket_connect("/ws") as websocket:
            websocket.send_text("Hello, world!")
            _text = websocket.receive_text()
            assert _text == "Message text was: Hello, world!"
    
        with pytest.raises(RuntimeError):
            with client.websocket_connect("/ws") as websocket:
>               websocket.send_bytes(b"Hello world")
E               Failed: DID NOT RAISE <class 'RuntimeError'>

tests/test_endpoints.py:134: Failed
================== 1 failed, 878 passed, 4 xfailed in 13.38s ===================
```

graingert:
ah I think it's a race condition here:

https://github.com/encode/starlette/blob/29b9c42273ad5e3a4a8120b3a1a8f7b578be894a/starlette/testclient.py#L132-L136

the other thread could be busy doing stuff moments before trying to put a message on the queue, so the queue appears empty when there's a message pending being put

You need to use queue.shutdown() 3.13+ or send an explicit EOF object when the queue is done with

graingert:
I think this is a pre-existing issue, just VERY rare

starlette/testclient.py
```diff
@@ -150,10 +162,13 @@ async def run_app(tg: anyio.abc.TaskGroup) -> None:
...
-            tg.start_soon(run_app, tg)
-            await self.should_close.wait()
-            tg.cancel_scope.cancel()
+        try:
+            async with anyio.create_task_group() as tg:
+                tg.start_soon(run_app, tg)
+                await self.should_close.wait()
+                tg.cancel_scope.cancel()
+        finally:
+            self._send_queue.put(EOF)
```
```suggestion
            self._send_queue.put(EOF)  # TODO: use self._send_queue.shutdown() on 3.13+
```

starlette/middleware/base.py
```diff
@@ -3,7 +3,7 @@
 import typing
 
 import anyio
-from anyio.abc import ObjectReceiveStream, ObjectSendStream
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
```
graingert:
```suggestion
```

starlette/testclient.py
```diff
@@ -85,6 +85,14 @@ class WebSocketDenialResponse(  # type: ignore[misc]
     """
 
 
+class _Eof(enum.Enum):
+    EOF = enum.auto()
+
+
+EOF: typing.Final = _Eof.EOF
+Eof = typing.Literal[_Eof.EOF]
```
Kludex: The problem this solves is not related to this PR then?

graingert: yes probably best to split into two PRs and add a changelog entry for the lost shutdown exception issue

graingert:
omg it's still possible to fail!?

Kludex:
Hmmm... Then it's not the queue shutdown thingy?

graingert:
rather than continue to hijack your PR I've opened a new one https://github.com/encode/starlette/pull/2814

graingert:
truly bizarre

graingert (approved):

Kludex:
I'll make a release when I get home.

linbrandon4:
Good catch/fix. Closing MemoryObjectReceiveStream when a middleware child throws is one of those annoying “only shows up under exceptions / CI” issues, and it makes sense to handle it in BaseHTTPMiddleware so you don’t leak streams and trigger resource warnings.

The review thread about the flaky websocket test was useful too. It really does look like there was a separate race in the testclient/queue shutdown path (rare, but real), so I like that it got split into a different PR instead of mixing concerns here.

Only things I’d double-check:
- the stream gets closed on all exit paths (normal response, exception, cancellation)
- we’re not accidentally masking the original exception (close should be best-effort)
- tests cover the exception path well enough so this doesn’t regress later

PR #2814, opened 2024-12-26
collect errors more reliably from websocket test client

graingert:
…iddleware' children

Co-authored-by: Thomas Grainger <413772+graingert@users.noreply.github.com>
Co-authored-by: Nikita Gashkov <8746283+nikitagashkov@users.noreply.github.com>

# Summary

# Checklist


graingert:
@Kludex I think this fixed it

Kludex:
What was the issue?

graingert:
The exception bubbling out of the _run was not being collected properly, and I think the queue EOF/shutdown thing also

starlette/testclient.py
```diff
@@ -16,10 +17,9 @@
 import anyio
 import anyio.abc
 import anyio.from_thread
-from anyio.abc import ObjectReceiveStream, ObjectSendStream
 from anyio.streams.stapled import StapledObjectStream
```
```suggestion
from anyio.abc import ObjectReceiveStream, ObjectSendStream
from anyio.streams.stapled import StapledObjectStream
```

starlette/testclient.py
```diff
@@ -97,63 +105,53 @@ def __init__(
...
         The sub-thread in which the websocket session runs.
         """
-
-        async def run_app(tg: anyio.abc.TaskGroup) -> None:
+        try:
             try:
-                await self.app(self.scope, self._asgi_receive, self._asgi_send)
-            except anyio.get_cancelled_exc_class():
-                ...
+                self.should_close = anyio.Event()
```
graingert: @Kludex this `self.should_close` is left as tests check for `self.should_close.is_set()` should I remove this property? it's odd that it's public

graingert:
I'd recommend reviewing this in side by side view: https://github.com/encode/starlette/pull/2814/files?diff=split

starlette/testclient.py
Kludex: You can remove the `should_close`. It was meant to be used on the `wait` part.

starlette/testclient.py
```diff
@@ -97,63 +105,53 @@ def __init__(
...
             finally:
-                tg.cancel_scope.cancel()
-
-        async with anyio.create_task_group() as tg:
-            tg.start_soon(run_app, tg)
-            await self.should_close.wait()
-            tg.cancel_scope.cancel()
+                self.should_close.set()
+        finally:
+            self._send_queue.put(EOF)  # TODO: use self._send_queue.shutdown() on 3.13+
```
Kludex, line 148: Should we use the `if sys.version_info` here?

starlette/testclient.py
```diff
@@ -97,63 +105,53 @@ def __init__(
...
-            self.close(1000)
-        finally:
-            self.portal.start_task_soon(self._notify_close)
-            self.exit_stack.close()
-        while not self._send_queue.empty():
+        self.exit_stack.close()
+
+        while True:
             message = self._send_queue.get()
+            if message is EOF:
```
Kludex, line 131: Is there an analogous to EOF from the standard library on 3.13?

Kludex (commented):
As long as this passes the tests, I'm fine with whatever refactor you think is the best here.

starlette/testclient.py
graingert: It raises an exception

starlette/testclient.py
graingert: I think it's better to stick to the EOF approach until someone puts up a Queue.shutdown backport, or we can use a MemoryObjectStream with portal

starlette/testclient.py
```diff
@@ -85,6 +85,29 @@ class WebSocketDenialResponse(  # type: ignore[misc]
...
+    async def __call__(self, /, *, task_status: anyio.abc.TaskStatus[_T_co]) -> None: ...
+
+
+@contextlib.contextmanager
+def _handle_task(portal: anyio.abc.BlockingPortal, async_fn: StartableAsyncFn[_T_co]) -> typing.Generator[_T_co]:
+    fut, result = portal.start_task(async_fn)
+    try:
+        yield result
+    finally:
+        fut.result()
```
```suggestion
        # can't raise an exception from stack.callback on Python 3.8
        # due to https://github.com/python/cpython/issues/69968
        fut.result()
```

tests/test_testclient.py
```diff
@@ -254,19 +254,25 @@ async def asgi(receive: Receive, send: Send) -> None:
...
-            while True:
-                await anyio.sleep(0.1)
+            nonlocal cancelled
+            try:
+                websocket = WebSocket(scope, receive=receive, send=send)
+                await websocket.accept()
+                await anyio.sleep_forever()
+            except anyio.get_cancelled_exc_class():
+                cancelled = True
+                raise
```
Kludex, line 269: Why is the cancelled exception here? Is this because this except was removed?

The addition of that `except` was on purpose, if I recall correctly.

graingert: this is instead of websocket.should_close.is_set() as it no longer exists - so we need to find out that the async function ran and was cancelled

graingert: did you mean to ask about this one: https://github.com/encode/starlette/pull/2814/files#diff-aca25e5f16c4fd49338ccdf3631f72455309335fee1e27f3d3b6016fa94ecedfL145 ?

this is because it's no longer possible to get a intentional cancelled_exc here because it will be caught by the CancelScope
if you do get an cancelled_exc here it's because someone incorrectly issued a native asyncio cancel or manually raised a trio cancel, both of which should be propagated out of the websocket_connect cmgr

Kludex: Got it.

So the point is: since the WebSocket doesn't call `send` it can't receive a disconnect exception, and since it doesn't call `receive`, it doesn't receive a `websocket.disconnected` message. Then, the `TestClient` here will propagate the cancellation exception here, but the user shouldn't worry about it because the `TestClient` will catch it anyway.

Right?

graingert: yeah, we use a CancelScope passed out with task_status.started(cs), this is then used to cancel the task on completion and collect a result. The CancelScope catches the cancellation that it causes to raise in the coro, unless there's another reason for cancellation which is then propagated, this is a fatal case and so the user should be notified

starlette/testclient.py
```diff
@@ -97,63 +105,47 @@ def __init__(
...
-        while not self._send_queue.empty():
+        self.exit_stack.close()
+
+        while True:
             message = self._send_queue.get()
+            if message is EOF:
+                break
             if isinstance(message, BaseException):
-                raise message
+                raise message  # pragma: no cover (defensive, should be impossible)
```
Kludex, line 134: Why it should be impossible?

The `except BaseException as exc` below doesn't have a `pragma: no cover`, so I assume it's being hit?

graingert: this is impossible because the exit stack will raise the exception out of fut.result() and so the queue won't be consumed.

This is only possible to be hit if `ws.receive()` is interrupted (eg with a KI) while waiting for an exception or message to be placed on the queue.

I'm currently sketching out another slight refactor that uses MemoryObjectStreams here instead that should clean this up a bit

tests/test_testclient.py
Kludex: So if there's a cancelled exception that is not from the TestClient, the user will see it, right?

I'm good with this.

Kludex (approved):

tests/test_testclient.py
graingert: Yeah the only way is if the user was calling .cancel() directly or manually raising constructing a trio cancel (bypassing the private constructor protections)

PR #2829, opened 2024-12-29
use a pair of memory object streams instead of two queues

graingert:
# Summary

# Checklist

(discussed on previous PR, no tests needed as this is a plain refactor and existing tests cover it, no doc changes as this is a transparent refactor)


Kludex (approved):
Amazing! 🙏

PR #2830, opened 2024-12-29
avoid collapsing exception groups from user code

graingert:
# Summary

If user code is raising ExceptionGroups eg from TaskGroups or otherwise, they should get ExceptionGroups raised. There could also be important notes attached.

# Checklist

No docs changes needed, probably should have made an issue for this.


tests/middleware/test_base.py
```diff
@@ -21,6 +22,9 @@
 from starlette.websockets import WebSocket
 from tests.types import TestClientFactory
 
+if sys.version_info < (3, 11):
```
graingert:
```suggestion
if sys.version_info < (3, 11):  # pragma: no cover
```

starlette/_utils.py
```diff
@@ -71,15 +75,33 @@ async def __aexit__(self, *args: typing.Any) -> None | bool:
...
     try:
         yield
-    except BaseException as exc:
-        if has_exceptiongroups:  # pragma: no cover
-            while isinstance(exc, BaseExceptionGroup) and len(exc.exceptions) == 1:
-                exc = exc.exceptions[0]
-
-        raise exc
+    except BaseExceptionGroup as excs:
+        if len(excs.exceptions) != 1:
```
graingert, line 88: Ah this branch needs coverage now

tests/test__utils.py
```diff
@@ -1,11 +1,15 @@
...
 from typing import Any
 
 import pytest
 
-from starlette._utils import get_route_path, is_async_callable
+from starlette._utils import create_collapsing_task_group, get_route_path, is_async_callable
 from starlette.types import Scope
 
+if sys.version_info < (3, 11):  # pragma: no cover
+    from exceptiongroups import ExceptionGroup
```
graingert:
```suggestion
    from exceptiongroup import ExceptionGroup
```

pyproject.toml
```diff
@@ -68,7 +68,6 @@ combine-as-imports = true
 
 [tool.mypy]
 strict = true
-python_version = "3.9"
```
graingert: This breaks running mypy on different versions of python in CI

In this case it breaks finding the exceptiongroup backport module on `sys.version_info < (3, 11)`

starlette/responses.py
```diff
@@ -258,7 +259,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
             except OSError:
                 raise ClientDisconnect()
         else:
-            async with anyio.create_task_group() as task_group:
+            async with create_collapsing_task_group() as task_group:
```
Kludex, line 273: Do we also have an issue with `StreamingResponse` standalone?

With my comment on the other PR I was trying to avoid the changes here. 🤔

graingert: Yeah it's the same sort of issue, but it's wrapped in two TaskGroups before the user gets the exception. Collapsing is ok here because if wait_for_disconnect raises an exception it is always 'catastrophic' eg won't be caught

hasB4K:
I have encountered this issue with starlette. I would love to have this PR merged tbh :)
Amazing work

starlette/_utils.py
```diff
@@ -70,16 +74,28 @@ async def __aexit__(self, *args: typing.Any) -> None | bool:
...
+            raise
+
+        exc = excs.exceptions[0]
+        context = exc.__context__
+        tb = exc.__traceback__
+        cause = exc.__cause__
+        sc = exc.__suppress_context__
+        try:
+            raise exc
```
alexmojaki:
```suggestion
            raise exc from exc.__cause__ or exc.__context__
```

This will prevent weird traceback repetition as in https://github.com/pydantic/logfire/issues/1279#issuecomment-3137505961 and also means that existing `__context__` won't be lost from the traceback and replaced with the exception group as the context instead, although it will become `__cause__` instead.

Kludex (approved):

PR #2833, opened 2024-12-31
fix create_memory_object_stream not subscriptable at runtime in old anyio versions

graingert:
# Summary

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 dependencies = [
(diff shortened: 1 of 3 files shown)


graingert:
@Kludex probably this needs a test matrix for each anyio version you want to support?

Kludex:
Were you able to test against anyio 3.4.0?

graingert:
3.6.2 is the oldest version I could get running

Kludex:
We should change the pinned version then

graingert:
@Kludex bumped

Kludex (approved):

PR #2836, opened 2025-01-01
Drop trio from test suite

Kludex:
The motivation here is that I don't believe having trio on the test suite is necessary.

`anyio` is the compatibility layer, and testing that both behave the same is supposed to be on that side.


Kludex:
I don't understand why 3.11 fails but 3.10 doesn't.

Kludex:
why only 3.11? o.o

graingert:
* this saves like 1 second in CI? seems not worth it.
* It's also useful to be able to run on both trio and asyncio to see if something's an anyio bug - because usually it will pass in trio and fail strangely on asyncio.
* Also I just like having trio around as a test dep here, I think it's an important message

Kludex:
> this saves like 1 second in CI? seems not worth it.

It's more because conceptually, `anyio` the compatibility layer.

Although locally, having 2 outputs for each test is annoying.

---

If I open this PR, it's because it bothered me to the point to open it. But I don't think it's worth it.

PR #2840, opened 2025-01-04
Add deprecated warnings to `TestClient` on use of `timeout` argument

### LINKED ISSUE #1108: TestClient.request does not honor timeout

iudeen:
Implements the solution in https://github.com/encode/starlette/issues/1108#issuecomment-2570876216

# Checklist

Edit by @kludex:
- Closes https://github.com/encode/starlette/issues/1108


iudeen:
@lealre I like the way you had created the test! I added that part of it from your PR. 😄

Kludex (approved):
Thanks. 🙏

PR #2841, opened 2025-01-04
Add deprecation warning to `timeout` in `TestClient`

lealre:
# Summary

Related to #1108

# Checklist


lealre:
Closing

I did not see that there was already a PR related to (#2840).

PR #2847, opened 2025-01-17
Upgrade `python-multipart` to >=0.0.19

DanielKusyDev:
# Summary
Upgrade `python-multipart` to avoid versions affected by [GHSA-59g5-xgcq-4qw3](https://github.com/advisories/GHSA-59g5-xgcq-4qw3)

# Checklist


Kludex:
It's already on the range without vulnerabilities.

PR #2849, opened 2025-01-21
use predefined status code instead of raw in authentication middleware

snosratiershad:
P.S.: very small change so I didn't opened a discussion for it.


adriangb (approved):
Looks good to me!

Kludex:
We don't really use HTTP status code or WebSocket close code internally, even if we have the status module.

This is a single place, but there are many others that would need to change.

For that reason, considering there's no great win, and increases verbosity, I'll be closing this PR. Thanks tho!

PR #2856, opened 2025-01-29
Do not process FileNotFoundError in FileResponse, resolves #979

### LINKED ISSUE #979: FileResponse is catching a FileNotFoundError to raise a RuntimeError
Found this in FileResponse:
```python
except FileNotFoundError:
    raise RuntimeError(f"File at path {self.path} does not exist.")
```
I feel this is catching one exception just to raise a less meaningful one. For my unit tests, I would prefer to expect the FileNotFoundError is raised.

toxadx:
FileNotFoundError can be processed in exception_handlers

# Summary
Avoid handling FileNotFoundError in FileResponse.
This allows FileNotFoundError to be properly handled using exception_handlers, for example, by returning a 404 error when a file is not found.

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -341,7 +341,7 @@ class FileResponse(Response):
                 stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
                 self.set_stat_headers(stat_result)
             except FileNotFoundError:
-                raise RuntimeError(f"File at path {self.path} does not exist.")
+                pass
             else:
                 mode = stat_result.st_mode
                 if not stat.S_ISREG(mode):
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -299,9 +299,8 @@ def test_file_response_with_missing_file_raises_error(tmp_path: Path, test_clien
     path = tmp_path / "404.txt"
     app = FileResponse(path=path, filename="404.txt")
     client = test_client_factory(app)
-    with pytest.raises(RuntimeError) as exc_info:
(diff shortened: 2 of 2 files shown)


starlette/responses.py
```diff
@@ -341,7 +341,7 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
                 stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
                 self.set_stat_headers(stat_result)
             except FileNotFoundError:
-                raise RuntimeError(f"File at path {self.path} does not exist.")
+                pass
             else:
                 mode = stat_result.st_mode
```
Kludex, line 346: 2 lines below, we have another `RuntimeError`.

Also, this is a breaking change.

Kludex (changes requested):
Can you show me a real example of what you are trying to do?

The client asks for a file, and you use `FileResponse`? What about `StaticFiles`?

starlette/responses.py
alex-oleshkevich: We can raise a new exception derived from both `RuntimeError` and `FileNotFoundError`. This way we can keep backward compatibility.
```python
class NotFoundError(FileNotFoundError, RuntimeError): ...
```

Kludex: Yeah, I think something on those lines makes more sense.

Kludex:
Closing, as I don't think `pass` is the behavior we want. But check my comment above.

PR #2858, opened 2025-02-01
Detect blocking calls in coroutines using BlockBuster

cbornet:
# Summary

This PR uses the [blockbuster](https://github.com/cbornet/blockbuster) library to detect blocking calls made in the asyncio event loop during unit tests.
Avoiding blocking calls is hard as these can be deeply buried in the code or made in 3rd party libraries.
Blockbuster makes it easier to detect them by raising an exception when a call is made to a known blocking function (eg: `time.sleep`).

# Checklist


starlette/datastructures.py
```diff
@@ -438,7 +438,7 @@ async def write(self, data: bytes) -> None:
         if self.size is not None:
             self.size += len(data)
 
-        if self._in_memory:
+        if self._in_memory and self.file.tell() + len(data) <= self.file._max_size:
```
cbornet: The `SpooledTemporaryFile` `rollover` is blocking

starlette/testclient.py
```diff
@@ -288,7 +288,7 @@ async def receive() -> Message:
                     await response_complete.wait()
                 return {"type": "http.disconnect"}
 
-            body = request.read()
+            body = await anyio.to_thread.run_sync(request.read)
```
cbornet, line 292: Httpx `request.read` can be blocking (eg: reading multipart file).
`await request.aread()` can't be used as-is since the request `stream` is a `SyncByteStream` not an `AsyncByteStream`.

tests/test_templates.py
```diff
@@ -23,7 +23,7 @@ def test_templates(tmpdir: Path, test_client_factory: TestClientFactory) -> None
     with open(path, "w") as file:
         file.write("<html>Hello, <a href='{{ url_for('homepage') }}'>world</a></html>")
 
-    async def homepage(request: Request) -> Response:
+    def homepage(request: Request) -> Response:
```
cbornet, line 26: Is it OK to change ?
`templates.TemplateResponse` is blocking.

tests/conftest.py
```diff
@@ -4,11 +4,19 @@
...
+from blockbuster import blockbuster_ctx
 
 from starlette.testclient import TestClient
 from tests.types import TestClientFactory
 
 
+@pytest.fixture(autouse=True)
+def blockbuster(request):
+    with blockbuster_ctx("starlette") as bb:
+        bb.functions["os.stat"].can_block_in("/mimetypes.py", "init")
```
cbornet, line 17: `FileResponse`'s constructor calls `mimetypes .guess_type` which is blocking the first time it is called.

.gitignore
```diff
@@ -11,3 +11,4 @@ venv*/
 .python-version
 build/
 dist/
+.idea/
```
Kludex: You can add this to your global `.gitignore`.
```suggestion
```

starlette/datastructures.py
```diff
@@ -438,7 +439,7 @@ async def write(self, data: bytes) -> None:
         if self.size is not None:
             self.size += len(data)
 
-        if self._in_memory:
+        if self._in_memory and self.file.tell() + len(data) <= getattr(self.file, "_max_size", sys.maxsize):
```
Kludex, line 442: what's this about?

tests/middleware/test_base.py
```diff
@@ -468,7 +468,7 @@ async def cancel_on_disconnect(
 
             # A timeout is set for 0.1 second in order to ensure that
             # we never deadlock the test run in an infinite loop
-            with anyio.move_on_after(0.1):
+            with anyio.move_on_after(0.2):
```
Kludex, line 494: Why?

tests/test_templates.py
Kludex: Why is `templates.TemplateResponse` blocking?

Kludex (commented):
I'm not keen to add a dependency, but seems a cool library.

starlette/datastructures.py
cbornet: If the data to write makes the file exceed the `SpooledTemporaryFile` `_max_size`, `self.file.write` will do a blocking `rollover` operation.

tests/middleware/test_base.py
cbornet: BlockBuster has a small toll on performance.
In CI, the test machines are already slow, the cancellation sometimes arrives too late and the test becomes flaky (I had no issue on my computer, only in CI).

tests/test_templates.py
cbornet: It loads the `index.html` from the FS using `jinja2.FileSystemLoader.get_source()` which does blocking calls.
The path is:
```sh
test_templates.py:27: in homepage
    return templates.TemplateResponse(request, "index.html")
../starlette/templating.py:208: in TemplateResponse
    template = self.get_template(name)
../starlette/templating.py:131: in get_template
    return self.env.get_template(name)
../venv/lib/python3.11/site-packages/jinja2/environment.py:1016: in get_template
    return self._load_template(name, globals)
../venv/lib/python3.11/site-packages/jinja2/environment.py:975: in _load_template
    template = self.loader.load(self, name, self.make_globals(globals))
../venv/lib/python3.11/site-packages/jinja2/loaders.py:126: in load
    source, filename, uptodate = self.get_source(environment, name)
../venv/lib/python3.11/site-packages/jinja2/loaders.py:204: in get_source
    if os.path.isfile(filename):
```
`path.isfile` is a blocking call as it does an `os.stat`.
The `get_source` code also does a `file.read()` blocking call later.

adriangb:
@cbornet I'm curious about the choice to call out file IO as blocking. In my experience file IO is tricky:
1. The issue with blocking calls is not if they are theoretically blocking or not, the issue is how long they block for. E.g. sorting a list is blocking work but as long as it's <1ms it's probably fine. When you start getting into multi ms blocks is when things get ugly since IO protocols that expect a regular tick and such may start timing out, causing cascading failures.
2. File IO can of course block for a long time (e.g. reading GBs of data in one go) but it can also be very fast (small operation on an SSD disk). And it's hard to know ahead of time which one it will be.
3. The overhead of starting up threads is non negligible, and anyio's `to_thread` has a relatively low semaphore (40 last time I checked) which is easy to accidentally exhaust, possibly leading to deadlocks.

Because of this my general approach has been to be conservative with chucking things into threads, especially file IO since it's often not a problem in practice.

cbornet:
@adriangb thanks for your feedback.
I agree with you that the impact of blocking calls depends a lot on the time it blocks. But how to know in advance ? Things can run perfectly well on your laptop or CI with SSDs and fall apart when you deploy to AWS with EFS slow network disks (bad recent personal experience).
I’ve seen that in general file ops are deferred to threads (or if using Linux, you can use aiofile which has true async support).
And I see that starlette already does it in a bunch of places (eg: UploadFile.write)
Starting a thread has a cost but I think anyio uses a thread pool ?
Anyway, if you don’t want to use threads, it’s possible to:
* set exemptions in the places detected by blockbuster. This way, you still have a kind of warning that something could be better.
* or completely disable file IO blockings detection by blockbuster.

starlette/datastructures.py
Kludex: Should come soon: https://github.com/agronholm/anyio/pull/867

cbornet:
Hi. Could you provide guidance on what I should do for this PR ?

starlette/datastructures.py
adriangb: Other than accessing the private attribute this makes sense to me. Insofar as it's an optimization that fails gracefully I think it would be okay to do this.

starlette/testclient.py
alex-oleshkevich: IMO, this is not a problem in unit tests and we can neglect it.

tests/test_templates.py
alex-oleshkevich: Jinja loads it only once, and then it is served from the in-memory cache.

cbornet: Yes. For production code, it’s up to the user to see if it’s a perf issue or not.
Here I’m modifying the test code so that the test passes with BlockBuster on.
I think it doesn’t change the test itself and what it is actually testing. Is it OK?

cbornet:
> Hi. Could you provide guidance on what I should do for this PR ?

Hi. I rebased. Is there anything I can do to make progress on this PR ?

Kludex:
Sorry for taking so long here @cbornet .

Thanks for your work on BlockBuster. I've also added it to https://github.com/pydantic/pydantic-ai. 🙏

PR #2865, opened 2025-02-06
fix(gzip): Make sure Vary header is always added if a response can be compressed

mattmess1221:
Fixes #2863

# Summary

Since the presense or absense of the `Accept-Encoding` header can change the response body, `Vary` should always be added whether the client requested gzip encoding or not. Otherwise some caching implementations might return a non-compressed response even when a compressed one was requested.

This change may seem a bit complex, but it reworks the `GZipMiddleware` to behave the same no matter what the compression is (or lack of). Basic changes include:

1. Move all gzip compression to a method `apply_compression` and don't update the `content-encoding` or `content-length` headers if the body didn't change.
2. Create a base class `IdentityResponder` from `GZipResponder` which will be used to respond to non-gzip requests. Its `apply_compression` method will return the body unchanged.
3. Coincidentally, it's now easier to implement new compression algorithms like deflate, brotli, or zstd. This PR does not implement those.

# Checklist

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -15,17 +15,24 @@ class GZipMiddleware:
         self.compresslevel = compresslevel
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if scope["type"] == "http":  # pragma: no branch
-            headers = Headers(scope=scope)
-            if "gzip" in headers.get("Accept-Encoding", ""):
-                responder = GZipResponder(self.app, self.minimum_size, compresslevel=self.compresslevel)
-                await responder(scope, receive, send)
-                return
-        await self.app(scope, receive, send)
+        if scope["type"] != "http":  # pragma: no cover
+            await self.app(scope, receive, send)
+            return
+
+        headers = Headers(scope=scope)
+        responder: ASGIApp
+        if "gzip" in headers.get("Accept-Encoding", ""):
+            responder = GZipResponder(self.app, self.minimum_size, compresslevel=self.compresslevel)
+        else:
+            responder = IdentityResponder(self.app, self.minimum_size)
+
+        await responder(scope, receive, send)
 
 
-class GZipResponder:
-    def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 9) -> None:
+class IdentityResponder:
+    content_encoding: str
+
+    def __init__(self, app: ASGIApp, minimum_size: int) -> None:
         self.app = app
         self.minimum_size = minimum_size
         self.send: Send = unattached_send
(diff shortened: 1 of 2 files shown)


Kludex:
couldn't we just move the `add_vary_header` up?

mattmess1221:
Unfortunately, it also needs to be added when the accept-encoding header is *not* gzip.

Kludex:
> Unfortunately, it also needs to be added when the accept-encoding header is _not_ gzip.

I understand that, that's the goal of this PR. Can't we really avoid this huge diff just to move that up?

mattmess1221:
My thinking was I would need to implement a responder for non-gzip encodings, and I also wanted to use the same logic for determining if I *shouldn't* add the Vary header (body is too small, or streaming response with content-encoding already set).

Anyway, I'll try to get it smaller.

Kludex:
hold, I'm going to review carefully.

No need to try anything for now. Thanks for the PR.

Kludex:
This PR is actually changing the behavior on when we do the compression. It actually implements what is proposed here: https://github.com/encode/starlette/pull/2753

mattmess1221:
Without flushing, the tests would fail.

Kludex:
> Without flushing, the tests would fail.

That's a change in behavior. I don't think it should flush on every chunk.

starlette/middleware/gzip.py
```diff
@@ -60,49 +64,76 @@ async def send_with_gzip(self, message: Message) -> None:
...
+
+    def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
+        self.gzip_file.write(body)
+        self.gzip_file.flush()
+        body = self.gzip_buffer.getvalue()
+
+        if not more_body:
+            self.gzip_file.close()
+
+        self.gzip_buffer.seek(0)
```
```suggestion
        self.gzip_file.write(body)
        if not more_body:
            self.gzip_file.close()

        body = self.gzip_buffer.getvalue()
        self.gzip_buffer.seek(0)
```

tests/middleware/test_gzip.py
```diff
@@ -57,10 +63,12 @@ def homepage(request: Request) -> PlainTextResponse:
     assert response.status_code == 200
     assert response.text == "OK"
     assert "Content-Encoding" not in response.headers
+    assert "Vary" not in response.headers
     assert int(response.headers["Content-Length"]) == 2
 
 
-def test_gzip_streaming_response(test_client_factory: TestClientFactory) -> None:
+@pytest.mark.parametrize("encoding", ["gzip", None])
+def test_gzip_streaming_response(encoding: str | None, test_client_factory: TestClientFactory) -> None:
```
Kludex: Can you split this test in two, please?

mattmess1221: Tests split

Kludex (approved):
Thanks. 🙏

PR #2810, opened 2024-12-26
Add `client` parameter to `TestClient`

iudeen:
# Summary
Fixes https://github.com/encode/starlette/issues/2295

Added client and port parameter to TestClient

PR #2870, opened 2025-02-15
Use `localhost` instead of `http://localhost` in `TestClient(client=...)` example

WilliamDEdwards:
# Summary

With https://github.com/encode/starlette/pull/2810, it is possible to override the client host (i.e. the remote address; see https://www.starlette.io/requests/#client-address).

The default value for `TestClient` is `testclient`, but an IP address in most circumstances.

The documentation example for overriding the client host uses a full URI, which makes no sense: it's the remote address.

# Checklist


Kludex (approved):

Kludex:
Thanks 🙏

PR #2872, opened 2025-02-17
Expose `url_path_for` on `HTTPConnection`

judahrand:
# Summary

Based on the conclusion of #604 which can be found [here](https://github.com/encode/starlette/issues/604#issuecomment-578895744) the consensus seems to be that trying to munge host headers when located behind a reverse proxy isn't the best approach and that instead absolute paths should be used in redirects. The same, I imagine, should apply in tempates.

This is currently possible but would be made easier/neater if the `url_path_for` method which exists on `Starlette` was exposed on the `HTTPConnection` and therefore the `Request`. Additionally, adding a `url_path_for` `jinja` function would allow for this to be achieved in templates much easier.

I leave this PR here for review with the knowledge that if this approach is approved of docs will also need to be added.

# Checklist

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -189,7 +189,9 @@ class HTTPConnection(typing.Mapping[str, typing.Any]):
     def url_path_for(self, name: str, /, **path_params: typing.Any) -> URLPath:
         url_path_provider: Router | Starlette | None = self.scope.get("router") or self.scope.get("app")
         if url_path_provider is None:
-            raise RuntimeError("The `url_for` method can only be used inside a Starlette application or with a router.")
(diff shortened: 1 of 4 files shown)


alex-oleshkevich:
`self.url_for` returns and instance of `URL` that you can use in routes/templates to get a path: `self.url_for(name).path` (you have it already done in the PR). 
I don't think we need another method.

The difference between `Starlette.url_path_for` and `HTTPConnection.url_for` is that Starlette app can be used in non-HTTP contexts (like CLI or background tasks) and therefore has no `Host` concept. In all other cases I advise using `url_for`.

PR #2884, opened 2025-02-23
Feat/upload file max size disk

khadrawy:
# Summary

Add check for uploaded file size on disk against configurable limit.

# Checklist


khadrawy:
@Kludex after https://github.com/encode/starlette/pull/2718 I propose this change to apply limits to uploaded file size on disk.

PR #2889, opened 2025-02-26
requests.py: remove wrong assert scope["type"] == "http" in Request

tonjo:
It differs from parent class.
Parent class HTTPConnection's `__init__` already asserts scope type is `"http"` or `"websocket"`, but derived Request class only checks `"http"`, so it still raises an error in case of `"websocket"`

# Summary

# Checklist

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -200,7 +200,6 @@ class Request(HTTPConnection):
 
     def __init__(self, scope: Scope, receive: Receive = empty_receive, send: Send = empty_send):
         super().__init__(scope)
-        assert scope["type"] == "http"
         self._receive = receive
         self._send = send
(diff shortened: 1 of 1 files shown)


Kludex:
Yeah, this is correct.

tonjo:
I didn't understand, if my correction was correct, why it has not been merged :) ?

Kludex:
I meant the code that is in master is correct.

tonjo:
Ok, thahks, but why?
It always throws errors with websockets...
HTTPConnection accepts websockets also as scope type, why Request does not?

Kludex:
Because the Request object is built for HTTP. The WebSocket object is built for WebSockets.

PR #2896, opened 2025-03-05
Allow relative directory path when `follow_symlinks=True`

davipt:
# Summary

Fix for StaticFiles(follow_symlinks=True, directory="some relative path") that stopped working with commit eee4cdc.

When using StaticFiles with a relative path (e.g. a static folder included with the source code deployment) and follow_symlinks is set to True, the commit at eee4cdc removed the code that translates the relative folder into an absolute folder, and then the system call os.path.commonpath fails because it expects the arguments to either be all relative or all absolute.

This MR fixes this issue by using the same pattern as the full_path - os.abspath for follow_symlinks=True, and os.realpath when False. 

# Checklist

- [-] I've updated the documentation accordingly.

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -153,6 +153,7 @@ class StaticFiles:
             joined_path = os.path.join(directory, path)
             if self.follow_symlink:
                 full_path = os.path.abspath(joined_path)
+                directory = os.path.abspath(directory)
             else:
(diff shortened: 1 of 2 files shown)


Kludex (approved):
Thanks. I left only the test that was failing in master.

davipt:
Awesome, thanks!

PR #2867, opened 2025-02-07
chore: improve type hints

waketzheng:
# Summary

Only improve type hints, does not change any logic
1. typing.xxx --> from typing import xxx
2. Run `ruff check --fix starlette tests`
3. `from typing import ContextManager` --> `from contextlib import AbstractContextManager`

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly.

PR #2903, opened 2025-03-11
Make annotation for `methods` more permissive

jonathanberthias:
# Summary

The type annotation for `methods` in the `Route` constructor restricts the input to lists of strings when it can accept any iterable. The `methods` on the `Route` itself are stored as a set.

I had the issue when I was trying to build new routes from other preexisting routes:
```python
new_route = Route(
    route.path.replace("/prefix1", "/prefix2"),
    route.endpoint,
    methods=route.methods,  # type error: Argument "methods" to "Route" has incompatible type "set[str] | None"; expected "list[str] | None"
    name=f"new_{route.name}",
)
```

I changed the annotation to [`Collection`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection), though the current implementation would work with any [`Iterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable). `Collection` allows lists, sets, tuples, and `Iterable` would also allow generators. Let me know what you prefer.

# Checklist


starlette/routing.py
```diff
@@ -211,7 +211,7 @@ def __init__(
         path: str,
         endpoint: typing.Callable[..., typing.Any],
         *,
-        methods: list[str] | None = None,
+        methods: typing.Collection[str] | None = None,
```
alex-oleshkevich: We typically use `typing.Sequence` across the project.

jonathanberthias: Indeed, but in this case it makes sense to allow sets in addition to list and tuples

alv2017: Reminder: https://docs.python.org/3/library/typing.html#deprecated-aliases
Shouldn't we start applying new standards?

jonathanberthias: I followed the current convention of importing everything from the typing module, but if #2867 gets merged before this PR, then I'll adjust 😄

jonathanberthias:
The tests on 3.14 are failing due to a change in `argparse` that errors within MyPy :shrug:

starlette/routing.py
jonathanberthias: I changed to import from `collections.abc`

Kludex (approved):

PR #2910, opened 2025-03-16
refactoring: `tests/middleware/test_middleware.py`

alv2017:
# Summary

This PR 
a) refactors the code of the `tests/middleware/test_middleware.py`. 
b) decorates `_MiddlewareFactory` with `@runtime_checkable` decorator. This enables usage of `isinstance()` with the protocol class.

It offers:
- more readable and more flexible code structure
- the tested middleware object is instantiated once per test class, and then re-used by test methods of the class.
- additional test added `test_created_instance()`, it shows that the created middleware object is an instance of `Middleware` and its `cls` attribute is equal to `CustomMiddleware`.
- additional test added `test_class_matches_middleware_protocol()`, showing that `CustomMiddleware` class is compliant with the `_MiddlewareFactory` protocol class 

# Checklist

### DIFF
--- a/tests/middleware/test_middleware.py
+++ b/tests/middleware/test_middleware.py
@@ -12,11 +12,21 @@ class CustomMiddleware:  # pragma: no cover
         await self.app(scope, receive, send)
 
 
-def test_middleware_repr() -> None:
-    middleware = Middleware(CustomMiddleware, "foo", bar=123)
-    assert repr(middleware) == "Middleware(CustomMiddleware, 'foo', bar=123)"
+class TestMiddleware:
+    custom_middleware = Middleware(CustomMiddleware, "foo", bar=123, baz="custom")
 
+    def test_created_instance(self) -> None:
+        custom_middleware = self.custom_middleware
+        assert isinstance(custom_middleware, Middleware)
+        assert custom_middleware.cls == CustomMiddleware
 
-def test_middleware_iter() -> None:
-    cls, args, kwargs = Middleware(CustomMiddleware, "foo", bar=123)
-    assert (cls, args, kwargs) == (CustomMiddleware, ("foo",), {"bar": 123})
+    def test_repr(self) -> None:
+        custom_middleware = self.custom_middleware
+        assert repr(custom_middleware) == "Middleware(CustomMiddleware, 'foo', bar=123, baz='custom')"
+
+    def test_iter(self) -> None:
+        custom_middleware = self.custom_middleware
+        cls, args, kwargs = custom_middleware
+        assert cls == custom_middleware.cls
+        assert args == ("foo",)
+        assert kwargs == {"bar": 123, "baz": "custom"}
--- a/tests/middleware/test_middleware.py
+++ b/tests/middleware/test_middleware.py
@@ -1,4 +1,4 @@
-from starlette.middleware import Middleware
+from starlette.middleware import Middleware, _MiddlewareFactory
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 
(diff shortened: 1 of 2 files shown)


Kludex:
I'm not sure if I see much benefit on this... When would you want to do this?

alv2017:
@Kludex That is true, but this is all about incremental improvements and with the time being the benefits will start showing up.
For example, have a look at `tests/middleware/test_base.py`, this file has already over the 1000 lines of code, it becomes difficult to follow up and difficult to maintain. It is about time to start refactoring it and giving  it a better structure.

tests/middleware/test_middleware.py
```diff
@@ -12,11 +12,25 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         await self.app(scope, receive, send)
 
 
-def test_middleware_repr() -> None:
-    middleware = Middleware(CustomMiddleware, "foo", bar=123)
-    assert repr(middleware) == "Middleware(CustomMiddleware, 'foo', bar=123)"
+def test_class_matches_middleware_protocol() -> None:
+    assert isinstance(CustomMiddleware, _MiddlewareFactory)
```
adriangb, line 16: Why are you deleting the old test and adding a new seemingly unrelated test? I don't think this test is valuable. If you wanted to check this do it via type checking: `foo: _MiddlewareFactory = CustomMiddleware`. But I don't think that is very valuable: `CustomMiddleware` is a test only class, we don't need to test it directly only indirectly via it's involvement in other tests

tests/middleware/test_middleware.py
```diff
@@ -12,11 +12,25 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
 
 
-def test_middleware_iter() -> None:
-    cls, args, kwargs = Middleware(CustomMiddleware, "foo", bar=123)
-    assert (cls, args, kwargs) == (CustomMiddleware, ("foo",), {"bar": 123})
+class TestMiddleware:
```
adriangb, line 19: Why this refactor? We have an existing test layout, modules with functions in them. I'm not going to argue what test layout is best, there's always tradeoffs, but changing the existing test layout of an established project like this is going to require a strong reason to do so.

adriangb (changes requested):
Sorry but I don't see the value in this PR or a long term story being proposed here.

> For example, have a look at tests/middleware/test_base.py, this file has already over the 1000 lines of code, it becomes difficult to follow up and difficult to maintain. It is about time to start refactoring it and giving it a better structure.

I agree with you that files with thousands of lines of code and many, many test can be hard to grok and maintain. But bad organization / abstractions (e.g. making many small arbitrary modules) is worse than no organization / abstraction.
We'd be open to a PR that re-organizes / renames / etc. while keeping the current structure and without changing what is being tested, but it would need to show an obvious improvement to the codebase.

tests/middleware/test_middleware.py
alv2017: No, this is a completely new test, and it demonstrates, how @runtime_checkable is applied.

tests/middleware/test_middleware.py
alv2017: The first problem with the initial tests is creation of a new middleware object in each test. This is not needed. Doing this in thousands of tests results in performance issues (and as a result,  increased enviromental damage)

The second problem: difficulty with adding new tests, and as a result loss of control of the code and loss of value of the tests (I'm referring here `tests/middleware/test_base.py`). I think this is a strong reason. It is not wise to wait untill the things go completely out of control.

adriangb: > as a result, increased enviromental damage

I think you are exaggerating here. Starlette's test suite is relatively small and quick to run, especially compared to other things in our industry nowadays like asking an AI to do things for you. And you could always just assign it to a global and call it a day, no need to introduce a new class

alv2017: @adriangb: Class was introduced for grouping and readability.

alv2017: all in all I'm OK with closing this PR, but lets see a bigger picture. I will be messing around with Starlette for a while :smile: And I will picking up a lot of stuff along the way, it is unlikely that I will not be introducing significant improvements to the code base, but small improvements is a good thing as well. :blush:

PR #2913, opened 2025-03-28
Fix link label in the documentation for contextvars.ContextVar

andri1gryniv:
# Summary

Fixing a typo in the documentation page: https://www.starlette.io/middleware/#limitations

The title for a link https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar is not correct.

Current value: `contextlib.ContextVar`
Fixing to: `contextvars.ContextVar`

# Checklist


Kludex (approved):

Kludex:
Thanks!

PR #2815, opened 2024-12-26
Add `max_part_size` parameter to `MultiPartParser`

iudeen:
Fixes #2785 

Also addresses issue discussed in [FastAPI Repo](https://github.com/fastapi/fastapi/discussions/12961#discussioncomment-11336476).

PR #2915, opened 2025-04-01
fix: make MultiPartParser read default max_part_size from the class attr

reith:
# Summary

 In FastAPI, the only way to change the maximum part size in a multipart request [is to override](https://github.com/fastapi/fastapi/discussions/12961) `MultiPartParser.max_part_size` globally. That has stopped working since the changes introduced by  [this commit](https://github.com/encode/starlette/pull/2815). I think the change in this PR is benign to keep. Even without considering compatibility with FastAPI, I think falling back to the class attribute to read the default value is a good idea but I don't mind if you think that problem should be addressed separately by FastAPI.

# Checklist

### DIFF
--- a/starlette/formparsers.py
+++ b/starlette/formparsers.py
@@ -134,7 +134,7 @@ class MultiPartParser:
         *,
         max_files: int | float = 1000,
         max_fields: int | float = 1000,
-        max_part_size: int = 1024 * 1024,  # 1MB
+        max_part_size: int | None = None,
     ) -> None:
         assert multipart is not None, "The `python-multipart` library must be installed to use form parsing."
         self.headers = headers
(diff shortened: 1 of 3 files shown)


Kludex:
The class attributes should be removed, they are still there by mistake.

FastAPI should be smart about this.

Kludex:
That said... I understand it can be a bit annoying to have to wait for FastAPI...

I think we need a constant `DEFAULT_MAX_SIZE_...` or something like that, which will allow users to change the behavior while waiting for FastAPI.

reith:
I ended up getting the raw request in my handler and using the Starlette's Request API there to change the part size limit. That’s better than changing the default since I perform authorization in a middleware, and it makes more sense not to raise the limit for unauthenticated requests.

Thanks for being considerate, though!

PR #2919, opened 2025-04-12
Include Python 3.14 to the pipeline

Kludex:


mkdocs.yml
```diff
@@ -2,6 +2,12 @@ site_name: Starlette
 site_description: The little ASGI library that shines.
 site_url: https://www.starlette.io
 
+repo_name: encode/starlette
+repo_url: https://github.com/encode/starlette
+edit_uri: edit/master/docs/
+
+strict: true
```
Kludex, line 9: The reference to `sponsorship.md` was wrong. We should be strict.

PR #2922, opened 2025-04-13
Mark http_exception as async to prevent thread creation.

danlapid:
# Summary

In https://github.com/encode/starlette/blob/master/starlette/_exception_handler.py#L61 we check if the exception_handler is an async function and if not we execute it in a different thread. 
This is meant to ensure that long-running sync IO operations don't block the main async eventloop. 
However, in the case of [http_exception](https://github.com/encode/starlette/blob/master/starlette/middleware/exceptions.py#L64), a simple function that does not do any io at all - the aforementioned logic spawns a thread to run the function. 
Marking this function async is a simple way of making sure this function does not accidently cause a thread to be created.

# Checklist

### DIFF
--- a/starlette/middleware/exceptions.py
+++ b/starlette/middleware/exceptions.py
@@ -61,7 +61,7 @@ class ExceptionMiddleware:
 
         await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
 
-    def http_exception(self, request: Request, exc: Exception) -> Response:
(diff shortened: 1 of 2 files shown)


Kludex (approved):
Thanks :)

I'm not sure how no one saw this...

danlapid:
@Kludex no worries 😄 
Do you mind hitting the merge button as well? I don't have permissions 🙏

danlapid:
@Kludex bumping this, would appreciate some help hitting the merge button 😄

PR #2925, opened 2025-04-16
fix: Use anyio.SpooledTemporaryFile in UploadFile

11kkw:
# Summary

This PR updates the internal implementation of `UploadFile` to use `anyio.SpooledTemporaryFile[bytes]` instead of a generic `BinaryIO` interface.

By using `anyio.SpooledTemporaryFile`, all file operations (`read`, `write`, `seek`) are now fully asynchronous and consistent across both in-memory and disk-based file handling. This eliminates the need for `run_in_threadpool` and simplifies the logic for determining whether the file is in memory or on disk.

This is my first contribution to Starlette. If any part of this PR doesn't follow the project's conventions or needs to be revised or split, I’m more than happy to make the necessary changes.

# Checklist


starlette/datastructures.py
```diff
@@ -413,7 +414,7 @@ class UploadFile:
 
     def __init__(
         self,
-        file: typing.BinaryIO,
+        file: anyio.SpooledTemporaryFile[bytes],
```
Kludex: Is there a base class analogous to BinaryIO? This was not a spoiled temporary file type before on purpose.

Kludex: Because that shouldn't matter for the UploadFile.

11kkw: > Because that shouldn't matter for the UploadFile.

Yeah, that makes sense — the original use of BinaryIO was definitely meant to keep things implementation-agnostic, and I get why switching directly to SpooledTemporaryFile might feel too specific.

If you're open to it, I think one possible middle ground could be introducing a small AsyncFile protocol — something simple that defines the expected async methods (read, write, seek, etc). That way, we can preserve the flexibility while still ensuring everything stays fully async.

Happy to adjust the PR in that direction if it sounds reasonable to you!

Kludex: Sounds reasonable, but maybe that structure already exists in anyio? 🤔

11kkw: > Sounds reasonable, but maybe that structure already exists in anyio? 🤔

Yes, it seems like anyio provides a type called AsyncFile for async file operations

graingert: AsyncFile is a concrete implementation not an abc, probably uh "AsyncFileIO[bytes]" protocol would make sense here.

If it's a protocol we can move it or have it in multiple places and have everything still work

starlette/formparsers.py
```diff
@@ -268,7 +269,7 @@ async def parse(self) -> FormData:
         except MultiPartException as exc:
             # Close all the files if there was an error.
             for file in self._files_to_close_on_error:
-                file.close()
```
graingert, line 273: this should be in an AsyncExitStack - if any aclose is cancelled we should continue to run the remaining ones

abersheeran (approved):
Good change!

starlette/datastructures.py
abersheeran: In fact, I don't think the file parameter should be exposed here. It seems to give users the power to customize the file, but in fact the other parts of UploadFile rely heavily on the implementation of file. If we need to make some abstractions to allow users to customize the file parsing results of multipart, we can use a design similar to [baize](https://github.com/abersheeran/baize/blob/cbb3854e8632e4f92fa0453d4d11cf065c08ccef/baize/multipart_helper.py#L26-L32). But at least in this PR, I think there is no problem with modifying the type in this way.

starlette/formparsers.py
```diff
@@ -250,30 +250,29 @@ async def parse(self) -> FormData:
...
+                    await part.file.write(data)
+                for part in self._file_parts_to_finish:
+                    assert part.file  # for type checkers
+                    await part.file.seek(0)
+                self._file_parts_to_write.clear()
+                self._file_parts_to_finish.clear()
+        except MultiPartException as exc:
+            # Close all the files if there was an error.
+            async with AsyncExitStack() as stack:
+                for f in self._files_to_close_on_error:
```
graingert: * I mean that _files_to_close_on_error would just be an exit stack
* these should be closed on any error including BaseException

starlette/formparsers.py
```diff
@@ -265,10 +267,9 @@ async def parse(self) -> FormData:
                     await part.file.seek(0)
                 self._file_parts_to_write.clear()
                 self._file_parts_to_finish.clear()
-        except MultiPartException as exc:
+        except BaseException as exc:
```
Kludex: Why did this change?

11kkw: > Why did this change?

I changed it to `BaseException` assuming it would help ensure cleanup even for cases like `CancelledError` or `KeyboardInterrupt`, beyond just `MultiPartException`.

But perhaps I misunderstood — please let me know if I’m thinking about this the wrong way!

graingert: CancelledError being the most likely one

starlette/formparsers.py
```diff
@@ -267,10 +269,9 @@ async def parse(self) -> FormData:
                 self._file_parts_to_write.clear()
                 self._file_parts_to_finish.clear()
             parser.finalize()
-        except MultiPartException as exc:
+        except BaseException as exc:
             # Close all the files if there was an error.
-            for file in self._files_to_close_on_error:
-                file.close()
+            await self._files_to_close_on_error.aclose()
             raise exc
```
graingert:
```suggestion
            raise
```

starlette/formparsers.py
```diff
@@ -267,10 +269,9 @@ async def parse(self) -> FormData:
                 self._file_parts_to_write.clear()
                 self._file_parts_to_finish.clear()
             parser.finalize()
-        except MultiPartException as exc:
+        except BaseException as exc:
```
graingert:
```suggestion
        except BaseException:
```

starlette/datastructures.py
graingert: The `UploadFile` constructor is not part of Starlette's public API — its signature is undocumented, and the docs only describe `UploadFile` as something received from `request.form()`. The sole internal caller (`formparsers.py`) is already updated by this PR. Changing the `file` parameter type from `BinaryIO` to `AsyncFileIO` is therefore not a breaking change imho.

graingert: `self.file = file` is public so maybe that should be considered a breaking change

graingert: if we're going to make a breaking change like this I'd recommend storing it on `self._file = file`

this `self.file` attribute is documented along with filename, content_type, size, and headers are documented

starlette/datastructures.py
```diff
@@ -422,27 +423,44 @@ def __init__(
...
         self.headers = headers or Headers()
 
     @property
     def content_type(self) -> str | None:
         return self.headers.get("content-type", None)
 
-    async def write(self, data: bytes) -> None:
+    @property
+    def file(self) -> AsyncFileIO:
+        warnings.warn(
```
graingert, line 443: do we want to use typing_extensions.deprecated ?

starlette/datastructures.py
```diff
@@ -422,27 +423,44 @@ def __init__(
...
+        )
+        return self._file
+
+    async def _write(self, data: bytes) -> None:
         if self.size is not None:
             self.size += len(data)
-        await self.file.write(data)
+        await self._file.write(data)
+
+    async def write(self, data: bytes) -> None:
```
graingert, line 456: do we want to use typing_extensions.deprecated?

graingert:
Starlette depends on typing_extensions>=4.10.0; python_version < '3.13'

so we can use

```py
import sys

if sys.version_info >= (3, 13):
    from warnings import deprecated
else:
    from typing_extensions import deprecated
```

PR #2934, opened 2025-05-11
docs: note about websocket support in uvicorn

sminnee:
# Summary

Docs update.

This small gotcha stumped me for a few hours and I thought it worth adding to the documentation. The separation of responsibility between uvicorn and Starlette wasn't self-evident.

# Checklist

- N/A I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.


sminnee:
Not sure why a docs update is failing the build, sorry!

AllenSH12:
Not a problem with this change, was able to reproduce this locally on the master branch using pyenv. Will dig around some more and try to submit a PR to fix.

```shell
$ git rev-parse --short HEAD
4acf1d1

(venv) $ python3 --version
Python 3.14.0b1

(venv) $ ./scripts/check
...
TypeError: AugmentedHelpFormatter.__init__() got an unexpected keyword argument 'prefix_chars'
```

Switch to 3.13.1 and it passes:
```shell
$ rm -rf venv/

$ pyenv shell system
$ python3 --version
Python 3.13.1

$ python3 -m venv venv
$ ./scripts/install

$ source venv/bin/activate
(venv) $ python3 --version
Python 3.13.1

(venv) $ ./scripts/check
...
All checks passed!
```

AllenSH12:
@Kludex perhaps it is not worth including 3.14 yet (or marking it optional)? 

It took 5 months to get from [v3.13.0b1](https://github.com/python/cpython/releases/tag/v3.13.0b1) to [v3.13.0](https://github.com/python/cpython/releases/tag/v3.13.0) and mypy is [not testing](https://github.com/python/mypy/blob/master/.github/workflows/test.yml#L69-L75) against 3.14 yet, would be a shame to have failing tests for such a potentially long time, even if 3.14.0a7 initially worked...

docs/websockets.md
```diff
@@ -2,6 +2,10 @@
 Starlette includes a `WebSocket` class that fulfils a similar role
 to the HTTP request, but that allows sending and receiving data on a websocket.
 
+**Note:** Starlette doesn't implement the websocket protocol, this must be handled
+by your ASGI web server. For example, uvicorn will need either the `wsproto` or
+`websockets` PyPI package installed.
```
Kludex: I don't think the sentence is very clear. 🤔 

Starlette does implement the WebSocket protocol.

Kludex:
> @Kludex perhaps it is not worth including 3.14 yet (or marking it optional)?
> 
> It took 5 months to get from [v3.13.0b1](https://github.com/python/cpython/releases/tag/v3.13.0b1) to [v3.13.0](https://github.com/python/cpython/releases/tag/v3.13.0) and mypy is [not testing](https://github.com/python/mypy/blob/master/.github/workflows/test.yml?rgh-link-date=2025-05-14T18%3A55%3A46Z#L69-L75) against 3.14 yet, would be a shame to have failing tests for such a potentially long time, even if 3.14.0a7 initially worked...

We can skip the type checker on 3.14, yep.

docs/websockets.md
sminnee: How would you describe the subsystem that Starlette requires uvicorn to provide? That main thing is that I was running uvicorn without websockets installed and it had me very confused! I had assumed that all the websocket support was built into Starlette, but there was a piece missing.

sminnee: Perhaps this is better?

> For websockets to work, the ASGI web server that calls Starlette will also need support for websockets. For example, uvicorn will need either the `wsproto` or `websockets` PyPI package installed. Without this, Starlette will respond with 404 errors on the websocket endpoints.

Kludex:
I don't think this comment is needed. It seems a problem with the ASGI server - it should provide a better error message!

Happy to discuss this over Uvicorn.

sminnee:
I was trying to help the next person who hit the snag I found, with a few sentences of documentation, but this is too hard.

I'm not sure how the AGSI server not configured to support websockets will know that not supporting websockets is an error.

Just in case it was missed (it had no reply), I had redrafted the proposed change:

> For websockets to work, the ASGI web server that calls Starlette will also need support for websockets. For example, uvicorn will need either the wsproto or websockets PyPI package installed. Without this, Starlette will respond with 404 errors on the websocket endpoints.

PR #2936, opened 2025-05-14
CI: skips type check until mypy ready for 3.14

AllenSH12:
# Summary

Skips mypy type check for 3.14 since they are [not testing](https://github.com/python/mypy/blob/master/.github/workflows/test.yml#L69-L75) against it yet and the build in this repo is [failing](https://github.com/encode/starlette/actions/runs/14960245809/job/42021194140) as a result.

More discussion in #2934

# Checklist

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -26,6 +26,7 @@ jobs:
         run: "scripts/install"
       - name: "Run linting checks"
         run: "scripts/check"
+        if: { { matrix.python-version != '3.14' } }
       - name: "Build package & docs"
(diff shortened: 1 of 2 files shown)


AllenSH12:
Hey @Kludex how does this look to you? Happy to make changes if needed

Kludex:
Can we skip this on the CI instead?

AllenSH12:
Sure thing, how does that look?

Kludex (approved):

Kludex:
Thanks! :)

AllenSH12:
You're welcome!

PR #2937, opened 2025-05-15
Add async test client

davidbrochart:
# Summary

This PR adds `AsyncTestClient`, an async equivalent of [TestClient](https://github.com/encode/starlette/blob/135e3af00a507714a4e391ab8d11b835399aa1a4/starlette/testclient.py#L360). The `TestClient` runs in a background thread in order to let the application run in the main thread, while the `AsyncTestClient` runs in a background task in the same thread as the application.
Currently, `AsyncTestClient` is almost a duplication of `TestClient`, with the replacement of the [BlockingPortal](https://anyio.readthedocs.io/en/stable/api.html#anyio.from_thread.BlockingPortal) with a [TaskGroup](https://anyio.readthedocs.io/en/stable/api.html#anyio.abc.TaskGroup). I'm not sure if/how some code could be shared.
See https://github.com/encode/starlette/discussions/2935.

# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -1017,14 +1017,11 @@ class AsyncTestClient(httpx.AsyncClient):
         base_url: str = "http://testserver",
         raise_server_exceptions: bool = True,
         root_path: str = "",
-        backend: typing.Literal["asyncio", "trio"] = "asyncio",
-        backend_options: dict[str, typing.Any] | None = None,
         cookies: httpx._types.CookieTypes | None = None,
         headers: dict[str, str] | None = None,
         follow_redirects: bool = True,
         client: tuple[str, int] = ("testclient", 50000),
     ) -> None:
-        self.async_backend = _AsyncBackend(backend=backend, backend_options=backend_options or {})
         if _is_asgi3(app):
             asgi_app = app
         else:
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -766,7 +766,7 @@ class AsyncWebSocketTestSession:
             self.extra_headers = message.get("headers", None)
             stack.push_async_callback(self.aclose, 1000)
             self.exit_stack = stack.pop_all()
-            return self
+        return self
 
     async def __aexit__(self, *args: typing.Any) -> bool | None:
         return await self.exit_stack.__aexit__(*args)
@@ -1113,7 +1113,7 @@ class AsyncTestClient(httpx.AsyncClient):
             extensions=extensions,
         )
 
-    def options(  # type: ignore[override]
+    async def options(  # type: ignore[override]
         self,
         url: httpx._types.URLTypes,
         *,
@@ -1125,7 +1125,7 @@ class AsyncTestClient(httpx.AsyncClient):
         timeout: httpx._types.TimeoutTypes | httpx._client.UseClientDefault = httpx._client.USE_CLIENT_DEFAULT,
         extensions: dict[str, typing.Any] | None = None,
     ) -> httpx.Response:
-        return super().options(
+        return await super().options(
             url,
             params=params,
             headers=headers,
@@ -1313,7 +1313,7 @@ class AsyncTestClient(httpx.AsyncClient):
             self.stream_receive = StapledObjectStream(*receive)
             self.task_done = anyio.Event()
 
-            async def lifespan():
+            async def lifespan() -> None:
                 await self.lifespan()
                 self.task_done.set()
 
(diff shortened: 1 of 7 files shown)


Kludex:
Nice! 😁

davidbrochart:
Should all tests also be run with the `AsyncTestClient`?

davidbrochart:
Any idea about the best way to factorize code between `TestClient` and `AsyncTestClient`?

Kludex:
> Should all tests also be run with the `AsyncTestClient`?

No, that would be too huge of a PR (but also, I don't think we should do it). You just need to test the client itself.

tests/conftest.py
```diff
@@ -5,8 +5,22 @@
...
+@pytest.fixture
+def async_test_client_factory(
+    anyio_backend_name: Literal["asyncio", "trio"],
+    anyio_backend_options: dict[str, Any],
+) -> AsyncTestClientFactory:
+    # anyio_backend_name defined by:
+    # https://anyio.readthedocs.io/en/stable/testing.html#specifying-the-backends-to-run-on
+    return functools.partial(
+        AsyncTestClient,
+        backend=anyio_backend_name,
```
graingert: Why would AsyncTestClient need the backend? Surely it would sniffio it from the first async call?

starlette/testclient.py
```diff
@@ -729,3 +734,642 @@ async def receive() -> typing.Any:
...
+        raise_server_exceptions: bool = True,
+        root_path: str = "",
+        backend: typing.Literal["asyncio", "trio"] = "asyncio",
+        backend_options: dict[str, typing.Any] | None = None,
+        cookies: httpx._types.CookieTypes | None = None,
+        headers: dict[str, str] | None = None,
+        follow_redirects: bool = True,
+        client: tuple[str, int] = ("testclient", 50000),
+    ) -> None:
+        self.async_backend = _AsyncBackend(backend=backend, backend_options=backend_options or {})
```
graingert: Unused attribute?

davidbrochart: Thanks.

tests/conftest.py
davidbrochart: Right.

davidbrochart:
> No, that would be too huge of a PR (but also, I don't think we should do it).

If we can parameterize tests so that they use both `TestClient` and `AsyncTestClient`, there won't be too many changes, just the extra `await` for `AsyncTestClient`.

> You just need to test the client itself.

The thing is that we don't get enough code coverage for `AsyncTestClient`, with only `test_asynctestclient.py`:
```
Name                      Stmts   Miss  Cover   Missing
-------------------------------------------------------
starlette/testclient.py     574     54    91%   757, 790, 792-801, 807, 810, 813-817, 826-828, 841, 871-872, 895, 939-962, 982-984, 989-991, 995-996, 1006-1007, 1125, 1148, 1175, 1206, 1237, 1264, 1287, 1362, 1372
-------------------------------------------------------
TOTAL                      9911     54    99%
```
I checked that it was the same for `TestClient`, where these lines were not covered in `test_testclient.py` only:
```
Name                      Stmts   Miss  Cover   Missing
-------------------------------------------------------
starlette/testclient.py     576    284    51%   49, 73, 141, 143-152, 158, 161, 164-168, 177-179, 192, 224-225, 248, 292-315, 335-337, 343-345, 349-350, 360-361, 493, 516, 543, 574, 605, 632, 655, 726, 736...
```
The other tests bring full code coverage.

Factorizing code between `TestClient` and `AsyncTestClient` will raise code coverage and has to be done anyway, but I don't think it will be enough. Maybe we can port some tests to async to get full code coverage.

davidbrochart:
I was thinking about mixin classes that the test client and the async test client would use, but I'm not sure it's worth it. It would make the code more DRY, but not that much and at the cost of more complexity. Maybe the async test client should live in a separate `asynctestclient.py` file, and every future change to `testclient.py` should be ported to `asynctestclient.py`?
Alternatively, if the async test client is not used in starlette's tests then it should live in a different package? I could port starlette's tests there and get full code coverage. Maybe also give it a "better" name than `AsyncTestClient`, since it's really an HTTP-over-ASGI client.
What do you think?

davidbrochart:
Hmm it looks like https://github.com/vinissimus/async-asgi-testclient is just what I needed.

tests/conftest.py
```diff
@@ -5,8 +5,15 @@
...
 
-from starlette.testclient import TestClient
-from tests.types import TestClientFactory
+from starlette.testclient import AsyncTestClient, TestClient
+from tests.types import AsyncTestClientFactory, TestClientFactory
+
+
+@pytest.fixture
+def async_test_client_factory() -> AsyncTestClientFactory:
+    return functools.partial(
```
graingert, line 14: This partial is redundant (and probably the whole fixture too)

PR #2939, opened 2025-05-18
feat: implement rfc9110 http status names

mbeijen:
# Summary

rfc9110 obsoletes the earlier rfc 7231. This document also includes some status codes that were previously only used for WebDAV and assigns more generic names to these status codes.

ref: https://www.rfc-editor.org/rfc/rfc9110.html#name-changes-from-rfc-7231

The status code constants are updated to reflect the new naming and the old constants are kept for backwards compatibility

 - HTTP_413_CONTENT_TOO_LARGE, previously HTTP_413_REQUEST_ENTITY_TOO_LARGE
 - HTTP_414_URI_TOO_LONG, previously HTTP_414_REQUEST_URI_TOO_LONG
 - HTTP_416_RANGE_NOT_SATISFIABLE, previously HTTP_416_REQUESTED_RANGE_NOT_SATISFIABLE
 - HTTP_422_UNPROCESSABLE_CONTENT, previously HTTP_422_UNPROCESSABLE_ENTITY

# Checklist


starlette/status.py
```diff
@@ -72,6 +72,11 @@
 HTTP_510_NOT_EXTENDED = 510
 HTTP_511_NETWORK_AUTHENTICATION_REQUIRED = 511
 
+# for backwards compatibility
+HTTP_413_REQUEST_ENTITY_TOO_LARGE = HTTP_413_CONTENT_TOO_LARGE
+HTTP_414_REQUEST_URI_TOO_LONG = HTTP_414_URI_TOO_LONG
+HTTP_416_REQUESTED_RANGE_NOT_SATISFIABLE = HTTP_416_RANGE_NOT_SATISFIABLE
+HTTP_422_UNPROCESSABLE_ENTITY = HTTP_422_UNPROCESSABLE_CONTENT
```
Kludex: We need to use a `__getattr__` on the module to check those. Please check the history on this file, you'll find how to implement it.

Kludex (changes requested):

Kludex:
We need to do this: https://github.com/Kludex/starlette/commit/b133ab45ad649cc91634dbe770f33c722dccd78b

Kludex (approved):

mbeijen:
Thanks! And sorry i did not fix the PR up myself. Typical drive-by-committer 😉

PR #2944, opened 2025-05-27
Allow to pass any object as lifespan state

Kludex:
This is a breaking change.

This is here first for discussion.


tests/test_routing.py
```diff
@@ -757,14 +757,14 @@ class State(TypedDict):
         count: int
         items: list[int]
 
-    async def hello_world(request: Request) -> Response:
-        # modifications to the state should not leak across requests
-        assert request.state.count == 0
+    async def hello_world(request: Request[State]) -> Response:
```
Kludex: This test needs to be refactored - I just changed to show the behavior changes.

alex-oleshkevich:
It breaks probably one of the most popular patterns: `request.state.DEPNAME`
For example, a database session middleware may set current session like that 
```python
class DbMiddleware:
    def __init__(self, app):
        self.app = app

        scope["state"]['dbsession'] = "database_connection"
```
and request it in an endpoint like
```python
async def view(request: Request[LifespanState]):
    return JSONResponse(
        {
            "request_state": request.state.dbsession,
        }
    )
```
Now it throws `AttributeError: 'State' object has no attribute 'dbsession'`.
IMO, this breaks too many third-party integrations.

alex-oleshkevich:
What if we bind `LifespanStateT` to `State` to require all child states to inherit from State? This way we can maintain b/w compatibility and users can provide types?
```python
_LifespanStateT = TypeVar("_LifespanStateT", default=State, bind=State)

class MyState(State):
    key: str
    dbsession: Any

class DbMiddleware:
        scope["state"]['dbsession'] = "database_connection"

async def lifespan_state():
    yield MyState(key='value')

async def view(request: Request[MyState]):
    print(request.state.dbsession) # works
    print(request.state.key) # also works
```

Kludex:
@alex-oleshkevich Would there be a way to move away from the `State` class? But yeah, I agree that what you are proposing is better for the ecosystem.

alex-oleshkevich:
> @alex-oleshkevich Would there be a way to move away from the `State` class? But yeah, I agree that what you are proposing is better for the ecosystem.

`State` can be a protocol, but this is not very convenient. In short, we only need a common interface to set/get values from the state.

Kludex:
I've applied your comment, but this doesn't provide the DX that I'm aiming. State is initialized with a `dict`, which is not type safe.

abersheeran:
Due to the limitations of Python's type system, I don't think there will be an elegant way to implement this feature. For now, using TypedDict/dataclass is a good compromise design.

Kludex:
I agree @abersheeran , let me try implementing it focusing on `TypedDict`.

PR #2949, opened 2025-06-02
Allow async exception handlers to type-check

jonathanberthias:
# Summary

There has been a lot of discussion around the exception handlers' type annotations, e.g., #2403 and #2048, but there is a simple case that isn't currently handled and can be fixed easily.

When building the `ExceptionMiddleware` directly, only sync handlers are accepted, even though the code can perfectly well handle async handlers. I think this was not noticed before because not many users instantiate this middleware directly, and when they do, the errors linked to the exceptions' variance always led to adding `# type: ignore`s, missing this simple error.

The fix is just to reuse the existing `ExceptionHandler` type alias.

I wasn't sure how to add tests for these changes. I added some code in `test_exceptions.py` that is just used to check there is not type error, let me know if there is another preferred approach.

While using the `ExceptionHandler` alias in a few places, it unearthed another error that can be solved by using [`TypeIs`](https://typing.python.org/en/latest/spec/narrowing.html#typeis) instead of [`TypeGuard`](https://typing.python.org/en/latest/spec/narrowing.html#typeguard) in `is_async_callable`. This fixes a few false positives from MyPy, though is not directly related to the initial change in this PR so I totally understand if you prefer I leave it out.

# Checklist

### DIFF
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -48,14 +48,6 @@ async def handler_that_reads_body(request: Request, exc: BadBodyException) -> JS
     return JSONResponse(status_code=422, content={"body": body.decode()})
 
 
-async def async_catch_all_handler(request: Request, exc: Exception) -> JSONResponse:
-    raise NotImplementedError
-
-
-def sync_catch_all_handler(request: Request, exc: Exception) -> JSONResponse:
-    raise NotImplementedError
-
-
 class HandledExcAfterResponse:
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         response = PlainTextResponse("OK", status_code=200)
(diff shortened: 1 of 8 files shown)


tests/test_exceptions.py
```diff
@@ -74,6 +82,9 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
     handlers={BadBodyException: handler_that_reads_body},  # type: ignore[dict-item]
 )
 
+sync_catch_all = ExceptionMiddleware(router, handlers={Exception: sync_catch_all_handler})
+async_catch_all = ExceptionMiddleware(router, handlers={Exception: async_catch_all_handler})
```
jonathanberthias: Before the changes in this PR, this line led to:
```
tests/test_exceptions.py:86: error: Dict entry 0 has incompatible type "type[Exception]": "Callable[[Request, Exception], Coroutine[Any, Any, JSONResponse]]"; expected "Any": "Callable[[Request, Exception], Response]"  [dict-item]
```

Kludex: Can we wrap this into a test? It seems pretty easy to see this code again and think this is not being used and remove it.

Kludex (approved):

Kludex:
Thanks! 🙏

PR #2953, opened 2025-06-25
Yield message is a http response start type on BaseHTTPMiddleware

jamesev15:
# Summary

I've been working with MCP servers using SSE transport methods. It turns out that the first message of a stream is of the `http.response.start` type, which raises an assertion since the current code expects an `http.response.body`. This PR adds a condition to handle that case: instead of raising an assertion, it yields the same message.

# Checklist

- [] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [] I've updated the documentation accordingly.

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -163,6 +163,10 @@ class BaseHTTPMiddleware:
 
             async def body_stream() -> BodyStreamGenerator:
                 async for message in recv_stream:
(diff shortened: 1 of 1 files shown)


Kludex:
They are probably sending 2 start messages, which is wrong.

Can you please create an issue on the MCP Python-sdk side, with an MRE? I'll try to check later.

PR #2955, opened 2025-06-25
Support returning Not Modified responses in FileResponse

paulo-raca:
# Summary

Looking at the souce for [StaticFiles](https://github.com/encode/starlette/blob/fa5355442753f794965ae1af0f87f9fec1b9a3de/starlette/staticfiles.py#L178-L180), it verifies the timestamp/etag and returns `HTTP 304 - Not Modified` when appropriate.

IMHO it would be simpler / more reusable if this was part of FileResponse.

(I brought this up a few hours ago in #2954, and decided to give it a try)

I haven't written new tests: There already is a ton of them and there is good coverage.

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -10,7 +10,7 @@ import sys
 import warnings
 from collections.abc import AsyncIterable, Awaitable, Iterable, Mapping, Sequence
 from datetime import datetime
-from email.utils import format_datetime, formatdate
+from email.utils import format_datetime, formatdate, parsedate
 from functools import partial
 from mimetypes import guess_type
 from secrets import token_hex
@@ -297,6 +297,15 @@ _RANGE_PATTERN = re.compile(r"(\d*)-(\d*)")
 class FileResponse(Response):
     chunk_size = 64 * 1024
 
+    NOT_MODIFIED_HEADERS = {
+        b"cache-control",
+        b"content-location",
+        b"date",
+        b"etag",
+        b"expires",
+        b"vary",
+    }
+
     def __init__(
         self,
         path: str | os.PathLike[str],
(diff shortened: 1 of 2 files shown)


Kludex:
@abersheeran I know you've played a lot with this. Does this PR make sense?

starlette/responses.py
```diff
@@ -452,8 +470,30 @@ async def _handle_multiple_ranges(
...
 
-    def _should_use_range(self, http_if_range: str) -> bool:
-        return http_if_range == self.headers["last-modified"] or http_if_range == self.headers["etag"]
+    def _is_not_modified(self, http_if_none_match: str | None, http_if_modified_since: str | None) -> bool:
+        """
+        Given the request and response headers, return `True` if an HTTP
+        "Not Modified" response could be returned instead.
+        """
+        if http_if_none_match is not None:
+            match = [tag.strip(" W/") for tag in http_if_none_match.split(",")]
```
abersheeran, line 479: There are some issues here; the strong and weak semantics of etag should not be handled this way.

abersheeran (changes requested):

Smixi-syn:
Hello, is this related to https://github.com/encode/starlette/pull/2891 ?

starlette/responses.py
Kludex: What's wrong with this?

abersheeran: > W/ (case-sensitive) indicates that a [weak validator](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Conditional_requests#weak_validation) is used. Weak ETags are easy to generate, but are far less useful for comparisons. Strong validators are ideal for comparisons but can be very difficult to generate efficiently. Weak ETag values of two representations of the same resources might be semantically equivalent, but not byte-for-byte identical. This means weak ETags prevent caching when [byte range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Ranges) are used, but strong ETags mean range requests can still be cached.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag

When requesting a range, different handling is required. However, this difference has been ignored here.

PR #2962, opened 2025-07-10
Make UploadFile check for future rollover

HonakerM:
# Summary

This PR ensures that during multi-part form parsing any slow disk IO happens in a background. This specifically fixes a case where a file would rollover during a `write()` call as talked about in [this discussion.](https://github.com/encode/starlette/discussions/2927#discussioncomment-13721403)

# Checklist

### DIFF
--- a/tests/test_formparsers.py
+++ b/tests/test_formparsers.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
 import os
+import threading
 from contextlib import AbstractContextManager, nullcontext as does_not_raise
 from io import BytesIO
 from pathlib import Path
 from tempfile import SpooledTemporaryFile
-import threading
 from typing import Any
 from unittest import mock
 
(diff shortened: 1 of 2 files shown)


tests/test_formparsers.py
```diff
@@ -303,6 +304,16 @@ def test_multipart_request_mixed_files_and_data(tmpdir: Path, test_client_factor
...
 
 
+def test_multipart_request_large_file(tmpdir: Path, test_client_factory: TestClientFactory) -> None:
+    data = BytesIO(b" " * MultiPartParser.spool_max_size * 2)
+    client = test_client_factory(app)
+    response = client.post(
+        "/",
+        files=[("test_large", data)],
+    )
+    assert response.status_code == 200
```
Kludex: This test passes on master. We want some condition that proves this PR actually solves the issue.

defnull: This type of issue is hard to verify without profiling or adding breakpoints and checking if a write operation that caused a rollover happened on the main thread. I do not know if it is feasable or even possible to provide a unit test that fails if the blocking rollover happens on the wrong thread, that would require assert statements and additional logic in the actual code.

Kludex: Then we can remove the test, can't we?

HonakerM: @Kludex yeah in it's current state it's not doing a whole lot. Let met try editing the test to use a monkey patched SpooledTemporaryFile to ensure the IO thread is separate from the event thread

defnull: It does not ensure that there is no blocking call on the event loop, but it does cover the new code paths that would otherwise not be covered.

HonakerM: @Kludex Just pushed the new test! I ended up creating a new class/app to track where rollover was happening. I'm still new to starlette development so please let me know if you have any code-quality suggestions.

defnull:
Not for this PR, but at this point I would just move the rollover logic into `UploadFile` and not use `SpooledTemporaryFile` at all. That would also save some expensive `getattr()` calls in the hot path. Maybe an idea for a future PR.

HonakerM:
@defnull @Kludex mentioned this elsewhere but [PR2925](https://github.com/encode/starlette/pull/2925) is probably the long term solution here. It moves all this logic to anyio and gets it out of the hot path

HonakerM:
@Kludex  sorry for the ping/re-review request but I'm wondering if we could get this merged soon? It should be ready to go. Please let me know if you have any more comments/suggestions/requirements

Kludex:
I'll release it today.

Kludex (approved):

HonakerM:
@Kludex since you published the security bulletin I'm no longer able to comment so I'm going to follow-up here:

1. wai25 is one of my co-workers who brought the issue to my attention and helped me narrow down the problem to starlette.
2. Could we request a low CVE for this issue? It seems the other advisories had them created

PR #2976, opened 2025-08-05
Do not pollute exception context in Middleware

adam-sikora:
# Summary

Avoid passing `anyio.EndOfStream` as exception context when reraising the exception in `BaseHTTPMiddleware`. Instead try to preserve the original context.

# Checklist


Kludex:
I think we'd want to merge https://github.com/encode/starlette/pull/2830 instead. Can you try that PR?

starlette/middleware/base.py
```diff
@@ -156,7 +156,16 @@ async def coro() -> None:
...
+                    # Prevent `anyio.EndOfStream` from polluting app exception context.
+                    # If both cause and context are None then the context is suppressed
+                    # and `anyio.EndOfStream` is not present in the exception traceback.
+                    # If exception cause is not None then it is propagated with
+                    # reraising here.
+                    # If exception has no cause but has context set then the context is
+                    # propagated as a cause with the reraise. This is necessary in order
+                    # to prevent `anyio.EndOfStream` from polluting the exception
+                    # context.
+                    raise app_exc from app_exc.__cause__ or app_exc.__context__
```
adriangb, line 168: I guess what happens if the exception has no cause or context? Then does `anyio.EndOfStream` become the new cause / context?

adam-sikora:
> I think we'd want to merge #2830 instead. Can you try that PR?

I can not test it directly at the moment but I believe that the mentioned PR does not change the exception behavior for the part that was changed in this PR. Furthermore the passing of the context there was is still just a suggestion in one comment.

starlette/middleware/base.py
adam-sikora: If neither the context nor cause are present then this is equivalent to `raise app_exc from None` which suppresses the exception context and just shows the raised exception. This means that `anyio.EndOfStream` does not show up in the traceback.

Should I expand the comment to make this more clear?

This is a desired solution for me since when the `anyio.EndOfStream` was propagated in traceback as the innermost exception (context of app_exc) - it caused our Error Reporting (tool from Google cloud that aggregates encountered errors from logs) to group all our API errors into single group as the innermost exception was always the same.

adam-sikora:
> > I think we'd want to merge #2830 instead. Can you try that PR?
> 
> I can not test it directly at the moment but I believe that the mentioned PR does not change the exception behavior for the part that was changed in this PR. Furthermore the passing of the context there was is still just a suggestion in one comment.

I managed to test the #2830 PR even with the last suggestion there to add context/cause to the exception and I can confirm that it does not cover the issue solved in this PR

adam-sikora:
@Kludex @adriangb can we please move forward with this PR? As stated above the #2830 in its current state does not fix the issue that this PR addresses.

Kludex (approved):
Makes sense. Great PR.

Sorry the delay!

PR #2981, opened 2025-08-19
Option to allow private network access in CORS middleware.

alxfed:
# Summary

The versions of Chrome browser after 130 require an additional header "Access-Control-Allow-Private-Network": "true" in PREFLIGHT interaction if you are trying to access f.i. a https://localhost resources, where uvicorn / starlette application is running. They promised to start blocking these requests altogether, but didn't, it's just a warning, but a very annoying one.
This is important for the current generation of AI applications if a local server with a Starlette application (not the MCP with FastApi) is used for interactions with a web-based AI application.

The allow_private_network parameter has been added to cors.py and the doc.

# Checklist

- [ *] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [ *] I've updated the documentation accordingly.


Kludex:
This PR doesn't have tests, docs, and the implementation doesn't check the `"access-control-request-private-network"` header.

For those reasons, I'll be closing this PR.

PR #2984, opened 2025-08-21
Use `asyncio.iscoroutinefunction` for Python 3.12 and older

mjpieters:
# Summary

inpsect.iscoroutinefunction still has issues with mocked coroutines, for example, until Python 3.13.

Resolves https://github.com/encode/starlette/discussions/2983

# Checklist

### DIFF
--- a/starlette/_utils.py
+++ b/starlette/_utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import functools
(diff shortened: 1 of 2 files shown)


Kludex (approved):

Kludex:
Thanks @mjpieters 🙏

PR #2986, opened 2025-08-25
Add `digest_method` to `SessionMiddleware`

bcarroll:
# Summary

itsdangerous includes a patch for this issue, but the TimestampSigner used in SessionMiddleware does not allow the digest_method to be changed from the itsdangerous default, which is SHA1, and causes SessionMiddleware to generate an exception in a ubi8/ubi9 RedHat container.  This commit allows the Starlette Session Middleware to support FIPS builds without SHA-1 by adding a digest_method parameter to the constructor.

# Checklist


starlette/middleware/sessions.py
```diff
@@ -23,9 +23,11 @@ def __init__(
         same_site: Literal["lax", "strict", "none"] = "lax",
         https_only: bool = False,
         domain: str | None = None,
+        digest_method: Any | None = None,
```
Kludex, line 27: What's the right type for this `digest_method`?

bcarroll: I am not sure what the correct type is, I just borrowed the `Any` type from the itsdangerous code.
https://github.com/pallets/itsdangerous/blob/672971d66a2ef9f85151e53283113f33d642dabd/src/itsdangerous/signer.py#L60

jimmyhedstr0m:
What is the status of this one? To be able to add digest method is highly appreciated.

PR #2989, opened 2025-08-30
Add json_encoder argument to JSONResponse class

gregorio-clara:
# Summary

# Checklist

I opened the discussion of this idea [here](https://github.com/encode/starlette/discussions/2990).

### DIFF
--- a/docs/responses.md
+++ b/docs/responses.md
@@ -95,8 +95,26 @@ async def app(scope, receive, send):
 
 #### Custom JSON serialization
 
-If you need fine-grained control over JSON serialization, you can subclass
-`JSONResponse` and override the `render` method.
+If you need fine-grained control over JSON serialization, you can pass a custom `json.JSONEncoder` to the `json_encoder` argument.
+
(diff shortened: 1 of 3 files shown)


Kludex:
Please search for similar issues/discussions/PRs before opening new ones.

Regarding this PR, as a user you should subclass the JSONResponse to handle this case.

gregorio-clara:
@Kludex I understand that currently we are supposed to subclass the JSONResponse class in these cases. However, I don't think it makes sense when not using a third party JSON library. The standard JSON library already uses 1 single argument (cls in dumps) for that, why not adding the functionality to just pass it as well?

I did read the discussions, the last one regarding JSON serialization was opened 6 years ago https://github.com/encode/starlette/discussions/1517 and last replied 2 years ago (actually proposing exactly the addition of the cls argument)

Moreover, I opened a discussion of my own https://github.com/encode/starlette/discussions/2990 along this PR so that we can have a conversation about it :)

PR #2996, opened 2025-09-04
Set encodings when reading config files

secrett2633:
# Summary
Fixes UnicodeDecodeError by allowing explicit encoding selection when reading .env files.

Resolves https://github.com/encode/starlette/discussions/2016

## Description
Modified the _read_file method in starlette/config.py to accept an encoding parameter.
This ensures .env files can be correctly decoded regardless of the system's default locale encoding.

**Key changes:**
- Added encoding parameter (default: "utf-8") to _read_file
- Passed encoding explicitly to open() when reading .env files
- Retained existing parsing logic (key/value handling unchanged)

## Motivation
The current implementation uses open(file_name) without specifying an encoding, which defaults to the system’s locale (e.g., GBK on Windows in Chinese locales).
This leads to UnicodeDecodeError when the .env file is encoded in UTF-8 (a common format).

## Expected Outcome
After this fix:
- ✅ UTF-8 .env files can be read on Windows systems using GBK/CP949 locales
- ✅ Developers have flexibility to override encoding if needed
- ✅ Backward compatibility preserved (utf-8 is default)
- ✅ No additional runtime cost introduced

# Checklist


Kludex:
Hi @secrett2633 ,

Thanks for the sponsorship! :)

Kludex (commented):
This doesn't seem all that is needed.

I guess we need to pass `encoding` from the `Config` class, otherwise this can't ever be different.

Although, I'm inclined to suggest users to use `PYTHONUTF8=1` on their environment.

secrett2633:
Thanks! Happy to help. 😊 @Kludex 

You're right that without passing encoding from the `Config` class it wouldn't be configurable.

I've adjusted the implementation based on that feedback.

starlette/config.py
```diff
@@ -107,9 +108,9 @@ def get(
             return self._perform_cast(key, default, cast)
         raise KeyError(f"Config '{key}' is missing, and has no default.")
 
-    def _read_file(self, file_name: str | Path) -> dict[str, str]:
+    def _read_file(self, file_name: str | Path, encoding: str = "utf-8") -> dict[str, str]:
```
Kludex:
```suggestion
    def _read_file(self, file_name: str | Path, encoding: str) -> dict[str, str]:
```

Kludex (approved):

PR #2998, opened 2025-09-04
Update all links from encode/starlette to Kludex/starlette

tarsil:
@Kludex from the message of this morning, this can also be updated, I believe or is it not in the plans?


Kludex (approved):

Kludex:
Thanks! I've updated all references.

PR #2999, opened 2025-09-06
Use uv instead of plain pip

Kludex:


pyproject.toml
```diff
@@ -39,6 +39,30 @@ full = [
...
+    "starlette[full]",
+    "coverage==7.8.2",
+    "importlib-metadata==8.7.0",
+    "mypy==1.16.1",
+    "ruff==0.12.1",
+    "types-PyYAML==6.0.12.20250516",
+    "pytest==8.4.1",
+    "trio==0.30.0",
+    # Check dist
+    "twine==6.1.0"
```
Kludex, line 54: Since we have a lockfile, we can just restrict the minimum here. I'll do this after.

zanieb:
Please poke me if you have any problems!

Kludex:
> Please poke me if you have any problems!

Sure! Thanks :)

pyproject.toml
```diff
@@ -39,6 +39,30 @@ full = [
     "httpx>=0.27.0,<0.29.0",
 ]
 
+[dependency-groups]
+dev = [
+    # We add starlette[full] so `uv sync` considers the extras.
+    "starlette[full]",
```
Kludex, line 45: @zanieb I had to do this - which is actually the first time I do it...

This is to avoid `uv sync --all-extras` (I can just do `uv sync`). Is there a way to add it to `default-groups` instead?

zanieb: I think you want https://github.com/astral-sh/uv/pull/12965, which we aren't sure what to do with.

This is an appropriate workaround.

Kludex: Funny. Copilot actually suggested me to use the `default-extras` 😅 

All good then. Thanks!

Issue #2977, opened 2025-08-05
Request Header cookie parsing doesn't follow RFC 7540 spec


brian316:
Only the first cookie in a header gets parsed. but multiple "cookie" headers may be sent and the server should concatenate them per [RFC 7540](https://datatracker.ietf.org/doc/html/rfc7540)

```
...
   If there are multiple Cookie header fields after
   decompression, these MUST be concatenated into a single octet string
   using the two-octet delimiter of 0x3B, 0x20 (the ASCII string "; ")
   before being passed into a non-HTTP/2 context, such as an HTTP/1.1
   connection, or a generic HTTP server application.

   Therefore, the following two lists of Cookie header fields are
   semantically equivalent.

     cookie: a=b; c=d; e=f

     cookie: a=b
     cookie: c=d
     cookie: e=f
```

The issue lies here
https://github.com/encode/starlette/blob/6ee94f2cac955eeae68d2898a8dec8cf17b48736/starlette/requests.py#L143

---

- [ ] Initially raised as discussion #2916

Kludex:
It seems we need to use `getlist` instead there. Weird that I didn't see this before.

It seems the code we have was inspired by Django 3.1.0, so maybe we can check if Django changed something.

Pravindgholap:
Hi, I’d like to work on this issue.

From my understanding, the fix would involve updating


`cookie_header = self.headers.get("cookie")`
to use self.headers.getlist("cookie"), and then concatenating the values with "; " in accordance with RFC 7540 §8.1.2.5.

I can also add a test case to cover scenarios where multiple Cookie headers are sent, ensuring the concatenated form matches expectations.

Please let me know if this approach sounds good before I proceed.

PR #3002, opened 2025-09-10
fix: #2977 -- Join multiple Cookie headers per RFC 7540.

11kkw:
# Summary
This PR updates `Request.cookies` to handle multiple `Cookie` headers correctly in compliance with RFC 7540 §8.1.2.5.

- Uses `headers.getlist("cookie")` instead of `get()`
- Strips trailing semicolons from individual values
- Joins values with `"; "` before parsing

This approach was inspired by Django’s fix for [ticket-36399](https://code.djangoproject.com/ticket/36399).  
Although another contributor expressed interest in working on [#2977](https://github.com/encode/starlette/issues/2977), no PR has been opened for over a month, so I’ve submitted this patch to move the issue forward

# Checklist

- [] I've updated the documentation accordingly.


starlette/requests.py
```diff
@@ -140,10 +140,12 @@ def path_params(self) -> dict[str, Any]:
     def cookies(self) -> dict[str, str]:
         if not hasattr(self, "_cookies"):
             cookies: dict[str, str] = {}
-            cookie_header = self.headers.get("cookie")
+            cookie_headers = self.headers.getlist("cookie")
+
+            if cookie_headers:
+                joined = "; ".join(value.rstrip("; ") for value in cookie_headers if value)
+                cookies = cookie_parser(joined)
```
Kludex: Instead of joining, to separate later, can we change the `cookie_parser` function to accept a list?

PR #3016, opened 2025-09-13
ci: use uv run on docs deploy

Kludex:


scripts/publish
Kludex, line 1: This file is not needed. We only publish via pipeline.

TheWesDias: please, remove the reference here
https://github.com/Kludex/starlette/blob/f2e3ab426d70f5ebe2a669808e4ce3f187cc2225/scripts/README.md?plain=1#L9

PR #3027, opened 2025-09-20
Use `Literal` type for `WebSocketEndpoint` encoding values

gmos2104:
# Summary

Use [Literal](https://docs.python.org/3/library/typing.html#typing.Literal) type to better type the accepted values for encoding in WebSocketEndpoint

The [documentation](https://www.starlette.dev/endpoints/) already states that it only accepts 'text', 'bytes' or 'json' so this only enforces it at typing level

# Checklist

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 from collections.abc import Generator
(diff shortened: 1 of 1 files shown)


Kludex (approved):

Kludex:
Thanks!

PR #3029, opened 2025-09-20
Support multiple cookie headers in `Request.cookies`

### LINKED ISSUE #2977: Request Header cookie parsing doesn't follow RFC 7540 spec

Kludex:
- Closes https://github.com/Kludex/starlette/pull/3002
- Closes https://github.com/Kludex/starlette/issues/2977

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -140,10 +140,11 @@ class HTTPConnection(Mapping[str, Any]):
     def cookies(self) -> dict[str, str]:
         if not hasattr(self, "_cookies"):
             cookies: dict[str, str] = {}
-            cookie_header = self.headers.get("cookie")
+            cookie_headers = self.headers.getlist("cookie")
 
-            if cookie_header:
-                cookies = cookie_parser(cookie_header)
+            if cookie_headers:
+                for header in cookie_headers:
+                    cookies.update(cookie_parser(header))
             self._cookies = cookies
         return self._cookies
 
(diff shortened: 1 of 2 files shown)


Kludex:
@11kkw Can you confirm my PR solves the issue?

---

I've added you as co-author on the commit here.

11kkw:
> @11kkw Can you confirm my PR solves the issue?
> 
> I've added you as co-author on the commit here.

The if key or val: part already handles cases with a trailing semicolon (;) without any extra changes. I completely missed that. Really elegant solution, thanks! Learned a lot from this.

PR #3036, opened 2025-10-09
Turn `State` into a `Mapping`

### LINKED ISSUE #3005: Make request.state generic
### Discussed in https://github.com/Kludex/starlette/discussions/2562

<sup>Originally posted by **patrick91** April  3, 2024</sup>
Hi there! 

I was playing around with `request.state`, and I was wondering if we could make it generic, so that users can get better autocomplete on it (and maybe a bit of type safety too).

The way this would work is where you accept a `Request` you could also do `Request[MyStateType]`, and `request.state` would be of type `MyStateType`

I think we can implement this in a backwards compatible way if we use `TypeVar` from typing extensions with a default value of any, like this:

```python
from typing import Any
from typing_extensions import TypeVar

RequestState = TypeVar("RequestState", default=Any)  # this could also be `State`
```

This means that any current code wouldn't throw an error, and power users could type the state var 😊

The only issue I see at the moment is that `State` is a wrapper on a dict, but doesn't provide `getitem`, and my assumption is that types for State would be `TypedDicts` 😊 but is should be easy to add support for that. But I'm also saying this without knowing the reason why the state is a class on top of a dict, so I might be wrong :D </div>

Kludex:
- Closes https://github.com/Kludex/starlette/issues/3005


starlette/requests.py
```diff
@@ -197,7 +201,7 @@ async def empty_send(message: Message) -> NoReturn:
     raise RuntimeError("Send channel has not been made available")
 
 
-class Request(HTTPConnection):
+class Request(HTTPConnection[StateT]):
```
Kludex, line 204: We are missing the same in `WebSockets`.

.github/workflows/main.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
Kludex, line 17: Just so the pipeline can pass, but we shouldn't merge this here.

Kludex:
@alex-oleshkevich @abersheeran Can you check this, please?

alex-oleshkevich:
> @alex-oleshkevich @abersheeran Can you check this, please?

All good.

Kludex:
> > @alex-oleshkevich @abersheeran Can you check this, please?
> 
> All good.

In the sense that this implementation makes sense, or something different? 🤔

alex-oleshkevich:
> > > @alex-oleshkevich @abersheeran Can you check this, please?
> > 
> > 
> > All good.
> 
> In the sense that this implementation makes sense, or something different? 🤔

We want to have a typed request state and keep backward compatibility. That was achieved, so all is good :)

patrick91:
This was exactly what I had in mind :D thanks @Kludex!

starlette/requests.py
```diff
@@ -68,7 +69,10 @@ class ClientDisconnect(Exception):
     pass
 
 
-class HTTPConnection(Mapping[str, Any]):
+StateT = TypeVar("StateT", bound=Mapping[str, Any] | State, default=State)
```
abersheeran, line 72: I'm not sure if this will pass the TypedDict check in some static type checkers.

Kludex: Which one you don't think it will not? We test with mypy here, but pyright doesn't seem to be failing (I use it on my environment).

abersheeran: It's great to be able to pass the inspection.

PerumallaGiridhar:
Hi there! 👋
I’ve added tests to help increase coverage around typed `request.state` in #3049. 
This is my first contribution to Starlette (and to open source in general)
I’d love to keep contributing in this area to help unblock and accelerate improvements.

adriangb (commented):
Can you add some tests? Some example usage in docstrings? Updated docs that show what users can achieve now, why the new patterns are better, etc.?

Kludex:
> Can you add some tests? Some example usage in docstrings? Updated docs that show what users can achieve now, why the new patterns are better, etc.?

I've added documentation, and tests, but I believe most of background for this is in the issue that motivated this PR: https://github.com/Kludex/starlette/issues/3005

adriangb (approved):

PR #3040, opened 2025-10-15
refactor: make GZipMiddleware excluded content types configurable

shreyas-dev:
# Summary
This PR refactors the `GZipMiddleware` to make excluded content types configurable via a constructor parameter instead of using a hardcoded constant.

# Checklist


Kludex:
What is the issue you are having?

petoknm:
The `DEFAULT_EXCLUDED_CONTENT_TYPES` variable is sometimes not what users want. In my case, I want event-stream to be compressed because I'm using data-star and returning a lot of very compressible html.

A workaround I've been using for now has been to either edit the variable directly in the package in my virtual enviroment or:
```python
from fastapi.middleware.gzip import GZipMiddleware
from starlette.middleware import gzip

app = FastAPI()
app.add_middleware(GZipMiddleware, minimum_size=1000, compresslevel=5)
gzip.DEFAULT_EXCLUDED_CONTENT_TYPES = ()
```
which is not pretty IMO... would like this to be configurable the same way `minimum_size` and `compresslevel` are.

Kludex:
> The `DEFAULT_EXCLUDED_CONTENT_TYPES` variable is sometimes not what users want. In my case, I want event-stream to be compressed because I'm using data-star and returning a lot of very compressible html.
> 
> A workaround I've been using for now has been to either edit the variable directly in the package in my virtual enviroment or:
> 
> ```python
> from fastapi.middleware.gzip import GZipMiddleware
> from starlette.middleware import gzip
> 
> app = FastAPI()
> app.add_middleware(GZipMiddleware, minimum_size=1000, compresslevel=5)
> gzip.DEFAULT_EXCLUDED_CONTENT_TYPES = ()
> ```
> 
> which is not pretty IMO... would like this to be configurable the same way `minimum_size` and `compresslevel` are.

Your workaround still waits for the whole content to be sent before compressing, I don't think that's what you want anyway.

petoknm:
Yes, you're correct. Currently it needs to wait for the whole response to compress it but in my case that happened to work out because I was just sending one big SSE response (which I admit is not what most SSE usage looks like).

But besides my concrete need being a bit unusual, the general idea of making this configurable sounds pretty valid to me.

shreyas-dev:
@Kludex The issue is excluded_content_types is hardcoded and there may be instances where the user doesn't want GZipMiddleware for particular content types especially for video and audio types. I also made it configurable so that user can add or remove.

dolfinus:
Superseded by #3418 and #3421, can be closed

Issue #3048, opened 2025-10-29
Can not use `send_denial_response` with `StreamingResponse`


Kludex:
### Discussed in https://github.com/Kludex/starlette/discussions/2566

<div type='discussions-op-text'>

<sup>Originally posted by **WSH032** April  6, 2024</sup>
## Describe the bug

If we use `FileResponse` or `StreamingResponse` when sending the Websocket Denial Response, a `RuntimeError` will occur.

## To Reproduce

```py
from fastapi import FastAPI, WebSocket
from fastapi.responses import StreamingResponse
from uvicorn import run

app = FastAPI()


@app.websocket("/")
async def _(ws: WebSocket):
    async def foo():
        yield "hello"
        yield "world"

    resp = StreamingResponse(content=foo())
    await ws.send_denial_response(resp)


run(app)
```

1. Launch the server
2. Try to establish the websocket connection
3. You will see following error

```console
  |   File "e:\GitHub\fastapi-proxy\Untitled-8.py", line 16, in _
  |     await ws.send_denial_response(resp)
  |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\starlette\websockets.py", line 209, in send_denial_response        
  |     await response(self.scope, self.receive, self.send)
  |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\starlette\responses.py", line 258, in __call__
  |     async with anyio.create_task_group() as task_group:
  |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\anyio\_backends\_asyncio.py", line 678, in __aexit__
  |     raise BaseExceptionGroup(
  | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\starlette\responses.py", line 261, in wrap
    |     await func()
    |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\starlette\responses.py", line 243, in stream_response
    |     await send(
    |   File "E:\GitHub\fastapi-proxy\.venv\lib\site-packages\starlette\websockets.py", line 75, in send
    |     raise RuntimeError(
    | RuntimeError: Expected ASGI message "websocket.accept","websocket.close" or "websocket.http.response.start",but got 'http.response.start'
    +------------------------------------
```

## Expected behavior

[...]

PR #3051, opened 2025-10-29
fix: add websocket prefix to StreamingResponse and FileResponse denia…

martin-joshy:
…l (#3048)

Hi, this is my first time contribution to open source ever, welcome to all kind feedbacks 

# Summary

`StreamingResponse` and `FileResponse` sent raw `http.response.*` messages during `send_denial_response`, but WebSocket context requires `websocket.http.response.*` prefix.

This caused:
```python
RuntimeError: Expected ASGI message "websocket.accept","websocket.close" or "websocket.http.response.start",but got 'http.response.start'

# Checklist

- [x] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!) → Discussion already exists: #3048
- [x] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change → Added
- [x] I've updated the documentation accordingly. → No doc changes needed — behavior is now correct as intended

### CODE REVIEW

**Comment by Kludex** [comment]
I think what we want is a `send` wrapper that will wrap the current `send`, and check if it's a websocket type.

```py
async def _send_wrap(send: Send, is_websocket_denial: bool):
    
    async def wrapped(message: Message):
        if is_websocket_denial:
            message["type"] = "websocket." + message["type"]
        await send(message)
        
    return wrapped
```

---

The idea is to avoid the additional parameter everywhere.

martin-joshy:
Hi, thanks for the snippet it made it lot clearer. Below is the code which I tried to run test with, but I am not get 100 % coverage. Not really sure what I should be doing.

```python
    async def send_denial_response(self, response: Response) -> None:
        if "websocket.http.response" in self.scope.get("extensions", {}):
            wrapped_send = self._send_wrap(self.send)
            await response(self.scope, self.receive, wrapped_send)
        else:
            raise RuntimeError("The server doesn't support the Websocket Denial Response extension.")

    @staticmethod
    def _send_wrap(send: Send) -> Send:
        async def wrapped(message: Message) -> None:
            message_type = message["type"]
            if message_type.startswith("http."):
                message["type"] = "websocket." + message_type

            await send(message)

        return wrapped
```

PR #3054, opened 2025-10-31
specify typing_extensions in the dev dependency group since it's needed by tests

dutyrok:
# Summary
The typing_extensions module is imported within the tests/test_config.py file.

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dev = [
     "types-PyYAML==6.0.12.20250516",
     "pytest==8.4.1",
     "trio==0.30.0",
+    "typing_extensions>=4.10.0",
     # Check dist
     "twine==6.1.0"
 ]


Kludex:
How is the pipeline passing tho? 🤔

Kludex:
It's a dependency of `anyio`, so we install the right version.

I don't think adding this is necessary, but thanks for the PR. 🙏

dutyrok:
Thank you for you answer.

I think that relying on another package is not a good idea. In my experience as an ALT Linux package maintainer, I've had issues when the packages as Anyio or others stopped bringing some of their dependencies with them during installation. I had to fix and rebuild them for the Sisyphus repository. The same thing happened with your Starlette, because typing_extensions will be needed for anyio if the Python version is below 3.13, but your tests import it regardless of the Python version.

Kludex:
> Thank you for you answer.
> 
> I think that relying on another package is not a good idea. In my experience as an ALT Linux package maintainer, I've had issues when the packages as Anyio or others stopped bringing some of their dependencies with them during installation. I had to fix and rebuild them for the Sisyphus repository. The same thing happened with your Starlette, because typing_extensions will be needed for anyio if the Python version is below 3.13, but your tests import it regardless of the Python version.

I'm relying on the test suite. The installed packages in all versions already include `typing_extensions`, and if pipeline fails, we should add it.

PR #3057, opened 2025-11-01
Ignore `if-modified-since` if `if-none-match` is present

### LINKED ISSUE #3052: staticfile caching breaks if mtimes are not available even though etags differ.
I have spent the day debugging into a caching problem that is already described in #2891 .

This problem might not appear so bad, because usually when you change the file, the mtime of the file changes as well. I am, however currently in an environment where the mtime of each static file is always set to unix epoch zero and i have no influence on changing that.

This leads to the following behavior:

1. check wheter if-none-match equals etag
2. get resulting difference, falsely leading to the if-modified-since check instead of returning false
3. comparing if-modified-since (which in my case is unix epoch zero) which the file's mtime (which is always unix epoch zero)
4. falsely determining that the files must be the same thus falsely returning 304

If this was implemented according to RFC7232, this would work without a problem even for my weird case.

Please consider merging #2891 . The problem is already solved.

Kludex:
- Closes https://github.com/Kludex/starlette/pull/2891
- Closes https://github.com/Kludex/starlette/issues/3052
- Closes https://github.com/Kludex/starlette/pull/3053

I've added both @Smixi-syn and @grindhold as co-authors. Thanks for the help!

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -201,13 +201,9 @@ class StaticFiles:
         Given the request and response headers, return `True` if an HTTP
         "Not Modified" response could be returned instead.
         """
-        try:
-            if_none_match = request_headers["if-none-match"]
+        if if_none_match := request_headers.get("if-none-match"):
             etag = response_headers["etag"]
-            if etag in [tag.strip(" W/") for tag in if_none_match.split(",")]:
-                return True
-        except KeyError:
-            pass
+            return etag in [tag.strip(" W/") for tag in if_none_match.split(",")]
 
         try:
             if_modified_since = parsedate(request_headers["if-modified-since"])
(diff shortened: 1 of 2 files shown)


grindhold:
awesome! thanks for taking action so quickly :)

Smixi-syn:
Hi, thanks for the merge :) !
Just a small note, as mentionned, if people overide the FileResponse/StaticFiles and didn't set the etag in the dictionnary, then it will crash when the file is retrieved. I don't know if this can be considered a breaking change ?

Kludex:
> Hi, thanks for the merge :) ! Just a small note, as mentionned, if people overide the FileResponse/StaticFiles and didn't set the etag in the dictionnary, then it will crash when the file is retrieved. I don't know if this can be considered a breaking change ?

It's not. I've added a note there - people would be able to track the problem easily if they do something weird.

PR #3060, opened 2025-11-01
Don't rely on httpx private types

Kludex:


Kludex:
I think I need to wait to see what happens with httpx and the stable release.

PR #3065, opened 2025-11-01
Add `allow_private_network` in `CORSMiddleware`

### LINKED ISSUE #3064: CORSMiddleware with Access-Control-Allow-Private-Network header
### Discussed in https://github.com/Kludex/starlette/discussions/2509

<sup>Originally posted by **Impro02** February 14, 2024</sup>
Is it possible to add the allow_private_network parameter to the CORSMiddleware __init__ like the following example ?

````
class CORSMiddleware:
    def __init__(
        self,
        app: ASGIApp,
        allow_origins: typing.Sequence[str] = (),
        allow_methods: typing.Sequence[str] = ("GET",),
        allow_headers: typing.Sequence[str] = (),
        allow_credentials: bool = False,
        allow_private_network: bool = False,
        allow_origin_regex: typing.Optional[str] = None,
        expose_headers: typing.Sequence[str] = (),
        max_age: int = 600,
    ) -> None:
        if "*" in allow_methods:
            allow_methods = ALL_METHODS

        compiled_allow_origin_regex = None
        if allow_origin_regex is not None:
            compiled_allow_origin_regex = re.compile(allow_origin_regex)

        allow_all_origins = "*" in allow_origins
        allow_all_headers = "*" in allow_headers
        preflight_explicit_allow_origin = not allow_all_origins or allow_credentials

        simple_headers = {}
        if allow_all_origins:
            simple_headers["Access-Control-Allow-Origin"] = "*"
        if allow_credentials:
            simple_headers["Access-Control-Allow-Credentials"] = "true"
        if allow_private_network:
            simple_headers["Access-Control-Allow-Private-Network"] = "true"
        if expose_headers:
            simple_headers["Access-Control-Expose-Headers"] = ", ".join(expose_headers)

        preflight_headers = {}
        if preflight_explicit_allow_origin:
            # The origin value will be set in preflight_response() if it is allowed.
            preflight_headers["Vary"] = "Origin"
        else:
            preflight_headers["Access-Control-Allow-Origin"] = "*"
        preflight_headers.update(
            {
                "Access-Control-Allow-Methods": ", ".join(allow_methods),
                "Access-Control-Max-Age": str(max_age),
            }
        )
        allow_headers = sorted(SAFELISTED_HEADERS | set(allow_headers))
        if allow_headers and not allow_all_headers:
            preflight_headers["Access-Control-Allow-Headers"] = ", ".join(allow_headers)
        if allow_credentials:
            preflight_headers["Access-Control-Allow-Credentials"] = "true"
        if allow_private_network:
            preflight_headers["Access-Control-Allow-Private-Network"] = "true"

        self.app = app
        self.allow_origins = allow_origins
        self.allow_methods = allow_methods
        self.allow_headers = [h.lower() for h in allow_headers]
        self.allow_all_origins = allow_all_origins
        self.allow_all_headers = allow_all_headers
        self.preflight_explicit_allow_origin = preflight_explicit_allow_origin
        self.allow_origin_regex = compiled_allow_origin_regex
        self.simple_headers = simple_headers
        self.preflight_headers = preflight_headers
````</div>

### PR DESCRIPTION
**Pull request by Kludex**
- Closes https://github.com/Kludex/starlette/issues/3064

Reference used for this PR:

- https://github.com/Kludex/starlette/pull/2621/
- https://github.com/adamchainz/django-cors-headers/pull/833/
- https://developer.chrome.com/blog/private-network-access-preflight

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -76,7 +76,7 @@ The following arguments are supported:
 * `allow_methods` - A list of HTTP methods that should be allowed for cross-origin requests. Defaults to `['GET']`. You can use `['*']` to allow all standard methods.
 * `allow_headers` - A list of HTTP request headers that should be supported for cross-origin requests. Defaults to `[]`. You can use `['*']` to allow all headers. The `Accept`, `Accept-Language`, `Content-Language` and `Content-Type` headers are always allowed for CORS requests.
 * `allow_credentials` - Indicate that cookies should be supported for cross-origin requests. Defaults to `False`. Also, `allow_origins`, `allow_methods` and `allow_headers` cannot be set to `['*']` for credentials to be allowed, all of them must be explicitly specified.
-* `allow_private_network` - Indicate that private network access should be allowed for cross-origin requests. Defaults to `False`.
+* `allow_private_network` - Indicates whether to accept cross-origin requests over a private network. Defaults to `False`.
 * `expose_headers` - Indicate any response headers that should be made accessible to the browser. Defaults to `[]`.
 * `max_age` - Sets a maximum time in seconds for browsers to cache CORS responses. Defaults to `600`.
 
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -76,6 +76,7 @@ The following arguments are supported:
 * `allow_methods` - A list of HTTP methods that should be allowed for cross-origin requests. Defaults to `['GET']`. You can use `['*']` to allow all standard methods.
 * `allow_headers` - A list of HTTP request headers that should be supported for cross-origin requests. Defaults to `[]`. You can use `['*']` to allow all headers. The `Accept`, `Accept-Language`, `Content-Language` and `Content-Type` headers are always allowed for CORS requests.
 * `allow_credentials` - Indicate that cookies should be supported for cross-origin requests. Defaults to `False`. Also, `allow_origins`, `allow_methods` and `allow_headers` cannot be set to `['*']` for credentials to be allowed, all of them must be explicitly specified.
+* `allow_private_network` - Indicate that private network access should be allowed for cross-origin requests. Defaults to `False`.
 * `expose_headers` - Indicate any response headers that should be made accessible to the browser. Defaults to `[]`.
 * `max_age` - Sets a maximum time in seconds for browsers to cache CORS responses. Defaults to `600`.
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: starlette/middleware/cors.py [hunk 1]
```diff
@@ -130,6 +133,13 @@ def preflight_response(self, request_headers: Headers) -> Response:
                     failures.append("headers")
                     break
 
+        # Handle Private Network Access
```
Kludex:
```suggestion
```

Zaczero (changes requested):
I think we want to Vary: Access-Control-Request-Private-Network, just like we currently do for Origin. Conditionally added whenever allow_private_network is set.

Kludex:
> I think we want to Vary: Access-Control-Request-Private-Network, just like we currently do for Origin. Conditionally added whenever allow_private_network is set.

But then, are you saying we should also add for the `access-control-request-header` and `access-control-request-method` as well?

The Django and Flask CORS do not add it to the Vary - this doesn't mean they are right, just a data point. But can you find a reference or justify it?

Zaczero:
My logic is that Vary makes it explicit to the caches that the same request will have a different response based on the headers sent. And this is the case here. When the request header is sent, it will have different response headers, so the response varies on that.

> Including a Vary header ensures that responses are separately cached based on the headers listed in the Vary field.
*https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary*

Also found this:
https://stackoverflow.com/questions/65675056/why-add-vary-on-access-control-request-method-acrh-and-origin-when-options-meth

To my limited knowledge on this topic, Vary seems like it will make the implementation more correct (doesn't mean it's necessary). We already Vary on Origin so the intention is that we want to support caching OPTIONS responses. May as well expand Vary to cover all varying headers.

Kludex:
But then, should the other access headers be also included?

I'm having an inclination to not do it, unless someone did it in any other CORS implementation (in any framework from any language).

Zaczero:
From the stackoverflow it explicitly mentions https://github.com/rs/cors as having this behavior. I haven't searched more.

If this decision was up to me, I would ensure Vary contains all varying headers correctly. The cost is little, and when someone decides to cache OPTIONS responses the setup will be simpler and less bug-prone. It would make the app more explicit about cache handling which is a good thing.

AI research on this: https://claude.ai/public/artifacts/860f4896-69c0-4416-8066-ff9bd73c1a37

> The official specifications provide minimal guidance on Vary headers for CORS, creating a gap between spec and practice. Vary: Origin is recommended but not strictly required by specifications, yet is absolutely critical in production environments with CDNs or caching layers. Vary: Access-Control-Request-Headers appears in no specification but prevents real cache poisoning issues.

docs/middleware.md
```diff
@@ -76,6 +76,7 @@ The following arguments are supported:
 * `allow_methods` - A list of HTTP methods that should be allowed for cross-origin requests. Defaults to `['GET']`. You can use `['*']` to allow all standard methods.
 * `allow_headers` - A list of HTTP request headers that should be supported for cross-origin requests. Defaults to `[]`. You can use `['*']` to allow all headers. The `Accept`, `Accept-Language`, `Content-Language` and `Content-Type` headers are always allowed for CORS requests.
 * `allow_credentials` - Indicate that cookies should be supported for cross-origin requests. Defaults to `False`. Also, `allow_origins`, `allow_methods` and `allow_headers` cannot be set to `['*']` for credentials to be allowed, all of them must be explicitly specified.
+* `allow_private_network` - Indicate that private network access should be allowed for cross-origin requests. Defaults to `False`.
```
```suggestion
* `allow_private_network` - Indicates whether to accept cross-origin requests over a private network. Defaults to `False`.
```

Kludex:
@Zaczero I'll merge this as is, because the `"Vary"` is actually an issue with the other request headers as well. Would you like to contribute with that?

This https://github.com/rs/cors has a beautiful code. Super easy to understand. 👍

PR #3067, opened 2025-11-05
CORS middleware: Vary handling, perf

Zaczero:
# Summary

Follow up to https://github.com/Kludex/starlette/pull/3065#issuecomment-3484250919

- Reviewed and revised Vary header handling.
- Swapped Sequences for Collections that are more generic, and allow for use of sets that have faster membership checks (common in CORSMiddleware).

# Checklist

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -2,13 +2,13 @@ from __future__ import annotations
 
 import functools
 import re
-from collections.abc import Sequence
+from collections.abc import Collection
 
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.responses import PlainTextResponse, Response
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
-ALL_METHODS = ("DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT")
+ALL_METHODS = {"DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"}
 SAFELISTED_HEADERS = {"Accept", "Accept-Language", "Content-Language", "Content-Type"}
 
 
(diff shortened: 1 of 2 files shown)


starlette/middleware/cors.py
```diff
@@ -2,74 +2,71 @@
...
-        )
-        allow_headers = sorted(SAFELISTED_HEADERS | set(allow_headers))
-        if allow_headers and not allow_all_headers:
-            preflight_headers["Access-Control-Allow-Headers"] = ", ".join(allow_headers)
+        if not allow_all_headers:
+            preflight_headers["Access-Control-Allow-Headers"] = ", ".join(sorted(allow_headers))
         if allow_credentials:
             preflight_headers["Access-Control-Allow-Credentials"] = "true"
 
+        preflight_vary: list[str] = ["Access-Control-Request-Headers", "Access-Control-Request-Private-Network"]
```
Zaczero, line 55: Normally Access-Control-Request-Private-Network wouldn't be needed if not allow_private_network, but we return 400.

starlette/middleware/cors.py
```diff
@@ -123,13 +119,12 @@ def preflight_response(self, request_headers: Headers) -> Response:
...
-        # If we allow all headers, then we have to mirror back any requested
-        # headers in the response.
+        # When allow_headers is wildcard, mirror any requested headers.
         if self.allow_all_headers and requested_headers is not None:
             headers["Access-Control-Allow-Headers"] = requested_headers
         elif requested_headers is not None:
-            for header in [h.lower() for h in requested_headers.split(",")]:
-                if header.strip() not in self.allow_headers:
+            for header in requested_headers.split(","):
+                if header.strip().lower() not in self.allow_headers:
```
Zaczero, line 127: Avoided allocation of list

starlette/middleware/cors.py
```diff
@@ -2,74 +2,71 @@
...
 
 
 class CORSMiddleware:
     def __init__(
         self,
         app: ASGIApp,
-        allow_origins: Sequence[str] = (),
-        allow_methods: Sequence[str] = ("GET",),
-        allow_headers: Sequence[str] = (),
+        allow_origins: Collection[str] = {},
```
dolfinus, line 19: Mutable defaults are not safe. Please use `frozenset` instead

PR #3068, opened 2025-11-05
Remove itsdangerous dependency

Zaczero:
# Summary

I've always disliked installing itsdangerous dependency just for basic SessionMiddleware functionality. The itsdangerous has lots of features but the amount of functionality needed by starlette is minimal. I decided to prepare starlette-optimized TimestampSigner implementation that has several benefits:

- No dependency, better UX
- Simple, no-fuss implementation
- Uses SHA-256/128 HMAC instead of SHA-1 HMAC
- Smaller signed tokens
- Faster

# Checklist

---

I also prepared a small benchmark script:

[benchmark_signing.py](https://github.com/user-attachments/files/23351791/benchmark_signing.py)

`uv run --with itsdangerous python benchmark_signing.py`

```txt
Input: {'user_id': 12345, 'username': 'alice', 'roles': ['admin', 'user']}

itsdangerous:
eyJ1c2VyX2lkIjogMTIzNDUsICJ1c2VybmFtZSI6ICJhbGljZSIsICJyb2xlcyI6IFsiYWRtaW4iLCAidXNlciJdfQ==.aQroow.EK2o0veXJdxMfN724EJaqjwjsQo

starlette.signing:
eyJ1c2VyX2lkIjogMTIzNDUsICJ1c2VybmFtZSI6ICJhbGljZSIsICJyb2xlcyI6IFsiYWRtaW4iLCAidXNlciJdfQBpCuijLtC0_yYQrfToZiZsjFCBzA_

Size: 127b → 119b (6.3% smaller)

Sign:          4.26 μs → 2.58 μs (1.65x faster)
Unsign (valid): 6.61 μs → 3.98 μs (1.66x faster)
Unsign (invalid): 2.47 μs → 0.07 μs (33.60x faster)
```

### DIFF
--- a/starlette/signing.py
+++ b/starlette/signing.py
@@ -1,0 +1,102 @@
+from __future__ import annotations
+
+import base64
+import hashlib
+import hmac
+import time
+
+from starlette.datastructures import Secret
+
+
+class TimestampSigner:
+    """
+    Signs and unsigns data with HMAC-SHA256 and timestamp validation.
+
+    Format: <b64(payload+timestamp)><b64(signature)>_
+    """
+
+    def __init__(self, secret: str | Secret) -> None:
+        """
+        Initialize signer with a secret key.
+
+        Args:
+            secret: Secret key for HMAC signing
+        """
+        self._secret = str(secret).encode("utf-8")
+
+    def sign(self, data: bytes) -> bytes:
+        """
+        Sign data with current timestamp.
+
+        Args:
+            data: Raw data to sign
+
+        Returns:
+            Signed token as bytes
+        """
+        timestamp_bytes = int(time.time()).to_bytes(5, "big")
+
+        combined = data + timestamp_bytes
+        combined_encoded = _b64_encode(combined)
+
+        signature = hmac.HMAC(self._secret, combined_encoded, hashlib.sha256).digest()[:16]
+        signature_encoded = _b64_encode(signature)
+
+        return combined_encoded + (signature_encoded + b"_")
+
+    def unsign(self, signed_data: bytes, max_age: int | None = None) -> bytes | None:
+        """
+        Verify and extract data from signed token.
+
+        Args:
+            signed_data: Signed token
+            max_age: Maximum age in seconds (optional)
+
+        Returns:
+            Payload bytes on success, None on failure
+        """
+        # Quick pre-checks
+        if len(signed_data) < 30 or signed_data[-1] != 95:
+            return None
+
+        signature_encoded = signed_data[-23:-1]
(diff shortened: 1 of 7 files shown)


Zaczero:
This change will invalidate existing sessions. If it's a big enough issue, I can patch it to attempt the itsdangerous dependency import, and when successful, attempt itsdangerous unsign after starlette.signing failure. I don't know if that's worth the added complexity.

PS. The starlette.signing ending marker, in addition to speeding up pre-validation, can be used to easily support multiple token encoding versions if there's ever a need for it.

starlette/signing.py
```diff
@@ -0,0 +1,102 @@
...
+
+        combined = _b64_decode(combined_encoded)
+        if combined is None:  # pragma: no cover
+            return None
+
+        # Check timestamp age if max_age is set
+        if max_age is not None:
+            timestamp_bytes = combined[-5:]
+            timestamp = int.from_bytes(timestamp_bytes, "big")
+            if time.time() - timestamp > max_age:
```
dolfinus, line 80: `time.time()` can go backwards if OS time is changed. `time.monotonic()` is a better solution

Zaczero: agreed

starlette/signing.py
```diff
@@ -0,0 +1,102 @@
...
+        if len(signed_data) < 30 or signed_data[-1] != 95:
+            return None
+
+        signature_encoded = signed_data[-23:-1]
+        signature = _b64_decode(signature_encoded)
+        if signature is None or len(signature) != 16:
+            return None
+
+        combined_encoded = signed_data[:-23]
+        expected_signature = hmac.HMAC(self._secret, combined_encoded, hashlib.sha256).digest()[:16]
```
dolfinus, line 68: Instead of hardcoding `hashlib.sha256`, hashing algorithm can be passed via `__init__` argument. Also `SessionMiddleware` can accept `algorithm: Literal["SHA-1", "SHA-128", "SHA-256"] = "SHA-1"` to keep backward compatibility. What do you think?

Zaczero: I'm open to improving, but what's the chance of getting this merged after? It seems like Starlette is dead. I already gave up on it, honestly, but I can come back if there are some guarantees so I don't waste time :D

Kludex: I understand the frustration of not having a PR reviewed in a reasonably timely manner. I'm sorry for that.

--- 

I think a reasonable next step here would be to first replace TimestampSigner by our own implementation without changing the behavior, and then opening smaller PRs to change the behavior.

PR #3069, opened 2025-11-06
Reduce the frequency of set-cookie in SessionMiddleware

### LINKED ISSUE #2019: SessionMiddleware sends a new set-cookie for every request, with unintended results

Zaczero:
Closes #2019, Requires #3068

# Summary

Utilize the new TimestampSigner to obtain timestamp information, and in combination with a simple equality check, reduce the frequency of set-cookie responses. The set-cookie is sent whenever:

- Session data changes
- Session data is older than 1/4 of max_age

I have additionally corrected the handling of max_age so in the unrealistic edge case of `0` it will treat it as configured value. This is also how itsdangerous handles it, so the current implementation is buggy in that sense.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,6 @@ Starlette only requires `anyio`, and the following are optional:
 * [`httpx`][httpx] - Required if you want to use the `TestClient`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
 * [`python-multipart`][python-multipart] - Required if you want to support form parsing, with `request.form()`.
-* [`itsdangerous`][itsdangerous] - Required for `SessionMiddleware` support.
 * [`pyyaml`][pyyaml] - Required for `SchemaGenerator` support.
 
 You can install all of these with `pip install starlette[full]`.
@@ -135,7 +134,6 @@ in isolation.
 [httpx]: https://www.python-httpx.org/
 [jinja2]: https://jinja.palletsprojects.com/
 [python-multipart]: https://multipart.fastapiexpert.com/
-[itsdangerous]: https://itsdangerous.palletsprojects.com/
 [sqlalchemy]: https://www.sqlalchemy.org
 [pyyaml]: https://pyyaml.org/wiki/PyYAMLDocumentation
 [techempower]: https://www.techempower.com/benchmarks/#hw=ph&test=fortune&l=zijzen-sf
--- a/docs/index.md
+++ b/docs/index.md
@@ -109,7 +109,6 @@ Starlette only requires `anyio`, and the following dependencies are optional:
 * [`httpx`][httpx] - Required if you want to use the `TestClient`.
 * [`jinja2`][jinja2] - Required if you want to use `Jinja2Templates`.
 * [`python-multipart`][python-multipart] - Required if you want to support form parsing, with `request.form()`.
-* [`itsdangerous`][itsdangerous] - Required for `SessionMiddleware` support.
 * [`pyyaml`][pyyaml] - Required for `SchemaGenerator` support.
 
 You can install all of these with `pip install starlette[full]`.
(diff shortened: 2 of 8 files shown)


dolfinus:
Superseded by #3166

PR #3074, opened 2025-11-26
Fix: Raise WebSocketDisconnect instead of RuntimeError on closed conn…

Manas-sin:
# Summary

Replaces the generic `RuntimeError` with [WebSocketDisconnect](cci:2://file:///c:/Users/manas/Downloads/createpr/starlette/starlette/websockets.py:19:0-22:34) (code 1006) when attempting to [send](cci:1://file:///c:/Users/manas/Downloads/createpr/starlette/starlette/websockets.py:58:4-97:48) or [receive](cci:1://file:///c:/Users/manas/Downloads/createpr/starlette/starlette/websockets.py:34:4-56:48) messages on a WebSocket connection that has already been closed.

This improves error handling by allowing applications to catch a specific exception type instead of relying on string matching for `RuntimeError`.

Fixes #2767.

# Checklist

- [x] I've updated the documentation accordingly. (No doc changes needed for this internal fix, but tests are updated)

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -54,7 +54,7 @@ class WebSocket(HTTPConnection):
                 self.client_state = WebSocketState.DISCONNECTED
             return message
         else:
-            raise RuntimeError('Cannot call "receive" once a disconnect message has been received.')
+            raise WebSocketDisconnect(code=1006)
 
     async def send(self, message: Message) -> None:
         """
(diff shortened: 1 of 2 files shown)


Kludex:
But this doesn't make sense; the client isn't supposed to see those error messages.

Let's continue the discussion on the other PR.

PR #3079, opened 2025-12-13
Reduce GC pressure by replacing closures with classes

kevmo314:
# Summary

Replace closure-based send wrappers with class-based wrappers to avoid reference cycles between closure cells, frames, and exception tracebacks.

The closure pattern with `nonlocal` creates cells that form cycles when exceptions are caught, because:
- Exception tracebacks hold frame references
- Frames hold local variables including closures
- Closures hold cells referencing frame locals

The motivation of this change is to reduce client-controlled garbage collection pressure. Specifically, if a client is able to hit a streaming endpoint and then disconnect, they can induce pressure on garbage collection and cause pauses that otherwise wouldn't exist through normal operation. By making the code reference-cycle-free, we rely on ref counting and the closures created don't induce garbage collection pressure.

# Checklist

### DIFF
--- a/starlette/_exception_handler.py
+++ b/starlette/_exception_handler.py
@@ -20,6 +20,25 @@ def _lookup_exception_handler(exc_handlers: ExceptionHandlers, exc: Exception) -
     return None
 
 
+class _ResponseSender:
+    """Wrapper for send that tracks if response has started.
+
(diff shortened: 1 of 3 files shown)


Kludex:
I'm not willing to accept refactor PRs that don't have data to sustain their changes.

In any case, thanks for the PR!

PR #3091, opened 2026-01-01
Add `excluded_content_types` parameter to `GZipMiddleware`

Kludex:


vin:
Back in #2753, at the time, we were of the opinion that a new configuration knob was undesirable. Now that we're proposing adding parameters anyway, are we open to revisiting the idea for a list of content types for which to disable buffering (for streaming responses) rather than only having the option to disable all compression?

Kludex:
> Back in #2753, at the time, we were of the opinion that a new configuration knob was undesirable. Now that we're proposing adding parameters anyway, are we open to revisiting the idea for a list of content types for which to disable buffering (for streaming responses) rather than only having the option to disable all compression?


Yes.

PR #3093, opened 2026-01-01
Fix stream caching in _CachedRequest

kuyugama:
# Summary

I noticed that `_CachedRequest` at `starlette.middleware.base` had uncomplete `stream()` generator caching (`_wrapped_rc_stream`), so I fixed that

# Checklist


Kludex:
Can you give me an MRE that shows the problem?

kuyugama:
I'm going to bed rn, so a bit later. To be clear I was just reading source code and noticed it, so I don't have MRE prepared.

kuyugama:
It seems, that there's no problem with this, so I am closing this pr

PR #3105, opened 2026-01-12
Fix range requests

viccie30:
# Summary

`FileResponse` does not follow [the specification for range requests](https://httpwg.org/specs/rfc9110.html#range.requests). The problems are:
1. Omitting `bytes` in the value of the `Content-Range` header when a range request is not satisfiable.
2. When multiple parts are transferred, the value of the `Content-Type` header should be `multipart/byteranges; boundary=<BOUNDARY>` and `Content-Range` should be unset. `FileResponse` instead sets the `Content-Range` header to `multipart/byteranges` and sends the original `Content-Type`.
3. When multiple parts are transferred, all header lines should be separated by CRLFs, not LFs. A final CRLF after the last boundary is allowed, but not required. It is seen as the start of an optional epilogue.

These 3 issues are addressed by the first 3 commits. The last commit lets `FileResponse` make a copy of the headers before modifying them for a range response. This means that it is possible to use the same `FileResponse` instance multiple times. Not being able to use a `FileResponse` multiple times was the original reason I started looking into this, but it is not documented either way and it isn't technically a bug fix because of that, I think.

Open questions:
- [ ] Should `FileResponse`'s constructor strip out a `Content-Range` header if it is passed in `headers`?

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -398,9 +398,10 @@ class FileResponse(Response):
     async def _handle_single_range(
         self, send: Send, start: int, end: int, file_size: int, send_header_only: bool
     ) -> None:
-        self.headers["content-range"] = f"bytes {start}-{end - 1}/{file_size}"
-        self.headers["content-length"] = str(end - start)
-        await send({"type": "http.response.start", "status": 206, "headers": self.raw_headers})
+        headers = MutableHeaders(raw=list(self.raw_headers))
+        headers["content-range"] = f"bytes {start}-{end - 1}/{file_size}"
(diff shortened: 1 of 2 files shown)


Kludex:
I've merged every commit in a separate PR, I think it's clear like this. Thanks for this, once again! :)

Kludex:
I've kept you as author on every commit, FYI.

viccie30:
Thank you! They really were separate issues, but they all came up while trying to do one thing, so I filed them as 1 PR originally.

PR #3123, opened 2026-02-01
Headers

ademmenh:
# Pull Request: Optimize Header Handling for Server Push

## Description
This pull request introduces a new `get_raw_list` method to the `Headers` datastructure and leverages it to optimize the `send_push_promise` function in the `Request` class. 

The primary goal is to reduce unnecessary decoding and encoding overhead when dealing with raw header bytes during server push operations.

### Core Changes
- **`starlette/datastructures.py`**:
    - Added `Headers.get_raw_list(key: str) -> list[bytes]`: Retrieves the raw byte values for a given header key. 

- **`starlette/requests.py`**:
    - Refactored `send_push_promise` (internal logic within `Request`) to use `get_raw_list`.
    - Improved performance by encoding header names once and avoiding the redundant `decode("latin-1")` -> `encode("latin-1")` cycle for header values.

### Tests
- **`tests/test_datastructures.py`**:
    - Added test case to verify that `get_raw_list` correctly returns byte values for multi-value headers.


Kludex:
Thanks for the contribution, @ademmenh !

I'm going to close this one — HTTP/2 Server Push has been removed from all major browsers, so `send_push_promise` is effectively dead code. Optimizing it and expanding the `Headers` API surface for it isn't worth the trade-off.

PR #3127, opened 2026-02-03
Fix CORSMiddleware to return explicit origin for Authorization header

### LINKED ISSUE #1832: CORSMiddleware does not provide explicit origin although Authorization header is present

KeepALifeUS:
## Summary
- Fixes #1832
- When using `allow_origins=["*"]`, the middleware now correctly returns the specific origin instead of `"*"` when the request includes an `Authorization` header
- This matches the existing behavior for `Cookie` headers (line 164-169)

## Changes
- Added check for `"authorization" in request_headers` alongside the existing cookie check
- Updated comment to reflect the expanded scope
- Added test `test_cors_authorization_header_returns_specific_origin` that verifies the fix

## Background
Per the CORS specification, wildcard `*` origin is not valid for credentialed requests. The middleware already handled this for requests with cookies, but was missing the case when `Authorization` header is present. Token-based authentication (Bearer tokens, etc.) is widely used and requires the same handling.

## Test plan
- [x] Added new test for Authorization header handling
- [x] Existing tests should continue to pass (cookie behavior unchanged)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -162,10 +162,11 @@ class CORSMiddleware:
         headers.update(self.simple_headers)
         origin = request_headers["Origin"]
         has_cookie = "cookie" in request_headers
+        has_authorization = "authorization" in request_headers
 
-        # If request includes any cookie headers, then we must respond
-        # with the specific origin instead of '*'.
-        if self.allow_all_origins and has_cookie:
+        # If request includes any cookie or authorization headers, then we must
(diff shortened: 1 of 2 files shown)


Kludex:
I've merged https://github.com/Kludex/starlette/pull/3137 instead, but thank you for the PR! :)

PR #3130, opened 2026-02-05
Expose `http.response.debug` info via response extensions

y2kbugger:
# Summary

Replace hard-coded `template` and `context` attributes with arbitrary info from `http.response.debug`

Includes a check to make sure we don't overwrite existing attributes

https://asgi.readthedocs.io/en/latest/extensions.html#debug

Developers extending Starlette beyond what the default `templating.py` provides might want to include addition keys to `http.response.debug`'s info

# Checklist

  - existing test should suffice for regressions. https://github.com/Kludex/starlette/pull/1991/changes
  - Looking for guidance on this. does this need to stated somewhere or is it more of, if you are doing this, you already know the ASGI extension?


Kludex:
What's the concrete use case you're trying to solve? Today the only thing that sends `http.response.debug` is `_TemplateResponse`, and it always sends `template` + `context`. Are you building a custom response class that needs to pass additional debug info through the TestClient?

y2kbugger:
Exactly. Specifically, I use a Jinja fragments library to render to
specific blocks. I have a couple custom methods to render a block of
multiple contexts, a context with multiple blocks, a generator of contexts
to give a HTTP stream of blocks fragments. Etc.

Kludex:
@y2kbugger Is this good for you?

y2kbugger:
@Kludex 

Why exclude the full `debug_info` when special casing template?

```
            if "template" in debug_info:
                response.template = debug_info["template"]  # type: ignore[attr-defined]
                response.context = debug_info["context"]  # type: ignore[attr-defined]
            else:
                response.extensions["http.response.debug"] = debug_info
```

Some of my classes use template key also.

I'm working on an update to this to allow debug info regardless.

are you against `response.debug_info` instead of `response.extensions["http.response.debug"]`?

y2kbugger:
ok I updated and added a test for it to unconditionally populate `response.extensions["http.response.debug"]`, even if the `info` has `'template'` as a key.

I assume there is a reason to prefer `response.extensions["http.response.debug"]` over `response.debug_info` so I ommited any changes pertaining to that, at least for now.

y2kbugger:
> @y2kbugger Is this good for you?


@Kludex 
Made some changes, and rebased for 1.0, I think this is good now.

docs/templates.md
```diff
@@ -137,6 +137,20 @@ def test_homepage():
     assert "request" in response.context
 ```
 
+## Customizing Jinja2 Environment
```
Kludex: Is this supposed to be in this PR?

Kludex (approved):
Thanks!

PR #3134, opened 2026-02-08
Return explicit origin for CORS requests with Authorization header

### LINKED ISSUE #1832: CORSMiddleware does not provide explicit origin although Authorization header is present

veeceey:
## Summary

- **Fixes #1832**: `CORSMiddleware` now treats the `Authorization` header as a credential, matching the existing behavior for `Cookie` headers.
- When `allow_origins=["*"]` and a request includes an `Authorization` header, the response returns the explicit request origin in `Access-Control-Allow-Origin` instead of `*`.
- This complies with the [Fetch specification](https://fetch.spec.whatwg.org/#cors-protocol-and-credentials), which defines credentials as including cookies, TLS client certificates, **and authentication headers**.

### Problem

Browsers reject CORS responses that return `Access-Control-Allow-Origin: *` when the request includes credentials. The middleware already handled this for `Cookie` headers, but not for `Authorization` headers:

```python
# Before (only checked cookies):
if self.allow_all_origins and has_cookie:
    self.allow_explicit_origin(headers, origin)

# After (also checks authorization):
if self.allow_all_origins and (has_cookie or has_authorization):
    self.allow_explicit_origin(headers, origin)
```

### Changes

- `starlette/middleware/cors.py`: Added `has_authorization` check alongside the existing `has_cookie` check in the `send` method.
- `tests/middleware/test_cors.py`: Added 4 new test functions covering:
  - Authorization header returns explicit origin (with and without `allow_credentials`)
  - Origin does not leak between authorization and non-authorization requests
  - `Vary: Origin` header is properly set for authorization requests

## Test plan

- [x] All 42 CORS tests pass (21 asyncio + 21 trio), including 8 new test runs
- [x] Existing tests remain unmodified and passing
- [x] Manual verification with a browser making credentialed `fetch()` calls with `Authorization` header

### DIFF
--- a/starlette/middleware/cors.py
+++ b/starlette/middleware/cors.py
@@ -162,10 +162,11 @@ class CORSMiddleware:
         headers.update(self.simple_headers)
         origin = request_headers["Origin"]
         has_cookie = "cookie" in request_headers
+        has_authorization = "authorization" in request_headers
 
-        # If request includes any cookie headers, then we must respond
-        # with the specific origin instead of '*'.
-        if self.allow_all_origins and has_cookie:
+        # If request includes any cookie or authorization headers, then we
+        # must respond with the specific origin instead of '*'.
+        if self.allow_all_origins and (has_cookie or has_authorization):
             self.allow_explicit_origin(headers, origin)
 
         # If we only allow specific origins, then we have to mirror back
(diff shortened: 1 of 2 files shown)


Kludex:
I've merged https://github.com/Kludex/starlette/pull/3137 instead, but thank you for the PR! :)

PR #3135, opened 2026-02-10
✨ Add `max_body_size` parameter to `Request.stream()` and related methods

tiangolo:
✨ Add `max_body_size` parameter to `Request.stream()` and related methods

https://github.com/Kludex/starlette/pull/3151 might be a better approach, implementing things as suggested by @Kludex 

# Summary

When a body size is too large, raise an exception of "413 Content Too Large".

This is done in `Request.stream()`, so it errors out before consuming the data (and storing stuff in memory).

It also errors out even when the data is already cached for consistency, e.g. `Request.body()` cached the data once with a large body and `Request.body(max_body_size=5)` is called right after.

This would hopefully fix / related to https://github.com/Kludex/starlette/issues/2155

Having it in this place in the `Request` (instead of a middleware) would let me use it in FastAPI how I need it, would allow me to set parameters in the app, router, and path operation levels.

## Questions

I don't know if Starlette would like to have other parameters, e.g. at the app level, or some middleware that sets a parameter, and then it could be read by the request from the ASGI scope.

## AI disclaimer

I used AI to help write the code and tests, then I manually reviewed carefully all the code and adjusted, reverted (and unreverted :sweat_smile: ) some changes.

The original prompt:

```
Implement a `max_body_size` in `Request.stream`

If the stream is beyond the size in bytes, raise an HTTPException

Pass the same `max_body_size` to `Request.body`, `Request.form` and `Request.json`

Implement tests for it

Use the venv in .venv
```

Using Claude Opus 4.5 (and 4.6).

There were some manual adjustments and additional prompts, etc.

# Checklist

    - I will update the docs after we confirm this makes sense, and/or add any other configs needed.

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -225,13 +225,16 @@ class Request(HTTPConnection[StateT]):
     def receive(self) -> Receive:
         return self._receive
 
-    async def stream(self) -> AsyncGenerator[bytes, None]:
+    async def stream(self, max_body_size: int | None = None) -> AsyncGenerator[bytes, None]:
         if hasattr(self, "_body"):
+            if max_body_size is not None and len(self._body) > max_body_size:
+                raise HTTPException(status_code=413, detail="Content Too Large")
             yield self._body
             yield b""
             return
         if self._stream_consumed:
             raise RuntimeError("Stream consumed")
+        total_size = 0
         while not self._stream_consumed:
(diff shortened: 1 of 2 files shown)


Kludex:
- Closed by https://github.com/Kludex/starlette/pull/3431

PR #3146, opened 2026-02-15
Enable autoescape by default in Jinja2Templates

RinZ27:
# Summary

Noticed that `Jinja2Templates` currently defaults to having `autoescape` disabled. It puts the burden on the developer to remember to enable it, which I've seen lead to XSS vulnerabilities in applications that render user-provided content.

While investigating, I also discovered that `Jinja2Templates` was missing support for `**env_options` in its constructor, making it difficult to customize the environment even though the documentation suggested it was supported.

Updated the implementation to:
- Use `jinja2.select_autoescape()` by default for HTML/XML files to ensure a secure baseline.
- Add support for `**env_options` in the `Jinja2Templates` constructor, allowing full customization of the Jinja2 environment.

### Changes
- Modified `Jinja2Templates.__init__` to accept and use `**env_options`.
- Enabled `autoescape` by default using `select_autoescape()` if not explicitly provided.
- Added test cases in `tests/test_templates.py` to verify both the secure default and the ability to override it.

# Checklist

- [x] Understood that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [x] Added a test for each change that was introduced, while I tried as much as possible to make a single atomic change.
- [x] Updated the documentation accordingly.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -129,7 +129,8 @@ def test_homepage():
 ## Customizing Jinja2 Environment
 
 `Jinja2Templates` accepts all options supported by Jinja2 `Environment`.
-This will allow more control over the `Environment` instance created by Starlette.
+Enabling `autoescape` by default for HTML/XML files using `jinja2.select_autoescape()` ensures a secure baseline against XSS vulnerabilities.
+These options can be customized or overridden by passing them as keyword arguments.
 
 For the list of options available to `Environment` you can check Jinja2 documentation [here](https://jinja.palletsprojects.com/en/3.0.x/api/#jinja2.Environment)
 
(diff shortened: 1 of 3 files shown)


RinZ27:
@Kludex, noticed #3148 covers the autoescape change. Moving forward with that is cool, though I'm happy to sync this PR with the new docs refactor if keeping the `**env_options` support makes sense.

Missing implementation for `**env_options` was something I spotted while checking the docs, so I've included it here.

Kludex:
> **env_options support makes sense.

We just dropped that.

RinZ27:
@Kludex got it. Since #3148 handles the autoescape and env_options is out, I'll close this. Thanks!

PR #3149, opened 2026-02-15
Add `BaseResponse` class and `ContentT` generic to `Response`

Kludex:
Continuation of #2547.

Extract `BaseResponse` from `Response` to separate responses that use `render()` (`Response`, `HTMLResponse`, `JSONResponse`, etc.) from those that don't (`StreamingResponse`, `FileResponse`). Currently `StreamingResponse` and `FileResponse` inherit from `Response` but never call `render()` — they only use shared infrastructure (headers, cookies, status code).

`Response` is now `Response(BaseResponse, Generic[ContentT])`, and `JSONResponse` is `JSONResponse(Response[ContentT])`.

`StreamingResponse` and `FileResponse` inherit from `BaseResponse` directly, and call `super().__init__()` instead of manually setting attributes.

Breaking changes:
- `StreamingResponse` and `FileResponse` no longer inherit from `Response`.
- Type aliases that used `Response` now use `BaseResponse` (e.g. `HTTPExceptionHandler`, `RequestResponseEndpoint`).

### DIFF
--- a/starlette/applications.py
+++ b/starlette/applications.py
@@ -8,7 +8,7 @@ from starlette.middleware import Middleware, _MiddlewareFactory
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.middleware.exceptions import ExceptionMiddleware
 from starlette.requests import Request
-from starlette.responses import Response
+from starlette.responses import BaseResponse
 from starlette.routing import BaseRoute, Router
 from starlette.types import ASGIApp, ExceptionHandler, Lifespan, Receive, Scope, Send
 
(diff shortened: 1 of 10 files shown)


Kludex:
@abersheeran @adriangb is this what we had in mind? (I don't think we should add this for V1 tho.)

abersheeran:
I'm not sure if this will break `isinstance` check that in `fastapi` or other codes. From my personal perspective, not changing the name of Response but instead creating a new base class might be more forward-compatible.

Kludex:
Let's close this for the time being. Happy to review any human PR related to this from a previous contributor.

Issue #3113, opened 2026-01-20
An API that reads the request body for a specified number of bytes?


marunguy:
Is there an API that reads a specified number of bytes from the request body?

https://docs.python.org/3/library/io.html#io.Reader
https://docs.aiohttp.org/en/stable/streams.html#aiohttp.StreamReader.read

This is needed to save portions of the body to multiple files while parsing a large request body.

While it's possible to save the entire request body to a file and then parse the saved file, this functionality is desired to reduce disk I/O.

Multipart is also a possible method, but unfortunately, we are currently unable to change the agent.

marcelomarkus:
If you have the flexibility to define how data is packaged at the source, take a look at b-fast. 
https://github.com/marcelomarkus/b-fast
B-FAST is a very high-performance binary serialization protocol for critical routes where latency, CPU usage, and bandwidth represent bottlenecks.

KeeganOP:
You can stream the body in chunks but not specific byte sizes as far as I know. Will this work for you:

```python 
async def homepage(request: Request):
    files = []

    async for chunk in request.stream():
        tmp_file = tempfile.TemporaryFile(mode='w+b')
        tmp_file.write(chunk)
        tmp_file.seek(0)
        files.append(tmp_file)

    for file in files:
        print(file.read())

```
Lifted from the documentation at https://starlette.dev/requests/ under the Body section

PR #3164, opened 2026-02-28
feat: add Request.read(size) method for partial body reading

Jagdish1115:
# Summary

Adds a read(size) method to the Request class that reads exactly size bytes from the request body at a time. Currently there is no way to read a specific number of bytes from the request body. The only options are body() which loads everything into memory at once, or stream() which gives server-decided chunk sizes.
This is useful for processing large request bodies in chunks without loading everything into memory — for example, saving portions of a large body to multiple files.

Relates to #3113 
Example usage:
chunk1 = await request.read(1024) # first 1KB
chunk2 = await request.read(1024) # next 1KB
chunk3 = await request.read(-1) # rest of body

# Checklist

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -215,6 +215,7 @@ class Request(HTTPConnection[StateT]):
         self._send = send
         self._stream_consumed = False
         self._is_disconnected = False
+        self._read_buffer = b""
         self._form = None
 
     @property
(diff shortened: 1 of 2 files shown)


Jagdish1115:
Hi,

This is my first open source contribution, so I appreciate any feedback or guidance.
I've implemented the `read(size)` method on the `Request` class as discussed in #3113.  The method reads up to `size` bytes from the request body at a time using an internal buffer, avoiding loading the entire body into memory at once.
Regarding the coverage failure — I noticed the main branch is already at 89% coverage before my changes. I've added tests for the main scenarios but I'm still working on fully covering line 259. Any guidance would be greatly appreciated.

PR #3182, opened 2026-03-11
Fix precedence of explicit HEAD routes over implicit HEAD

Hedingber:
# Summary

Fixes the routing bug described in [Discussion #3128](https://github.com/Kludex/starlette/discussions/3128#discussion-9424965).

Previously, `Route` added `"HEAD"` to its `methods` list when initialized with `methods=["GET"]`. Because routing resolves sequentially via `Match.FULL`, a `HEAD` request would trigger the implicit `HEAD` fallback on a default `GET` route rather than continuing the search for an explicitly tailored `HEAD` handler mapped later in the `Router`.

This PR:
- Adds an internal tracking property `self.explicit_methods` to log methods genuinely passed during route initialization by the user.
- Falls back to `Match.PARTIAL` for `HEAD` requests occurring on routes that didn't intentionally list it in `explicit_methods`. This politely passes routing scope further down the line to ensure real `HEAD` handlers still take precedence!
- Adds a regression test case.

# Checklist


Kludex:
Thanks for the PR. I’m closing this because prioritizing explicit HEAD routes regardless of order adds complexity and risks regressions in fallback routing. Registering the HEAD route before the GET route already supports this use case and follows the documented route priority rules.

PR #3193, opened 2026-03-19
Use `removeprefix` to strip weak ETag indicator in `is_not_modified`

gnosyslambda:
`is_not_modified()` uses `tag.strip(" W/")` to remove the weak ETag prefix, but `strip()` works on a character set — it strips any combination of `' '`, `'W'`, and `'/'` from both ends of the string. This means an ETag value ending in `W` or `/` would get incorrectly trimmed.

Replaced with `tag.strip().removeprefix("W/")` which strips whitespace first, then removes the literal `W/` prefix only. `removeprefix` is available since Python 3.9 and Starlette requires 3.10+.

In practice this is unlikely to cause issues with the current MD5-based ETags from FileResponse, but the fix makes the code do what it actually intends to do.


themavik (commented):
Good catch on `strip(" W/")` — it can remove unrelated characters at the ends of a strong etag. `strip()` plus `removeprefix("W/")` matches the weak-etag prefix case without over-stripping the quoted value.

themavik (commented):
`tag.strip().removeprefix("W/")` fixes weak ETag handling compared to `strip(" W/")`, which could eat unrelated leading characters.

Kludex (approved):
Thanks!

PR #3198, opened 2026-03-20
Deprecate HTTP_306_RESERVED status code

### LINKED ISSUE #3192: Remove or deprecate HTTP_306_RESERVED — status code 306 is (Unused) per IANA and RFC 9110
The starting point for issues should usually be a discussion...

https://github.com/Kludex/starlette/discussions

Possible bugs may be raised as a "Potential Issue" discussion, feature requests may be raised as an "Ideas" discussion. We can then determine if the discussion needs to be escalated into an "Issue" or not.

This will help us ensure that the "Issues" list properly reflects ongoing or needed work on the project.

---

**Summary**

`starlette/status.py` defines `HTTP_306_RESERVED = 306`, but status code 306 is marked as **(Unused)** in the IANA HTTP Status Code Registry and RFC 9110.

**References**

- **IANA Registry** (https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml): 306 is listed as "(Unused)" between 305 (Use Proxy) and 307 (Temporary Redirect).
- **RFC 9110, Section 15.4.7** (https://www.rfc-editor.org/rfc/rfc9110#section-15.4.7): *"The 306 status code was defined in a previous version of this specification, is no longer used, and the code is reserved."*

**Current state in Starlette**

- `HTTP_306_RESERVED` is exported in `__all__` and defined as `306`.
- The constant name suggests it is reserved, but the spec explicitly states the code is no longer used.

**Proposal**

Consider removing `HTTP_306_RESERVED` from `starlette/status.py`, or deprecating it (e.g. via `__deprecated__` and a deprecation warning) to align with RFC 9110 and the IANA registry.

felmonon:
Closes #3192.

Status code 306 is listed as **(Unused)** in the [IANA HTTP Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml) and [RFC 9110 Section 15.4.7](https://www.rfc-editor.org/rfc/rfc9110#section-15.4.7) states it is no longer used and the code is reserved.

This PR moves `HTTP_306_RESERVED` from `__all__` into the existing `__deprecated__` dict so that existing code continues to work but receives a `DeprecationWarning`. The `__getattr__` logic was updated to handle deprecated constants that have no replacement (as opposed to the existing renamed constants like `HTTP_413_REQUEST_ENTITY_TOO_LARGE` → `HTTP_413_CONTENT_TOO_LARGE`).

**Changes:**
- Removed `HTTP_306_RESERVED` from `__all__` and its direct assignment
- Added `HTTP_306_RESERVED` to the `__deprecated__` dict
- Updated `__getattr__` to support deprecated constants without a replacement name
- Added test case for the new deprecation warning

### DIFF
--- a/starlette/status.py
+++ b/starlette/status.py
@@ -31,7 +31,6 @@ __all__ = [
     "HTTP_303_SEE_OTHER",
     "HTTP_304_NOT_MODIFIED",
     "HTTP_305_USE_PROXY",
-    "HTTP_306_RESERVED",
     "HTTP_307_TEMPORARY_REDIRECT",
     "HTTP_308_PERMANENT_REDIRECT",
     "HTTP_400_BAD_REQUEST",
@@ -111,7 +110,6 @@ HTTP_302_FOUND = 302
 HTTP_303_SEE_OTHER = 303
 HTTP_304_NOT_MODIFIED = 304
 HTTP_305_USE_PROXY = 305
-HTTP_306_RESERVED = 306
 HTTP_307_TEMPORARY_REDIRECT = 307
 HTTP_308_PERMANENT_REDIRECT = 308
 HTTP_400_BAD_REQUEST = 400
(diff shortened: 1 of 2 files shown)


Kludex:
Same as https://github.com/Kludex/starlette/issues/3192#issuecomment-4106457628.

Thanks for the PR.

PR #3200, opened 2026-03-22
Add is_disconnected property to WebSocket

r266-tech:
## Description

Adds a synchronous `is_disconnected` property to the `WebSocket` class that returns `True` when either the client or application state is `DISCONNECTED`.

Closes #2766

### Motivation
When handling WebSocket connections with multiple coroutines, it's common for more than one coroutine to attempt closing the connection. Without a way to check the connection state synchronously, this leads to `RuntimeError` exceptions.

This property enables clean, exception-free state checking:

```python
if not websocket.is_disconnected:
    await websocket.close()
```

### Implementation
The property simply checks whether either `client_state` or `application_state` equals `WebSocketState.DISCONNECTED` — both of which are already tracked by the existing state machine in `starlette/websockets.py`.

### Changes
- Added `is_disconnected` property to `WebSocket` class
- Added unit test `test_is_disconnected_property`

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -32,6 +32,13 @@ class WebSocket(HTTPConnection[StateT]):
         self.client_state = WebSocketState.CONNECTING
         self.application_state = WebSocketState.CONNECTING
 
+    @property
+    def is_disconnected(self) -> bool:
+        return (
+            self.client_state == WebSocketState.DISCONNECTED
+            or self.application_state == WebSocketState.DISCONNECTED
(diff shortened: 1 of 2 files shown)


themavik (commented):
test assigns `websocket.client_state` directly after `receive()`. That skips the normal transition you are trying to observe and could drift if `receive()` ever stops leaving the object in a different state. Prefer driving a real disconnect/close path only, or add a short comment on why the mutation is required.

PR #3201, opened 2026-03-22
Handle standalone If-None-Match wildcards in StaticFiles

r266-tech:
`StaticFiles` currently returns `200` for an existing file requested with `If-None-Match: *`. Return `304` for `GET` and `HEAD` when the header is a standalone wildcard, allowing surrounding whitespace, as required by [RFC 9110 §13.1.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.2).

Keep quoted ETags such as `"foo,*,bar"` on the normal comparison path. Splitting the header to look for wildcards incorrectly matches these tags; mixed wildcard lists are syntactically invalid under the RFC. The wildcard test covers both methods and surrounding whitespace. The existing mismatch test also covers `"*"` and `"foo,*,bar"`.

Validation: 939 tests passed, 2 skipped, 2 expected failures; 100% coverage. Ruff lint, formatting, and mypy 1.16.1 checks passed.

## AI Disclaimer

### DIFF
--- a/starlette/staticfiles.py
+++ b/starlette/staticfiles.py
@@ -207,6 +207,9 @@ class StaticFiles:
         if if_none_match := request_headers.get("if-none-match"):
             # The "etag" header is added by FileResponse, so it's always present.
             etag = response_headers["etag"]
+            # Per RFC 7232 §3.2, "If-None-Match: *" matches any current representation.
+            if if_none_match.strip() == "*":
+                return True
             return etag in [tag.strip(" W/") for tag in if_none_match.split(",")]
 
         try:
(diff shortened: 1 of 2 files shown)


themavik (commented):
nit: only treats the header as a wildcard when the whole value strips to `*`. If you want closer RFC behavior for combined lists (`*, "etag"`), consider splitting on commas and matching any segment that is `*` after strip.

themavik (commented):
`StaticFiles.is_not_modified` short-circuits `If-None-Match: *` per RFC 7232; test covers the straight `*` case. Values like `foo, *` still go through the comma split path.

Kludex:
Thanks for this PR! :)

PR #3202, opened 2026-03-22
Replace RuntimeError with WebSocketDisconnect when socket is disconnected

r266-tech:
Closes #2766

## Changes
- Replace `RuntimeError` with `WebSocketDisconnect` in `WebSocket.receive()` when called after disconnect message has been received
- Replace `RuntimeError` with `WebSocketDisconnect` in `WebSocket.send()` when called after close message has been sent  
- Add `is_disconnected` property to `WebSocket` for checking connection state (as discussed in [#2762](https://github.com/encode/starlette/discussions/2762))
- Update tests to expect `WebSocketDisconnect` instead of `RuntimeError`
- Add tests for `is_disconnected` property

## Motivation
When handling WebSocket connections in concurrent scenarios (e.g., multiple coroutines managing the same connection), it's common for one coroutine to close the connection while another attempts to send/receive. Currently this raises a generic `RuntimeError`, but `WebSocketDisconnect` is more semantically appropriate and allows users to catch a single exception type.

## Testing
- All 290 relevant tests pass (websocket, routing, applications, endpoints, testclient)
- Added 2 new tests for `is_disconnected` property

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -32,6 +32,13 @@ class WebSocket(HTTPConnection[StateT]):
         self.client_state = WebSocketState.CONNECTING
         self.application_state = WebSocketState.CONNECTING
 
+    @property
+    def is_disconnected(self) -> bool:
+        return (
+            self.client_state == WebSocketState.DISCONNECTED
+            or self.application_state == WebSocketState.DISCONNECTED
+        )
+
     async def receive(self) -> Message:
         """
         Receive ASGI websocket messages, ensuring valid state transitions.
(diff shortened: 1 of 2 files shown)


themavik (commented):
probe

themavik:
Ignore my earlier one-word review — bad local API check, not real feedback.

themavik (commented):
Swapping the stale `receive`/`send` RuntimeErrors for `WebSocketDisconnect` is a nicer ASGI signal, though `code=1000` on programmer misuse can look like a normal close to code that only inspects `code`. `is_disconnected` matches the usual "either side closed" guard people want before sending.

PR #3219, opened 2026-04-04
Improve typing for add_exception_handler

Ragug:
Allow `add_exception_handler` to accept HTTP and WebSocket handlers whose exception parameter matches the registered exception class. Generic overloads accept correctly paired handlers and reject handlers that cannot handle the registered exception type, while preserving runtime behavior.

Make the HTTP and WebSocket handler aliases generic with `Exception` as the default, so existing bare annotations retain their meaning. Reuse `ExceptionHandler[_E]` for class registrations and `ExceptionHandler[Exception]` for integer registrations, since `500` handles arbitrary exceptions. The `exception_handlers` constructor mapping is unchanged. This follows the narrower scope discussed in [#2403](https://github.com/Kludex/starlette/pull/2403#issuecomment-1925728367).

Validation: add one compact strict typing test for accepted and rejected registrations, including synchronous and asynchronous HTTP handlers, WebSocket handlers, `500`, and the default alias types. Existing runtime tests remain unchanged. The full suite passes locally on Python 3.14: 928 passed, 2 skipped, 2 expected failures, and 100% coverage. Strict mypy and Ruff checks pass.

## AI Disclaimer


Ragug:
Hi @Kludex, I’d appreciate it if you could review this and let me know if any changes are needed. Looking forward to your feedback.

PR #3223, opened 2026-04-05
docs: add docstrings to background and concurrency modules

mastash3ff:
# Summary

Add docstrings to the public classes and functions in `starlette/background.py` and `starlette/concurrency.py`. These modules are part of the core public API and were missing any inline documentation.

Changes:
- `BackgroundTask` — class docstring describing purpose and async/sync dispatch behaviour
- `BackgroundTasks` — class docstring describing sequential execution semantics
- `BackgroundTasks.add_task` — one-line method docstring
- `run_in_threadpool` — one-line function docstring
- `iterate_in_threadpool` — multi-line docstring describing the thread-pool wrapping behaviour

No logic, behaviour, or test changes.

# Checklist

### DIFF
--- a/starlette/background.py
+++ b/starlette/background.py
@@ -10,6 +10,13 @@ P = ParamSpec("P")
 
 
 class BackgroundTask:
+    """A task to run in the background after a response has been sent.
+
+    Accepts a callable along with any positional and keyword arguments.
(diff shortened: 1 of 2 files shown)


Kludex:
Stop.

PR #3230, opened 2026-04-10
Fix GZipMiddleware buffering streaming response chunks

janbjorge:
# Summary
GZipResponder.apply_compression() never called gzip_file.flush() for intermediate streaming chunks (more_body=True), causing zlib to hold compressed bytes in its internal buffer until ~32KB accumulated or the stream closed. Added self.gzip_file.flush() (Z_SYNC_FLUSH) so compressed bytes are emitted immediately for each chunk.

# Checklist


Kludex:
Please create a discussion with an MRE.

PR #3240, opened 2026-04-16
Add multipart header limit options to Request.form

Kludex:
## Summary

Add support for `python-multipart`'s multipart header limit kwargs from Starlette's request form API.

This threads `max_header_count` and `max_header_size` through:

- `Request.form()`
- `Request._get_form()`
- `MultiPartParser`
- `python_multipart.MultipartParser(...)`

The defaults match the current `python-multipart` master branch defaults:

- `max_header_count=8`
- `max_header_size=4096 + 128`

## Why

`python-multipart` added header-count and header-size limits on multipart parts, but Starlette did not yet expose those knobs from `request.form()`.

This keeps the API consistent with the existing multipart parsing limit options like `max_files`, `max_fields`, and `max_part_size`.

## Notes

- Existing defaults and behavior are preserved.
- If a caller sets non-default header limits while running against an older installed `python-multipart` that does not accept those kwargs yet, Starlette raises a clear runtime error instead of silently ignoring the configuration.

## Tests

- Added coverage that the new kwargs are passed through to the underlying multipart parser.
- Added enforcement coverage for header-count and header-size limits.
- Updated docs and release notes.

## Validation

- `ruff check starlette/formparsers.py starlette/requests.py tests/test_formparsers.py`
- `python -m pytest tests/test_formparsers.py -q`


docs/release-notes.md
```diff
@@ -2,6 +2,12 @@
 toc_depth: 2
 ---
 
+## Unreleased
+
+#### Added
+
+* Add `max_header_count` and `max_header_size` parameters to `Request.form()`.
```
Kludex, line 9: Remove this.

docs/requests.md
```diff
@@ -113,12 +113,18 @@ state with `disconnected = await request.is_disconnected()`.
 
 Request files are normally sent as multipart form data (`multipart/form-data`).
 
-Signature: `request.form(max_files=1000, max_fields=1000, max_part_size=1024*1024)`
+Signature: `request.form(max_files=1000, max_fields=1000, max_part_size=1024*1024, max_header_count=8, max_header_size=4096 + 128)`
 
-You can configure the number of maximum fields or files with the parameters `max_files` and `max_fields`; and part size using `max_part_size`:
+You can configure the number of maximum fields or files with the parameters `max_files` and `max_fields`; part size using `max_part_size`; and multipart header limits using `max_header_count` and `max_header_size`:
```
Kludex, line 118: On each of the previous parameters we slightly explain what they do, we should do the same with `max_header_count` and `max_header_size`.

starlette/formparsers.py
```diff
@@ -127,6 +141,10 @@ class MultiPartParser:
     """The maximum size of the spooled temporary file used to store file data."""
     max_part_size = 1024 * 1024  # 1MB
     """The maximum size of a part in the multipart request."""
+    max_header_count = DEFAULT_MAX_MULTIPART_HEADER_COUNT
+    """The maximum number of headers per multipart part."""
+    max_header_size = DEFAULT_MAX_MULTIPART_HEADER_SIZE
+    """The maximum size of a multipart header line."""
```
Kludex, line 147: We should not include those here. They should be only object attributes, not class.

starlette/formparsers.py
```diff
@@ -136,12 +154,16 @@ def __init__(
         max_files: int | float = 1000,
         max_fields: int | float = 1000,
         max_part_size: int = 1024 * 1024,  # 1MB
+        max_header_count: int = DEFAULT_MAX_MULTIPART_HEADER_COUNT,
```
Kludex, line 157: We should call it `max_headers` instead - aligned with `max_fields` and `max_files`.

starlette/formparsers.py
```diff
@@ -248,7 +270,21 @@ async def parse(self) -> FormData:
...
+            }
+        elif (
+            self.max_header_count != DEFAULT_MAX_MULTIPART_HEADER_COUNT
+            or self.max_header_size != DEFAULT_MAX_MULTIPART_HEADER_SIZE
+        ):
+            raise RuntimeError(
+                "The installed `python-multipart` version does not support "
+                "`max_header_count` or `max_header_size`. Upgrade `python-multipart` to configure them."
+            )
+        parser = multipart.MultipartParser(boundary, callbacks, **parser_kwargs)
```
Kludex, line 287: I think this is overly complicated.

PR #3241, opened 2026-04-16
Add tests for request caching (body, JSON, form) and stream consumption guarantees

diskeu:
# Summary

Adds tests for request body caching behavior in Starlette.

This includes coverage for:
- raw body caching
- JSON-to-body caching (`request.json()`)
- form parsing caching (both `application/x-www-form-urlencoded` and `multipart/form-data`)
- stream gets consumed only once guarantees (`request.stream()`)
 
To improve test clarity and reuse, two fixtures were introduced:
- `scope`: provides a basic HTTP ASGI scope
- `receive` tracker: wraps the ASGI receive callable and tracks the call count to verify that the receive function only gets called once
# Checklist

- [ ] I've updated the documentation accordingly. (no need to update)

### DIFF
--- a/tests/test_requests.py
+++ b/tests/test_requests.py
@@ -14,6 +14,12 @@ from starlette.types import Message, Receive, Scope, Send
 from tests.types import TestClientFactory
 
 
+@pytest.fixture()
+def scope() -> dict[str, Any]:
+    scope = {"type": "http", "path": "/", "method": "GET", "headers": [(b"content-type", b"application/json")]}
+    return scope
+
+
 class ReceiveTracker:
     def __init__(self) -> None:
         self._count = 0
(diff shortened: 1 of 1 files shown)


Kludex:
What's the motivation here?

diskeu:
> What's the motivation here?

Without strict guarantees around request body caching, it is easy to accidentally consume the receive stream multiple times.

There is also no straightforward way to reliably test whether caching is actually working without explicitly tracking receive calls. This makes changes to caching logic risky, as correctness must otherwise be assumed rather than verified.

These issues are subtle because they depend on internal request access patterns and are not immediately visible from the outside behavior.

Adding explicit tests and receive tracking improves confidence in the implementation and helps ensure correct behavior even as new caching-related features are added in the future.

PR #3243, opened 2026-04-18
Describe disconnected-after-response behavior in test docstring

diskeu:
The current docstring implies that the test verifies whether a request is
marked as disconnected when a client disconnect occurs after reading
the request body.

However, the test actually verifies that the is_disconnected state
remains consistent after reading the request body and sending a response.

This PR updates the docstring to match the actual behavior of the test.

# Checklist


Kludex (approved):
Thanks!

PR #3244, opened 2026-04-18
Add strict JSON Content-Type validation

diskeu:
# Summary

Introduces a `strict` parameter for `Request.json()` that validates the request's
`Content-Type` before parsing.

When `strict=True`, a `RuntimeError` is raised if the `Content-Type` is
not `application/json` or does not match `application/*+json`.

This makes JSON parsing consistent with `Request.form()` which also checks
the `Content-Type` header.

# Checklist

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -258,8 +258,22 @@ class Request(HTTPConnection[StateT]):
             self._body = b"".join(chunks)
         return self._body
 
-    async def json(self) -> Any:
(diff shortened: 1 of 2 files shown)


hashim2266 (approved):

Kludex:
I don't think the parameter is necessary, but maybe we can add a `try..except` around the `json.loads`?

Kludex:
Closing as stale, and I don't think this is necessary, you can read the body yourself.

PR #3262, opened 2026-05-11
Support reading config files from FIFO (Unix)

j-labbe:
# Summary

- Adds support to read config vars from FIFOs (Unix).

# Checklist


Kludex:
I'm happy to maintain `Config` for some time as a simple config class, but I'm not willing to add new features, since I don't use it myself.

Thanks for the PR.

PR #3271, opened 2026-05-15
Implement `identity` on `SimpleUser` and `UnauthenticatedUser`

navidnabavi:
fix(authentication): implement identity property on SimpleUser and UnauthenticatedUser

BaseUser defined identity as abstract but neither concrete subclass implemented it, causing NotImplementedError at runtime.

- SimpleUser.identity returns username
- UnauthenticatedUser.identity returns ""

# Summary

BaseUser defines identity as an abstract property (raises NotImplementedError).
Neither SimpleUser nor UnauthenticatedUser implement it.
[Discussion](https://github.com/Kludex/starlette/discussions/3270)

# Checklist

- [ x ] I understand that this PR may be closed in case there was no previous discussion. (This doesn't apply to typos!)
- [ x ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ x ] I've updated the documentation accordingly.


Kludex (approved):

Kludex:
Thanks!

PR #3279, opened 2026-05-21
Ignore malformed `Host` header when constructing `request.url`

Kludex:
When the `Host` header contains characters that are invalid per RFC 9110 §7.2 (`/`, `?`, `#`, `@`, `\`, spaces, ...), `urlsplit` can produce a `URL` whose `path` differs from `scope["path"]`. This makes `request.url.path` unreliable for code that inspects it.

This PR validates the `Host` header against an allowlist regex (matching Werkzeug's and Django's approach) before using it - only `[a-zA-Z0-9.-]` for domains, valid IPv6 in brackets, and optional port numbers. If the header doesn't match, we fall back to the `server` tuple from scope, the same as when no `Host` header is present.

## AI Disclaimer

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -1,14 +1,9 @@
 from __future__ import annotations
 
+import re
 from collections.abc import ItemsView, Iterable, Iterator, KeysView, Mapping, MutableMapping, Sequence, ValuesView
 from shlex import shlex
-from typing import (
-    Any,
-    BinaryIO,
-    NamedTuple,
-    TypeVar,
-    cast,
-)
+from typing import Any, BinaryIO, NamedTuple, TypeVar, cast
 from urllib.parse import SplitResult, parse_qsl, urlencode, urlsplit
 
 from starlette.concurrency import run_in_threadpool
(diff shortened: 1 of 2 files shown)


aeris:
@Kludex This PR seems AI hallucination and dangerous

Host header can be UTF-8 percent encoded, or IDNA
https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2.2
```
The reg-name syntax allows percent-encoded octets in order to
   represent non-ASCII registered names in a uniform way that is
   independent of the underlying name resolution technology.  Non-ASCII
   characters must first be encoded according to UTF-8 [[STD63](https://www.rfc-editor.org/info/rfc3986/#ref-STD63)], and then
   each octet of the corresponding UTF-8 sequence must be percent-
   encoded to be represented as URI characters.
 ```

You seems rejecting valid host with such very strict ASCII-only regex

Worse, weird host using (% or @ or anything now not included on the regex should be also good indicator for an attack, but your PR just mask totally the attack with fallback to server host. We lose critical information in such case.

Kludex:
> @Kludex This PR seems AI hallucination and dangerous

Your comment is dangerous. No AI hallucination here.

---

I don't discuss vulnerabilities in public, but happy to reject your points with more context in a security advisory, as per our [security policy](https://github.com/Kludex/starlette?tab=security-ov-file).

Issue #1027, opened 2020-08-11
Wrong content_type for UploadFile


elsimqoz:
```
python 3.8
fastapi==0.60.1
pydantic==1.6.1
starlette==0.13.6
pytest==5.4.3
```

test:
```
@pytest.fixture()
def app():
    return create_app()


@pytest.fixture()
def web_client(app):
    return TestClient(app=app)


@pytest.fixture()
def file():
    f = NamedTemporaryFile(suffix='.csv')
    f.write(b'123;4\n5;67')
    f.seek(0)
    return f


def test_import(web_client, file):
    files = {'file': open(file.name, 'rb')}
    web_client.post('/import', files=files)
```

view:
```
import mimetypes

from fastapi import APIRouter, File, UploadFile


router = APIRouter()


@router.post('/import')
def import_file(file: UploadFile = File(...)):
    assert mimetypes.guess_type(file.filename)[0] == 'text/csv'  # ok
    assert file.content_type == 'text/csv'  # fail
```
Why `file.content_type == ''`?

erewok:
If you include the headers to instruct the server what your content type is, then it will use those. Here's an example using Starlette:

```python
async def import_file(request):
    form = await request.form()
    filename = form["upload_file"].filename
    content_type = form["upload_file"].content_type
    return JSONResponse({"content_type": content_type, "filename": filename})
```

```python
In [1]: import requests

In [2]: with open("test.csv", "r") as fl:
   ...:     resp = requests.post("http://localhost:8000/i", files={"upload_file": ('test.csv', fl, 'text/csv')})
   ...: resp.json()
Out[2]: {'content_type': 'text/csv', 'filename': 'test.csv'}
```

elsimqoz:
@erewok , thank you so much, it works.

What do you think about feature request? Smth like this:
```
class UploadFile:

    def __init__(
        self, filename: str, file: typing.IO = None, content_type: str = ""
    ) -> None:
        self.filename = filename
        self.content_type = content_type if content_type else mimetypes.guess_type(filename)[0]
        ...
```

erewok:
In these cases we have typically chosen to hew closely to the API offered by Django or Flask. I believe in the Django case (but it's been awhile since I looked at the code) that they fallback to "text/plain" if the content type has not been provided. (There are sometimes surprising security implications in areas such as this.)

Further, this is a simple thing to include in your own handlers where presumably you know a lot more about the types of files your server is expecting?

PR #3283, opened 2026-05-22
Use `"application/octet-stream"` as the `FileResponse` media type fallback

ATOM00blue:
# Summary

`FileResponse` falls back to `text/plain` when the media type cannot be guessed from the filename or path:

```python
if media_type is None:
    media_type = guess_type(filename or path)[0] or "text/plain"
```

For files with an unrecognized extension this means the response is served as `text/plain`, so browsers render the bytes inline as text instead of downloading the file. This was reported in discussion #3106 (an Outlook `.msg` file rendered as text rather than triggering a download).

`application/octet-stream` is the correct default for unknown content per RFC 2046 ("MIME implementations must at a minimum treat any unrecognized subtypes as being equivalent to `application/octet-stream`") and the MDN guidance. It's also what Django and Werkzeug use:

- Django: [`content_type = content_type or "application/octet-stream"`](https://github.com/django/django/blob/main/django/views/static.py)
- Werkzeug `send_file` falls back to `application/octet-stream`

The earlier choice of `text/plain` (#1027) was made on the assumption that Django used `text/plain`, which isn't the case.

This affects `FileResponse` directly and `StaticFiles`, which serves files through `FileResponse`. Passing an explicit `media_type` is unchanged.

# Checklist

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -312,7 +312,7 @@ class FileResponse(Response):
         self.status_code = status_code
         self.filename = filename
         if media_type is None:
-            media_type = guess_type(filename or path)[0] or "text/plain"
+            media_type = guess_type(filename or path)[0] or "application/octet-stream"
         self.media_type = media_type
         self.background = background
         self.init_headers(headers)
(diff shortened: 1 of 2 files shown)


Kludex (approved):
Makes sense.

Thanks. 🙏

PR #3285, opened 2026-05-23
Annotate URLPath protocol parameter with Literal

Chang-LeHung:
- Define an `ASGIProtocol = Literal["http", "websocket", ""]` type alias in `starlette/types.py`.
- Replace the plain `str` annotation on `URLPath.__new__` and `URLPath.__init__`'s `protocol` parameter with `ASGIProtocol`, so static type checkers can catch invalid values at analysis time.
- The existing runtime `assert` is kept as a safety net for dynamic usage.


Kludex:
Where is empty string passed?

Chang-LeHung:
The empty string "" is the default value for the protocol parameter. It's used when callers omit protocol, for example in Mount.url_path_for and Host.url_path_for:


https://github.com/Kludex/starlette/blob/348f86dc92fed6e3c0314154e38deec90a6bcabd/starlette/routing.py#L420-L427

https://github.com/Kludex/starlette/blob/348f86dc92fed6e3c0314154e38deec90a6bcabd/starlette/routing.py#L486-L492

In these cases the protocol is unknown at that point — it gets resolved later when a child route (like Route or WebSocketRoute) provides "http" or "websocket".

Chang-LeHung:
> The empty string "" is the default value for the protocol parameter. It's used when callers omit protocol, for example in Mount.url_path_for and Host.url_path_for:
> 
> https://github.com/Kludex/starlette/blob/348f86dc92fed6e3c0314154e38deec90a6bcabd/starlette/routing.py#L420-L427
> 
> https://github.com/Kludex/starlette/blob/348f86dc92fed6e3c0314154e38deec90a6bcabd/starlette/routing.py#L486-L492
> 
> In these cases the protocol is unknown at that point — it gets resolved later when a child route (like Route or WebSocketRoute) provides "http" or "websocket".

@Kludex

Kludex (approved):
I don't think there's much benefit from this, but I guess it's fine.

Kludex:
Thanks anyway. 🙏

PR #3288, opened 2026-05-23
Auto-detect async backend in TestClient

Zac-HD:
## Summary

Previously `TestClient` defaulted to `backend="asyncio"`, forcing trio users to repeatedly pass `backend="trio"` or build a custom client factory; and anyio users have to figure out backend detection.

This patch teaches `TestClient` itself to figure out which backend is appropriate in most cases, with techniques falling through from `sniffio` to pytest config to `backend_options` inspection to `sys.modules`, ultimately warning if the question remains ambiguous.

I proposed something very similar in agronholm/anyio#1151 last week, which resolved that the feature should be added directly to downstream libraries.  So here we go!

## Checklist

        (it seemed about as easy to open a PR for discussion as for an issue though, so here we are 🙂)

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -95,7 +95,25 @@ client = TestClient(app, client=('localhost', 8000))
 These options are passed to `anyio.start_blocking_portal()`.
 See the [anyio documentation](https://anyio.readthedocs.io/en/stable/basics.html#backend-options)
 for more information about the accepted backend options.
-By default, `asyncio` is used with default options.
+When `backend` is not specified (the default), the active async library
+is auto-detected via [sniffio](https://github.com/python-trio/sniffio),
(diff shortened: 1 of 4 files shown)


Kludex:
> I proposed something very similar in https://github.com/agronholm/anyio/pull/1151 last week, which resolved that the feature should be added directly to downstream libraries. So here we go!

I could be wrong, but that's not the conclusion that I got from what @agronholm said... 🤔

In fact, our test suite has what you want: https://github.com/Kludex/starlette/blob/a4ff83b1aacaf8955e241dd00444c27823be163a/tests/conftest.py#L12-L23

---

Thanks for the PR, but I don't think the added complexity outweighs the quality-of-life benefit for trio users.

agronholm:
I didn't have a chance to look deeper, but I did think a simple solution should already exist to cover this use case.

PR #3291, opened 2026-05-25
Support httpx2 in the test client

Kludex:
`starlette.testclient` now prefers `httpx2` when installed, falling back to `httpx`. The fallback emits a `StarletteDeprecationWarning` urging migration; if neither is installed, the existing `RuntimeError` is raised pointing at `httpx2`.

Non-breaking: users with only `httpx` installed keep working, just warned.

- Adds `StarletteDeprecationWarning(UserWarning)` to `starlette/exceptions.py` (from #3119) - inherits from `UserWarning` so it's visible by default.
- `TYPE_CHECKING` guard keeps mypy typing against `httpx` (httpx2 has no stubs yet).

## AI Disclaimer


Kludex:
Being a bit more transparent about this, instead of just the LLM generated description:

`httpx` has become somehow unmaintained (closing issues, discussions, and no release since 2024), and https://github.com/pydantic/httpx2 is the least annoying path forward for every consumer of that package.

edgarrmondragon:
Should the readme be updated to reflect the preference for httpx2?

PR #3294, opened 2026-05-26
fix: replace assert with ValueError in Response.set_cookie()

Utkarshbarapatre:
Using `assert` for user-facing validation is unsafe because Python disables assertions when run with the `-O` (optimise) flag. This silently skips the `samesite` check and allows an invalid value to be set in the cookie header.

This PR replaces the `assert` statement with an explicit check and a `ValueError`, which is consistent with the validation already used for `partitioned` cookies in the same method.

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -117,11 +117,8 @@ class Response:
         if httponly:
             cookie[key]["httponly"] = True
         if samesite is not None:
-            assert samesite.lower() in [
-                "strict",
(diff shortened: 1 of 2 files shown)


Kludex:
The assertion exists because `samesite` was just `str`, and not a `Literal` before. I don't think we want to raise anything here, but maybe we could drop the assertion.

Anyway, thanks for the PR.

PR #3301, opened 2026-05-29
Guard empty hostname in `URL.replace()`

koriyoshi2041:
A `URL` with no netloc raises `IndexError: string index out of range` from `URL.replace()` when `port`/`username`/`password` is given without a `hostname`:

```python
URL("/path").replace(port=8080)        # IndexError
URL("").replace(port=8080)             # IndexError
URL("http://").replace(username="u")   # IndexError
```

In `replace()`, when no `hostname` kwarg is passed it derives the hostname via `netloc.rpartition("@")`, which yields an empty string for these URLs. The subsequent `hostname[-1] != "]"` check then indexes an empty string and blows up. The trailing-`]` check (IPv6 literal) is only meaningful for a non-empty hostname, so guarding the index is enough:

```python
if hostname and hostname[-1] != "]":
```

Added `test_url_replace_with_empty_netloc`, which fails with `IndexError` before the fix and passes after, covering the relative-path, empty-string and scheme-only cases. This is the same one-line fix as #3154, which was closed for not having a test.

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -123,7 +123,7 @@ class URL:
                 netloc = self.netloc
                 _, _, hostname = netloc.rpartition("@")
 
-                if hostname[-1] != "]":
+                if hostname and hostname[-1] != "]":
                     hostname = hostname.rsplit(":", 1)[0]
 
             netloc = hostname
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -62,6 +62,15 @@ def test_url() -> None:
     assert url.replace(username="u") == URL("http://u@host:80")
 
 
+def test_url_replace_with_empty_netloc() -> None:
+    # URLs with no netloc (e.g. relative URLs) used to raise IndexError in
+    # replace() because the empty hostname was indexed with hostname[-1].
+    assert URL("/path").replace(port=8080) == URL("//:8080/path")
+    assert URL("").replace(port=8080) == URL("//:8080")
+    assert URL("http://").replace(port=8080) == URL("http://:8080")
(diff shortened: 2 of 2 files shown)


Kludex:
What's the real case scenario that you need this to work?

koriyoshi2041:
Honestly, it's an API-robustness fix rather than something a specific production flow forced on me. `URL.replace()` is public, and calling it with `port`/`username`/`password` but no `hostname` on a URL that has no authority component leaks a low-level `IndexError` from internal string indexing:

```python
URL("/path").replace(port=8080)   # IndexError: string index out of range
```

You hit this when building a URL incrementally from a relative one. The expectation for a public method is a sensible result (here `//:8080/path`) or a clear error — not `hostname[-1]` blowing up. The one-char guard only skips the IPv6-bracket check when there's no hostname, which is the one case where that check isn't meaningful anyway.

I won't oversell it — there's no burning real-world scenario, it's an edge case. It's the same fix as #3154, which you closed for lacking a test; this time it's RED/GREEN (`test_url_replace_with_empty_netloc` fails with the IndexError on `main`, passes with the guard). If you'd rather not carry the edge case I'm happy to close it — just wanted to give it a test-backed second look.

Kludex:
@koriyoshi2041 are you OpenClaw, Hermes or a human?

koriyoshi2041:
I'm Rios, an OpenClaw assistant working through this account; not Hermes. If you'd rather not take assisted PRs in Starlette, that's totally fine — I can close this.

Kludex:
That saves me so much time, thanks!

PR #3304, opened 2026-05-30
Use same module (httpx|httpx2) for type checking as for runtime

leifwar:
# Summary

Using different module when type checking (httpx) compared to runtime (httpx2) will cause type
checkers like pyright to fail. This PR solves this issue mentioned in #3302

# Checklist

Use same module (httpx|httpx2) for type checking as for runtime

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -37,7 +37,7 @@ else:  # pragma: no cover
     from typing_extensions import Self
 
 if TYPE_CHECKING:
-    import httpx
+    import httpx2 as httpx
 else:
     try:
         import httpx2 as httpx


Kludex (approved):

Kludex:
Yes, this is what I meant.

alicederyn:
This seems to have lost typing information: the return type of `TestClient.get` (according to mypy at least) is now `Any`, not `httpx._models.Response`. Is there something that needs to be done differently to get the typing information back? If so, can it be added to the release notes?

alicederyn:
It seems like I could add the `starlette[full]` extra to pull in `httpx2` &mdash; is that the recommendation? Ideally I'd have preferred a `starlette[testclient]` extra with just `httpx2`, if that's possible and something you'd be willing to support going forwards.

_edit_ Oh, that's only on `main`, too, not in the released version. I see https://github.com/Kludex/starlette/pull/3323 / https://github.com/Kludex/starlette/discussions/3320 are pertinent, I'll move to the discussion.

PR #3309, opened 2026-05-31
Fix/assert to typeerror

NEFORCEO:
# Summary

Replace `assert` statements with explicit `TypeError` raises in `@requires` decorator (`authentication.py`) and `OpenAPIResponse.render` (`schemas.py`).

`assert` statements are silently removed when Python runs with the `-O` (optimize) flag, turning helpful type mismatch messages into cryptic `AttributeError`s deeper in the call stack. Using `raise TypeError` ensures the check always runs regardless of optimization level.

Changed files:
- `starlette/authentication.py` — 3 places: `websocket_wrapper`, `async_wrapper`, `sync_wrapper`
- `starlette/schemas.py` — `OpenAPIResponse.render`

Added tests:
- `tests/test_authentication.py` — 3 new tests verifying `TypeError` is raised when sync/async request wrapper and websocket wrapper receive wrong argument types
- `tests/test_schemas.py` — 1 new test verifying `TypeError` is raised when `OpenAPIResponse` receives a non-dict

# Checklist


KeeganOP:
related to: https://github.com/Kludex/starlette/pull/3298#issuecomment-4585177417

KeeganOP (approved):

NEFORCEO:
> related to: https://github.com/Kludex/starlette/pull/3298#issuecomment-4585177417

Done. I've changed it  to `raise TypeError` as suggested and added the tests. Please take another look when you have a chance. Thanks!

Kludex:
I'd actually prefer to not have this kind of plumbing for the time being.

Thanks for the PR.

PR #3318, opened 2026-06-04
Refactor testclient imports

waketzheng:
# Summary

Remove a duplicated `import httpx2 as httpx` line.

  This is a small cleanup with no behavior change.

# Checklist

- [x] I've added a test for each change that was introduced. N/A, existing tests cover this cleanup; no new test is needed. I tried as much as possible to make a single atomic change.
- [x] I've updated the documentation accordingly. N/A, this is not a documentation change.


Kludex:
You had to do `# type:ignore[no-redef]`, that's what the current code avoids. Also, I use pyright strict mode locally, and that's more strict that mypy (this repo has - maybe I should switch it).

But thanks for the PR.

PR #3319, opened 2026-06-05
docs: fix TestClient examples and default client host

leestana01:
A few issues in `docs/testclient.md`:

- The two `def test_app()` examples under **Selecting the Async backend** are missing the trailing colon, so the snippets are not valid Python.
- The **Change client address** section states the default client host is `"testserver"`, but `TestClient` defaults `client` to `("testclient", 50000)` (`"testserver"` is the `base_url`). `request.client` is therefore `("testclient", 50000)`, as asserted in `tests/test_testclient.py`. The docs now say `"testclient"`.

Docs only — no behavior change.

### DIFF
--- a/docs/testclient.md
+++ b/docs/testclient.md
@@ -81,7 +81,7 @@ case you should use `client = TestClient(app, raise_server_exceptions=False)`.
 
 ### Change client address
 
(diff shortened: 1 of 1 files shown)


Kludex (approved):

Kludex:
Thanks!

PR #3322, opened 2026-06-07
Adjust testclient typing and warnings

waketzheng:
# Summary

  - Move the `Self` import behind `TYPE_CHECKING` in `testclient`
  - Add clearer warning attribution for `TestClient` timeout deprecation warnings
  - Suppress the underlying `ModuleNotFoundError` chain when `httpx2`/`httpx` is missing
  - Keep `WebSocketTestSession.__enter__` typed as `Self`
  
# Checklist

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -11,14 +11,7 @@ from collections.abc import Awaitable, Callable, Generator, Iterable, Mapping, M
 from concurrent.futures import Future
 from contextlib import AbstractContextManager
 from types import GeneratorType
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Literal,
-    TypedDict,
-    TypeGuard,
-    cast,
-)
+from typing import TYPE_CHECKING, Any, Literal, TypedDict, TypeGuard, cast
 from urllib.parse import unquote, urljoin
 
 import anyio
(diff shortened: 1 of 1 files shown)


starlette/testclient.py
```diff
@@ -454,6 +446,7 @@ def request(  # type: ignore[override]
                 "You should not use the 'timeout' argument with the TestClient. "
                 "See https://github.com/Kludex/starlette/issues/1108 for more information.",
                 DeprecationWarning,
+                stacklevel=2,
```
Kludex, line 451: This makes sense.

starlette/testclient.py
```diff
@@ -49,13 +43,14 @@
                 "The starlette.testclient module requires the httpx2 package to be installed.\n"
                 "You can install this with:\n"
                 "    $ pip install httpx2\n"
-            )
+            ) from None
```
Kludex, line 45: This makes sense.

Kludex (commented):
The rest that I didn't mark as "This makes sense" I don't think makes sense.

waketzheng:
@Kludex Updated.

Kludex (approved):

Kludex:
Thanks!

PR #3336, opened 2026-06-17
Fix URL.include_query_params() reordering existing query parameters

JSap0914:
# Summary

`URL.include_query_params()` silently reorders existing query parameters when a caller updates a key that is not the last one in the query string.

**Root cause:** the implementation delegates to `MultiDict.update()`, which removes the key from its current position and appends the new value at the end.

**Minimal reproduction:**
```python
from starlette.datastructures import URL

u = URL("https://example.org/search?page=2&sort=name&order=asc")
print(u.include_query_params(sort="date").query)
# Expected: page=2&sort=date&order=asc
# Actual:   page=2&order=asc&sort=date   ← sort moved to end
```

This breaks applications that rely on a stable URL shape — e.g. cache keys, cursor-based pagination links, and test assertions against URLs with multiple parameters.

**Fix:** rewrite `include_query_params` to traverse the original key=value list once, replacing the first occurrence of each updated key in place (and dropping any duplicate occurrences for that key), then appending any keys that were not already present. `replace_query_params` and `remove_query_params` are unaffected.

# Checklist

**Verification:**
```
pytest tests/test_datastructures.py -v
# 45 passed
```

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -141,9 +141,21 @@ class URL:
         return self.__class__(components.geturl())
 
     def include_query_params(self, **kwargs: Any) -> URL:
-        params = MultiDict(parse_qsl(self.query, keep_blank_values=True))
-        params.update({str(key): str(value) for key, value in kwargs.items()})
-        query = urlencode(params.multi_items())
+        str_kwargs = {str(key): str(value) for key, value in kwargs.items()}
+        # Rebuild the list in-place so that existing keys retain their position.
+        merged: set[str] = set()
+        new_params: list[tuple[str, str]] = []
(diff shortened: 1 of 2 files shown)


Kludex:
I don't think this is worth it. You can perform that logic in your application if you need to.

Thanks for the PR.

PR #3342, opened 2026-06-24
docs: correct GZipMiddleware compresslevel range to 0-9

patrickwehbe:
The GZip middleware docs say `compresslevel` is "an integer ranging from 1 to 9". `GZipMiddleware` passes the value straight into `gzip.GzipFile(..., compresslevel=compresslevel)`, which accepts 0 through 9, where 0 means no compression. So 0 is a valid level the current docs exclude.

This updates the range to 0 to 9. Docs-only change.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3342?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -240,7 +240,7 @@ app = Starlette(routes=routes, middleware=middleware)
 The following arguments are supported:
 
 * `minimum_size` - Do not GZip responses that are smaller than this minimum size in bytes. Defaults to `500`.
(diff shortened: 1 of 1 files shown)


Kludex:
But if you are going to use 0, then don't use it... Right? 😅

PR #3343, opened 2026-06-25
feat: add `TestClient.query()` helper for the HTTP QUERY method

ItamarShalev:
## Summary

Add a `query()` convenience method to `TestClient`, mirroring the existing body-carrying helpers (`post`, `put`, `patch`). `QUERY` is a *new* HTTP method standardized in [RFC 10008 - "The HTTP QUERY Method"](https://datatracker.ietf.org/doc/rfc10008/) (June 2026): a safe, idempotent method that carries a request body. `httpx` exposes no dedicated helper for it, so `query()` dispatches through the generic `request()` API with method `"QUERY"`.

## Why

Starlette already routes the `QUERY` method (the router has no method whitelist), so this is purely an ergonomic test-client addition - `client.query("/path", json=...)` instead of `client.request("QUERY", "/path", json=...)`. It complements first-class `QUERY` support being added in FastAPI.

## Tests

- `test_query_method`: a `QUERY` request with a JSON body is routed and the body is read back.
- `test_query_method_not_allowed`: a non-`QUERY` request to a `QUERY`-only route returns `405` with `QUERY` in the `Allow` header.

Both run on the asyncio and trio backends. `ruff check`/`format` pass and the full `tests/test_testclient.py` suite is green.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -629,6 +629,41 @@ class TestClient(httpx.Client):
             extensions=extensions,
         )
 
+    def query(
+        self,
+        url: httpx._types.URLTypes,
+        *,
+        content: httpx._types.RequestContent | None = None,
+        data: _RequestData | None = None,
+        files: httpx._types.RequestFiles | None = None,
+        json: Any = None,
+        params: httpx._types.QueryParamTypes | None = None,
+        headers: httpx._types.HeaderTypes | None = None,
+        cookies: httpx._types.CookieTypes | None = None,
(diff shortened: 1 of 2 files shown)


Kludex:
That's actually already supported if you have the latest httpx2 installed.

PR #3346, opened 2026-06-28
ServerErrorMiddleware reimplemented using TraceRite.

LeoVasanko:
# Summary

Implemented cleaner traceback rendering with the ServerErrorMiddleware completely rewritten. Maintains compatibility with the old one, but uses the lightweight plain-Python tracerite module to do the formatting.

In addition to HTML and text output, this also offers optional JSON output (disabled by default to maintain earlier functionality and because that might still be preferred especially with curl).

The motivation to this is that [TraceRite](https://github.com/sanic-org/tracerite) makes error messages to the point, short and easy to read with just the relevant information visible. Refer to its page for further information and examples. The module is maintained by Sanic and has been well battle tested.

I left out console pretty exception logging, also supported by the module, to keep the changes minimal. It may be separately loaded if needed.

# Checklist

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3346?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ classifiers = [
 ]
 dependencies = [
     "anyio>=3.6.2,<5",
+    "tracerite>=2.3.1",
     "typing_extensions>=4.10.0; python_version < '3.13'",
 ]
 
--- a/starlette/middleware/errors.py
+++ b/starlette/middleware/errors.py
@@ -96,16 +96,9 @@ class ServerErrorMiddleware:
         summary = build_chain_header(chain)
         doc = Document(summary, lang="en")
         font_stack = (
-            "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', "
-            "Roboto, 'Helvetica Neue', Arial, sans-serif"
+            "system-ui, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif"
         )
-        doc.style(
-            "body {"
-            f"    font-family: {font_stack};"
-            "    line-height: 1.4;"
-            "    margin: 1.5rem;"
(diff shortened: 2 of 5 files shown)


LeoVasanko:
image
TraceRite on the left, the current implementation on the right.

LeoVasanko:
image
Text output with curl.

Kludex:
We can make this page cuter, but I don't want to add another dependency for this.

LeoVasanko:
I ask to consider reopening. The dependency is very light and does something that nobody else is doing. It is NOT styling, but serious data processing that it does.

LeoVasanko:
The processing includes:
- Processing of the whole exception chain, as opposed to only showing the most recent exception
- Finding out where in user code the problem is, showing relevant frames of each stack frame
- Of each frame, showing relevant lines, not just some fixed number of lines
- Cursor position highlights on that snippet (py 3.11+, when available)
- Variable inspector, again with serious pretty printing and filtering (only showing relevant things, not all locals)

And my favorite:
- Reordering reported stack frames over the chain in chronological order (built-in tracebacks do NOT follow order of execution and don't mark returning to `except` specially to see what is going on).

This is not something you can reproduce without years of work. And still, it is a small package with reduction of code in Starlette itself.

But I do see your point in that if the package is installed with no extras, it only depends on anyio and idna (albeit both themselves heavier than the suggested addition), and that exception formatting is an optional feature that not everybody might need.

Issue #3357, opened 2026-07-03
TrustedHostMiddleware Doesn't handle IPv6 properly


gopackgo90:
- Initially raised as discussion #3321 

Say I have the below Starlette app,
```python
# main.py
from starlette.applications import Starlette
from starlette.middleware import Middleware
from starlette.middleware.trustedhost import TrustedHostMiddleware
from starlette.responses import JSONResponse
from starlette.routing import Route


async def homepage(request):
    return JSONResponse({'hello': 'world'})

routes = [
    Route("/", endpoint=homepage)
]

middleware = [
    Middleware(TrustedHostMiddleware, allowed_hosts=["[::1]"])
]

app = Starlette(debug=True, routes=routes, middleware=middleware)
```

and run it with `uvicorn main:app --host '::1'`. When I go to http://[::1]:8000 in my browser, I get the "Invalid host header" response. This appears to be because Starlette is thinking the host is `"["` based on the splitting logic in the middleware, https://github.com/Kludex/starlette/blob/70a3bd4276712f0ee11b161c8b14225a84316df1/starlette/middleware/trustedhost.py#L40.
When I remove the `TrustedHostMiddleware`, I get the hello world response as expected.

gopackgo90:
I think the PR I opened, https://github.com/Kludex/starlette/pull/3358, should be preferred as it’s implementation has already been well-tested in Django.

aryansk:
Working on a fix for the IPv6 bracket handling in TrustedHostMiddleware (host extraction for `[::1]:8000`). Will open a PR against `main` shortly — PR #3444 already covers this with the minimal bracket-aware split fix.

Gambit-Checkmate:
Hi, I’d like to investigate the IPv6 handling in TrustedHostMiddleware. I can add a minimal IPv6 host test, compare it with the duplicate report, and keep the fix limited to host normalization. Is this issue still open for community work?

Mukller:
I ran all six open PRs for this issue locally (CPython 3.13 / Windows) against the reported repro plus a probe matrix of malformed Host headers, so the differences are easier to see than from six parallel diffs.

Every branch fixes the reported case (`Host: [::1]:8000` with `allowed_hosts=["[::1]"]`). They diverge on malformed input:

| PR | host extraction | `[::1]junk` | `[::1]:abc` | `[::1` (unclosed) | fixes TestClient netloc too | targeted tests |
|---|---|---|---|---|---|---|
[...]

PR #3368, opened 2026-07-06
Fix TestClient IPv6 base_url parsing

krrishapatel:
Related to #3357 (same class of bug — `:` splitting breaks IPv6).

---

`TestClient` with an IPv6 `base_url` like `http://[::1]:8000` crashes with `ValueError: invalid literal for int() with base 10: ':1]'` because the netloc is split on `:` without accounting for bracket notation.

This adds IPv6-aware parsing that extracts the bare address and port from bracketed notation before falling through to the existing split logic.

Assisted-by: Claude Opus 4.6

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/testclient.py
+++ b/starlette/testclient.py
@@ -231,7 +231,15 @@ class _TestClientTransport(httpx.BaseTransport):
 
         default_port = {"http": 80, "ws": 80, "https": 443, "wss": 443}[scheme]
 
-        if ":" in netloc:
+        if netloc.startswith("["):
+            # IPv6: "[::1]" or "[::1]:port"
+            bracket_end = netloc.index("]")
+            host = netloc[1:bracket_end]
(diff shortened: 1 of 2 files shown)


Kludex:
Thanks for the contribution. Closing in favor of #3360, which now includes the TestClient IPv6 fix along with the corresponding TrustedHostMiddleware fix. #3360 supersedes this PR.

PR #3370, opened 2026-07-06
Add support for the QUERY HTTP method

Kfir-G:
# Summary

Adds support for the `QUERY` HTTP method (draft spec: https://http.dev/query). Follow-up to the discussion here: [link to your discussion].

Routing itself already worked with no changes, since `Route`/`Router` don't hardcode a fixed list of HTTP verbs - `Route("/search", search, methods=["QUERY"])` already works today. The actual changes needed were small:

- `starlette/endpoints.py` - added `"QUERY"` to the verbs `HTTPEndpoint` checks when building `_allowed_methods`, so a `.query()` handler on a class-based endpoint is recognized (and shows up correctly in the `Allow` header on 405 responses).
- `starlette/schemas.py` - added `"query"` to the method list `SchemaGenerator` checks on class-based endpoints, so `.query()` handlers are picked up in generated OpenAPI schemas.
- `docs/routing.md` and `docs/endpoints.md` - short additions showing `QUERY` as a supported method.

This is purely additive - `QUERY` is only reachable if a route explicitly opts in via `methods=[...]`, same as any other verb, so it shouldn't interact with the recent `HTTPEndpoint` dispatch advisory (GHSA-x746-7m8f-x49c).

Ran `scripts/check` and `scripts/test` locally - all passing, 100% coverage maintained.

This is my first PR here, so happy to adjust based on feedback or conventions I'm missing.

# Checklist

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


marcp54:
TestClient needs a query function that follows the same pattern as the other methods and  inherits from Httpx2 Client which support for query was added a day ago. 

- Release: https://github.com/pydantic/httpx2/releases#release-v2.6.0
- PR: https://github.com/pydantic/httpx2/pull/1055

However the issue, starlette's `pyproject.toml` restricts uv new builds up to 7 days. 
https://github.com/Kludex/starlette/blob/5174d4c8358a6f06aa8056bafd14c2272dab8dd1/pyproject.toml#L73

I don't think the maintainers will want to reduce the restriction so this might sit for a few days until that window expires.

Mukller (approved):
Verified locally on the PR branch (Python 3.13): `pytest tests/test_endpoints.py tests/test_routing.py tests/test_schemas.py` — **113/113 green**, including the new QUERY dispatch test and the updated `allow: GET, QUERY` header assertion.

Checked the substance beyond tests:

1. The addition to `_allowed_methods` in `HTTPEndpoint` is gated by handler presence (`getattr(self, method.lower(), None)`), so classes without a `query` method are unaffected — no behavior change for existing code.
2. `schemas.BaseSchema` generation now includes `query` alongside the others, keeping OpenAPI output consistent with routable methods.
3. Method-negotation edge cases still hold: 405 responses list QUERY correctly, and non-verb pseudo-methods are not dispatched (existing test covers that).
4. Docs additions (routing `methods=["QUERY"]`, endpoint class example) match the implementation exactly.

Context for maintainers: HTTP QUERY (safe method with a body) is currently a draft (draft-ietf-httpbis-safe-method-w-body), but since starlette routes already accept arbitrary method strings, this change only makes class-based endpoints and schema generation consistent with what function-based routes could always do. Low risk, coherent docs.

Nothing to fix from my side.

Kludex:
Superseded by [#3489](https://github.com/Kludex/starlette/pull/3489). The replacement preserves the original commits and adds the review fixes on top.

PR #3376, opened 2026-07-13
Support `partitioned` in `Response.delete_cookie`

sungbin1015:
## Summary

`Response.set_cookie` accepts a `partitioned` flag, but its sibling `delete_cookie` does not. Deleting a partitioned ([CHIPS](https://developer.mozilla.org/en-US/docs/Web/Privacy/Privacy_sandbox/Partitioned_cookies)) cookie requires the clearing `Set-Cookie` to also carry the `Partitioned` attribute — otherwise the browser will not match and remove the cookie.

This adds `partitioned: bool = False` to `delete_cookie`, forwarding it to `set_cookie` alongside the other attributes so the two methods stay in parity. The existing Python 3.14+ gating lives in `set_cookie`, so no new gating is needed here.

## Changes

- `starlette/responses.py` — add `partitioned` param to `delete_cookie`, forward to `set_cookie`.
- `tests/test_responses.py` — add `test_delete_cookie_partitioned` (version-safe: asserts `Partitioned` on 3.14+, absent otherwise, mirroring `test_set_cookie`).
- `docs/responses.md` — update the `delete_cookie` signature (it was stale, also missing `secure`/`httponly`/`samesite`).

## Testing

```
$ pytest tests/test_responses.py -k cookie
18 passed

$ ruff format --check starlette tests   # already formatted
$ ruff check starlette tests            # All checks passed
$ mypy starlette                        # no issues
```

Local interpreter is Python 3.13, so the `< 3.14` path is exercised locally; the `>= 3.14` path is covered by the CI matrix.

---
*This PR was prepared with AI assistance (Claude); every changed line was reviewed and the tests above were run locally.*

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3376?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sungbin1015:
Thanks for the review!

**P3 (docs):** Fixed in 6beb4d9 — added the `partitioned` description (including the Python 3.14+ constraint) to the `delete_cookie` docs, mirroring `set_cookie`.

**P1 (coerce `secure` when `partitioned=True`):** I would prefer to leave this as-is. `delete_cookie` intentionally forwards every argument straight to `set_cookie`, and `set_cookie(partitioned=True)` does **not** coerce `secure` either — the whole purpose of this PR is behavioral parity between the two sibling methods. Starlette also already follows this "pass the attributes through faithfully" convention elsewhere (e.g. it does not force `secure` for `samesite=\"none\"`, which browsers likewise require). Auto-setting `secure` in `delete_cookie` only would make the two methods diverge. If forcing `secure` for partitioned cookies is desired, I think that belongs in a separate change that applies to `set_cookie` too — happy to open a follow-up if you would like.

> 🤖 This PR was prepared with AI assistance (Claude Code); all changes were reviewed by me.

Samielakkad (commented):
Retracted.

tests/test_responses.py
```diff
@@ -512,6 +512,22 @@ async def app(scope: Scope, receive: Receive, send: Send) -> None:
...
+            "mycookie",
+            partitioned=True if sys.version_info >= (3, 14) else False,
+        )
+        await response(scope, receive, send)
+
+    partitioned_text = "Partitioned" if sys.version_info >= (3, 14) else ""
+
+    client = test_client_factory(app)
+    response = client.get("/")
+    assert partitioned_text in response.headers["set-cookie"]
```
Sanjays2402, line 528: on <3.14 this reduces to `assert "" in response.headers["set-cookie"]`, which passes no matter what delete_cookie does. maybe skipif the whole test on 3.14 and assert the attribute is actually absent otherwise, so the two paths both mean something.

Mukller (approved):

**Tests**: `pytest tests/test_responses.py -k cookie` — 18/18 green.

**Docs claim checked empirically** — the new line *"Only available for Python 3.14+, otherwise an error will be raised"* is accurate:

```python
Response("x").delete_cookie("mycookie", partitioned=True)
# ValueError: Partitioned cookies are only supported in Python 3.14 and above.
# (raised by set_cookie's existing sys.version_info guard)

Response("x").delete_cookie("mycookie")
# mycookie=""; expires=...; Max-Age=0; Path=/; SameSite=lax   (unchanged)
```

One point in favor that's worth stating explicitly: forwarding `partitioned` through `delete_cookie` is not just API symmetry — per CHIPS/RFC6265bis handling, a user agent stores the `Partitioned` attribute as part of the cookie's identity, so a deletion (`Set-Cookie` with `Max-Age=0`) that omits `Partitioned` will not match a previously stored partitioned cookie. So for cookies set with `partitioned=True`, deletion must echo the attribute — which is exactly what this enables. Good catch.

No issues found.

Issue #823, opened 2020-02-09
Mount requires trailing '/', while Route does not.


jwtrhs:
Mounted subroutes require the trailing `/` to hit their 'root' subroute, whereas regular `Route`s do not.

E.g.

```python
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import PlainTextResponse
from starlette.routing import Mount, Route

async def homepage(request):
    return PlainTextResponse('homepage')

async def subroute(request):
    return PlainTextResponse('subroute')

routes = [
    Route('/', homepage),
    Mount('/sub', routes=[Route('/', subroute)]),
]

app = Starlette(debug=True, routes=routes)
```

`localhost:8000` and `localhost:8000/` both return "hompage", whereas `localhost:8000/sub` returns a 307 temp redirect to `localhost:8000/sub/` which then returns "subroute".

Small issue and not saying either approach is correct, but I guess I would expect both `Mount` and `Route` to behave the same here.

Using version 13.1.

lovelydinosaur:
> Mounted subroutes require the trailing / to hit their 'root' subroute, whereas regular Routes do not.

Indeed. That's intentional, because a mount point should always be in the form `/path-to-mount/{...}`, so eg all the following are valid...

* `/{...}`
* `/some-path/{...}`
* `/some/path/{...}`

Wheras `/some-path` isn't, because we wouldn't want to match `/some-path{...}`

(Also The HTTP spec mandates that the empty `` path should be treated as equivelent to `/`.)

jwtrhs:
Ah I see, thanks!

So that would mean these two routes would operate a little differently, correct?

```python
routes = [
    Route('/', homepage),
    Mount('/sub', routes=[Route('/', subroute)]),
]

routes = [
    Route('/', homepage),
    Route('/sub', subroute),
]
```

Really I'm just trying to structure my routes/mounts so that they all treat trailing slashes the same. I think my solution is to put the trailing `/` in all my route paths.

```python
routes = [
    Route('/', homepage),
    Route('/sub1/', subroute),
    Mount('/sub2', routes=[Route('/', subroute)]),
]
```

and3rson:
Same issue here: we have REST & WebSocket routes mixed, and since WebSocket app needs to be mounted, we end up having to add slashes at the end of some routes, which is very confusing and error-prone. Example:

```py
def handle_rest(request):
    ...

def handle_websocket(scope, receive, send):
    ...

routes=[
[...]

PR #3377, opened 2026-07-15
fix(routing): normalize trailing slashes on mounted routes

### LINKED ISSUE #869: Handling trailing slashes properly with mounted routes
Hello there!

In the documentation, there is [this example about large applications](https://www.starlette.io/routing/#submounting-routes). If we make it a little more complete, we can get something like this:

```python
import uvicorn


async def homepage(request):
    return JSONResponse(status_code=200)


async def users(request):
    return JSONResponse([{"username": "bob"}, {"username": "alice"}])


async def user(request):
    username = request.path_params["username"]
    if username == "bob":
        return JSONResponse({"username": "bob"})
    if username == "alice":
        return JSONResponse({"username": "alice"})
    return JSONResponse(status_code=404)


routes = [
    Route('/', homepage),
    Mount('/users', routes=[
        Route('/{username}', user),
    ])
]

uvicorn.run(app)
```

But the problem with this example is that all routes are behaving differently:
* `curl localhost:8000` answers as expected
* `curl localhost:8000/` answers as expected
* `curl localhost:8000/users` emits a 307
* `curl localhost:8000/users/` answers as expected
* `curl localhost:8000/users/bob` answers as expected
* `curl localhost:8000/users/bob/` emits a 307

So one route answers as expected whether you give it a trailing slash or not, another one only answers directly if you give it the trailing slash, and another one only answers directly if you do **not** give it the trailing slash.

Of course, the fact that the homepage is answering with or without slash is only due to how HTTP works, because you can't send a GET without a route, and all HTTP clients will convert this to `GET /`.

At this point I'm just paraphrasing #823, where the discussion ended with: 

> That's intentional, because a mount point should always be in the form `/path-to-mount/{...}`

Alright, let's take that for a fact, but then... what if I want all my mounted routes to answer without a trailing slash? This is particularly important for API development, where you want to provide consistent endpoints.

I started by trying to give an empty route, i.e. `Route('', users, methods=['GET', 'POST'])`, but that simply does not work: `AssertionError: Routed paths must start with '/'`.

Reconsidering the answer in #823, I then tried to mount on `/` directly, i.e. doing this, thinking that it would work around my problem:

```python
routes = [
    Route('/', homepage),
    Mount('/', routes=[
        Route('/users', users, methods=['GET', 'POST']),
        Route('/users/{username}', user),
    ])
]
```
That seems weird here, because you could simply add the routes at the top level, but in my real use-case, I'm using routers that I import and mount. And that works great... until you add another `Mount`, that will never be resolved, because the routing stops at the first mount point with a matching prefix, as explained in https://github.com/encode/starlette/issues/380#issuecomment-462257965.

Then I saw https://github.com/encode/starlette/issues/633#issuecomment-530595163, and tried to add `/?` everywhere, just like this:

```python
routes = [
    Route('/', homepage),
    Mount('/users', routes=[
        Route('/?', users, methods=['GET', 'POST']),
        Route('/{username}/?', user),
    ])
]
```

This is not user-friendly at all, but it helped on the `/users/{username}` route, which now answers as expected whether you give it a trailing slash or not. Alas, the `/users` route still emits a 307. So basically, at this point, I just don't know how to handle this, and I'm pretty frustrated by spending so much time on such a trivial thing.

What I ended up doing, since I'm using routers, is defining the whole path in my routes, with a big hack at the application level:

```python
routes = (
    users_router.routes,
    + other_router.routes
)
```

If there isn't something obvious that I missed and would resolve my problem in a simpler way, please consider one or more of the following (by my preference order):
* Making all routes accessible without redirection with or without the trailing slash
* "Denormalizing" all routes when they're mounted, so that you only have a list of all paths and do not stop on the first mount point that matches (basically what my hack is doing)
* Making the developer in complete control of trailing slashes by not emitting any implicit 307
* Having this behavior configurable and documented
* Handling empty routes
* Handling `/?`

Sorry for the long text, I couldn't find a simpler way to expose the whole problem and thinking behind it.

> [!IMPORTANT]

<a href="https://polar.sh/encode/starlette/issues/869">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/encode/starlette/issues/869/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

pctablet505:
Fixes #869

`Mount('/users', routes=[Route('/', users), Route('/{username}', user)])` currently only matches when the request path ends with a slash. `GET /users` returns a 307 redirect to `/users/`, and `GET /users/bob/` returns a 307 redirect to `/users/bob`, while a top-level `Route('/', ...)` responds to both `/` and an empty path. This asymmetry between `Route` and `Mount` has been reported repeatedly: external webhook callers that don't follow redirects fail, Ariadne's GraphQL mount is unreachable at `/graphql`, and people have resorted to setting `app.router.redirect_slashes = False` or adding middleware to rewrite the path.

In #823 the maintainer explained that requiring the slash was intentional, because a mount point is meant to be `/path-to-mount/{...}` and `/some-path` should not match `/some-path{...}`. That constraint is still respected here: `/usersa` still 404s.

This change rewrites the catch-all `path` regex that `Mount` builds from `compile_path(self.path + "/{path:path}")` so the trailing slash is optional. When the mount point is hit without a slash, the router appends the slash to `scope["path"]` internally before dispatching to the mounted app, so `Route('/', ...)` resolves directly and `Route('/{username}', ...)` works with or without a trailing slash. The existing `test_mount_urls` assertion is updated to expect `/users` to stay `/users` instead of redirecting to `/users/`, and a new `test_mount_trailing_slash_with_child_routes` covers the child-route cases.

The main trade-off is that `/users` no longer redirects to `/users/`; it now serves the mounted content directly. That removes the 307 pain for mounts but changes behavior for anyone relying on the old redirect. Other `Route` redirects and `redirect_slashes` behavior are unaffected.

Verified by running `pytest tests/test_routing.py`.

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -379,6 +379,16 @@ class Mount(BaseRoute):
                 self.app = cls(self.app, *args, **kwargs)
         self.name = name
         self.path_regex, self.path_format, self.param_convertors = compile_path(self.path + "/{path:path}")
+        if self.path:
+            # A mount path of `/users` should match both `/users` and `/users/`.
+            # Rewrite the catch-all `path` parameter so that the trailing slash
+            # is optional, while `/usersa` still does not match.
+            path_regex = self.path_regex.pattern
+            path_convertor_regex = self.param_convertors["path"].regex
+            suffix = f"/(?P<path>{path_convertor_regex})$"
+            if path_regex.endswith(suffix):
+                path_regex = path_regex[: -len(suffix)] + f"(?P<path>(?:/.*)?)$"
+                self.path_regex = re.compile(path_regex)
 
     @property
     def routes(self) -> list[BaseRoute]:
@@ -394,8 +404,17 @@ class Mount(BaseRoute):
                 matched_params = match.groupdict()
                 for key, value in matched_params.items():
                     matched_params[key] = self.param_convertors[key].convert(value)
-                remaining_path = "/" + matched_params.pop("path")
-                matched_path = route_path[: -len(remaining_path)]
+                path_value = matched_params.pop("path")
+                if not route_path.endswith("/") and path_value in (None, ""):
+                    # The request path matched the mount point without a trailing
+                    # slash (e.g. `/users`). Route it the same as `/users/` so
+                    # child routes like `Route("/", ...)` resolve correctly.
+                    matched_path = route_path
+                    scope["path"] = scope["path"] + "/"
+                    remaining_path = "/"
+                else:
+                    remaining_path = path_value if path_value else "/"
+                    matched_path = route_path[: -len(remaining_path)]
                 path_params = dict(scope.get("path_params", {}))
                 path_params.update(matched_params)
                 child_scope = {
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -387,7 +387,7 @@ class Mount(BaseRoute):
             path_convertor_regex = self.param_convertors["path"].regex
             suffix = f"/(?P<path>{path_convertor_regex})$"
             if path_regex.endswith(suffix):
-                path_regex = path_regex[: -len(suffix)] + f"(?P<path>(?:/.*)?)$"
+                path_regex = path_regex[: -len(suffix)] + r"(?P<path>(?:/.*)?)$"
                 self.path_regex = re.compile(path_regex)
 
     @property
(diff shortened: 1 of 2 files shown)


pctablet505:
Closing this for now — I want to rethink the approach before taking up review time here. Sorry for the noise.

PR #3381, opened 2026-07-20
Fix KeyError on empty text frame in WebSocketEndpoint with default encoding

Kropiunig:
## Summary

`WebSocketEndpoint.decode` raises `KeyError: 'bytes'` when a client sends an **empty text frame** and the endpoint uses the default `encoding = None`.

## Reproduction

```python
from starlette.endpoints import WebSocketEndpoint
from starlette.routing import WebSocketRoute

class Echo(WebSocketEndpoint):
    encoding = None  # default

    async def on_receive(self, websocket, data):
        await websocket.send_text(f"got:{data!r}")

client = TestClient(Starlette(routes=[WebSocketRoute("/ws", Echo)]))
with client.websocket_connect("/ws") as ws:
    ws.send_text("")          # empty text frame
    print(ws.receive_text())  # KeyError: 'bytes'
```

- **Actual:** `KeyError: 'bytes'` inside `decode`; `dispatch` catches it, closes with code `1011`, and re-raises. `on_receive` is never called.
- **Expected:** `on_receive` receives `""`.

## Root cause

```python
return message["text"] if message.get("text") else message["bytes"]
```

An empty text frame is `{"type": "websocket.receive", "text": ""}` with no `"bytes"` key. The truthiness check treats the valid empty string `""` as falsy, so it falls through to `message["bytes"]` and raises `KeyError`.

## Fix

Use `is not None`, matching the check already used in the `encoding == "json"` branch a few lines above:

```python
return message["text"] if message.get("text") is not None else message["bytes"]
```

This only changes the `text == ""` case (from `KeyError` to returning `""`); empty binary frames and all other cases are unchanged.

## Test

Added `test_websocket_endpoint_on_default_empty_text` in `tests/test_endpoints.py`, mirroring the existing `test_websocket_endpoint_on_default`. It fails before the change (`KeyError: 'bytes'`) and passes after. `scripts/check` (ruff format, ruff check, mypy) passes locally.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3381?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -90,13 +90,13 @@ class WebSocketEndpoint:
 
     async def decode(self, websocket: WebSocket, message: Message) -> Any:
         if self.encoding == "text":
-            if "text" not in message:
+            if message.get("text") is None:
                 await websocket.close(code=status.WS_1003_UNSUPPORTED_DATA)
                 raise RuntimeError("Expected text websocket messages, but got bytes")
             return message["text"]
 
         elif self.encoding == "bytes":
-            if "bytes" not in message:
+            if message.get("bytes") is None:
                 await websocket.close(code=status.WS_1003_UNSUPPORTED_DATA)
                 raise RuntimeError("Expected bytes websocket messages, but got text")
             return message["bytes"]
--- a/starlette/endpoints.py
+++ b/starlette/endpoints.py
@@ -114,7 +114,7 @@ class WebSocketEndpoint:
                 raise RuntimeError("Malformed JSON data received.")
 
         assert self.encoding is None, f"Unsupported 'encoding' attribute {self.encoding}"
-        return message["text"] if message.get("text") else message["bytes"]
+        return message["text"] if message.get("text") is not None else message["bytes"]
 
     async def on_connect(self, websocket: WebSocket) -> None:
         """Override to handle an incoming websocket connection"""
--- a/tests/test_endpoints.py
+++ b/tests/test_endpoints.py
@@ -165,6 +165,22 @@ def test_websocket_endpoint_on_default(test_client_factory: TestClientFactory) -
         assert _text == "Message text was: Hello, world!"
 
 
+def test_websocket_endpoint_on_default_empty_text(
+    test_client_factory: TestClientFactory,
+) -> None:
+    class WebSocketApp(WebSocketEndpoint):
+        encoding = None
+
+        async def on_receive(self, websocket: WebSocket, data: str) -> None:
+            await websocket.send_text(f"Message text was: {data!r}")
+
+    client = test_client_factory(WebSocketApp)
+    with client.websocket_connect("/ws") as websocket:
+        websocket.send_text("")
+        _text = websocket.receive_text()
+        assert _text == "Message text was: ''"
+
+
 def test_websocket_endpoint_on_disconnect(
     test_client_factory: TestClientFactory,
 ) -> None:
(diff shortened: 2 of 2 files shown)


Sanjays2402:
the two branches above have the same shape of problem: `if "text" not in message` only works for servers that omit the unused key. hypercorn (and the ASGI spec example) always send both, with one set to None, so an `encoding = "text"` endpoint receiving a binary frame passes that check and returns None instead of closing with 1003. worth switching those to `message.get("text") is not None` in the same pass?

Kropiunig:
Good catch, you're right. Verified against the spec and the two servers:

- ASGI spec, `websocket.receive`: `bytes`/`text` are "Optional; if missing, it is equivalent to `None`" and "Exactly one of `bytes` or `text` must be non-`None`. One or both keys may be present, however."
- hypercorn `ws_stream.py` builds every receive message with both keys — `{"bytes": ... if BytesIO else None, "text": ... if StringIO else None}`.
- uvicorn (`websockets_impl`, `wsproto_impl`, `websockets_sansio_impl`) omits the unused key, which is why the existing tests never caught it.

So on hypercorn an `encoding = "text"` endpoint receiving a binary frame passed `"text" not in message`, and `decode` returned `None` to `on_receive` instead of closing with 1003. Same for the `bytes` branch.

Switched both guards to `message.get(...) is None`, matching the `json` and default branches:

```python
if self.encoding == "text":
    if message.get("text") is None:
        ...

elif self.encoding == "bytes":
    if message.get("bytes") is None:
        ...
```

Empty frames still decode correctly, since `""` and `b""` are not `None`.

Added `test_websocket_endpoint_encoding_mismatch_with_explicit_none`, parametrized over both encodings. The `TestClient` omits the unused key, so it can't produce this message shape — the test drives the endpoint as a raw ASGI app with a hypercorn-shaped `websocket.receive` and asserts both the `RuntimeError` and that `websocket.close` was sent with 1003. It fails on the previous commit with `DID NOT RAISE RuntimeError` and passes now.

`tests/test_endpoints.py`, `test_websockets.py`, `test_routing.py`, `test_testclient.py`: 239 passed. `ruff format`/`ruff check` clean; `mypy` clean apart from a pre-existing Windows-only `os.pathconf` error in `test_staticfiles.py`. The staticfiles/config failures in my local full run are pre-existing symlink-privilege and path issues on Windows, unrelated to this change.

Mukller (approved):

**Before/after with an empty text frame** (`WebSocketRoute` + `encoding = None`, `TestClient`, instrumented to capture the raw ASGI message):

```
message received: {'type': 'websocket.receive', 'text': ''}

main (before):   KeyError: 'bytes'
                 at endpoints.py:117 -> message["text"] if message.get("text") else message["bytes"]
branch (after):  on_receive gets '' and echoes fine
```

The falsy-`""` trap in the default branch was the real bug; the `is None` checks also make the `"text"`/`"bytes"` branches robust against explicit `None` values. `tests/test_endpoints.py`: 28/28 green including the new empty-text and parametrized mismatch cases.

**Heads-up for maintainers**: PR #3372 contains a byte-for-byte identical change to `endpoints.py` (same three hunks, near-identical tests). Both fix the same bug correctly; whichever you merge, the other can be closed as duplicate. My verification applies equally to both.

PR #3382, opened 2026-07-21
fix(websockets): raise WebSocketDisconnect after close on send

Solaris-star:
# Summary

After a WebSocket has already been closed / disconnected, further `send`/`close` used to raise `RuntimeError`. Per discussion #2762 / issue #2766, raise `WebSocketDisconnect` instead so concurrent close/send paths can treat this as a normal disconnect.

Updated `test_duplicate_close` accordingly.

# Checklist

Fixes #2766

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -95,7 +95,9 @@ class WebSocket(HTTPConnection[StateT]):
                 self.application_state = WebSocketState.DISCONNECTED
             await self._send(message)
         else:
-            raise RuntimeError('Cannot call "send" once a close message has been sent.')
+            # Already disconnected: raise WebSocketDisconnect so concurrent
+            # close/send paths can handle this like a normal disconnect (#2766).
+            raise WebSocketDisconnect()
 
     async def accept(
         self,
--- a/starlette/websockets.py
+++ b/starlette/websockets.py
@@ -15,6 +15,7 @@ class WebSocketState(enum.Enum):
     CONNECTED = 1
     DISCONNECTED = 2
     RESPONSE = 3
+    RESPONSE_COMPLETE = 4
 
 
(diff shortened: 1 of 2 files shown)


starlette/websockets.py
```diff
@@ -95,7 +95,9 @@ async def send(self, message: Message) -> None:
                 self.application_state = WebSocketState.DISCONNECTED
             await self._send(message)
         else:
-            raise RuntimeError('Cannot call "send" once a close message has been sent.')
+            # Already disconnected: raise WebSocketDisconnect so concurrent
+            # close/send paths can handle this like a normal disconnect (#2766).
+            raise WebSocketDisconnect()
```
Sanjays2402, line 103: This branch is also reached after `send_denial_response()` finishes its final `websocket.http.response.body`, so this changes a second send after a completed HTTP denial into `WebSocketDisconnect` too. That path never established a WebSocket connection; should the state distinguish a completed denial response from a sent close?

Solaris-star: Good catch. I pushed `77d0200` with a distinct `RESPONSE_COMPLETE` application state. A completed HTTP denial response now retains the original `RuntimeError` on a later send, while an actual WebSocket close still raises `WebSocketDisconnect`. Added a regression for the denial path; the full websocket test module passes (81 tests) with ruff/format clean.

PR #3396, opened 2026-07-28
Preserve messages polled by Request.is_disconnected()

davidpavlovschi:
# Summary

`Request.is_disconnected()` polls the ASGI receive channel inside a pre-cancelled scope. If the server already buffered a message, `receive()` can return without suspending, so the poll consumes a real `http.request` or `http.disconnect` event.

Starlette currently drops that event. Dropping a request event truncates the body or makes a later body read wait forever. Dropping the disconnect event can also leave a later body read waiting after the client has gone away.

This keeps polled events in a small FIFO buffer and makes `Request.stream()` drain that buffer before reading from the receive channel again. The body and disconnect events therefore reach the stream in their original order.

This follows up on [discussion #2215](https://github.com/Kludex/starlette/discussions/2215).

# Test plan

- Added regression coverage for a buffered body event.
- Added repeated-poll ordering coverage.
- Added partial-body-then-disconnect coverage.
- Added body-after-polled-disconnect coverage.
- `scripts/test`: 987 passed, 2 xfailed, Ruff and mypy clean, 100% branch coverage.

# Checklist

- [x] I understand that this PR may be closed if there was no previous discussion; discussion #2215 covers this bug.
- [x] I added tests for each changed behavior and kept the source change atomic.
- [x] No documentation update is needed because this restores the documented request-body and disconnect behavior without changing the public API.

# AI assistance

I used Claude Code with Claude Opus 5 to investigate, implement, and adversarially test this change. The commit includes the Claude co-author and session attribution.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -231,7 +231,15 @@ class Request(HTTPConnection[StateT]):
 
     @property
     def receive(self) -> Receive:
-        return self._receive
+        return self._receive_message
+
+    async def _receive_message(self) -> Message:
+        # `is_disconnected()` may have polled messages off the channel; hand those out
+        # before reading the channel again, so every consumer of `receive` — `stream()`
+        # or anything the request is delegated to — sees the channel in order.
+        if self._polled_messages:
+            return self._polled_messages.popleft()
+        return await self._receive()
 
     async def stream(self) -> AsyncGenerator[bytes, None]:
         if hasattr(self, "_body"):
@@ -241,7 +249,7 @@ class Request(HTTPConnection[StateT]):
         if self._stream_consumed:
             raise RuntimeError("Stream consumed")
         while not self._stream_consumed:
-            message = self._polled_messages.popleft() if self._polled_messages else await self._receive()
+            message = await self._receive_message()
             if message["type"] == "http.request":
                 body = message.get("body", b"")
                 if not message.get("more_body", False):
--- a/starlette/requests.py
+++ b/starlette/requests.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import sys
+from collections import deque
 from collections.abc import AsyncGenerator, Iterator, Mapping
 from http import cookies as http_cookies
 from typing import TYPE_CHECKING, Any, Generic, NoReturn, cast
@@ -221,6 +222,7 @@ class Request(HTTPConnection[StateT]):
         self._send = send
         self._stream_consumed = False
         self._is_disconnected = False
+        self._polled_messages: deque[Message] = deque()
         self._form = None
 
     @property
@@ -239,7 +241,7 @@ class Request(HTTPConnection[StateT]):
         if self._stream_consumed:
             raise RuntimeError("Stream consumed")
         while not self._stream_consumed:
-            message = await self._receive()
+            message = self._polled_messages.popleft() if self._polled_messages else await self._receive()
             if message["type"] == "http.request":
                 body = message.get("body", b"")
                 if not message.get("more_body", False):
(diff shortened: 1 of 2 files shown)


starlette/requests.py
```diff
@@ -334,8 +336,14 @@ async def is_disconnected(self) -> bool:
...
                 message = await self._receive()
 
-            if message.get("type") == "http.disconnect":
-                self._is_disconnected = True
+            if message:
+                # The server had a message ready. Hold on to it so `stream()` still sees
+                # the receive channel in order — dropping a body message loses part of
+                # the body, and dropping the disconnect loses the only thing that ends
+                # the stream, so a later `body()` waits for a message that never comes.
+                self._polled_messages.append(message)
```
Sanjays2402, line 352: the buffer only gets drained by `stream()`, but `Request.receive` hands out `self._receive` raw, so anything that takes the channel onward (`await response(request.scope, request.receive, send)`, a sub-app, a websocket upgrade) still misses the message `is_disconnected()` pulled. same lost chunk, one accessor over. worth having the property return a wrapper that pops `_polled_messages` first?

davidpavlovschi: Good catch — agreed, the buffer only helped consumers that went through `stream()`. Pushed 5d25bbd: the `receive` property now returns `_receive_message`, a bound method that pops `_polled_messages` before touching the raw channel, and `stream()` goes through the same method, so there's a single drain point and both paths see the channel in order. Nothing in the codebase relied on `request.receive` being identical to the raw ASGI callable (checked `endpoints.py`, middleware, responses — they all carry the constructor-injected `receive`, not the request property). Added a regression test that polls `is_disconnected()` and then consumes the body purely through `request.receive` handed to a delegated app — it fails on the previous revision and passes now. Full suite: 987 passed.

Mukller (approved):
Verified locally on the PR branch (Python 3.13) — this fixes a real data-loss bug, and I reproduced it before/after:

**Repro** (raw ASGI receive queue delivering a two-chunk body; handler calls `is_disconnected()` first):

```
receive channel: [{body: b"chunk1-", more_body: True}, {body: b"chunk2", more_body: False}]

await request.is_disconnected()   # polls chunk1 off the channel
await request.body()

main   (before): body == b'chunk2'        <- chunk1 silently lost
branch (after):  body == b'chunk1-chunk2' <- buffered and replayed in order
```

Instrumenting the poll confirms the mechanism: the immediate message is returned synchronously by `receive`, captured into `_polled_messages`, and `_receive_message` hands buffered messages out before touching the channel again — so every consumer (`stream()`, direct `.receive`) sees the original order.

**Tests**: `pytest tests/test_requests.py` on the branch — 95 passed. The 2 failures in that file are pre-existing on main too (headers assertion drift in my local httpx version, identical output both sides), not related to this change.

Design notes I checked:
- The `if message:` guard is correct for the timeout path (empty `{}` from a cancelled poll is neither buffered nor treated as disconnect).
- Buffering the disconnect message matters as much as buffering body chunks: without it, a disconnect consumed by `is_disconnected()` would leave a later `stream()` waiting forever.
- Routing all consumers through the `receive` property wrapper means subclasses/user code that reads `request.receive` also get ordering for free.

Approving — this is exactly the kind of subtle receive-channel bug that's painful to debug in production.

PR #3397, opened 2026-07-30
fix: reject empty byte ranges in FileResponse

### LINKED ISSUE #3388: FileResponse accepts inverted single-byte Range like bytes=5-4
### Bug

`FileResponse._parse_range_header` rejects a Range only when `start > end`, but `bytes=5-4` parses to the empty half-open range `(5, 5)` and slips past that guard. The result is a malformed `206 Partial Content` (`Content-Length: 0`, a backwards `Content-Range: bytes 5-4/...`) instead of a `400`.

`bytes=10-5` is caught (it parses to `(10, 6)`, so `10 > 6`), but the off-by-one boundary case where `end + 1 == start` is not.

### Reproduce

```python
from starlette.responses import FileResponse
FileResponse._parse_range_header("bytes=5-4", 100)   # -> [(5, 5)]  (accepted; should be 400)
FileResponse._parse_range_header("bytes=5-5", 100)   # -> [(5, 6)]  (valid one-byte range)
FileResponse._parse_range_header("bytes=10-5", 100)  # -> MalformedRangeHeader (already correct)
```

### Fix

The guard should be `start >= end` (an empty range is not satisfiable). A genuine one-byte request `bytes=5-5` parses to `(5, 6)` and is unaffected.

I have a fix + tests ready and opened it as a draft PR for reference. Happy to adjust or drop it if you'd prefer to handle it differently.

ltsyk:
## Summary
`FileResponse._parse_range_header` accepted empty half-open ranges such as `bytes=5-4` → `(5, 5)`, producing a malformed 206. One-byte `bytes=5-5` → `(5, 6)` is unaffected.

## Change
Reject ranges with `start >= end` instead of `start > end`.

Fixes #3388

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -474,8 +474,7 @@ class FileResponse(Response):
         if any(not (0 <= start < file_size) for start, _ in ranges):
             raise RangeNotSatisfiable(file_size)
 
-        # Half-open ranges: empty ranges (start == end) are unsatisfiable, e.g. bytes=5-4
-        # parses to (5, 5). One-byte range bytes=5-5 parses to (5, 6). Issue #3388.
+        # An empty half-open range is not a valid range.
         if any(start >= end for start, end in ranges):
             raise MalformedRangeHeader("Range header: start must be less than end")
 
--- a/starlette/responses.py
+++ b/starlette/responses.py
@@ -474,7 +474,9 @@ class FileResponse(Response):
         if any(not (0 <= start < file_size) for start, _ in ranges):
             raise RangeNotSatisfiable(file_size)
 
-        if any(start > end for start, end in ranges):
+        # Half-open ranges: empty ranges (start == end) are unsatisfiable, e.g. bytes=5-4
+        # parses to (5, 5). One-byte range bytes=5-5 parses to (5, 6). Issue #3388.
+        if any(start >= end for start, end in ranges):
             raise MalformedRangeHeader("Range header: start must be less than end")
 
         if len(ranges) == 1:
--- a/tests/test_responses.py
+++ b/tests/test_responses.py
@@ -810,10 +810,17 @@ def test_file_response_range_must_be_requested(file_response_client: TestClient)
     assert response.text == "Range header: range must be requested"
 
 
-def test_file_response_start_must_be_less_than_end(file_response_client: TestClient) -> None:
-    response = file_response_client.get("/", headers={"Range": "bytes=100-0"})
+@pytest.mark.parametrize("header", ["bytes=5-4", "bytes=0--1", "bytes=0-1, 5-4"])
+def test_file_response_empty_range(file_response_client: TestClient, header: str) -> None:
+    response = file_response_client.get("/", headers={"Range": header})
     assert response.status_code == 400
     assert response.text == "Range header: start must be less than end"
+
+
+def test_file_response_single_byte_range(file_response_client: TestClient) -> None:
+    response = file_response_client.get("/", headers={"Range": "bytes=5-5"})
+    assert response.status_code == 206
+    assert response.headers["content-length"] == "1"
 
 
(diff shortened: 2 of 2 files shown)


kdelay (commented):
I hit the same thing from the reporting side of #3388 and ran this branch locally (Python 3.12, macOS) to check the boundary. The change itself holds up, and it fixes a bit more than the issue describes.

Serving a 100-byte file through `FileResponse`, on `master` (5174d4c) vs this branch (c36ab12):

| `Range` | master | this branch |
| --- | --- | --- |
| `bytes=5-4` | `206`, `Content-Range: bytes 5-4/100`, `Content-Length: 0` | `400` |
| `bytes=0--1` | `206`, `Content-Range: bytes 0--1/100`, `Content-Length: 0` | `400` |
| `bytes=0-1, 5-4` | `206` multipart with a zero-length part | `400` |
| `bytes=5-5` | `206`, `Content-Length: 1` | `206`, `Content-Length: 1` |

The last two are worth calling out because they are not in the issue. `bytes=0--1` reaches `_parse_ranges` as `end_str="-1"`, so `int("-1") < file_size` holds and `end` becomes `0`; the multi-range case emits a multipart body with an empty part rather than a single malformed 206. Both are the same `start == end` case, and both are covered by this one-line change.

One gap: this branch has no test. I reverted just the `start >= end` hunk on top of this branch and the full suite is still green (977 passed, 2 skipped, 2 xfailed), so nothing here fixes the behaviour in place. The PR checklist asks for a test per change, and `tests/test_responses.py` already has `test_file_response_start_must_be_less_than_end` right next to it. Something like:

```python
@pytest.mark.parametrize("header", ["bytes=5-4", "bytes=0--1", "bytes=0-1, 5-4"])
def test_file_response_empty_range(file_response_client: TestClient, header: str) -> None:
    response = file_response_client.get("/", headers={"Range": header})
    assert response.status_code == 400
    assert response.text == "Range header: start must be less than end"


def test_file_response_single_byte_range(file_response_client: TestClient) -> None:
    response = file_response_client.get("/", headers={"Range": "bytes=5-5"})
    assert response.status_code == 206
    assert response.headers["content-length"] == "1"
```

I ran these on this branch: they pass, and with the hunk reverted the first one fails on `bytes=5-4` (`assert 206 == 400`), so they do pin the change. `ruff check`, `ruff format --check` and `mypy` are clean either way.

Also, #3389 (draft) carries tests for `bytes=5-4` and `bytes=5-5`; I applied its test file on top of this branch and they pass, so the two are compatible. Neither set covers the negative-`end` or multi-range variants above, so those seem worth folding in wherever the tests end up living.

Minor wording: the new comment says empty ranges "are unsatisfiable", but the branch it guards raises `MalformedRangeHeader` (400), while `RangeNotSatisfiable` (416) is the check just above it. Reading "unsatisfiable" next to a 400 is a little confusing — maybe "an empty range is not a valid range" instead.

ltsyk:
Thanks for the thorough boundary check and the suggested cases.

I added tests for `bytes=5-4`, `bytes=0--1`, and the multi-range empty part (`bytes=0-1, 5-4`), plus a single-byte `bytes=5-5` control that still returns 206. The comment now says an empty half-open range is **not a valid range** (400 / `MalformedRangeHeader`) rather than “unsatisfiable” (416), so it is not confused with the `RangeNotSatisfiable` branch above it.

CI is green with these tests in place.

PR #3407, opened 2026-08-04
Lazily allocate GZipMiddleware compression resources

Kludex:
## Summary

- lazily allocate `BytesIO` and `GzipFile` only when a response is compressed
- retain context-managed, reverse-order cleanup on normal completion and application errors
- add regression coverage for responses below `minimum_size`, existing `Content-Encoding`, `text/event-stream`, and `http.response.pathsend`
- update the gzip benchmark to use the lazy compression context

## Motivation

`GZipResponder` previously constructed the gzip buffer and compressor before it knew whether the response should be compressed. Bypassed responses therefore paid the compressor's allocation cost even though their body was forwarded unchanged.

This keeps the public `GZipMiddleware` configuration and response behavior unchanged while avoiding those allocations for bypassed responses. CodSpeed's memory-mode bypass benchmarks will measure the resulting allocation reduction.

## Validation

- `scripts/check`
- `uv run pytest` — 989 passed, 2 skipped, 2 xfailed
- focused gzip tests and representative CodSpeed benchmark cases on Python 3.14

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -1,5 +1,6 @@
 import gzip
 import io
+from contextlib import ExitStack
 from typing import NoReturn
 
 from starlette.datastructures import Headers, MutableHeaders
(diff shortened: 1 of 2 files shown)


starlette/middleware/gzip.py
```diff
@@ -122,14 +123,28 @@ class GZipResponder(IdentityResponder):
     def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 9) -> None:
         super().__init__(app, minimum_size)
 
-        self.gzip_buffer = io.BytesIO()
-        self.gzip_file = gzip.GzipFile(mode="wb", fileobj=self.gzip_buffer, compresslevel=compresslevel)
+        self.compresslevel = compresslevel
+        self.gzip_buffer: io.BytesIO | None = None
+        self.gzip_file: gzip.GzipFile | None = None
```
Kludex: Can't we have `self._gzip_bugfer` and then have a property that is `io.BytesIO` without None?

starlette/middleware/gzip.py
```diff
@@ -122,14 +123,28 @@ class GZipResponder(IdentityResponder):
...
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        with self.gzip_buffer, self.gzip_file:
+        with self._gzip_context:
             await super().__call__(scope, receive, send)
 
     def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
+        self._ensure_compression_started()
+        assert self.gzip_file is not None
+        assert self.gzip_buffer is not None
```
Kludex: let's remove this.

PR #3410, opened 2026-08-04
Offload large GZip compression

Kludex:
## Summary

- offload GZip body chunks of at least 128 KiB to an AnyIO worker thread
- add `thread_minimum_size` as an optional `GZipMiddleware` setting
- use a dedicated, private, per-event-loop capacity limiter (40 tokens, matching AnyIO's default worker-thread concurrency) so GZip work does not compete with the shared thread pool
- update the ASGI benchmark runner to support middleware calls that suspend (landed separately in #3412)

## Rationale

Compressing a multi-megabyte response inline blocks the event loop until zlib finishes — ~150–310 ms for a 10 MiB level-9 response, freezing every other request on the loop. zlib releases the GIL, so moving sufficiently large chunks to a worker thread lets unrelated requests continue while preserving the inexpensive inline path for smaller chunks.

Built on the `zlib.compressobj` refactor (#3411): `compress()` performs all deflate work inside the call, so the offloaded path is simply `compressor.compress(body)` running in the worker with no flush workarounds.

The dedicated limiter keeps GZip work from starving AnyIO's shared worker-thread capacity (used by `run_in_threadpool`, `FileResponse`, etc.).

## Threshold choice (128 KiB)

The worker-thread dispatch costs a fixed ~70 µs per chunk. Below 128 KiB that overhead is a meaningful fraction of the compression work (offloading 32 KiB streamed chunks measured up to +18% total time on realistic data), while compressing a chunk just under 128 KiB inline blocks the loop for at most ~1.7 ms (random data) to ~7 ms (level-9 worst case) — within normal scheduling jitter. At 128 KiB and above, dispatch overhead is ≤4% of the compression work.

## Compatibility

`thread_minimum_size` defaults to 128 KiB. Existing middleware configuration remains valid. Responses below `minimum_size`, responses with an existing `Content-Encoding`, server-sent events, and `http.response.pathsend` continue through their existing paths.

## Benchmark

The responsiveness benchmark (baseline in #3409) schedules a tiny bypass response behind a 10 MiB level-9 response: inline compression blocks the tiny response for the full ~190 ms; with offload it completes in ~0.25 ms (~750× improvement), while the large response's total completion time is unchanged.

Note: CodSpeed's instruction-counting simulation can only observe the dispatch cost of offloading, not the event-loop time it frees — the remaining ≥128 KiB single-chunk deltas are expected.

## Validation

- Full test suite and focused middleware/benchmark suites pass on Python 3.10 and 3.14
- `scripts/check`

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -18,9 +18,9 @@ def _get_gzip_capacity_limiter() -> anyio.CapacityLimiter:
     try:
         return _gzip_capacity_limiter.get()
     except LookupError:
-        # Inline compression is effectively serialized by the event loop today.
-        # Keep that bound while allowing unrelated event-loop work to continue.
-        limiter = anyio.CapacityLimiter(1)
+        # Keep gzip compression isolated from AnyIO's default worker-thread
+        # capacity limiter while matching its default concurrency.
+        limiter = anyio.CapacityLimiter(40)
         _gzip_capacity_limiter.set(limiter)
         return limiter
 
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -170,12 +170,8 @@ class GZipResponder(IdentityResponder):
     async def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
         if len(body) >= self.offload_to_thread_minimum_size:
             # Compressing large chunks inline would block the event loop.
-            return await anyio.to_thread.run_sync(
-                self._compress_body,
-                body,
-                more_body,
-                limiter=gzip_capacity_limiter(),
-            )
+            limiter = gzip_capacity_limiter()
+            return await anyio.to_thread.run_sync(self._compress_body, body, more_body, limiter=limiter)
         return self._compress_body(body, more_body)
 
     def _compress_body(self, body: bytes, more_body: bool) -> bytes:
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -31,12 +31,12 @@ class GZipMiddleware:
         app: ASGIApp,
         minimum_size: int = 500,
         compresslevel: int = 9,
-        offload_to_thread_minimum_size: int = 128 * 1024,  # 128 KiB
+        offload_minimum_size: int = 128 * 1024,  # 128 KiB
     ) -> None:
         self.app = app
         self.minimum_size = minimum_size
         self.compresslevel = compresslevel
-        self.offload_to_thread_minimum_size = offload_to_thread_minimum_size
+        self.offload_minimum_size = offload_minimum_size
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
         if scope["type"] != "http":  # pragma: no cover
@@ -50,7 +50,7 @@ class GZipMiddleware:
                 self.app,
                 self.minimum_size,
                 compresslevel=self.compresslevel,
-                offload_to_thread_minimum_size=self.offload_to_thread_minimum_size,
+                offload_minimum_size=self.offload_minimum_size,
             )
         else:
             responder = IdentityResponder(self.app, self.minimum_size)
@@ -153,12 +153,12 @@ class GZipResponder(IdentityResponder):
         minimum_size: int,
         compresslevel: int = 9,
         *,
-        offload_to_thread_minimum_size: int,
+        offload_minimum_size: int,
     ) -> None:
         super().__init__(app, minimum_size)
 
         self.compresslevel = compresslevel
-        self.offload_to_thread_minimum_size = offload_to_thread_minimum_size
+        self.offload_minimum_size = offload_minimum_size
         self._compressor: zlib._Compress | None = None
 
     @property
@@ -168,7 +168,7 @@ class GZipResponder(IdentityResponder):
         return self._compressor
 
     async def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
-        if len(body) >= self.offload_to_thread_minimum_size:
+        if len(body) >= self.offload_minimum_size:
             # Compressing large chunks inline would block the event loop.
             limiter = gzip_capacity_limiter()
             return await anyio.to_thread.run_sync(self._compress_body, body, more_body, limiter=limiter)
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -31,12 +31,12 @@ class GZipMiddleware:
         app: ASGIApp,
         minimum_size: int = 500,
         compresslevel: int = 9,
-        offload_minimum_size: int = 128 * 1024,  # 128 KiB
+        thread_minimum_size: int = 128 * 1024,  # 128 KiB
     ) -> None:
         self.app = app
         self.minimum_size = minimum_size
         self.compresslevel = compresslevel
-        self.offload_minimum_size = offload_minimum_size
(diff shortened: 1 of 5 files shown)


starlette/middleware/gzip.py
```diff
@@ -1,19 +1,44 @@
...
 from typing import NoReturn
 
+import anyio
+
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 DEFAULT_EXCLUDED_CONTENT_TYPES = ("text/event-stream",)
+DEFAULT_GZIP_OFFLOAD_MINIMUM_SIZE = 32 * 1024
+"""Minimum body-chunk size that is compressed outside the event loop."""
```
Kludex: Drop the docstring for this constant.
```suggestion
```

starlette/middleware/gzip.py
```diff
@@ -1,19 +1,44 @@
...
 from contextlib import ExitStack
 from typing import NoReturn
 
+import anyio
+
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
 DEFAULT_EXCLUDED_CONTENT_TYPES = ("text/event-stream",)
+DEFAULT_GZIP_OFFLOAD_MINIMUM_SIZE = 32 * 1024
```
```suggestion
DEFAULT_GZIP_OFFLOAD_MINIMUM_SIZE = 32 * 1024  # 32 Mb
```

starlette/middleware/gzip.py
```diff
@@ -1,19 +1,44 @@
...
+
+def _get_gzip_capacity_limiter() -> anyio.CapacityLimiter:
+    try:
+        return _gzip_capacity_limiter.get()
+    except LookupError:
+        # Inline compression is effectively serialized by the event loop today.
+        # Keep that bound while allowing unrelated event-loop work to continue.
+        limiter = anyio.CapacityLimiter(1)
+        _gzip_capacity_limiter.set(limiter)
+        return limiter
```
Kludex, line 25: Can't we just have a global `anyio.CapacityLimiter`? Why 1 tho? the default one has 100, right?

starlette/middleware/gzip.py
```diff
@@ -120,13 +150,21 @@ def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
 class GZipResponder(IdentityResponder):
     content_encoding = "gzip"
 
-    def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 9) -> None:
+    def __init__(
+        self,
+        app: ASGIApp,
+        minimum_size: int,
+        compresslevel: int = 9,
+        offload_minimum_size: int = DEFAULT_GZIP_OFFLOAD_MINIMUM_SIZE,
```
Kludex: Can we avoid having a default value here?

starlette/middleware/gzip.py
```diff
@@ -1,19 +1,44 @@
...
 
 
 class GZipMiddleware:
-    def __init__(self, app: ASGIApp, minimum_size: int = 500, compresslevel: int = 9) -> None:
+    def __init__(
+        self,
+        app: ASGIApp,
+        minimum_size: int = 500,
+        compresslevel: int = 9,
+        offload_minimum_size: int = DEFAULT_GZIP_OFFLOAD_MINIMUM_SIZE,
```
Kludex: can we actually inline the value of default gzip offload minimum size?

Kludex: also, the name of the parameter should be more clear that it's used for offloading to a thread.

starlette/middleware/gzip.py
```diff
@@ -99,14 +127,17 @@ async def send_with_compression(self, message: Message) -> None:
...
+            message["body"] = await self._apply_compression(body, more_body=more_body)
 
             await self.send(message)
         elif message_type == "http.response.pathsend":  # pragma: no branch
             # Don't apply GZip to pathsend responses
             await self.send(self.initial_message)
             await self.send(message)
 
+    async def _apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
+        return self.apply_compression(body, more_body=more_body)
```
Kludex: I don't think we need this.

Kludex:
Updated after #3411 landed: merged main and rebuilt `GZipResponder` on top of the `zlib.compressobj` refactor. This removes the `GzipFile`/`BytesIO` machinery and, importantly, the `Z_NO_FLUSH` workaround entirely — `compressobj.compress()` performs all deflate work inside the call, so the offloaded path is now simply `compressor.compress(body)` running in the worker thread. Net effect of the merge: −24 lines in the middleware. Tests (20) and CodSpeed benchmarks (68) pass, `scripts/check` clean.

starlette/middleware/gzip.py
```diff
@@ -131,7 +167,18 @@ def compressor(self) -> zlib._Compress:
...
-    def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
+    async def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
+        if len(body) >= self.offload_to_thread_minimum_size:
+            # Compressing large chunks inline would block the event loop.
+            return await anyio.to_thread.run_sync(
+                self._compress_body,
+                body,
+                more_body,
+                limiter=gzip_capacity_limiter(),
+            )
```
```suggestion
            limiter = gzip_capacity_limiter()
            return await anyio.to_thread.run_sync(
                self._compress_body, body, more_body, limiter=limiter
            )
```

Kludex: Applied in 0e22255 (kept the run_sync call on one line since it fits within the line length).

starlette/middleware/gzip.py
Kludex: Good questions — I checked both.

**Why not a plain global?** It actually works: `anyio.CapacityLimiter` created at module level returns a `CapacityLimiterAdapter` that lazily creates the real backend limiter on first use. However, I verified its state is **shared across event loops** (borrowed tokens accumulate across loops), so with multiple loops (e.g. uvicorn workers with `--loop` per-thread setups, or tests running many loops) they'd compete for the same tokens and a dead loop could strand tokens. The `RunVar` keeps the limiter per-event-loop, which matches how AnyIO scopes its own default thread limiter. Happy to switch to a plain global if you prefer the simplicity, though.

**Why 40 (not 1, not 100)?** The default AnyIO thread limiter has **40 tokens** (verified: `current_default_thread_limiter().total_tokens == 40`), not 100. An earlier revision used 1 to preserve the event loop's effective serialization of compression work, but that unnecessarily queues concurrent large responses; 40 matches AnyIO's default concurrency while keeping gzip from starving the shared pool that `run_in_threadpool`/`FileResponse` use. The description has been updated accordingly.

tests/middleware/test_gzip.py
```diff
@@ -93,6 +95,57 @@ async def generator(bytes: bytes, count: int) -> ContentStream:
     assert "Content-Length" not in response.headers
 
 
+def test_gzip_offloaded_compression(test_client_factory: TestClientFactory) -> None:
```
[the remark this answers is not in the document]
Kludex: Intentional: the exhaustion-based test was deemed too much machinery for the repo's testing style (saturating 40 tokens, fail_after as a deadlock detector). The dedicated-limiter property is enforced structurally — `run_sync` is only called with `limiter=gzip_capacity_limiter()`, and coverage requires that line to execute — and documented in docs/middleware.md.

Issue #3414, opened 2026-08-05
GZipResponder in 1.4.0 adds a required keyword-only argument, breaking Streamlit


lukasmasuch:
The Starlette 1.4.0 update [breaks Streamlit](https://github.com/streamlit/streamlit/issues/16341) :( We are overwriting the `GZipResponder` (is this considered internal?) which got a new required `thread_minimum_size` argument.  Streamlit's custom gzip middleware subclasses that responder and constructed it the old way, so on 1.4.0 every gzip response crashes at startup with:

```
TypeError: GZipResponder.__init__() missing 1 required keyword-only argument: 'thread_minimum_size'
```

Is `GZipResponder` considered internal? Is there a way to roll out patch that makes this parameter optional (e.g. via a default value)? 

<details><summary>AI Report</summary>

### Summary

Starlette 1.4.0 (["Offload large GZip co mpression", #3410](https://github.com/encode/starlette/pull/3410)) changed the constructor of `starlette.middleware.gzip.GZipResponder` in a backwards-incompatible way: it added a **required keyword-only** parameter `thread_minimum_size` with **no default value**. It also changed `IdentityResponder.apply_compression` / `GZipResponder.apply_compression` from a synchronous method to an `async` method.

`GZipResponder` and `IdentityResponder` are importable from `starlette.middleware.gzip` and are subclassed by downstream projects that want to customize GZip behavior (e.g. excluding certain content types). Any such subclass that instantiated `GZipResponder(app, minimum_size, compresslevel=...)` now crashes at runtime with:

```
TypeError: GZipResponder.__init__() missing 1 required keyword-only argument: 'thread_minimum_size'
```

This surfaces as a hard failure on the first gzip-eligible HTTP response, before user code runs.

### Version info

- Broke in: **starlette 1.4.0** (published 2026-08-05)
- Last working: **starlette 1.3.1**
- Introduced by: https://github.com/encode/starlette/pull/3410 (see also #3407, #3411)

### Signature diff

**1.3.1**
```python
class GZipResponder(IdentityResponder):
    def __init__(self, app: ASGIApp, minimum_size: int, compresslevel: int = 9) -> None: ...

    def apply_compression(self, body: bytes, *, more_body: bool) -> bytes: ...   # sync
```

**1.4.0**
```python
class GZipResponder(IdentityResponder):
    def __init__(
        self,
        app: ASGIApp,
        minimum_size: int,
        compresslevel: int = 9,
        *,
[...]

PR #3418, opened 2026-08-06
Add `exclude_content_types` parameter to `GZipMiddleware`

Kludex:
Add a keyword-only `exclude_content_types` parameter to `GZipMiddleware`, replacing the hardcoded exclusion check.

- A response is excluded when its `Content-Type` media type equals an entry, or matches a `type/*` wildcard entry (e.g. `"image/*"`). Matching is case-insensitive and ignores parameters like `charset`; configured values are normalized the same way.
- Defaults to `DEFAULT_EXCLUDED_CONTENT_TYPES` (`("text/event-stream",)`), which stays importable so the defaults can be extended rather than replaced.
- Both responders accept the parameter keyword-only with a default, so direct instantiation keeps working (same pattern as #3415).

Supersedes #3091 and #3040, and implements the feature requested in #3414.

Wire-visible deltas from the previous prefix check, for the release notes: matching is now case-insensitive, `text/event-stream-foo` no longer matches, and reassigning `DEFAULT_EXCLUDED_CONTENT_TYPES` at runtime no longer takes effect (the exclusions are read at construction).

## AI Disclaimer

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -9,6 +9,7 @@ import anyio.to_thread
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
+# TODO(v2): We should rename `DEFAULT_EXCLUDED_CONTENT_TYPES` to `DEFAULT_EXCLUDE_CONTENT_TYPES`.
 DEFAULT_EXCLUDED_CONTENT_TYPES = ("text/event-stream",)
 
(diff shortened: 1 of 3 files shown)


starlette/middleware/gzip.py
Kludex, line 13: The "EXCLUDED" was a mistake.

It should have been exclude.

PR #3419, opened 2026-08-06
Flush GZip output for each streamed chunk

Kludex:
Compress and emit output for every streamed chunk (`Z_SYNC_FLUSH`), instead of letting zlib buffer across chunks. Chunk boundaries are the application's intent - a compressor should not erase them.

This deliberately reverses the conclusion of #2753 and supersedes #3230 (see also discussion #3231). The precedent moved since #2753 was closed: Django now flushes every streamed chunk unconditionally (`12bb16da8f`, "Fixed #36293 -- Avoided buffering streaming responses in GZipMiddleware"), and Tomcat, Undertow, every Netty-based stack, Apache `mod_deflate`, h2o, Deno's runtime, and salvo all do the same, with no configuration knob.

- Non-streaming responses are unaffected (single-shot bodies were already finalized immediately).
- `text/event-stream` remains excluded by default. With this change, compressed *and* timely SSE is one setting away: `exclude_content_types=()`.
- The worker-thread path (#3410) is unchanged - the flush happens inside `_compress_body` on either path.
- Cost: compression ratio drops for streams that emit many small chunks (~5 bytes overhead per flush plus reduced entropy pooling). Applications wanting maximum ratio can batch their chunks.

Stacked on #3418.

## AI Disclaimer

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3419?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -191,7 +191,9 @@ class GZipResponder(IdentityResponder):
 
     def _compress_body(self, body: bytes, more_body: bool) -> bytes:
         if more_body:
-            return self.compressor.compress(body)
+            # Emit everything compressed so far: streamed chunks must reach the
+            # client as the application sends them, not when the buffer fills.
+            return self.compressor.compress(body) + self.compressor.flush(zlib.Z_SYNC_FLUSH)
         return self.compressor.compress(body) + self.compressor.flush()
 
 
--- a/starlette/middleware/gzip.py
+++ b/starlette/middleware/gzip.py
@@ -9,6 +9,7 @@ import anyio.to_thread
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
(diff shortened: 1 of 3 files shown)


starlette/middleware/gzip.py
```diff
@@ -192,7 +192,9 @@ async def apply_compression(self, body: bytes, *, more_body: bool) -> bytes:
 
     def _compress_body(self, body: bytes, more_body: bool) -> bytes:
         if more_body:
-            return self.compressor.compress(body)
+            # Emit everything compressed so far: streamed chunks must reach the
+            # client as the application sends them, not when the buffer fills.
```
Kludex: Drop this comment.

Kludex: Done.

PR #3426, opened 2026-08-08
Extract route resolution into `Router.resolve`

Kludex:
## Summary

`Router.app` inlines the match loop twice, once for the request and once for the trailing-slash retry. Anything that wants to change how a route is found has to reimplement dispatch around it.

```python
def resolve(self, scope: Scope) -> tuple[BaseRoute, Scope] | None:
    """Pick the route that should handle `scope`, and the scope it needs."""
```

It returns the first full match, or else the first partial one (which is how a `405` is produced), or `None`. `app()` becomes the four steps around it: resolve, handle, redirect, default.

> [!NOTE]
> Based on #3339, so the diff here is only the extraction.

## Why

FastAPI's `APIRouter.app` (`fastapi/routing.py`, since 0.138.0) is a copy of this method. Diffed against ours it is identical apart from cosmetics, plus one inserted block immediately before `await self.default(...)` for its low-priority (SPA fallback) tier. The copy also scans `self.routes` linearly, so a FastAPI app does not benefit from the narrowing in #3339.

Two things follow from this PR:

- Downstream routers can override `resolve()` to change route lookup and inherit the rest of dispatch, including the narrowing.
- FastAPI's low-priority tier sits exactly where `self.default` is invoked, so it can be expressed as a `default` callable rather than a copied `app()`. Verified against this branch: a custom `default` gives SPA fallback on a miss, while `redirect_slashes` still takes precedence and 405s are unaffected.

## Behaviour

Unchanged. The one difference worth noting is internal: the trailing-slash check previously returned on the first route with any match, and now asks `resolve()`, which prefers a full match over a partial. The redirect fires in exactly the same cases.

1058 tests pass, 100% coverage, mypy strict and ruff clean.

## AI Disclaimer

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3426?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -683,6 +683,25 @@ class Router:
             self._trie = trie
         return [routes[i] for i in self._trie.match_all(get_route_path(scope))]
 
+    def resolve(self, scope: Scope) -> tuple[BaseRoute, Scope] | None:
+        """Pick the route that should handle `scope`, and the scope it needs.
+
+        Returns the first route that fully matches, or else the first that
+        partially matches, which is how a `405 Method Not Allowed` is produced.
+        Returns `None` when nothing matches.
+
+        Override this to change how routes are found, without reimplementing the
+        rest of dispatch.
+        """
+        partial: tuple[BaseRoute, Scope] | None = None
+        for route in self._candidate_routes(scope):
+            match, child_scope = route.matches(scope)
+            if match == Match.FULL:
+                return route, child_scope
+            if match == Match.PARTIAL and partial is None:
+                partial = (route, child_scope)
+        return partial
+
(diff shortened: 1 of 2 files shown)


Kludex:
FastAPI can use Starlette’s existing `default` callback and inherit the trie-backed dispatcher directly through [fastapi/fastapi#16236](https://github.com/fastapi/fastapi/pull/16236). There is no longer a concrete need for this additional routing extension point.

PR #3436, opened 2026-08-10
Track session mutations from popitem() and |=

nyxst4ck:
# Summary

`Session` tracks mutations so that `SessionMiddleware` knows when to emit an updated cookie. `dict.popitem()` and the in-place union operator (`|=`) were not overridden, so those mutations left `Session.modified` false.

A request handler could observe the changed mapping, but the response omitted `Set-Cookie`, and the mutation disappeared on the next request.

This change overrides both mutators and adds middleware-level persistence regression coverage.

# Reproduction

1. Store at least one value in `request.session`.
2. In a later request, call `request.session.popitem()`, or assign `session = request.session` and then call `session |= values`.
3. Observe the changed mapping in that request.
4. Make another request and observe that the previous cookie value is restored because the mutating response did not emit `Set-Cookie`.

# Tests

- `uv run pytest tests/middleware/test_session.py` - 23 passed
- `uv run ruff check starlette/middleware/sessions.py tests/middleware/test_session.py` - passed
- `uv run ruff format --check starlette/middleware/sessions.py tests/middleware/test_session.py` - passed
- `uv run mypy starlette/middleware/sessions.py tests/middleware/test_session.py` - passed

# Checklist

- [x] Documentation is unchanged because this fixes existing mapping mutation behavior without introducing a new public API.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## AI Disclaimer

### DIFF
--- a/starlette/middleware/sessions.py
+++ b/starlette/middleware/sessions.py
@@ -115,6 +115,11 @@ class Session(dict[str, typing.Any]):
         self.modified = self.modified or key in self
         return super().pop(key, *args)
 
+    def popitem(self) -> tuple[str, typing.Any]:
+        item = super().popitem()
+        self.mark_modified()
+        return item
+
     def setdefault(self, key: str, default: typing.Any = None) -> typing.Any:
         if key not in self:
             self.mark_modified()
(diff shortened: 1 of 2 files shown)


Mukller (approved):

**Tests**: `pytest tests/middleware/test_session.py` — 23/23 green, including the new `|=` and `popitem` cases.

**Behavior probes**:

```python
s = Session({"a": 1}); s.popitem()
# s.modified -> True   (previously stayed False -> session not persisted)

s2 = Session({"a": 1}); s2 |= {"b": 2}
# s2.modified -> True, content correct
```

Both are genuine gaps: `dict.popitem` and `dict.__ior__` mutate without ever hitting `__setitem__`, so a session updated only through them was silently treated as unmodified and dropped from the cookie. The unconditional `mark_modified()` in both matches the existing style of `update()`/`clear()`, and the `type: ignore[override,misc]` with the explanatory comment about the widening `dict.__or__` overloads is the right call — a `dict` subclass genuinely can't satisfy mypy here.

Nothing to fix.

Kludex (approved):

PR #3438, opened 2026-08-10
Add OpenTelemetry middleware

### LINKED ISSUE #685: Include routing information for middleware

Kludex:
## Summary

- add an explicit `OpenTelemetryMiddleware` for HTTP server spans
- expose the matched route through `scope["route"]`
- derive span names and `http.route` from actual router dispatch, including mounts and hosts
- add `opentelemetry-api` as an optional dependency

```python
from starlette.middleware.opentelemetry import OpenTelemetryMiddleware

app = Starlette(middleware=[Middleware(OpenTelemetryMiddleware)])
```

The middleware uses the global OpenTelemetry provider at request time. Starlette does not enable tracing automatically.

Closes #685.

## Validation

- `./scripts/check`
- `uv run pytest tests/middleware/test_opentelemetry.py`

## AI Disclaimer


starlette/middleware/opentelemetry.py
```diff
@@ -0,0 +1,128 @@
...
+class _OpenTelemetryMiddleware:
+    """Create OpenTelemetry server spans for incoming HTTP requests."""
+
+    def __init__(self, app: ASGIApp) -> None:
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http" or _tracing_request.get():
+            await self.app(scope, receive, send)
+            return
```
Kludex:
```suggestion
            return await self.app(scope, receive, send)
```

starlette/middleware/opentelemetry.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _set_route(span: Span, scope: Scope, method: str) -> None:
+    route_path = scope.get("starlette.route_path")
+    if isinstance(route_path, str):
+        span.update_name(f"{method} {route_path}")
+        span.set_attribute("http.route", route_path)
+
+
+class _OpenTelemetryMiddleware:
```
Kludex: I think this should be public. Is there any other middleware that needs specific imports? What do they do?

starlette/middleware/opentelemetry.py
```diff
@@ -0,0 +1,128 @@
...
+from opentelemetry import propagate, trace
+from opentelemetry.propagators.textmap import Getter
+from opentelemetry.trace import Span, SpanKind, Status, StatusCode
+
+from starlette import __version__
+from starlette.datastructures import URL
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+_KNOWN_HTTP_METHODS = {"CONNECT", "DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT", "TRACE"}
+_tracing_request: ContextVar[bool] = ContextVar("starlette_tracing_request", default=False)
```
Kludex: Is this so we don't do it twice? Instead of a ContextVar, can't we just pass it through the state or something that survives the request-response cycle middleware stack?

starlette/applications.py
```diff
@@ -77,9 +77,15 @@ def build_middleware_stack(self) -> ASGIApp:
         middleware += self.user_middleware
         middleware.append(Middleware(ExceptionMiddleware, handlers=exception_handlers, debug=debug))
 
-        app = self.router
+        app: ASGIApp = self.router
```
```suggestion
        app = self.router
```

starlette/middleware/opentelemetry.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import annotations
+
+from opentelemetry import propagate, trace
+from opentelemetry.trace import SpanKind, Status, StatusCode
```
Kludex:
```suggestion
try:
    from opentelemetry import propagate, trace
    from opentelemetry.trace import SpanKind, Status, StatusCode
except ImportError:
    raise ImportError("The `opentelemetry-api` package is required to use `OpenTelemetryMiddleware`.")
```

docs/middleware.md
```diff
@@ -47,6 +47,58 @@ application would look like this:
...
+Creates an OpenTelemetry server span for every incoming HTTP request. The span follows the
+OpenTelemetry HTTP semantic conventions, extracts distributed trace context from the request
+headers, and uses the matched route template for its name and `http.route` attribute.
+
+Install the optional API dependency with `pip install opentelemetry-api`, or as part of
+`pip install "starlette[full]"`. Starlette only uses the OpenTelemetry API. Your application
+chooses and configures the SDK and exporter. If no tracer provider is configured, Starlette
+skips tracing.
+
+Logfire configures a global OpenTelemetry provider, so no Starlette-specific instrumentation
```
```suggestion
**Pydantic Logfire** configures a global OpenTelemetry provider, so no Starlette-specific instrumentation
```

starlette/applications.py
```diff
@@ -80,7 +82,15 @@ def build_middleware_stack(self) -> ASGIApp:
         app = self.router
         for cls, args, kwargs in reversed(middleware):
             app = cls(app, *args, **kwargs)
-        return app
+
+        if self._is_instrumented_by_opentelemetry:
+            return app
```
Kludex: Instead of this flag, I think I prefer a parameter into `Starlette` e.g. `enable_opentelemetry`. Or maybe just `opentelemetry` so we could perhaps add configuration into the setting after? 🤔

docs/middleware.md
```diff
@@ -47,6 +47,65 @@ application would look like this:
...
+```
+
+Multiple native middleware instances on the same request create only one span.
+
+When using another ASGI instrumentor, disable Starlette's native instrumentation to avoid
+creating duplicate spans:
+
+```python
+app = Starlette(opentelemetry=False)
+```
```
Kludex, line 107: Instead of this, can't this be done via env var according to otel, or am I wrong?

starlette/middleware/opentelemetry.py
```diff
@@ -0,0 +1,111 @@
...
+from starlette.datastructures import URL
+from starlette.routing import Mount
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+_KNOWN_HTTP_METHODS = frozenset(
+    os.environ.get(
+        "OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS",
+        "CONNECT,DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT,QUERY,TRACE",
+    ).split(",")
+)
```
Kludex: I don't think this bullshit is necessary. There are no other HTTP methods.

marcelomarkus (approved):
nice ✨

PR #3446, opened 2026-08-19
fix(datastructures): safely handle non-string and non-latin1 keys in Headers.__contains__

aoright:
### Problem
`Headers.__contains__` directly called `key.lower().encode("latin-1")`. When `key` was not a string (e.g. `123 in headers`) or contained characters not encodable in latin-1 (e.g. `"\u20ac" in headers`), it raised unhandled `AttributeError` or `UnicodeEncodeError` instead of returning `False` as expected by the standard Python `Mapping` interface.

### Solution
- In `Headers.__contains__`, check `isinstance(key, str)` and catch `UnicodeEncodeError`, returning `False`.
- Add assertions in `test_headers` in `tests/test_datastructures.py`.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/datastructures.py
+++ b/starlette/datastructures.py
@@ -549,7 +549,12 @@ class Headers(Mapping[str, str]):
         raise KeyError(key)
 
     def __contains__(self, key: Any) -> bool:
-        get_header_key = key.lower().encode("latin-1")
+        if not isinstance(key, str):
+            return False
+        try:
+            get_header_key = key.lower().encode("latin-1")
+        except UnicodeEncodeError:
+            return False
         for header_key, header_value in self._list:
             if header_key == get_header_key:
                 return True
--- a/tests/test_datastructures.py
+++ b/tests/test_datastructures.py
@@ -230,6 +230,9 @@ def test_headers() -> None:
     assert "b" in h
     assert "B" in h
     assert "c" not in h
+    assert 123 not in h
+    assert None not in h
+    assert "\u20ac" not in h
     assert h["a"] == "123"
     assert h.get("a") == "123"
     assert h.get("nope", None) is None


Mukller (approved):
Verified locally on the PR branch (Python 3.13, `git checkout` of the head commit):

**Tests**: full `tests/test_datastructures.py` — 44/44 green.

**Behavior matrix**:

```python
h = Headers(raw=[(b"a", b"1")])
123 in h        # False
None in h       # False
"\u20ac" in h   # False   (non-latin-1 encodable)
"A" in h        # True    (case-insensitivity preserved)
"a" in h        # True
```

Returning `False` for wrong-typed keys matches Python container convention (`"x" in [1, 2]` → `False`, no raise), and swallowing only the `UnicodeEncodeError` (rather than all exceptions) keeps the guard tight. Tests cover exactly these three classes of input.

Two non-blocking notes:

1. `Headers.get(123)` / `Headers.__getitem__(None)` still raise `AttributeError` (`key.lower()`), i.e. the asymmetry now exists between `in` and accessors. I think that's fine and idiomatic — just flagging in case maintainers want a follow-up for `__getitem__`/`get` to raise a cleaner `TypeError`.
2. CI: both failing/cancelled jobs are lint/infra ("Run linting checks", exit 1 in ~13s). Locally on the touched files: `ruff check`, `ruff format --check` and `mypy starlette/datastructures.py` all pass clean, so the red check looks unrelated to this diff — worth a re-run.

PR #3468, opened 2026-08-23
fix(base): run inner app background tasks after the response is fully sent

### LINKED ISSUE #3458: BackgroundTask starts before the response has been sent when BaseHTTPMiddleware is in the stack
With a `BaseHTTPMiddleware` anywhere in the stack, a `BackgroundTask` attached to a response starts running **before the last byte of the response has been sent** to the client. The [background docs](https://www.starlette.io/background/) state:

> A background task should be attached to a response, and will run only once the response has been sent.

**Version:** starlette 1.6.0
**Python:** 3.13.14

## Repro

Raw ASGI harness with a deliberately slow `send` (simulates a slow client). A pass-through `BaseHTTPMiddleware` is enough to trigger it; without any middleware the docs are honored.

```python
import asyncio, warnings
warnings.simplefilter("ignore")

events = []

async def bg():
    events.append(("background-done", asyncio.get_running_loop().time()))

async def homepage(request):
    return PlainTextResponse("hello", background=BackgroundTask(bg))

app = Starlette(routes=[Route("/", homepage)])
app = BaseHTTPMiddleware(app, dispatch=passthrough)

async def main():
    scope = {"type": "http", "method": "GET", "path": "/", "headers": [],
             "query_string": b"", "root_path": "", "scheme": "http",
             "server": ["t", 80], "client": ("t", 1), "http_version": "1.1"}
    async def receive():
        await asyncio.sleep(3600)
    async def send(message):
        events.append((message["type"], asyncio.get_running_loop().time()))
        await asyncio.sleep(0.15)   # slow client
    for name, t in sorted(events, key=lambda e: e[1]):
        print(f"  {name:<30} @{t:.3f}")

asyncio.run(main())
```

Actual ordering (deterministic across runs):

```
http.response.start        @t+0.000
http.response.body(more=True)  @t+0.155
background-done            @t+0.155      <-- task finished HERE
http.response.body(more=False) @t+0.310  <-- last byte only here
```

## Why it happens

When `BaseHTTPMiddleware` wraps the app, the inner `Response.__call__` hands its final message into the middleware's memory stream and then runs the background task immediately — while the outer `_StreamingResponse` created from the buffered messages is still forwarding bytes to the real `send`. Without the middleware, the inner response *is* the outermost sender and runs the task after the final message.

## Impact

Tasks that assume "response fully delivered" (e.g. logging request completion, cache invalidation after streaming, metrics that should not overlap body transfer) start early under this common middleware setup. With large/slow bodies the gap can be significant.

Suggested direction: defer `await response(background=...)`-style execution in `BaseHTTPMiddleware._StreamingResponse` until its own final `send` completes (or document the caveat explicitly).

Happy to work on a fix PR if maintainers consider this worth aligning with the documented contract.

webdevsamran:
## Description

With `BaseHTTPMiddleware` anywhere in the stack, a `BackgroundTask` attached by the inner application starts running **before the last byte of the response has been sent** to the client, violating the documented contract that background tasks "run only once the response has been sent".

Deterministic repro (slow client simulates delivery latency):

```
http.response.start             @t+0.000
http.response.body(more=True)   @t+0.155
background-done                 @t+0.155   <-- task finished here
http.response.body(more=False)  @t+0.310   <-- last byte only here
```

### Cause

`BaseHTTPMiddleware.coro()` runs the inner app to completion — including its background tasks — as soon as its messages have been handed off into the internal memory stream, independently of when `_StreamingResponse` actually forwards them to the server.

### Fix

In `send_no_error`, after handing off the final body message (`more_body: false`), hold the inner application until `response_sent` is set. This serializes: inner app finishes sending → outer response fully streamed → inner app's background tasks run. There is no deadlock: `response_sent` is set by the outer flow after `await response(...)` returns, which does not depend on the inner app completing.

## Testing

New test `test_background_task_runs_after_response_fully_sent` drives the raw ASGI callable with a deliberately slow `send` and asserts the exact event ordering (start, chunk, final body, then background-done).

- Red/green verified: the test fails on unpatched `main` (`'background-done' != 'http.response.body'` at index 1) and passes with the fix.
- Full `tests/middleware/test_base.py`: **61 passed, 2 xfailed** — no regressions.

Fixes #3458

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3468?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/middleware/base.py
+++ b/starlette/middleware/base.py
@@ -132,6 +132,12 @@ class BaseHTTPMiddleware:
             async def send_no_error(message: Message) -> None:
                 try:
                     await send_stream.send(message)
+                    if message["type"] == "http.response.body" and not message.get("more_body", False):
+                        # Hold the inner application until the response has
+                        # actually been sent, so background tasks it attaches
+                        # only run once the last byte has reached the client,
+                        # as documented in starlette.io/background.
+                        await response_sent.wait()
                 except anyio.BrokenResourceError:
                     # recv_stream has been closed, i.e. response_sent has been set.
                     return
--- a/tests/middleware/test_base.py
+++ b/tests/middleware/test_base.py
@@ -1316,3 +1316,52 @@ def test_error_context_propagation(test_client_factory: TestClientFactory) -> No
     assert str(ctx.value) == "Outer exception"
     assert ctx.value.__cause__ is not None
     assert str(ctx.value.__cause__) == "Inner exception"
+
+
+# https://github.com/encode/starlette/issues/3458
+@pytest.mark.anyio
+async def test_background_task_runs_after_response_fully_sent() -> None:
+    """A background task attached by the inner app must run only after the last
+    byte of the response has been sent, even with BaseHTTPMiddleware stacked."""
+
+    async def bg() -> None:
+        events.append("background-done")
+
+    events: list[str] = []
+
+    async def homepage(request: Request) -> PlainTextResponse:
+        return PlainTextResponse("hello", background=BackgroundTask(bg))
+
+    async def passthrough(
+        request: Request, call_next: RequestResponseEndpoint
+    ) -> Response:
+        return await call_next(request)
+
+    app: ASGIApp = Starlette(routes=[Route("/", homepage)])
+    app = BaseHTTPMiddleware(app, dispatch=passthrough)
+
+    async def receive() -> Message:
+        return {"type": "http.request"}
+
+    async def send(message: Message) -> None:
+        # A deliberately slow client: without the fix, the inner app's
+        # background task runs while the response is still being streamed.
+        await anyio.sleep(0.05)
+        events.append(str(message["type"]))
+
+    scope: Scope = {
+        "type": "http",
+        "method": "GET",
+        "path": "/",
+        "headers": [],
+        "query_string": b"",
+        "root_path": "",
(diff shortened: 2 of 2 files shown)


official-burak (commented):
The happy path is right: waiting in `send_no_error` after the inner last body keeps the inner app (and its `BackgroundTask`) from finishing until `__call__` has actually written to the client.

`response_sent.set()` only runs if `response()` returns. If `send` raises (client gone, write error), the event is never set. `send_no_error` is then blocked in `wait()`, not in `send_stream.send`, so `BrokenResourceError` does not wake it. The inner task sits there until the collapsing task group cancels it on the way out. Inner background work never runs, and the original send error gets mixed with a cancellation.

Set the event in a `finally` so the inner send unblocks either way:

```python
try:
    await response(scope, wrapped_receive, send)
finally:
    response_sent.set()
    recv_stream.close()
```

The existing test only covers a slow but successful client. A `send` that raises would catch this.

PR #3523, opened 2026-09-06
Stop file transfers when HTTP clients disconnect

Kludex:
`FileResponse` can continue reading an entire file after the client disconnects on ASGI versions before 2.4. Run its existing file handler alongside an `http.disconnect` listener and shield file closure from cancellation, so aborted full-file and range transfers stop and close the file before background cleanup.

Alternative to [#3390](https://github.com/Kludex/starlette/pull/3390/), following the focused approach in [libratechw's proposal](https://github.com/Kludex/starlette/pull/3390#issuecomment-5548572632): preserve `StreamingResponse`, handler signatures, and the direct HEAD, pathsend, WebSocket denial, and ASGI 2.4+ paths.

Validation: `scripts/test` (1,295 passed, 100% coverage), `scripts/build`, and the four existing file-response benchmark cases passed; regression tests cover both asyncio and Trio, and fail against both the original implementation and an unshielded `aclosing()` variant.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## AI Disclaimer


starlette/responses.py
```diff
@@ -376,24 +377,52 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
 
+    @asynccontextmanager
+    async def _open_file(self) -> AsyncIterator[anyio.AsyncFile[bytes]]:
+        file = await anyio.open_file(self.path, mode="rb")
+        try:
+            yield file
+        finally:
+            # Closing must finish even when the transfer is cancelled.
+            with anyio.CancelScope(shield=True):
+                await file.aclose()
```
Kludex, line 416: @agronholm Isn't there something builtin in anyio that I can use instead of my own wrapper?

agronholm: This particular issue was fixed in AnyIO v4.15.0.

Kludex: Are you sure? My coding agent seems to think only `TemporaryDirectory` was fixed. 🤔

agronholm: You're right. Somehow I was under the impression that it was a broader fix for open files, but it's not. The same issue is present in `AsyncFile` too. I'll get this fixed for the next patch release.

agronholm: https://github.com/agronholm/anyio/pull/1314

Shriprasad-P (commented):
Thanks for the PR. Looks focused from a first pass — please confirm tests cover the new path.

PR #3530, opened 2026-09-08
Handle non-standard status codes in HTTPException defaults

ryanchou1994:
# Summary

`HTTPException(499)` raises `ValueError` when no detail is supplied, even though the same status works with an explicit detail. Use an empty default detail for non-standard integer codes between 100 and 599. Out-of-range codes still raise `ValueError` when detail is omitted; explicitly supplied details and headers keep their existing behavior.

Refs https://github.com/Kludex/starlette/discussions/3499.

Tests exercise actual endpoint responses through asyncio and trio, plus invalid-code boundaries, invalid fallback types, and the existing explicit-detail behavior. Validation on macOS and Linux with Python 3.13: 1,263 passed, 2 expected failures, and 100% coverage. Ruff, mypy, package checks, and documentation builds pass on both platforms. The original code fails the four endpoint regression cases for 499 and 599.

# Checklist

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/exceptions.md
+++ b/docs/exceptions.md
@@ -121,7 +121,7 @@ The `ExceptionMiddleware` implementation defaults to returning plain-text HTTP r
 * `HTTPException(status_code, detail=None, headers=None)`
 
 When `detail` is not provided, it defaults to the standard phrase for the status code,
-or an empty string for a non-standard status code.
+or an empty string for a non-standard status code between 100 and 599.
 
 You should only raise `HTTPException` inside routing or endpoints.
 Middleware classes should instead just return appropriate responses directly.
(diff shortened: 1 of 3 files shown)


Kludex:
- Closed by https://github.com/Kludex/starlette/pull/3545

PR #3542, opened 2026-09-10
Match host names case-insensitively in `Host` routing and `TrustedHostMiddleware`

sachhg:
## Summary

Host names are case-insensitive ([RFC 9110, section 4.2.3](https://www.rfc-editor.org/rfc/rfc9110#section-4.2.3)), but Starlette compares them case-sensitively. A request carrying an uppercase or mixed-case `Host` header is rejected by `TrustedHostMiddleware` and fails to match a `Host` route, even though the header is valid and names the same host.

`_HOST_RE` already accepts these headers (it is compiled with `re.IGNORECASE`), so they pass parsing and then fail the comparison.

## Reproduction

On `main`:

```python
from starlette.middleware.trustedhost import TrustedHostMiddleware
from starlette.routing import Host, Route, Router

app = Starlette(
    routes=[Route("/", lambda r: PlainTextResponse("OK"))],
    middleware=[Middleware(TrustedHostMiddleware, allowed_hosts=["example.com"])],
)
client = TestClient(app)
print(client.get("/", headers={"host": "example.com"}).status_code)  # 200
print(client.get("/", headers={"host": "EXAMPLE.com"}).status_code)  # 400

router = Router(routes=[Host("example.org", app=PlainTextResponse("hello"))])
client = TestClient(router)
print(client.get("/", headers={"host": "example.org"}).status_code)  # 200
print(client.get("/", headers={"host": "EXAMPLE.org"}).status_code)  # 404
```

## Changes

`compile_path` already distinguishes host patterns from path patterns via `is_host`, so host regexes are now compiled with `re.IGNORECASE`. Path matching is untouched and stays case-sensitive.

`TrustedHostMiddleware` normalizes `allowed_hosts` once at construction and lowercases the parsed host before comparing, so the exact match, the wildcard suffix match, and the `www` redirect check are all case-insensitive.

## Notes

`URL` construction is deliberately unchanged: `request.url` still echoes the case the client sent, and the `www` redirect preserves it, so this only affects matching.

Lowercasing is safe against Unicode case-folding surprises here. Header values are decoded as latin-1, and no non-ASCII character reachable that way case-folds into the host character class, so characters such as U+212A KELVIN SIGN cannot be used to reach an allowed host.

Uppercase patterns now work as well: `allowed_hosts=["Example.com"]` and `Host("EXAMPLE.org")` previously matched nothing, because the parsed host can never contain uppercase after normalization.

## Testing

Added parametrized coverage for exact, wildcard, uppercase-pattern, port-suffix, and `www` redirect cases in `tests/middleware/test_trusted_host.py`, and for `Host` routing including a `{subdomain}` pattern in `tests/test_routing.py`. All 26 fail on `main` and pass with this change.

`scripts/test` is green: 1269 passed, 4 xfailed, 100% coverage, `ruff` and `mypy` clean. The four xfails are pre-existing and unchanged. Also verified on Python 3.10, the minimum supported version.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3542?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/starlette/middleware/trustedhost.py
+++ b/starlette/middleware/trustedhost.py
@@ -25,7 +25,8 @@ class TrustedHostMiddleware:
             if pattern.startswith("*") and pattern != "*":
                 assert pattern.startswith("*."), ENFORCE_DOMAIN_WILDCARD
         self.app = app
-        self.allowed_hosts = list(allowed_hosts)
+        # Host names are case-insensitive. See RFC 9110, section 4.2.3.
+        self.allowed_hosts = [pattern.lower() for pattern in allowed_hosts]
         self.allow_any = "*" in allowed_hosts
         self.www_redirect = www_redirect
 
@@ -42,7 +43,7 @@ class TrustedHostMiddleware:
         if parsed_host is None:
             await PlainTextResponse("Invalid host header", status_code=400)(scope, receive, send)
             return
-        host = parsed_host.host
+        host = parsed_host.host.lower()
 
         is_valid_host = False
         found_www_redirect = False
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -485,7 +485,9 @@ class Host(BaseRoute):
             parsed_host = parse_host_header(headers.get("host"))
             if parsed_host is None:
                 return Match.NONE, {}
-            host = parsed_host.host
+            # Match against the lowercased host so that captured host params are
+            # canonical, as they were when only lowercase hosts could match.
+            host = parsed_host.host.lower()
 
             match = self.host_regex.match(host)
             if match:
--- a/starlette/routing.py
+++ b/starlette/routing.py
@@ -163,7 +163,9 @@ def compile_path(
 
     path_format += path[idx:]
 
-    return re.compile(path_regex), path_format, param_convertors
+    # Host names are case-insensitive. See RFC 9110, section 4.2.3.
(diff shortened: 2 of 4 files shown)


Shriprasad-P (commented):
Thanks for the PR. Looks focused from a first pass — please confirm tests cover the new path.

sachhg:
Thanks @Shriprasad-P. Yes, every changed line runs under the new tests. Measured with coverage on this branch:

- `trustedhost.py:29`, lowercasing the allowed host patterns
- `trustedhost.py:46`, lowercasing the request host
- `routing.py:167-168`, compiling host patterns with `re.IGNORECASE`
- `routing.py:490`, matching against the lowercased host

The new tests are `test_trusted_host_middleware_is_case_insensitive` and `test_www_redirect_is_case_insensitive` in `tests/middleware/test_trusted_host.py`, and `test_host_routing_is_case_insensitive` and `test_host_routing_params_are_lowercased` in `tests/test_routing.py`. Against the current source all 28 of their cases fail, and with this change both modules pass in full, 177 tests.

PR #3568, opened 2026-09-20
Preserve FileResponse status for range requests

lowgame:
# Summary

`FileResponse` evaluated `Range` headers regardless of its configured status. As a result, `FileResponse(path, status_code=404)` returned `206 Partial Content` and a truncated body when the request included a valid byte range; this also affected custom `404.html` responses served by `StaticFiles`.

Range evaluation only applies when the selected response would otherwise be `200 OK`. This change sends the normal full response whenever `FileResponse.status_code` is not 200, preserving its explicit status and body. A regression test covers the caller-visible 404 case.

# Test plan

- `GITHUB_ACTIONS=1 scripts/test tests/test_responses.py` — 142 passed
- `scripts/test` — 1271 passed, 2 xfailed, 2 failed
  - Both failures were the environment-dependent root-permission cases in `test_staticfiles_with_invalid_dir_permissions_returns_401`.
  - `GITHUB_ACTIONS=1 scripts/test tests/test_staticfiles.py::test_staticfiles_with_invalid_dir_permissions_returns_401` produced the same two failures on the clean upstream commit.
- `scripts/check` (run by `scripts/test`) — Ruff formatting, mypy, and Ruff checks passed
- `git diff --check` — passed

# Issue number

No existing issue; this was found by code audit.

# Checklist

- [x] I've added a test for the behavior change and kept the change atomic.
- [x] Documentation is unchanged because this fixes existing HTTP response semantics without changing the public API.

<a href="https://cubic.dev/pr/Kludex/starlette/pull/3568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kludex (approved):
Thanks!

